PR #18, opened 2015-04-13
fix cache line index offset in cachesim.cc

wsong83:
When define the cache block size using S:W:B, the conversion from B to index offset (idx_shift in cache_sim_t) seems wrong. For example, when B=64, the calculated idx_shift is 7, but it should be 6. This causes an under-estimated miss rate in caches.


aswaterman:
Thanks.

PR #19, opened 2015-04-14
Added basic block tracking functionality required for creating simpoints

sherry151:
1) bb_tracker_t class added which has functions and variables needed for tracking basic blocks.
2) execute_insn() in processor.cc instrumented to dump basic block vector.
3) Added a configure flag "--enable-simpoint" to optionally enable compilation of basic block tracking support in spike. By default it is disabled.
4) Added the additional files in riscv.mk.in.


sherry151:
The configure scricpt must be regenerated with a new version of autoconf before releasing. I did not commit the configure script as I am running an older version of autoconf.

ccelio:
Can you provide an addition to the README on how to use this? We'd like to check this out and try it on our end.

sherry151:
Added README section. I will add a spike flag to enable/disable this functionality at runtime when I have some time.

ccelio:
Thanks for your contribution. However, we've decided the best course of action is to turn Spike into a library so that other users can wrap around it with their added functionality (simpoints, checkpoint creation, cache simulators, etc.), as opposed to pushing too much functionality into Spike.  Until this happens, you should maintain this on a different fork. Keep in touch though, this is functionality that's really helpful to the RISC-V ecosystem.

PR #20, opened 2015-05-13
various fixes to make riscv-isa-sim installable

palmer-dabbelt:
I want to build my extension library without touching the riscv-isa-sim source repo.  Without these changes it's kind of a pain to do so.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -55,14 +55,14 @@ enable_stow  := @enable_stow@
 
 ifeq ($(enable_stow),yes)
   stow_pkg_dir := $(prefix)/pkgs
-  DESTDIR ?= $(stow_pkg_dir)/$(project_name)-$(project_ver)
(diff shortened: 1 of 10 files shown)


rndmcnlly:
Oooh! I have an awkward fork of risc-isa-sim that exists only to host an extension I know is never going to be standardized. Making it easy to develop and maintain extension simulators outside of this tree would be a big deal for me.

palmer-dabbelt:
Well, that's what the PR is supposed to be for :).  @aswaterman if you don't CR this I'm just going to merge it.

@rndmcnlly My empty extension is available here

  https://github.com/palmer-dabbelt/hurricane-spike

in case you want to copy it.  Spike will load arbitrary shared objects with "--extlib", and then you can proceed to load extensions like normal (with "--extension").

aswaterman:
I skimmed this and nothing jumps out at me, so I'll go ahead and merge it.

PR #22, opened 2015-05-26
RV32: Fix bugs

heshamelmatary:


aswaterman:
The Spike convention is to keep 32-bit values sign-extended when stored in 64-bit registers; this mirrors what a HW implementation that supports both RV32 and RV64 would do.  This is also why the branch instructions don't need the cast to [u]int32_t when in 32-bit mode.

What code sequence failed that made you want to make these changes?

heshamelmatary:
For blt:

uint32_t blt_test(uint32_t x)
{
  while (!(x & 0x80000000U));
  return 0;
}

A blt_test(1); call will fail, 
and similar comparisons like (x <= x80000000U) will also fail. Mainly because the compiler emits bltz and bgez for such C statements. 

---

For auipc:

The "la x1, label" pseudo instruction with a label that has its address equals or higher than 0x80000000 will produce equivalent auipc code that's 64-bit sign extended, so a subsequent "jr x1" instruction will also fail, jumping to 64-bit address led with 0xFFFFFFFF80000000 (if pc = 0) for example. 

---

I understand these could be a compiler issues.

aswaterman:
On Wed, May 27, 2015 at 1:10 AM, Hesham ALMatary notifications@github.com
wrote:

> For blt:
> 
> uint32_t blt_test(uint32_t x)
> {
> while (!(x & 0x80000000U));
> return 0;
> }
> 
> A blt_test(1); call will fail,
> and similar comparisons like (x <= x80000000U) will also fail. Mainly
> because the compiler emits bltz and bgez for such C statements.

bgez is the right thing to emit here, because 32-bit types are stored
sign-extended in 64-bit registers--even if the 32-bit type is unsigned.  If
'x' is not sign-extended upon entry into blt_test, it is because there is a
bug elsewhere, perhaps in the function that calls blt_test.

See the bottom of page 23 of the ISA spec (
http://riscv.org/spec/riscv-spec-v2.0.pdf) for the rationale.

> ---
> 
> For auipc:
> 
> The "la x1, label" pseudo instruction with a label that has its address
> equals or higher than 0x80000000 will produce equivalent auipc code that's
> 64-bit sign extended, so a subsequent "jr x1" instruction will also fail,
> jumping to 64-bit address led with 0xFFFFFFFF80000000 (if pc = 0) for
> example.

The same reasoning applies here.

heshamelmatary:
Thanks for the detailed information and the reference! I've re-implemented my algorithms with this default 64-bit sign extension in mind, and all the code is working fine now.

FWIW, when debugging on spike and trying to use "until pc 0 90000000", this one fails to break, however "until pc 0 ffffffff90000000" works fine. This was kinda misleading for me.

aswaterman:
Glad to hear stuff's working.  I agree this behavior of the debug interface
doesn't make much sense; it should transparently do the right thing for
RV32.  I'll fix that soon.

PR #23, opened 2015-06-03
move interactive function init out of main loop

vapier:
These values don't change between runs, so there's no point in recreating
them on every instruction step.  Pull it out to the top to simplify.


aswaterman:
Thanks.

PR #24, opened 2015-06-04
improve interactive behavior a bit more

vapier:
Keep making more user friendly.


aswaterman:
Thanks, Mike.  This is quite helpful.

The interactive mode is kind of a historical aberration; I originally created it just to debug the simulator itself.  But I guess it's just barely useful enough that others want to use it, too :-)

vapier:
i find it useful to compare two different simulators and make sure they're behaving the same.  going through more layers just slows that process down :).

PR #27, opened 2015-07-11
fix clang compile error

sbeamer:
mmu.cc uses abort() but none of the included headers contain it. This mimics the include pattern of other headers in the project.


aswaterman:
thanks!

Issue #26, opened 2015-07-11
warning from clang


sbeamer:
When compiling with clang under OS X the following warning is generated:

g++ -fPIC -MMD -MP -I/usr/local/riscv/include  -Wall -Wno-unused -O2 -std=c++11 -DPREFIX=\"/usr/local/riscv\" -I. -I../riscv -I../hwacha -I../dummy_rocc -I../softfloat -I../spike_main -c ../riscv/interactive.cc
../riscv/interactive.cc:186:14: warning: if statement has empty body
      [-Wempty-body]
      if (0) ;
             ^
../riscv/interactive.cc:186:14: note: put the semicolon on a separate line to
      silence this warning

I would try to fix this but I'm not sure what the code in interactive.cc is attempting to do or what optimization it is leveraging.

PR #28, opened 2015-07-13
fixes clang warning (#26)

sbeamer:
The clang warning (#26) was caused by an empty if statement. Since the list of preprocessor generated if statements all have return clauses, the else isn't needed for the default case.


aswaterman:
Thanks.

PR #29, opened 2015-08-06
Enable execution log generation in non-interactive mode.

pmundkur:
There didn't seem to be an obvious way to do this using the existing options, so I added one.

Add an option (-l) to display a log of execution in non-interactive mode.   Interactive (-d) mode overrides this option when both are specified.


aswaterman:
Looks good to me, thanks.

PR #31, opened 2015-09-11
more portable generation with gsed to fix #30

### LINKED ISSUE #30: OS X build is broken
The OS X build breaks partway through due to the new commit 71d04ecd7ab6 with the following error:

```
In file included from ../riscv/processor.cc:566:
./insn_list.h:1:17: error: expected ';' at end of declaration
DEFINE_INSN(add addi addiw addw amoadd_d amoadd_w amoand_d amoand_w amomax_d amomaxu_d amom...
            ^
../riscv/processor.cc:565:25: note: expanded from macro 'DEFINE_INSN'
    REGISTER_INSN(this, name, opcodes[#name].first, opcodes[#name].second)
                        ^
../riscv/processor.h:141:23: note: expanded from macro 'REGISTER_INSN'
  extern reg_t rv32_##name(processor_t*, insn_t, reg_t); \
                      ^
```

It seems like this may be caused by a difference in GNU sed and BSD sed. I have gsed installed and changing riscv.mk fixes this bug, but it would be great to not have gsed be a dependence.

sbeamer:
Tested on OS X 10.10 and Linux.

### DIFF
--- a/riscv/riscv.mk.in
+++ b/riscv/riscv.mk.in
@@ -251,7 +251,7 @@ icache.h: mmu.h
 	mv $@.tmp $@
 
 insn_list.h: $(src_dir)/riscv/riscv.mk.in
-	echo $(riscv_insn_list) | sed 's/\s\+\|$$/\n/g' | sed '/^$$/d' | sed 's/\./_/g' | sed 's/\(.*\)/DEFINE_INSN(\1)/' > $@.tmp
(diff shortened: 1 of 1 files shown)


sbeamer:
Uh oh, slight data race. Looks like also fixed by 01189a9395 so there are choices.

a0u:
Oops, I didn't see this before opening #32.  I'll let @aswaterman decide which to merge.

aswaterman:
I merged #32.

PR #37, opened 2016-01-20
enable extlibs to register devices

jameyhicks:
This change enables shared libraries loaded via --extlib to register devices.


jerryz123:
This is now supported.

PR #40, opened 2016-05-16
Added missing header files to riscv.mk.in

acw1251:
This pull request fixes missing include files in `<prefix>/include/spike` after doing `make install`.


gkelly:
Looks good to me!

aswaterman:
thanks!

PR #44, opened 2016-05-21
htif: catch proper store exception

gkelly:
The sim's debug_mmu will generate a trap_load_access_fault if the load
fails, not a trap_store_access_fault.  This doesn't materially affect
exection, but results in a nicer log message.


aswaterman:
thanks

PR #51, opened 2016-06-07
make check: Fail if the tests failed

neuschaefer:
The goal is to integrate the spike tests into riscv-tools' .travis.yml and to properly detect test failures.


aswaterman:
(not specific to your PR, but) shouldn't we just be checking for spike return codes, rather than grepping logs?

neuschaefer:
That's probably a better approach, yes.

aswaterman:
OK, we'll cross that bridge later.  For now, it's merged.  Thanks.

PR #53, opened 2016-06-10
Implement a minimal MMIO 8250 UART

neuschaefer:
See the commit message for more information.

### DIFF
--- a/riscv/devices.h
+++ b/riscv/devices.h
@@ -76,4 +76,11 @@ class clint_t : public abstract_device_t {
   std::vector<mtimecmp_t> mtimecmp;
 };
 
+class uart_t : public abstract_device_t {
+ public:
+  bool load(reg_t addr, size_t len, uint8_t* bytes);
+  bool store(reg_t addr, size_t len, const uint8_t* bytes);
+  size_t size() { return UART_SIZE; }
+};
+
 #endif
--- a/riscv/dts.cc
+++ b/riscv/dts.cc
@@ -65,6 +65,18 @@ std::string make_dts(size_t insns_per_rtc_tick, size_t cpu_hz,
   s << std::hex << ">;\n"
          "      reg = <0x" << (clintbs >> 32) << " 0x" << (clintbs & (uint32_t)-1) <<
                      " 0x" << (clintsz >> 32) << " 0x" << (clintsz & (uint32_t)-1) << ">;\n"
+         "    };\n";
+
+  reg_t uartbs = UART_BASE;
+  reg_t uartsz = UART_SIZE;
+  s << std::hex <<
+         "    uart@" << UART_BASE << " {\n"
+         "      compatible = \"serial\";\n"
+         "      reg = <0x" << (uartbs >> 32) << " 0x" << (uartbs & (uint32_t)-1) <<
+                     " 0x" << (uartsz >> 32) << " 0x" << (uartsz & (uint32_t)-1) << ">;\n"
+         "      reg-shift = <0>;\n"
+         // TODO: interrupts = <...>;
+         // TODO: clock-frequency or clocks
          "    };\n"
          "  };\n"
          "  htif {\n"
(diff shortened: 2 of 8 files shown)


neuschaefer:
The pupose of this is to have a console interface in spike that's easier to use in coreboot than the HTIF, since the HTIF requires special ELF symbols, and coreboot compiles to a "ROM image" and is only later turned back into an ELF that spike can load.

aswaterman:
We're getting rid of the HTIF soon, in favor of the standard RISC-V Debug port.  The intent is that the debug module will expose serial ports.  These will show up as devices in the memory map, and will be the standard way of generating debug console output during boot.

Your approach makes sense to me, but I would prefer to see the debug serial stuff through before deciding to add another console device.  I'll try to minimize disruption so it's feasible to maintain this on a branch in the mean time.

neuschaefer:
That makes sense, thanks.

riscv/devices.cc
```diff
@@ -1,4 +1,5 @@
 #include "devices.h"
+#include <stdio.h>
```
neuschaefer: Oops, this line shouldn't be in the commit.

frantony:
@aswaterman  

> We're getting rid of the HTIF soon, in favor of the standard RISC-V Debug port. The intent is that the debug module will expose serial ports. These will show up as devices in the memory map, and will be the standard way of generating debug console output during boot.

At the moment HTIF is _still_ only serial port for riscv-sim-isa.
I can't find exhaustive debug module serial port description. The document [RISC-V External Debug Support Version 0.11 (sep 22 2016)](https://dev.sifive.com/documentation/risc-v-external-debug-support/) has draft status (e.g. Serial Send 0 register's address is not defined).
Is anybody working on adding _standard RISC-V Debug port_ support to riscv-sim-isa?

timsifive:
On Sun, Oct 30, 2016 at 5:11 PM, Antony Pavlov notifications@github.com
wrote:

> @aswaterman https://github.com/aswaterman
> 
> We're getting rid of the HTIF soon, in favor of the standard RISC-V Debug
> port. The intent is that the debug module will expose serial ports. These
> will show up as devices in the memory map, and will be the standard way of
> generating debug console output during boot.
> 
> At the moment HTIF is _still_ only serial port for riscv-sim-isa.
> I can't find exhaustive debug module serial port description. The document RISC-V
> External Debug Support Version 0.11 (sep 22 2016)
> https://dev.sifive.com/documentation/risc-v-external-debug-support/ has
> draft status (e.g. Serial Send 0 register's address is not defined).
> Is anybody working on adding _standard RISC-V Debug port_ support to
> riscv-sim-isa?
> 
> There is a RISC-V Debug Task Group, and I am its chair. We haven't
> officially met yet, but tentatively the group's goal is to decide on a
> standard debug interface. That will probably include some serial support.
> There's been some discussion on the mailing list (
> https://groups.google.com/a/groups.riscv.org/forum/#!forum/debug), but
> you're the first to express real interest in the serial support.

Tim

neuschaefer:
I, for one, don't care as much how the serial port works, but I do care that there is a serial port, which doesn't require complicated setup — i.e. it should be usable before DRAM is available, and shouldn't require any expensive extra hardware. A simple serial port can make low-level debugging a lot easier.

aswaterman:
The plan is to make Spike accept hardware devices as plugins, rather than baking a specific character device into the simulator.  When we do that, we'll also release a model of our platform UART (so the same software will also work on the FPGA models and ASIC implementations).

sorear:
@aswaterman You are referring to the part which sifive/riscv-pk#sifive uses for the SBI console?  FYI I have a model for that in QEMU ( https://github.com/riscv/riscv-qemu/blob/master/hw/riscv/sifive_uart.c ) which is complete enough for BBL.

aswaterman:
Yeah, that one, thanks.  Of course, as it's a prototype, we're not certain
whether we'll keep that exact register interface.

PR #56, opened 2016-06-28
Add gdb support for RV32

timsifive:
The debug code figures out what XLEN is by running a program in Debug RAM, just like we have to do on real hardware. Once figured out, the only real difference is to use lw/sw instead of ld/sd, but to make that nice required a bunch of refactoring.

I have testcases for this at https://github.com/timsifive/riscv-tests/tree/debug/debug

Once I get everything passing, I'll get that merged. Then I can remove the gdb tests from the riscv-isa-sim. They belong much more in riscv-tests since they test both gdb and spike. Additionally, the same tests are used to test gdb with openocd.

### DIFF
--- a/riscv/gdbserver.cc
+++ b/riscv/gdbserver.cc
@@ -50,22 +50,11 @@ enum {
   REG_END = 4161
 };
 
-// Return access size to use when writing length bytes to address, so that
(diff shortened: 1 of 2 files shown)


aswaterman:
danke

PR #59, opened 2016-07-28
Add support for virtual priv register.

timsifive:
Users can use this register to inspect and change the privilege level of
the core. It doesn't make any assumptions about the actual underlying
debug mechanism (as opposed to having the user change DCSR directly,
which may not exist in all debug implementations).


aswaterman:
lgtm

PR #61, opened 2016-08-16
remove old rvc directory

colinschmidt:
This directory was left around after the "rip out rvc" commit. Since it all lives in riscv/ now we can just delete it. @aswaterman


aswaterman:
oops, good call

PR #62, opened 2016-08-22
Implement address and data triggers.

timsifive:
So far I only have testcases for instruction and data address.
Not implemented is the mechanism that lets the debugger prevent a user
program from using triggers at all. I'll be adding that soonish.

The critical path is unchanged, but my experimenting shows the
simulation is slowed down about 8% by this code. Reducing the size of
trigger_match() (which is never called during my benchmark) fixes that,
but making it not be inlined has no effect. I suspect the slowdown comes
from cache alignment or something similar, and on a different CPU or
after more code changes the speed will come back.


aswaterman:
Can you make a corresponding PR to riscv-opcodes for encoding.h and the CSR listing?

timsifive:
Done at https://github.com/riscv/riscv-opcodes/pull/10

timsifive:
I'd like this merge soonish if possible. At worst the code doesn't break anything, and the branch was cut off from master quite some time ago.

PR #67, opened 2016-09-13
restore clang support by fixing printf identifiers

sbeamer:


riscv/debug_module.cc
```diff
@@ -19,7 +19,7 @@ bool debug_module_t::load(reg_t addr, size_t len, uint8_t* bytes)
     return true;
   }
 
-  fprintf(stderr, "ERROR: invalid load from debug module: %ld bytes at 0x%lx\n",
+  fprintf(stderr, "ERROR: invalid load from debug module: %ld bytes at 0x%llx\n",
```
aswaterman: reg_t (i.e. uint64_t) should use PRIx64 (i.e., "0x%" PRIx64 "\n") -- I think that's what we do elsewhere in Spike.  Some compilers complain about %llx to print a uint64_t.

sbeamer:
I just overwrote the commit to use the PRI identifiers.

aswaterman:
thanks!

PR #70, opened 2016-09-29
Update trigger behavior.

timsifive:
M-mode writes to tdata1 with dmode set are ignored instead of raising an
exception.
Add the same behavior for tdata2.


aswaterman:
thanks!

PR #75, opened 2016-11-12
Ensure that g++ knows it is building a PCH

bgamari:
It seems that g++ 5.4 doesn't realize that it is building a precompiled 
header unless you pass it -x c++-header.


aswaterman:
curious.  thanks!

PR #77, opened 2016-12-01
Added comments about the modified Duff's Device in execute.cc

acw1251:
I hope this isn't too many comments. This will hopefully provide a starting point for cleaning up the code a bit.


aswaterman:
Thanks Andy.

PR #81, opened 2016-12-17
Use correct format codes for reg_t and size_t

sorear:
Fixes 32-bit build.  I have not been able to test correct execution.


aswaterman:
Thanks.

sorear:
Confirmed, riscv64 Linux boots in spike compiled in a 32-bit container with this change.

PR #82, opened 2016-12-21
Fix gdb communication error

bacam:
When the send buffer wrapped around it would write the start of the packet again, rather than the end.


timsifive:
That change looks right. Thank you!

Have you tried to run the riscv-tests/riscv-tools tests after this change? The simplest way is:
```
cd riscv-tests/riscv-tools
make clean
make all
```

spike32 might fail if you don't have 32-bit tools built, but spike64 should pass.

bacam:
I've run the spike64 tests without problems.  I don't have the 32-bit tools at the moment (although I'll probably build them soon).

PR #83, opened 2016-12-21
Gdb protocol fixes

bacam:
These fix a couple of problems that don't seem to affect gdb, but can confuse other tools that attempt to use the gdb protocol.  One fixes reading S0 with `p8` (gdb uses `g`), and the other is a duplicate response, which gdb appears to ignore.

### DIFF
--- a/riscv/gdbserver.cc
+++ b/riscv/gdbserver.cc
@@ -592,12 +592,16 @@ class register_read_op_t : public operation_t
       switch (step) {
         case 0:
           if (reg >= REG_XPR0 && reg <= REG_XPR31) {
-            if (gs.xlen == 32) {
-              gs.dr_write32(0, sw(reg - REG_XPR0, 0, (uint16_t) DEBUG_RAM_START + 16));
-            } else {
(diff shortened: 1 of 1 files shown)


timsifive:
I have the exact same comment as for your other pull request. The code looks right. Did you run the test suite?

Thank you,
Tim

bacam:
Running the tests picks up issue #84, because the duplicate response is effectively hiding the error.  It's probably best not to merge this without sorting that out first.

timsifive:
What errors are you seeing? I just ran the testsuite against your branch and everything worked.

bacam:
I get failures for SimpleS1Test and WriteGprs:

    (gdb) p/x $s1=0xe520e31011238d8
    p/x $s1=0xe520e31011238d8
    Could not write register "x9"; remote failure reply 'E03'

and

    (gdb) p/x $x9=478198855569
    p/x $x9=478198855569
    Could not write register "x9"; remote failure reply 'E03'

respectively.  This is for 64 bit; I think 32 bit would work.  If I also apply #85 then it works for 64 bit.

bacam:
I just checked this against the current head, and the debug test suite passes on both 32 bit and 64 bit.

PR #85, opened 2016-12-30
Only read exception flag in gdb register read/write.

### LINKED ISSUE #84: Debug exceptions get overwritten by S1
The debug ROM writes 1s to the last word in the debug RAM to indicate an exception.  However, it also uses the last word for stashing the value of S1 before running the code in the debug RAM, so the exception information is lost.  The gdb support checks for exceptions in a couple of places and can misbehave when S1 is non-zero.  In particular, it will send an error code for `p` and `P` commands, although the latter is masked by an extra `OK` response (see pull request #83).

bacam:
The flag is 32 bits, and if we read 64/128 bits then we get fragments of
S1 too and can accidentally send an error.  Fixes #84.

### DIFF
--- a/riscv/gdbserver.cc
+++ b/riscv/gdbserver.cc
@@ -648,7 +648,7 @@ class register_read_op_t : public operation_t
 
         case 1:
           {
-            unsigned result = gs.dr_read(SLOT_DATA_LAST);
+            unsigned result = gs.dr_read32(DEBUG_RAM_SIZE / 4 - 1);
             if (result) {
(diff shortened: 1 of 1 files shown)


bacam:
I've successfully run the 64 bit tests on this.

timsifive:
Looks great! Thank you!

Andrew, can I merge things like this or would you like to do it?

aswaterman:
I'll merge it while I'm here (but you should feel free to merge similar PRs).  Thanks @bacam @timsifive

PR #86, opened 2016-12-31
Comply with GNU coding standards.

dvc94ch:
Currently the DESTDIR variable is not used correctly which leads to
bogus RUNPATH entries.

https://www.gnu.org/prep/standards/html_node/DESTDIR.html


palmer-dabbelt:
Thanks!

PR #90, opened 2017-01-17
Update MMIO hack

richardxia:
@aswaterman @yunsup 

This PR merges the latest master of riscv-isa-sim into the mmio-hack branch so that we can continue to use the mmio-hack with the more recent changes to riscv-tools.


richardxia:
FWIW, I apparently still have push access to `riscv/riscv-tools` but not `riscv/riscv-isa-sim`, so I actually already bumped the `mmio-hack` branch in `riscv-tools` to point to the HEAD of this PR.

richardxia:
It's slightly unfortunate that this repo is set up to rebase pull requests onto the branch they are merging onto, since now git thinks that `mmio-hack` is 22 commits ahead of master rather than just 2. For future reference, 9ed239dceae9b8674fe42a5f3cb8125ff96045e3 and 6a9fab647a98381f7f6beb9710470d0b7984b949 are the only two commits that are distinct in `mmio-hack`.

aswaterman:
Should we force rollback and merge instead of rebase?

richardxia:
I think it's fine to leave as-is, since we will hopefully throw away this branch in the future anyway. It might just get a little bit more complicated if we need to keep on maintaining this branch.

PR #96, opened 2017-04-26
Updates to match the latest debug spec

palmer-dabbelt:
With these, most of the tests pass against Spike.

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -205,19 +205,19 @@ bool debug_module_t::dmi_read(unsigned address, uint32_t *value)
   D(fprintf(stderr, "dmi_read(0x%x) -> ", address));
   if (address >= DMI_DATA0 && address < DMI_DATA0 + abstractcs.datacount) {
     unsigned i = address - DMI_DATA0;
-    result = read32(dmdata, i);
+    result = abstractcs.busy ? -1 : read32(dmdata, i);
 
     if (abstractcs.busy && abstractcs.cmderr == CMDERR_NONE) {
       abstractcs.cmderr = CMDERR_BUSY;
     }
 
-    if ((abstractauto.autoexecdata >> i) & 1){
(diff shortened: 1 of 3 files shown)


riscv/debug_module.cc
```diff
@@ -205,19 +205,19 @@ bool debug_module_t::dmi_read(unsigned address, uint32_t *value)
...
-    result = read32(dmdata, i);
+    result = abstractcs.busy ? -1 : read32(dmdata, i);
 
     if (abstractcs.busy && abstractcs.cmderr == CMDERR_NONE) {
       abstractcs.cmderr = CMDERR_BUSY;
     }
 
-    if ((abstractauto.autoexecdata >> i) & 1){
+    if (!abstractcs.busy && ((abstractauto.autoexecdata >> i) & 1)) {
       perform_abstract_command();
```
mwachs5, line 215: I thought "perform_abstract_command() also checked for busy, but maybe not.

riscv/debug_module.cc
```diff
@@ -396,23 +396,25 @@ bool debug_module_t::dmi_write(unsigned address, uint32_t value)
   D(fprintf(stderr, "dmi_write(0x%x, 0x%x)\n", address, value));
   if (address >= DMI_DATA0 && address < DMI_DATA0 + abstractcs.datacount) {
     unsigned i = address - DMI_DATA0;
-    write32(dmdata, address - DMI_DATA0, value);
+    if (!abstractcs.busy)
```
mwachs5, line 399: I don't think the HW prevents this sort of access to DATA while busy. Is it necessary for correct behavior? It just flags BUSY error if you do it, but doesn't actually block the write (I think).

mwachs5: Ok you convinced me. The spec should clarify this as well.

mwachs5 (approved):

PR #108, opened 2017-06-05
Configure should fail if device-tree-compiler is not installed

### LINKED ISSUE #107: spike fails unhelpfully without device tree compiler
After building spike without the device tree compiler installed, running spike yields the following message:
```Failed to run : No such file or directory```

It would be nice if the build failed if dtb isn't found.

aswaterman:
Fixes #107

### DIFF
--- a/configure
+++ b/configure
@@ -3409,6 +3409,7 @@ done
   done
 IFS=$as_save_IFS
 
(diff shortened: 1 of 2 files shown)


palmer-dabbelt (approved):
LGTM

PR #109, opened 2017-06-07
Forbid S-mode execution from user memory

aswaterman:
https://github.com/riscv/riscv-isa-manual/commit/285c81746fe664060b62ae0584865dbfa9f42e1a


palmer-dabbelt (approved):
Does Spike enforce this?  I'd like to make we didn't screw anything up in Linux.

aswaterman:
This is the patch that makes Spike enforce it.  Linux still boots.

PR #112, opened 2017-06-09
Return success on writes to abstractauto

timsifive:
This bug was exposed by newer OpenOCD which actually checks the result.


mwachs5:
👍

PR #114, opened 2017-06-14
Support 64-bit start PCs in reset vector.

timsifive:


aswaterman:
And the code's cleaner, too.

PR #116, opened 2017-07-17
disallow FETCH under SUM

DSHorner:
for U-pages disallow FETCH in S or M-modes.


DSHorner:
When can we expect priv-1.10 branch to be promoted to master?

aswaterman:
priv-1.10 was merged into master a while ago.

aswaterman:
It looks like I've already made this change, sorry for the confusion: 3e92055f8323729fdfa72e4ac463ed677424606c

PR #128, opened 2017-09-21
Fix debug reset.

timsifive:
ndmreset now resets all harts (instead of just the current hart), and
hartreset resets the selected hart (instead of being ignored).


mwachs5 (commented):
Is there anything besides `proc`s in spike that would need to be reset?

timsifive:
> Is there anything besides procs in spike that would need to be reset?

Probably, but it doesn't seem to be well-organized. Eg. sim_t::reset() ends up calling new, but I don't see where the pointer is ever cleaned up. Each abstract device may implement reset() as well.

I'm interested in spike as a platform to test external debug (especially automatically). Getting reset to really reset the simulator fully is less important to me.

timsifive:
Can somebody approve this request, or comment on what needs to be changed?

mwachs5 (approved):

timsifive:
Thank you!

PR #129, opened 2017-09-25
Implement Q extension

aswaterman:
I don't have a good way to test this, so I'm sure there are some bugs.  @kito-cheng did you make any tests for binutils/gcc?


kito-cheng:
GCC part: https://github.com/riscv/riscv-gcc/pull/99

palmer-dabbelt (approved):
I just skimmed the code.

PR #130, opened 2017-09-26
Add a timer devicetree node

palmer-dabbelt:
We tightly coupled a handful of drivers for ISA-defined devices (timer
and interrupts) to the RISC-V Linux port, but this isn't kosher so
upstream won't like it.  In order to decouple these drivers from the
RISC-V part of the part, we need device tree nodes so the drivers can
hook into those nodes to initialize themselves.

This adds device tree nodes for the ISA-mandated timer attached to each
hart, which I've also converted Linux to use.


terpstra:
Err. What? There's no reason we should have to put this in device tree. Did they actually push back on this or is this just in anticipation of push back?

PR #137, opened 2017-10-15
Include math.h for NAN

jar:
commit 85c40db208db3e26f507dc6a74a5dc540b504b5c introduced a NAN dependency but did not include the math.h header


aswaterman:
Thanks

Issue #135, opened 2017-10-13
Bug in devices.cc prevents bus_t from finding a device at address 0x0


evancox10:
The code for bus_t does a "clever" trick when storing/looking up devices by address. Instead of storing address, it stores -address (the negative of it). The idea is that this will store the devices in reverse order of address: -0x800, -0x700, ..., -0x100, -0x000. Then, when the lower_bound search is used with an example input of -0x780, the device with base address of 0x700 is returned, rather than 0x800 that would be returned if we stored the device by increasing address order.

There's just one problem with the ordering of my list, however: -0 is still 0! so the actual list order is 0x000, -0x800, -0x700, ..., -0x100. And if we try to search for the device with address between 0x0 and 0x100, e.g. address 0x80, lower_bound skips over the first device, which we wanted, and gives the highest address device instead (0x800).

The fix is to store devices with the regular order, do an _upper_bound_ search, and then decrement the iterator. Some bounds checking needs to occur to check for an empty list and whether we found any item with a base address <= addr.

I've uploaded the fix to a gist: https://gist.github.com/evancox10/c5e6b9ebfa3a8d38d27ad8d81f4eb4cb

I'm not able to verify the code at this moment, but I believe it will work.

Note: this bug is most likely the cause for this issue mentioned in decode.h:
```
// Seems that 0x0 doesn't work.
#define DEBUG_START             0x100
#define DEBUG_END                 (0x1000 - 1)
```

aswaterman:
Thanks, Evan - can you pull-request the fix once you've verified it?

PR #142, opened 2017-10-19
Fix bus_t bug with devices at 0x0

evancox10:
Fix a bug that prevented bus_t from storing to, loading from,
or finding a device that existed at address 0x0.

Resolves: #135


aswaterman:
Thanks.  Merged manually to fix the no-newline-at-end-of-file.

Issue #141, opened 2017-10-19
Commit-Log builds broken due to Q extension


evancox10:
The WRITE_FREG macro in decode.h is broken for builds with commit log enabled (`-DRISCV_ENABLE_COMMITLOG`)

```
#ifndef RISCV_ENABLE_COMMITLOG
# define WRITE_REG(reg, value) STATE.XPR.write(reg, value)
# define WRITE_FREG(reg, value) DO_WRITE_FREG(reg, freg(value))
#else
# define WRITE_REG(reg, value) ({ \
    reg_t wdata = (value); /* value may have side effects */ \
    STATE.log_reg_write = (commit_log_reg_t){(reg) << 1, wdata}; \
    STATE.XPR.write(reg, wdata); \
  })
# define WRITE_FREG(reg, value) ({ \
    freg_t wdata = freg(value); /* value may have side effects */ \
    STATE.log_reg_write = (commit_log_reg_t){((reg) << 1) | 1, wdata.v}; \
    DO_WRITE_FREG(reg, wdata); \
  })
#endif
```

The type of wdata is now `float128_t`, which is typdef'd in softfloat_types.h to `struct { uint64_t v[2]; }`. So, this code is trying to convert from `uint64_t*` to `uint64_t`, which isn't allowed.

To get things compiling you can change `wdata.v` to `wdata.v[0]`. However, I'm not sure if this actually gets you the right data for f32/64 types. I'm not doing any floating point work so can't test this easily. A simple solution for these types is probably available using the box/unbox convenience macros in decode.h .

However, to actually log `float128_t` writes in the commit log, more changes would be needed to expand commit log to support the bigger type.

aswaterman:
@ccelio I guess only printing the lower 64 bits is OK, since nothing that uses the commit log (besides spike) supports the Q extension?  Or should we actually print all 128 bits?

ccelio:
Thanks for the report @evancox10. 

although annoying, we should do things right and go with the least surprising behavior. I assume we can't know (max XLEN FLEN) and print based on that? Then we should do 128b. If space is an issue one could pipe to a compressor.

evancox10:
@aswaterman The commit log is useful as a general purpose tool for anyone implementing their own CPU or functional units. IMO we can't assume that just because the publicly available/known tools do not use it, that therefore nothing does. I'm sure that someone designing a 128-bit FPU from scratch would appreciate known-good output from the reference model (ie spike).

[...]

PR #143, opened 2017-10-20
Fix commit-log for Q extension, and for RV32

aswaterman:
The number of nibbles printed out now depends upon XLEN or FLEN,
as appropriate.


aswaterman:
#141

ccelio (approved):
Thanks for writing this up. Looks like a pretty clean implementation of the most general form (extra painful since misa is writeable). Hopefully the performance impact isn't too noticeable (since we're already writing to printf anyways).

PR #151, opened 2017-11-03
Put HTIF in the device tree

palmer-dabbelt:
I wanted to actually put the address of the HTIF into the DTS, but that
seems to be a bit too much work: since the HTIF addresses are just
defined in an ELF file it's a bit awkward to make that work.

Instead, I'm just putting a dummy HTIF key in the DTS.


terpstra (approved):
Try leaving out reg field, then good.

PR #156, opened 2017-11-15
Support for non-contiguous hartids

p12nGH-zz:


spike_main/spike.cc
```diff
@@ -116,7 +130,7 @@ int main(int argc, char** argv)
   if (mems.empty())
     mems = make_mems("2048");
 
-  sim_t s(isa, nprocs, halted, start_pc, mems, htif_args);
```
palmer-dabbelt, line 119: I wouldn't bother with `std::move` here.  If you just make it a `const vector&` you can elide the copy, but it probably doesn't matter.

palmer-dabbelt (approved):
LGTM

aswaterman:
There's at least one other place in Spike that assumes hart ids are contiguous: sim_t::make_dtb, where it prints cpu@ and _intc.

p12nGH-zz:
Also some commands that take processor id as a parameter don't work correctly as well

mwachs5:
You will probably need to change something in the debug module in here as well, as it assumes (I think) that the mapping is 1:1

mwachs5:
https://github.com/riscv/riscv-isa-sim/blob/58b6c187df1a5cc066c97a63483d297fd159f241/riscv/debug_module.cc#L149

Here, the core writes its `mhartid` to certain addresses, and the debug module uses that value to update internal state, and return `flags` read data. to index directly into "flags". This is basically assuming a 1:1 mapping between hartsel and mhartid, which I think you are changing in this PR?

mwachs5 (changes requested):
This needs some additional changes in the debug_module.cc

mwachs5 (approved):
I rescind my change request, since it doesn't break anything that doesn't use this flag, and doesn't use the debug module

aswaterman:
Merging now since this won't break anything that doesn't use the new flag, and under the promise that @p12nGH will fix the devicetree and debug module soon.

PR #158, opened 2017-11-17
Libreadline

p12nGH-zz:
Adds libreadline capabilities to spike command shell like history and search


aswaterman (changes requested):
Readline is GPL, so we cannot link it into Spike, which is BSD.

PR #162, opened 2017-11-20
Fix commitlog.

ccelio:
A regression caused any instruction with rd=x0 to not be emitted.


ccelio:
This is my shame for not being a better reviewer on #143 .

aswaterman:
whoops

PR #165, opened 2017-12-11
Add support for program buffer of size 2

timsifive:
This allows us to test that OpenOCD can work with that limitation.

The default program buffer size is now 2, but a command line option allows it to be set to any size.

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -16,9 +16,9 @@
 
 ///////////////////////// debug_module_t
 
-debug_module_t::debug_module_t(sim_t *sim, unsigned progsize) :
-  progsize(progsize),
-  program_buffer_bytes(4 + 4*progsize),
+debug_module_t::debug_module_t(sim_t *sim, unsigned progbufsize) :
+  progbufsize(progbufsize),
+  program_buffer_bytes(4 + 4*progbufsize),
   debug_progbuf_start(debug_data_start - program_buffer_bytes),
   debug_abstract_start(debug_progbuf_start - debug_abstract_size*4),
   sim(sim)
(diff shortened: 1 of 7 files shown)


mwachs5:
Which version of the "dmerr" layout does this use (even if irrelevant)?

timsifive:
Currently spike does not implement dmerr at all.

mwachs5:
right, but you are bumping the version of the defines.h file, just wondering if this is the latest or will just create further incompatibilities

timsifive:
I bumped it to get the definition for impebreak. I don't know if this is from before/after dmerr was moved.

mwachs5:
Can we keep it all consistent with the current spec?

mwachs5:
Cool, Looks good here. Presumably you are going to bump riscv-isa-sim and riscv-openocd together at some point in the near future? In that case can you bump riscv-openocd (and riscv-tests?) so that they are at least all consistent with the version of debug_defines they are using?

riscv/debug_module.cc
```diff
@@ -16,30 +16,46 @@
...
   abstractauto = {0};
 
+  program_buffer = new uint8_t[program_buffer_bytes];
+
   memset(halted, 0, sizeof(halted));
   memset(debug_rom_flags, 0, sizeof(debug_rom_flags));
   memset(resumeack, 0, sizeof(resumeack));
-  memset(program_buffer, 0, sizeof(program_buffer));
+  memset(program_buffer, 0, program_buffer_bytes);
+  program_buffer[4*progbufsize] = ebreak();
```
mwachs5, line 44: is it worth making this an option?

timsifive: What exactly are you referring to?

mwachs5: Whether or not implicit ebreak is in the code. You seem to be hard-coding it, but it is optional in the spec. Wouldn't it be beneficial to leave the optionality exposed in spike? To test SW paths to deal with implicit ebreak not being there?

timsifive: Yes. Ideally spike has command line options for every optional feature in the debug spec. We can open an issue for that if you like, but I have other things to work on first.

PR #168, opened 2018-01-08
mem_t: Throw an error if zero-sized memory is requested

neuschaefer:
This improves usability in cases where the user accidentally specifies a
memory size that's big enough to overflow to 0.


aswaterman:
I'm all for making this sort of change, but more robust would be to detect any overflow, rather than this special case.  I think both ```size``` variables in ```make_mems``` should be checked for ```size != (size_t)size```.

neuschaefer:
Good idea. I'll push an additional patch that checks for overflow in `make_mems`

neuschaefer:
(and adjust the commit message of the existing commit)

PR #171, opened 2018-01-30
Add support for debug bus mastering

timsifive:
Tested with all the gdbserver tests, and an OpenOCD that knows how to use this feature.

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -16,9 +16,10 @@
 
 ///////////////////////// debug_module_t
 
-debug_module_t::debug_module_t(sim_t *sim, unsigned progbufsize) :
+debug_module_t::debug_module_t(sim_t *sim, unsigned progbufsize, unsigned max_bus_master_bits) :
   progbufsize(progbufsize),
   program_buffer_bytes(4 + 4*progbufsize),
+  max_bus_master_bits(max_bus_master_bits),
   debug_progbuf_start(debug_data_start - program_buffer_bytes),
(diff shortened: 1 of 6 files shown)


riscv/debug_module.cc
```diff
@@ -78,6 +68,14 @@ void debug_module_t::reset()
   abstractcs.progbufsize = progbufsize;
 
   abstractauto = {0};
+
+  sbcs = {0};
```
mwachs5, line 73: Is system bus access a hard-coded feature? As well as the access types?

timsifive: Yes.

mwachs5: Can we make it optional with a command line arg? S.t. SW can be tested against one that doesn't have system bus?

timsifive: Sure. Do you have suggestions for an option name? I'm having a hard time thinking of something concise but descriptive. Maybe --debug-sba? What do you want the default to be?

Currently in my testing I rely on OpenOCD to prefer the program buffer if it exists.

mwachs5: I think --debug-sba=<N> , where N is the maximum size it supports and the default is 0? (I don't think it makes sense to support some size and not support something smaller than that).

timsifive: Done.

timsifive:
Can I merge this?

palmer-dabbelt (approved):

PR #172, opened 2018-02-13
Implement cycleh/instreth CSRs for RV32

aswaterman:


aswaterman:
@michaeljclark

larkmjc:
LGTM

larkmjc:
BTW I _unimplemented_ rdtime/rdtimeh in riscv-qemu (in system mode) so that it uses the same bbl emulation code.

aswaterman:
cool

PR #174, opened 2018-02-20
Allow an alternative top-level simulator implementation.

pmundkur:
processor_t and mmu_t currently directly refer to sim_t, and sim_t is far too broad an interface for someone wishing to write an alternative top-level simulator based on libriscv.  This patch introduces a simif_t interface that is used instead.  This also allows removal of the friend relationship between sim_t and processor_t.  This leads to an extra indirection in interactive.cc, where it seems acceptable.

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -2,7 +2,6 @@
 
 #include "processor.h"
 #include "mmu.h"
-#include "sim.h"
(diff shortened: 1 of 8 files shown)


palmer-dabbelt (approved):

pmundkur:
ping @aswaterman

PR #176, opened 2018-02-26
Fix installation missing a couple of header files.

pmundkur:
The header files installed by 'make install' were not complete and hence not usable from the installation include directory.  Since debug_rom is not a subproject, I moved one missed header from there to riscv, so that it gets installed as part of the riscv subproject install.


pmundkur:
Any chance for a review here?  It would be nice to fix the underlying issue.

aswaterman:
lgtm

PR #177, opened 2018-02-27
Add debug module authentication.

timsifive:
Off by default, enabled with --debug-auth.
The protocol is very simple (definitely not secure) to allow debuggers
to test their authentication feature. To authenticate a debugger must:
1. Read authdata
2. Write to authdata the value that it just read, plus 1

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -16,10 +16,12 @@
 
 ///////////////////////// debug_module_t
 
-debug_module_t::debug_module_t(sim_t *sim, unsigned progbufsize, unsigned max_bus_master_bits) :
+debug_module_t::debug_module_t(sim_t *sim, unsigned progbufsize, unsigned max_bus_master_bits,
+    bool require_authentication) :
   progbufsize(progbufsize),
   program_buffer_bytes(4 + 4*progbufsize),
   max_bus_master_bits(max_bus_master_bits),
+  require_authentication(require_authentication),
   debug_progbuf_start(debug_data_start - program_buffer_bytes),
   debug_abstract_start(debug_progbuf_start - debug_abstract_size*4),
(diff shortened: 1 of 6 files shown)


riscv/debug_module.cc
```diff
@@ -580,6 +592,8 @@ bool debug_module_t::dmi_write(unsigned address, uint32_t value)
           if (!dmcontrol.dmactive && get_field(value, DMI_DMCONTROL_DMACTIVE))
             reset();
           dmcontrol.dmactive = get_field(value, DMI_DMCONTROL_DMACTIVE);
+          if (!dmstatus.authenticated)
+            return true;
```
mwachs5, line 596: I'm lost in this code... what does this return `true` apply to?

mwachs5: Do I need to write dmactive before DMAUTHDATA is available? or must I be authenticated before writing DMACTIVE?

mwachs5 (commented):
Thanks for doing this. I had some questions I guess mostly about the intended behavior of the spec, which may be better addressed as questions on the spec.

riscv/debug_module.cc
timsifive: `return true` here indicates the DMI write was successful. But since the debugger hasn't authenticated yet, we don't allow any of the other bits in dmcontrol to be written.

In this implementation you need to write `dmactive` first, and authentication is reset when `dmactive` is cleared. The spec doesn't specify, and probably should. I'm not sold on these semantics, but they feel reasonable.

timsifive:
Once this has merged, PR https://github.com/riscv/riscv-tests/tree/debug_auth

timsifive:
Can I get a review on this?

palmer-dabbelt (approved):

PR #178, opened 2018-03-10
Fix single stepping csrrw instructions

timsifive:
This code is still a bit voodoo to me, but now we pass all the tests
again. (Stepping was broken by
4299874ad4b07ef457776513a64e5b2397a6a75e.)


aswaterman:
Thanks, I'll merge this now and think it over later.  Sorry for the headache.

PR #179, opened 2018-03-13
Provide an api get the cpu misa.

pmundkur:
misa is not part of state_t, so it needs a separate public api to read it.


aswaterman:
Is there any downside to moving misa into state_t instead?

pmundkur:
Other than a larger diff, I couldn't see any.

aswaterman:
Mind doing it that way instead?  I think the current factoring is a historical accident, as it predates the misa CSR.

PR #180, opened 2018-03-14
Fix a bug caused by moving misa into state_t.

pmundkur:
state->reset() clears misa, which causes it to lose its value assigned in the processor constructor.  One way to fix this by making state->reset() preserve misa.  A bit hacky; this might be the reason misa was outside state_t in the first place.


aswaterman:
I think it should actually reset misa to max_isa.

PR #181, opened 2018-03-16
Fix for issue #183: No illegal instruction exception for c.sxxi instructions encoded with zero shift amount

deepsrc:
RISC-V spec says that the shift amount must be non-zero for c.slli, c.srli and c.srai instructions in RV32C and RV64C. SPIKE was not reporting illegal instruction exception in case a zero shift amount was encoded. The diff being submitted below will fix that.


aswaterman:
Merged.  Thanks.

aswaterman:
Turns out this was incorrect after all; see #190.

deepsrc:
Thanks for the update. The documentation is somewhat confusing. I got my
test condition wrong as a result.

Regards,
Shubhodeep

PR #182, opened 2018-03-16
Implement debug havereset bits

timsifive:


mwachs5 (approved):
Is there any way for Spike to actually exercise this? Like is there some sort of SW-writable register that would cause a reset?

timsifive:
The debugger can cause a reset by writing ndmactive. I have a test for this that I'll be merging once the OpenOCD code is also merged.

PR #184, opened 2018-03-19
Fix spike-dasm.

timsifive:
It had been broken by 90bafe660b323250338fd564bb9ab4316576d59b.

If somebody wants to flip the travis switch on this repo, I will write the .travis.yml file that does a build and ensures that spike-dasm doesn't segfault on startup.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -155,7 +155,8 @@ void processor_t::reset()
   if (ext)
     ext->reset(); // reset the extension
 
-  sim->proc_reset(id);
+  if (sim)
+    sim->proc_reset(id);
 }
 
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

aswaterman:
Building this repo depends on riscv-fesvr and testing it meaningfully depends on riscv-tests.  How do you propose we enable travis without submoduling those?

I would support writing a 'make check' target that runs some ISA execution, debug, and disasm sanity checks, but this would require the other stuff be installed.  riscv-tools' travis could run it.

timsifive:
Oh, I didn't realize there were so many dependencies. :-(

PR #185, opened 2018-03-21
Fix the access exception during page-table walks to match the original access type, as specified in the manual.

pmundkur:
This fix corresponds to my reading of the spec.  Is it correct?  Interestingly, all the rv64-v- tests pass with with and without this fix.  So they don't seem to be testing this detail.


aswaterman:
Good catch!

Issue #144, opened 2017-10-21
Add ability to model ROM devices


evancox10:
Currently, there is no method to add an user-specified ROM device on the bus. spike has a debug ROM and a boot ROM, but these are not controlled by the user.

This is problematic because it means you can't simulate a device whose application code is stored in ROM. This is very common in embedded application. Sometimes devices will be shipped with a literal ROM that is baked into the mask set. Others may have some other form of embedded non-volatile memory (NVM), such as flash or EEPROM. These memories are "writable" but writing/erasing them usually involves accessing a dedicated peripheral on the memory map; you can't just issue a store to them.

Proposal: Add the ability for user to specify rom devices, just like the -m option for memory. These would be loaded once by spike at boot-time, but could not be changed later during code execution. The data for these devices could come from the primary ELF file, or more options could be added to load data from additional binary files, ELF or otherwise.

Issue #105, opened 2017-05-30
Memory mapped I/O support


prosaad:
I want to simulate a bare metal code on spike, which is using memory mapped I/O. Is there a way to add an I/O memory region to spike and then read/write to it from host or GDB?

PR #186, opened 2018-03-21
Implement infrastructure for MMIO device plugins.

vexingcodes:
This commit introduces the ability to write plugins in the form of shared
object files that allow user-defined Memory-Mapped-I/O behaviors. The overall
behavior of these plugins was modeled after the existing RoCC extensions which
also operate through shared objects and have a similar registration procedure.

All plugins must be a class type derived (directly or indirectly) from
`abstract_device_t` with the additional requirement that their constructors
take a single `const std::string&` containing arguments than can be passed to
the plugin to configure its behavior. Additionally, an `extern "C"` function
must be made that creates a heap-allocated instance of the plugin class. (This
memory is never freed until the spike process exits, much like `mem_t`
objects.) This `extern "C"` function should be passed to the
`REGISTER_PLUGIN_DEVICE` along with the friendly name that will be used on the
command-line to create instances of this device type.

The following code is a very simple example plugin that simply prints messages
when it is created or its memory is accessed.

```
#include <spike/devices.h>

class test_plugin_device_t : public abstract_device_t {
 public:
  test_plugin_device_t(const std::string& args) {
    printf("CREATE -- ARGS=%s\n", args.c_str());
  }

  bool load(reg_t addr, size_t len, uint8_t* bytes) {
    printf("LOAD -- ADDR=0x%lx LEN=%lu\n", addr, len);
    return true;
  }

  bool store(reg_t addr, size_t len, const uint8_t* bytes) {
    printf("STORE -- ADDR=0x%lx LEN=%lu\n", addr, len);
    return true;
  }
};

extern "C" {
  abstract_device_t* create_test_plugin_device(const std::string& args) {
    return new test_plugin_device_t(args);
  }
}

REGISTER_PLUGIN_DEVICE(test_dev, create_test_plugin_device);
```

Once this code is compiled into a shared object, it can be used as an MMIO
device at any valid base address using the proper spike invocation. To invoke
spike with a plugin device, first the `--extlib` flag must be passed with the
path to the shared object file containing the plugin code. Once the shared
object is loaded with `--extlib` the `--device` flag can be used to create an
instance of the plugin device. The argument to `--device` is a comma-separated
string of the format `type,base,args` where `type` is the same as the friendly
name used in the `REGISTER_PLUGIN_DEVICE` macro. The `base` is the starting
memory location where the plugin device should be mapped. This value can be in
decimal, hex (preceeded by 0x or 0X), or octal (preceeded by 0). Finally the
optional `args` are passed along to the plugin's constructor. The `args` may
contain commas, all characters after the second comma in the `--device`
parameter are passed unparsed to the plugin constructor. If `args` is not
provided then an empty string will be passed to the plugin constructor. Any
number of devices can be attached in this way, since `--extlib` and `--device`
may be used multiple times in the same command-line invocation.

As an example, consider the case where the code of the plugin above has been
compiled into `plugin.so` in the current directory, and there is a RISC-V
executable called `executable` also in the current directory. The following
spike invocation will attach a MMIO device at base address 0x10000000:

```
spike --extlib=./plugin.so --device=test_dev,0x10000000,argument executable
```

If the executable is set up to read and then store 4 bytes starting at memory
location 0x10000000 then the following stdout from the plugin would be
observed:

```
CREATE -- ARGS=argument
LOAD -- ADDR=0x0 LEN=4
STORE -- ADDR=0x0 LEN=4
```

From this framework it should be possible to emulate some useful MMIO devices
in spike without having to modify the spike source code directly. One current
limitation of this system is that plugins cannot interact with anything else on
the bus. A future enhancement could enable this, but it would require a decent
amount of refactoring for an unknown amount of benefit.

I have created an example plugin at https://github.com/vexingcodes/spike-plugin.
There's also an automated Vagrant/Ansible setup that compiles all of the
relevant repositories and runs the example plugin at
https://github.com/vexingcodes/spike-dev.

I believe this implementation is an answer to #105, and may help move #144
along.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -202,8 +202,9 @@ void vectorUnit_t::reset(){
   ELEN = get_elen();
   SLEN = get_slen(); // registers are simply concatenated
   reg_file = malloc(NVPR * (VLEN/8));
-  vtype = -1;
-  set_vl(-1, 0, 0); // vsew8, vlmul1
+
+  vtype = 0;
+  set_vl(-1, 0, -1); // default to illegal configuration
 }
 
 reg_t vectorUnit_t::set_vl(uint64_t regId, reg_t reqVL, reg_t newType){
@@ -215,8 +216,12 @@ reg_t vectorUnit_t::set_vl(uint64_t regId, reg_t reqVL, reg_t newType){
     vlmax = VLEN/vsew * vlmul;
     vmlen = vsew / vlmul;
     reg_mask = (NVPR-1) & ~(vlmul-1);
+
+    vill = vsew > e64 || vediv != 1 || (newType >> 7) != 0;
+    if (vill)
+      vlmax = 0;
   }
-  vl = reqVL <= vlmax ? (regId == 0)? vlmax: reqVL : vlmax;
+  vl = reqVL <= vlmax && regId != 0 ? reqVL : vlmax;
   vstart = 0;
   setvl_count++;
   return vl;


aswaterman:
I think this is a great feature addition.  I'm going to leave it to @palmer-dabbelt to give feedback on the command-line interface and APIs.

vexingcodes:
I realized the `extern "C"` part mentioned in the commit message isn't needed. That's a holdover from when I was using `dlsym` to look up symbols in the shared object and I needed non-mangled names. A static function inside the plugin class itself works just as well and looks a little cleaner. I updated the example plugin at https://github.com/vexingcodes/spike-plugin.

amirgon:
@vexingcodes The plugin infrastructure is very useful, thank you!  
There is one thing that I think is missing - a way to generate an interrupt.
Currently I need the riscv code to constantly poll the mmio in order to detect new events that need to be handled.

Any suggestions how to improve the plugin to allow it trigger interrupts?

PR #189, opened 2018-03-26
Add an api to get the name for a CSR.

pmundkur:
Useful to have this upstream.


pmundkur:
@aswaterman @palmer-dabbelt any opinion on adding this api?

palmer-dabbelt (approved):
This looks fine to me.

PR #197, opened 2018-04-27
Extract out device-tree generation and compilation into an exported api.

pmundkur:
Feedback appreciated.


pmundkur:
@aswaterman  @palmer-dabbelt  do you have any issues with exporting this api?  it would make it easier to reuse platform info in simulators that imitate spike or use spike as a backend verifier.

palmer-dabbelt:
I'm fine with it.

PR #213, opened 2018-06-25
Reset an extension when it is registered

heshamelmatary:
Right now processor_t::reset is being called when sim_t is created. That's before spike_main is able to register an extension.

This doesn't reset the extension (ext->reset()) as ext is still null at the time of resetting the processor.

This commit resets the extension when it is registered, and not when processor_t::reset is called

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -152,9 +152,6 @@ void processor_t::reset()
   halt_on_reset = false;
   set_csr(CSR_MSTATUS, state.mstatus);
 
-  if (ext)
-    ext->reset(); // reset the extension
-
   if (sim)
     sim->proc_reset(id);
 }
@@ -744,6 +741,9 @@ void processor_t::register_extension(extension_t* x)
     throw std::logic_error("only one extension may be registered");
   ext = x;
   x->set_processor(this);
+
+  /* Reset the extension */
+  x->reset();
 }
 
 void processor_t::register_base_instructions()


aswaterman:
Why not just call reset() on the extension at the instant it’s registered? That’s a lot less invasive.

heshamelmatary:
That's a good idea and actually, I tried to do so but I wanted to keep the semantic of resetting an extension part of resetting the processor, in case there are some assumptions about that. I'm happy to reset the extension when it's registered, that makes it way easier.

I have a related question about the current implementation of `extensions_t` in Spike. Is it assumed/intended that an `extension_t` is only part of the processor/ISA? If not, is there a reason why not to make an extension part of `sim_t` and register it with other subsystems that it might need access to (e.g. bus_t, mmu_t, cache_t, etc). This would also add a new `set_sim()` function to `extension_t` similar to `set_processor()`

aswaterman:
The thought was that extensions shouldn't need that wide an interface to the simulator.  Since it has a handle to the processor, it also has a handle to the mmu, which is enough to access the memory system and/or master the bus.

heshamelmatary:
I modified to PR to do a reset when registering an extension. 

> The thought was that extensions shouldn't need that wide an interface to the simulator. Since it has a > handle to the processor, it also has a handle to the mmu, which is enough to access the memory > system and/or master the bus.

I couldn't find an easy way for an extension, relying on the processor_t interface, to know about the DRAM's size (and base) without exposing the `bus_t`

aswaterman:
I'm surprised the extension needed to know the size/location of DRAM; I'd expect the location/extent of data structures to be provided at runtime from the instruction stream.

aswaterman:
@colinschmidt can you comment as to whether this proposed change is ok for hwacha etc.?

heshamelmatary:
> I'm surprised the extension needed to know the size/location of DRAM; I'd expect the location/extent of data structures to be provided at runtime from the instruction stream.

The extension has some instructions as well as tagged memory support, which I'm trying to keep it contained/separated from Spike's source code

colinschmidt:
So there are two methods to register and extension, via the ISA string and via the `--extension` parameter.
From my reading of the current code using the ISA string will result in the extension being registered at the creation of the processor which then resets the processor (and extensions).
In the second case with the `--extension` parameter this fix makes sense.

This would be fine for all the accelerators I'm aware/in control of. Thanks for checking.

colinschmidt (approved):

PR #217, opened 2018-07-09
Refactor and fix LR/SC implementation

### LINKED ISSUE #215: clearing load_reservation on xret
The load reservation is set on virtual addresses, but xRET doesn't yield the reservation as is done on traps.  
Isn't this a bug?

aswaterman:
- Use physical addresses to avoid homonym ambiguity (closes #215)

- Disallow LR/SC to non-memory-like regions

- Yield reservation on store-conditional (https://github.com/riscv/riscv-isa-manual/commit/03a5e722fc0fe7b94dd0a49f550ff7b41a63f612)

- Don't yield reservation on exceptions (it's no longer required).

### DIFF
--- a/riscv/insns/lr_d.h
+++ b/riscv/insns/lr_d.h
@@ -1,4 +1,4 @@
 require_extension('A');
 require_rv64;
-p->get_state()->load_reservation = RS1;
+MMU.acquire_load_reservation(RS1);
 WRITE_RD(MMU.load_int64(RS1));
--- a/riscv/insns/lr_w.h
+++ b/riscv/insns/lr_w.h
@@ -1,3 +1,3 @@
 require_extension('A');
-p->get_state()->load_reservation = RS1;
+MMU.acquire_load_reservation(RS1);
 WRITE_RD(MMU.load_int32(RS1));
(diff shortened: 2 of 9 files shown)


aswaterman:
attn @pmundkur

pmundkur:
What about the shared memory case, either between two processes or
between a process and the kernel?  It shouldn't be possible for an LR
in one address-space to result in an SC succeeding in another.  Say in
the case of buggy program A, where A does SC without a preceding LR,
but it succeeds since the LR was done in B, and B
context-switched/trapped/xret-ted to A, and A and B share memory.

Shouldn't privilege transitions also yield the reservation?

aswaterman:
Yeah. The OS software is responsible for this. Quoting the commentary,

“A store-conditional instruction to a scratch word of memory should be used during a preemptive context switch to forcibly yield any existing load reservation.”

pmundkur:
Thanks, I'd forgotten or missed that advisory.  But the patch now
makes sense to me.

Are there any portability issues though on the use of a scratch
address?

aswaterman:
OK, cool.  The spec now mandates that SC yield the reservation (this was part of the memory-consistency model), so there should be no portability issues.

PR #218, opened 2018-07-17
Clean up/optimize Debug ROM.

timsifive:
It's 3 instructions shorter now.


aswaterman (commented):
Given that the entry point 0x808 shows up in the middle of people’s processor implementations, this is sort of an API-breaking change. An internal API, to be sure, but it’s still more of a headache than simply changing the ROM.

I have a preference for eating the extra 4 bytes and leaving the exception entry point as-is. What do you think @mwachs5?

timsifive:
I was under the impression that this debug ROM was only used by spike, but if people are grabbing the code and using it with Rocket verbatim then this is probably not worth changing. (The right solution is probably to bundle debug ROM with the implementation that uses it, so there should be one to be used with Rocket in its repo.)

aswaterman:
People should provide their own debug ROMs, but we at least base ours on this one.

mwachs5:
rocket-chip has its own Debug ROM:

https://github.com/freechipsproject/rocket-chip/blob/master/scripts/debug_rom/debug_rom.S

Not sure if these have already diverged.

timsifive:
They have diverged. In spike we use fence instructions, and the main loop code uses different logic to achieve the same thing in Rocket as in spike.

PR #220, opened 2018-07-23
Fix using the uninitialized disassemble object.

zizztux:
This fixes runtime crash when custom extension registers its
disassembly.

Before this commit, disassembler_t is instantiated after parsing the ISA string.
So, if custom extension adds disasm_insn_t, at that time, disassembler_t object is not exist.
disassembler_t can be instantiated before parsing ISA string.


aswaterman:
Makes sense.

PR #229, opened 2018-08-10
Fix 2 trigger corner cases.

timsifive:
1. When hitting a trigger during a single step, dcsr.cause must reflect
the trigger not the step.
2. Also check for triggers on accesses that require a slow path fetch.


aswaterman:
lgtm

PR #233, opened 2018-08-23
Add dummy custom debug registers, to test OpenOCD.

timsifive:


timsifive:
The motivation is just to have a target that I can test this OpenOCD functionality against.

aswaterman:
Ah, OK, fine with me.

PR #241, opened 2018-09-24
Add "--log-cache-miss" option to generate a log of cache miss.

takeoverjp:
Thank you for your attention.
I'd like to add "--log-cache-miss" option to understand which instruction has caused the cache miss.

- When this option is true, spike outputs a log of cache miss (default: OFF).
- This option must be used with "--ic" and/or "--dc" options to enable cache simulation.
- This option is useful with "-l" option to understand  which instruction has caused the cache miss.

## Sample to use

```
$ spike --ic=64:4:64 --dc=64:4:64 --log-cache-miss -l pk sample
I$ read miss 0x1000 (4 bytes)
core   0: 0x0000000000001000 (0x00000297) auipc   t0, 0x0
core   0: 0x0000000000001004 (0x02028593) addi    a1, t0, 32
I$ read miss 0x80000000 (4 bytes)
core   0: 0x0000000080000000 (0x1e80006f) j       pc + 0x1e8
I$ read miss 0x800001e8 (4 bytes)
core   0: 0x00000000800001e8 (0x00000093) li      ra, 0
core   0: 0x00000000800001ec (0x00000113) li      sp, 0
core   0: 0x00000000800001f0 (0x00000193) li      gp, 0
core   0: 0x00000000800001f4 (0x00000213) li      tp, 0
core   0: 0x00000000800001f8 (0x00000293) li      t0, 0
core   0: 0x00000000800001fc (0x00000313) li      t1, 0
I$ read miss 0x80000200 (4 bytes)
...
```

### DIFF
--- a/riscv/cachesim.cc
+++ b/riscv/cachesim.cc
@@ -7,7 +7,7 @@
 #include <iomanip>
 
 cache_sim_t::cache_sim_t(size_t _sets, size_t _ways, size_t _linesz, const char* _name)
- : sets(_sets), ways(_ways), linesz(_linesz), name(_name)
+: sets(_sets), ways(_ways), linesz(_linesz), name(_name), log(false)
 {
   init();
 }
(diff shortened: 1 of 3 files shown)


aswaterman (commented):
I suggest removing the “ (4 bytes)” from the end of the string. Printing the size of the access that caused the miss isn’t very useful (especially since cache hits are not logged), and it increases the size of the log.

takeoverjp:
@aswaterman 
Thank you for your prompt reply, and I agree your suggestion.
I pushed the log format modification.
Could you review these again?

aswaterman (approved):

PR #242, opened 2018-09-25
[ADD] Public method to add device to the sim_t system bus.

jm4rtin:


riscv/sim.h
```diff
@@ -45,6 +45,17 @@ class sim_t : public htif_t, public simif_t
...
   void proc_reset(unsigned id);
 
+  /**
+   * Add a device onto the system bus.
+   * @param baseAddress Base address of the device.
+   * @param pDevice Pointer to the device to add to the bus.
+   * @note The bus_t will not free this device. You should not free
+   * the device until the sim_t object has been deleted.
+   */
+  void add_bus_device(reg_t baseAddress, abstract_device_t *pDevice) {
```
aswaterman: perhaps just add_device?  (At this level, the bus is an implementation detail).

But more generally I approve of the API addition.

jm4rtin: Changed.

Issue #246, opened 2018-10-18
Provide a noisy interactive until


heshamelmatary:
Right now there is the "until" command that's stopping the PC on a specific command. However, it runs silently i.e. without printing the trace log. 

In some debugging scenarios (only using Spike with no GDB), it would be useful to have a noisy version of "until" that prints out the trace as well. For example, to check if a specific instruction/path has been executed. That would mimic a start/end trace commands using a combination of until and until-noisy commands.

I am aware I can use "run" and dump the output trace to a file. However, files can get very large for non-terminating programs. 

If you think having a noisy until might be useful I'd be happy to submit a PR.

aswaterman:
Seems reasonable.

PR #247, opened 2018-10-18
Provide a noisy until interactive command

heshamelmatary:
This is useful for example when the trace until a PC value needs
to be extracted (#246)


palmer-dabbelt (approved):
This looks useful.  Thanks!

PR #251, opened 2018-10-23
 log memory accesses

ingallsj:


ingallsj:
a log of memory accesses (load/store, address, data) would be helpful, similar to the existing commit log (register number, data)

ingallsj:
updates in repository sifive/riscv-isa-sim branch feature-memory-log

PR #252, opened 2018-10-23
Feature memory hazards

ingallsj:


aswaterman:
This is probably too microarchitecturey for Spike and should be maintained as a separate fork. @palmer-dabbelt thoughts?

ingallsj:
> This is probably too microarchitecturey for Spike and should be maintained as a separate fork. @palmer-dabbelt thoughts?

yes, you're exactly right, and I already closed this because I accidentally merged it as a pull request here instead of in my own fork!

aswaterman:
Word.

Issue #266, opened 2018-12-19
progbufsize check in ebug_module_t::perform_abstract_command()


RolfyYu:
In file debug_module.cc,
function debug_module_t::perform_abstract_command()

line 516
if (regno < 0x1000 && progbufsize < 2) {
.... 
}

When I set progbufsize to 8 (with openOCD),
it could not do CSR abstract command.
I think the check of progbufsize should move to elsewhere.

PR #267, opened 2018-12-31
Add --debug-no-abstract-csr

timsifive:
This is used to make sure that OpenOCD can work on targets that don't
support abstract access to CSR registers. It replaces a simpler hack,
which caused #266.

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -32,13 +32,15 @@ static unsigned field_width(unsigned n)
 ///////////////////////// debug_module_t
 
 debug_module_t::debug_module_t(sim_t *sim, unsigned progbufsize, unsigned max_bus_master_bits,
-    bool require_authentication, unsigned abstract_rti, bool support_hasel) :
+    bool require_authentication, unsigned abstract_rti, bool support_hasel,
+    bool support_abstract_csr_access) :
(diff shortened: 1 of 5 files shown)


timsifive:
Can I get a review on this?

riscv/debug_module.cc
```diff
@@ -585,9 +587,7 @@ bool debug_module_t::perform_abstract_command()
     unsigned i = 0;
     if (get_field(command, AC_ACCESS_REGISTER_TRANSFER)) {
 
-      if (regno < 0x1000 && progbufsize < 2) {
-        // Make the debugger use the program buffer if it's available, so it
-        // can test both use cases.
+      if (regno < 0x1000 && support_abstract_csr_access) {
```
aswaterman, line 590: Does this need to be `(support_abstract_csr_access || progbufsize < 2)`?  Or perhaps earlier, `assert(support_abstract_csr_access || progbufsize >= 2);`

aswaterman (approved):

riscv/debug_module.cc
timsifive: No, this test is correct. It's testing that the value is <0x1000 (the CSRs have values from 0-0xfff), and that abstract CSR access is enabled.

PR #270, opened 2019-01-26
Fix PMP checks for partially-matching accesses

aswaterman:
PMP checks should unconditionally fail if the PMP matches part of, but
not all of, an access.  We got this right, but went too far: we checked
whether _any_ PMP matches in this manner.  In fact, only the first PMP
that maches any of the bytes should be checked in this manner.


aswaterman:
@srivatsa611y can you see if this addresses your problem?

srivatsa611y:
Yes this fixes the problem

PR #271, opened 2019-01-28
Ensure mstatus.FS is hardwired to zero in absence of floating-point.

pmundkur:
mstatus.FS was getting set even when --isa disabled F/D.  The spec says it should be hardwired to zero.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -380,10 +380,13 @@ void processor_t::set_csr(int which, reg_t val)
         mmu->flush_tlb();
 
       reg_t mask = MSTATUS_SIE | MSTATUS_SPIE | MSTATUS_MIE | MSTATUS_MPIE
-                 | MSTATUS_FS | MSTATUS_MPRV | MSTATUS_SUM
+                 | MSTATUS_MPRV | MSTATUS_SUM
                  | MSTATUS_MXR | MSTATUS_TW | MSTATUS_TVM
                  | MSTATUS_TSR | MSTATUS_UXL | MSTATUS_SXL |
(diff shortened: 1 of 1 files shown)


pmundkur:
I see, thanks.  How does one specify a system with S mode but with no native or emulated FP support?  Is this via a riscv,isa key in device-tree that omits 'fd'?

aswaterman:
Spike's device-tree reports what ISA is actually implemented in hardware.  It is the responsibility of the M-mode software to inject the 'fd' into S-mode's copy of the device-tree, if M-mode is going to emulate the FPU on S-mode's behalf.

pmundkur:
Ok.  So is it accurate to say that mstatus.FS is a WARL field?  If so, the spec is silent on this issue, unless I'm missing something.  (I'll close this pull request after I'm clear on FS.)

aswaterman:
Yeah.  Not marking it WARL is an oversight; I'll amend the spec as follows:

    diff --git a/src/machine.tex b/src/machine.tex
    index e4753cc..d17dc06 100644
    --- a/src/machine.tex
    +++ b/src/machine.tex
    @@ -742,7 +742,7 @@ arbitrary user-mode state extensions.
       additional state beyond the floating-point CSR and data registers.
     \end{commentary}
 
    -The FS[1:0] read/write field and the XS[1:0] read-only field are used
    +The FS[1:0] \warl\ field and the XS[1:0] read-only field are used
     to reduce the cost of context save and restore by setting and tracking
     the current state of the floating-point unit and any other user-mode
     extensions respectively.  The FS field encodes the status of the

pmundkur:
Great, thanks!

PR #280, opened 2019-02-26
Implement halt groups

timsifive:
This lets the debugger halt multiple harts near simultaneously.


aswaterman (approved):
I don't have time to properly review this, but nothing jumped out at me from a quick skim.

PR #284, opened 2019-03-06
Allow delegation of all interrupts

fintelia:
The RISC-V privileged spec places no restrictions on which interrupts implementations can allow to be delegated. Accordingly, this pull request allows delegation of all supported interrupt types, including MSIP and MTIP which were previously restricted.

It is unclear whether it is a good idea to delegate these interrupts. In particular, delegating these two machine mode interrupts could be better for performance and perhaps eliminate the need for any m-mode code at all. However, it involves m-mode code giving up substantial control of the runtime execution of the hart to s-mode code and the current SBI interface cannot be implemented with either of these delegated.

Regardless, I'd argue that this ISA simulator is the wrong place to make that decision and that this case should be supported to allow for experimentation.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -333,9 +333,10 @@ int processor_t::paddr_bits()
 void processor_t::set_csr(int which, reg_t val)
 {
   val = zext_xlen(val);
-  reg_t delegable_ints = MIP_SSIP | MIP_STIP | MIP_SEIP
+  reg_t all_ints = MIP_SSIP | MIP_STIP | MIP_SEIP
+                       | MIP_MSIP | MIP_MTIP
                        | ((ext != NULL) << IRQ_COP);
-  reg_t all_ints = delegable_ints | MIP_MSIP | MIP_MTIP;
+  reg_t delegable_ints = all_ints;
 
   if (which >= CSR_PMPADDR0 && which < CSR_PMPADDR0 + state.n_pmp) {
     size_t i = which - CSR_PMPADDR0;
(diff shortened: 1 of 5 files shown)


aswaterman:
The ISA simulator necessarily encodes some details of a particular platform, and the current (informal) RISC-V platforms hardwire these bits. So I’m not excited to make Spike divergent in this respect.

@palmer-dabbelt @kasanovic your thoughts?

fintelia:
@aswaterman In cases where there were contradictory options for platform decisions, I definitely understand the goal to mirror existing hardware. But unless I'm misunderstanding, shouldn't an ISA simulator attempt to support all features of the ISA and hardwire as little as possible? Not hardwiring these bits should have no impact on software that doesn't want to use them, it just makes it possible to write software that does (and in the process inform what features are valuable to include in hardware implementations).

aswaterman:
I appreciate that sentiment, but I think most people would prefer that Spike mirror what the hardware platforms actually do.

aswaterman:
I'm fine with supporting this with a configure-time or command-line option, but I really don't think the default should change.

fintelia:
That seems completely reasonable

palmer-dabbelt:
I like the command-line argument version, that's essentially what we're doing in QEMU.

fintelia:
@palmer-dabbelt do you have a link to the analogous change in QEMU?

palmer-dabbelt:
The QEMU thing wasn't specific to this feature, but was more a general practice.  The code is pretty much here https://github.com/qemu/qemu/blob/master/target/riscv/cpu.c#L122 , the general idea is that we have a feature bitmask that controls the emulation.  The various QEMU targets we support are then just different feature sets.

This makes a bit more sense as a command-line interface in QEMU because QEMU emulates whole systems and therefor naturally should contain a list of them.  Spike doesn't match any real full system, it just matches ISAs, so it makes more sense in Spike to just encode each feature as its own command-line argument.

PR #287, opened 2019-03-27
Implement debug hasel support

timsifive:
This lets a debugger simultaneously resume/halt multiple harts.

As part of this I redid how many instructions are executed in a batch in debug mode (now using wfi for that) to get simultaneous resume/halt within 5000 instructions of perfect.


debug_rom/debug_rom.S
```diff
@@ -13,8 +13,6 @@
 
 entry:
        jal zero, _entry
-resume:
-       jal zero, _resume
```
aswaterman: May I suggest again that we not change the debug entry point addresses, as it is an interface encoded inside the processor RTL?  (I do realize rocket-chip has its own debug ROM, but I am averse to this change.)

We can get the same code-size savings without changing this API by moving the FENCE from `_entry` to `entry:

    entry:
           // This fence is required because the execution may have written something
           // into the Abstract Data or Program Buffer registers.
           fence
           jal zero, _entry

    exception:
           jal zero, _exception

    _entry:

riscv/debug_rom_defines.h
```diff
@@ -18,6 +18,6 @@
 // These needs to match the link.ld         
 #define DEBUG_ROM_WHERETO 0x300
 #define DEBUG_ROM_ENTRY   0x800
-#define DEBUG_ROM_TVEC    0x808
+#define DEBUG_ROM_TVEC    0x804
```
aswaterman: If you accept my suggestion on debug_rom.S, this change would need to be backed out as well

riscv/execute.cc
```diff
@@ -244,6 +237,16 @@ void processor_t::step(size_t n)
...
     }
+    catch (wait_for_interrupt_t &t)
+    {
+      // Return to the outer simulation loop, which gives other devices/harts a
+      // chance to generate interrupts.
+      //
+      // In the debug ROM this prevents us from wasting time looping, but also
+      // allows us to switch to other threads only once per idle loop in case
+      // there is activity.
+      return;
```
aswaterman: This approach looks good, but I think this control path fails to update `state.minstret`.  Instead of doing a `return;` you can do `n = instret;`, just like the case for `catch(trap_t& t)` does.  This will cause the loop to terminate normally.

aswaterman (commented):
I didn't check the DM implementation against the spec, but otherwise it looks good

debug_rom/debug_rom.S
timsifive: Done. I don't understand the workflow that involves using this file for RTL, but for spike this cleanup is not really important.

riscv/execute.cc
timsifive: Good point. I've fixed that.

aswaterman (approved):

aswaterman:
Thanks, feel free to merge once it's ready to go

PR #297, opened 2019-05-02
restore macOS support (BSD ar)

sbeamer:
The current set of flags to ar (`rcs -o`), triggers the following error on when building on macOS: `ar: illegal option combination for -r`

The conflicting flag appears to be `-o`. After reading the man pages for both the GNU and BSD versions of ar, they agree that it should preserve the original times when extracting files, but it isn't clear what effect that flag should have when making an archive. Perhaps the GNU version is more tolerant of spurious flags.

Simply removing `-o` seems to fix the problem, and it builds correctly on both macOS (10.14) and linux (Ubuntu 18.04). If there is a need for the `-o`, I can look for another solution.


aswaterman:
Thanks.  I'm pretty sure the `-o` flag is just a mistake... it looks like the intent was to set the name of the output file, but of course that's not how `ar` works.

PR #299, opened 2019-05-07
Clean up debug module options.

timsifive:
1. Instead of passing each one a few levels deep, create
debug_module_config_t which contains them all.
2. Rename all those command line options so they start with --dm for
debug module.
3. Add --dm-no-halt-groups to disable halt group support.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,7 +1,13 @@
 Version 1.0.1-dev
 -----------------
 - Added `hasel` debug feature.
-- Added `--debug-no-abstract-csr` command-line option.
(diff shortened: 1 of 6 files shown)


aswaterman (approved):

aswaterman:
Good call.

I'm trying to maintain ChangeLog.md now that we are versioning Spike.  For changes to the command-line interface or new ISA features, can you add an entry?   I already added one for `--debug-no-abstract-csr`, which you should just change rather than adding a new one.  It's up to you to decide whether to be less terse than me :)

PR #303, opened 2019-06-11
v-spec 0.7.1 support

chihminchao:
Based on [0.7.1](https://github.com/riscv/riscv-v-spec/commit/0a24d0f61b5cd3f1f9265e8c40ab211daa865ede)

The implementation supports
by arch          
- RV32 and RV64
- Vector register length (VLEN): 32 bits to 4096 bits, power by 2 (doesn’t test more than 4096 bits)
- Vector element length (ELEN): 32 bits to VLEN, power by 2
- Vector strip length(SLEN): 32 bits to 512 bits, power by 2 (doesn’t test more than 512 bits)

by instruction category
- integer/fixed-point/mask/reduction/permutation
  sections: 12/13/15.1 ~ 15.2/16/17
  element size: 8/16/32/64
  support ediv: 1
- load/store
  section: 7
  element size: 8/16/32/64
- floating-point
  sections: 14/15.3 ~ 15.4
  element size: 32


aswaterman:
Cool.  Will review and provide feedback over the next day or so.

riscv/execute.cc
```diff
@@ -141,11 +145,71 @@ void processor_t::step(size_t n)
             state.single_step = state.STEP_STEPPED;
           }
 
+          if (debug && !state.prev_state) { // lazy init
```
aswaterman: I think this code should be deleted, or at least placed in  `#if 0`.

riscv/execute.cc
```diff
@@ -141,11 +145,71 @@ void processor_t::step(size_t n)
...
+            }
+            state.prev_state = saved;
+          }
+
           insn_fetch_t fetch = mmu->load_insn(pc);
           if (debug && !state.serialized)
             disasm(fetch.insn);
           pc = execute_insn(this, pc, fetch);
 
+          if (debug && !state.serialized) {
```
aswaterman: I think this code should be deleted, or at least placed in #if 0.

riscv/insns/csrrc.h
```diff
@@ -1,8 +1,16 @@
 bool write = insn.rs1() != 0;
 int csr = validate_csr(insn.csr(), write);
-reg_t old = p->get_csr(csr);
-if (write) {
-  p->set_csr(csr, old & ~RS1);
+reg_t old;
+if (check_vcsr(csr)) {
```
aswaterman: This file should not be modified, because demultiplexing CSRs shouldn't happen here.  Instead, the `get_csr` and `set_csr` routines should be modified.  This avoids repetition (as currently this logic is duplicated in 6 files).

riscv/insns/vfadd_vf.h
```diff
@@ -0,0 +1,13 @@
...
+VI_VFP_VF_LOOP
+({
+  switch(P.VU.vsew) {
+  case e32:
+    vd = f32_add(rs1, vs2);
+    break;
+  case e16:
+  case e8:
+  default:
+    softfloat_exceptionFlags = 1;
```
aswaterman: I think this should be `require(0);` (i.e., throw an illegal instruction exception), rather than setting the FP exception flags.  Same comment for all of these files.

riscv/insns/vfunary0_vv.h
```diff
@@ -0,0 +1,110 @@
+// VFUNARY0 encoding space
+if ((insn.rs1() & 0b11000) == 0b00000) {
```
aswaterman: I think this should be split up into 5 files, one per instruction.  I updated `opcodes-rvv` accordingly: https://github.com/riscv/riscv-opcodes/blob/a2bf04f1fa574e5d8045bf557550027e55d14d1a/opcodes-rvv#L324

riscv/insns/vfunary0_vv.h
```diff
@@ -0,0 +1,110 @@
+// VFUNARY0 encoding space
```
aswaterman: This should be expanded out into 5 files, one per instruction.  I've made the corresponding change to riscv-opcodes: https://github.com/riscv/riscv-opcodes/commit/0c06e7381689977480fd9ec6a1711f8eb48c6382#diff-b8de9572fc6a9a303c3300f15cf8199c

riscv/insns/vfunary1_vv.h
```diff
@@ -0,0 +1,18 @@
+// VFUNARY1 encoding space
```
aswaterman: This should be expanded out into 2 files, one per instruction.  I've made the corresponding change to riscv-opcodes: https://github.com/riscv/riscv-opcodes/commit/0c06e7381689977480fd9ec6a1711f8eb48c6382#diff-b8de9572fc6a9a303c3300f15cf8199c

riscv/decode.h
```diff
@@ -23,9 +23,12 @@
 
 typedef int64_t sreg_t;
 typedef uint64_t reg_t;
+typedef __int128 int128_t;
```
aswaterman, line 27: Using `uint128_t` breaks compilation on 32-bit host systems.  I think it's OK to require 64-bit when using the vector extension, but we shouldn't break backwards compatibility when not using the vector extension.

One possible workaround is to do something like

    #ifdef __SIZEOF_INT128__
    typedef ...
    #endif

and in the definition of instructions that depend on int128, something like

    #ifdef __SIZEOF_INT128__
    // define the instruction
    #else
    abort_need_int128();
    #endif

riscv/execute.cc
davetw: got it. will remove

riscv/execute.cc
davetw: reason: print the vtype and changed registers out after instruction been executed automatically is convenient for software debugging. 
add #if DEBUG or add an option to dump more debug information. Such more debug info option is default been disabled.

riscv/insns/csrrc.h
davetw: OK. We'll rollback.

riscv/insns/vfadd_vf.h
davetw: Got it.

riscv/insns/vfunary0_vv.h
davetw: OK

davetw: @chihminchao

riscv/insns/vfunary1_vv.h
davetw: @chihminchao

riscv/decode.h
davetw: OK

riscv/insns/vfunary0_vv.h
chihminchao: I will follow change in  riscv-opcodes.

riscv/insns/vfunary1_vv.h
chihminchao: I will follow change in riscv-opcodes.

riscv/decode.h
chihminchao: It is a little complicated issue. We uses double width type to handle widen integer instruction  or carry/borrow bit calculation.  128bit type is used to handle SEW =64 case.  For 32bit host, it means target can only support vector extension with ELAN = 32.

In my opinion,  for x86 host development environment, to drop i386 support is trend. Ubuntu have only provide x86_64 image since 17.10 and redhat also has remove it from RHEL 7.

It is reasonable to assume user's development is 64bit host.

aswaterman:
Okay. We can fix this later if people complain. Can you update the autoconf/configure scripts to check for a 64-bit host, and issue a useful error message if the host is 32-bit?

riscv/decode.h
chihminchao: I will add 64bit check to configure.

myftptoyman:
How to test these vector instructions ? Current assembler can't compile vector instructions

aswaterman:
Binutils port is also being released (sorry, I don’t have a link)

jim-wilson:
You can find rvv 0.7.1 binutils sources in the rvv branch at
    github.com/sifive/riscv-binutils-gdb
These have been publicly visible for a few months, though they haven't been "stable" until now.  Pass -march=rv64gcv or similar to the assembler to enable vector instruction support.

We should create a branch in riscv/riscv-gnu-toolchain to hold this, and some minimal gcc patches to support the vector extension, but I'm swamped with work and haven't gotten around to it yet.  Plus I wasn't sure what exactly was happening with the spike work.

There is an existing rvv branch in riscv/riscv-gnu-toolchain, but it holds rvv 0.6 support.  I don't know if that is useful to anyone anymore, so I don't know if it is safe to overwrite that or not.  I could alternatively create a rvv-0.7 branch for the new work to avoid conflicting with the old work and avoid conflicts with possible future work.

I'd rather not have non-upstream patches on riscv-gnu-toolchain master, and I'd rather not upstream patches that aren't an official standard yet, so I don't have any better suggestion than creating an rvv specific branch.

riscv/execute.cc
chihminchao: I remove this part and will send more complete implementation with memory trace function in future.

riscv/execute.cc
chihminchao: I remove this part and will send more complete implementation with memory trace function in future.

riscv/insns/csrrc.h
chihminchao: done

riscv/insns/vfunary0_vv.h
chihminchao: also fix vfclass.v opcode and has included it in my previous PR.
https://github.com/riscv/riscv-opcodes/pull/29

riscv/decode.h
chihminchao: at last,  I add an AC_TYPE_CHECK for __int128_t but not 64bit host checkg.

chihminchao:
Changes
  1. separate vfunary0/vfunary1
  2. check 128 bit integer support in autoconf
  3. rollback vector control register change
  4. remove some unused define
  5. remove register change feature in interactive mode

aswaterman:
Thank you for addressing all these requests!  I think it is @palmer-dabbelt's turn to review the code.

myftptoyman:
I have builld success the rvv Spike and binutils rvv 0.7.1 and run on gcc inline assembly. It is look very good. I will try some algorithm on this platform. Thank you.

palmer-dabbelt (dismissed):

riscv/decode.h
```diff
@@ -277,8 +299,1439 @@ inline freg_t f128_negate(freg_t a)
...
+  }else if(sew == e16){ \
+    VV_PARAMS(e16); \
+    BODY; \
+  }else if(sew == e32){ \
+    VV_PARAMS(e32); \
+    BODY; \
+  }else if(sew == e64){ \
+    VV_PARAMS(e64); \
+    BODY; \
+  } \
```
palmer-dabbelt, line 654: This should fail on an unknown SEW.  There's a bunch of these.

riscv/interactive.cc
```diff
@@ -218,6 +220,46 @@ freg_t sim_t::get_freg(const std::vector<std::string>& args)
...
+    int num_elem = vlen/elen;
+    fprintf(stderr, "VLEN=%d bits; ELEN=%d bits\n", vlen << 3, elen << 3);
+
+    for (int r = 0; r < NVPR; ++r) {
+        fprintf(stderr, "%-4s: ", vr_name[r]);
+        for (int e = num_elem-1; e >= 0; --e){
+            long unsigned val;
+            switch(elen){
+                case 8:
+                    val = ((long unsigned*)p->VU.reg_file + r*num_elem)[e];
```
palmer-dabbelt: This isn't 32-bit safe.  I know the int128_t stuff isn't supported by GCC on 32-bit targets, but at least that one will trigger a compiler error.

palmer-dabbelt (approved):

riscv/decode.h
davetw: we're checking the unsupported SEW at very begging of each instruction. Please reference  the decode.h:460.

riscv/interactive.cc
chihminchao: We have notice this and discuss the same issue [here](https://github.com/riscv/riscv-isa-sim/pull/303#pullrequestreview-248257156)
I add an __int128_t type check in autocond and report error at building stage.

kito-cheng: Maybe `uint64_t` instead of `long unsigned` could provide better portability and readability?

chihminchao: Agree with kito.  I will refine them with precise integer type.

chihminchao:
Changes:
   1. fix vext signed extension issue
   2. refine register checking
   3. add register overlapping checking for vslide1up and vslideup
   4. follow review's suggestion to refine register dump function.

aswaterman:
I think this version of the patch is mergeable; any objections?

palmer-dabbelt:
LGTM

aswaterman:
Thanks @chihminchao!

PR #311, opened 2019-07-16
Writing non-existent CSRs, access FPRs with mstatus.FS=0

timsifive:
2 changes:
1. When writing a non-existent CSR, don't clobber S0.
2. Allow abstract command access to FPRs even when mstatus.FS=0.


aswaterman (approved):
Presumably on systems without dscratch1 this is accomplished with extra commands to read mstatus into the debugger and later restore it?

timsifive:
I assume that systems that have a magic way to access FPRs for abstract commands can use that magic. If they use this model where everything is just code, the debugger will simply put the appropriate stuff in the program buffer.

Spike is a bit odd in that it uses instructions for everything yet still supports abstract command FPR access. The only reason it does is that it allows me to test the OpenOCD code paths that use that. In e.g. rocket I would expect abstract FPR access commands to not be implemented, and leave it to the debugger to put the correct thing into the program buffer.

aswaterman:
Makes sense.

PR #320, opened 2019-08-18
Implement support for big-endian hosts

zeldin:
This can also serve as a base for simulating big- or bi-endian RISC-V on little-endian hosts.

Tested on POWER9 using integer, floating point and vector operations.

### DIFF
--- a/configure
+++ b/configure
@@ -4130,7 +4130,8 @@ fi
 $as_echo "$ac_cv_c_bigendian" >&6; }
  case $ac_cv_c_bigendian in #(
    yes)
-     as_fn_error $? "Spike requires a little-endian host" "$LINENO" 5;; #(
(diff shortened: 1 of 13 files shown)


aswaterman:
Given that Spike doesn't actually support big-endian targets yet, the ELF loader should not be changed to accept big-endian ELFs.  Failing to issue an error for big-endian ELFs will lead to confusion.

zeldin:
@aswaterman A fair point.  (Although BE ELFs weren't exactly rejected before, they just gave the
loader a high chance of bombing out due to misinterpreting the fields. :smile_cat:)

I changed the assert to reject BE ELFs and removed the BE cases at the bottom.  It's easy enough to put them back when needed.

aswaterman:
Ah, you’re right. Well, thanks for improving the status quo :-)

zeldin:
Rebased and fixed the conflict in `mmu.h`

PR #324, opened 2019-08-24
Extends the commit log feature with memory writes.

dave-estes-syzexion:
This provides a little more information for debugging instruction
traces, allowing you to maintain the state of memory as the trace
is processed.

The following sample trace output illustrates the formatting of
the new memory writes. Each line contains the address and the
instruction bytes. If there is a register commited, then it is shown
with the register number and the value. If there is memory written
then it appears after 'mem' with the address and the value.

    0 0x80000130 (0x03d79863)
    0 0x80000134 (0x800005b7) x11 0x80000000
    0 0x8000010c (0x00a6a023) mem 0x80002000 0x80000000
    0 0x80000138 (0x00b6a72f) x14 0x7ffff800 mem 0x80002000 0xfffff800

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -18,6 +18,9 @@ static void commit_log_stash_privilege(processor_t* p)
 static void commit_log_print_value(int width, uint64_t hi, uint64_t lo)
 {
   switch (width) {
+    case 8:
+      fprintf(stderr, "0x%01" PRIx8, (uint8_t)lo);
+      break;
     case 16:
       fprintf(stderr, "0x%04" PRIx16, (uint16_t)lo);
       break;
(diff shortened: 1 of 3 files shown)


ccelio (commented):
I like the cuteness of squeezing “mem” in to distinguish stores from others, but how do AMOs work? Writing regs and mem look mutually exclusive.

dave-estes-syzexion:
Ping @ccelio. PR #323 has recently landed. Previously I asked that we hold off on landing this PR so that I could merge up changes from that PR #323 once it landed. Well, due to some feedback from @aswaterman on PR #323, I eliminated the need to check the commit logging flag. Therefore, this PR no longer has any conflicts with PR #323 and can land whenever it passes muster.

ccelio:
lgtm. 

I'll let @aswaterman decide to pull the trigger if it passes his muster.

aswaterman:
Fine with me.

PR #326, opened 2019-09-02
bugfix for macOS/BSD

sbeamer:
Recently (https://github.com/riscv/riscv-isa-sim/commit/d184cd4dbfb1c863ae9f1e7c4c263ac626706351), `--export-dynamic` was added to `LDFLAGS`, but it appears unnecessary on linux and build-breaking on macOS/BSD.

This PR was tested (and works) on Ubuntu 18.04.3 and macOS 10.14.6.

### DIFF
--- a/configure
+++ b/configure
@@ -4364,8 +4364,6 @@ CFLAGS="-Wall -Wno-unused -g -O2"
 
 CXXFLAGS="-Wall -Wno-unused -g -O2 -std=c++11"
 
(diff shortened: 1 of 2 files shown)


aswaterman:
I can't say for sure, but I think the --export-dynamic might be necessary for some `dlopen` use cases, which aren't exercised by building and running spike.  I will make a patch that excludes this flag when not supported.

PR #335, opened 2019-09-24
Fixed match trigger MATCH_NAPOT case.

fborisovskii:
Mask calculation was not in consistency with debug spec.
Watch debug spec. 5.2.7 match field overview and
debug spec. B.9 fourth example.
Mask should not cover LSB zero bit.

Also there is a way to make it simplier:
reg_t mask = ~(((~state.tdata2[i]) - 1) ^ ~state.tdata2[i]);

### DIFF
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -388,7 +388,7 @@ public:
           break;
         case MATCH_NAPOT:
           {
-            reg_t mask = ~((1 << cto(state.tdata2[i])) - 1);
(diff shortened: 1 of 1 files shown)


timsifive (approved):
Looks good. Thank you for catching this!

fborisovskii:
Hello. Should we wait somebody else to apply my changes?

aswaterman:
Merged, thanks again.

PR #351, opened 2019-10-22
Stop loading "past the end" of the vector.

nick-knight:
The [`VI_LD`](https://github.com/knightsifive/riscv-isa-sim/blob/fb27391bf65ed867b62ef40b726a21ac839fa37a/riscv/decode.h#L1427) macro should only issue memory operations corresponding to the elements with unmasked indices between `vstart` and `vl`-1 of each vector register in the group. The current code also issues memory operations for elements with indices between `vl` and `vlmax`-1, despite the loaded values being subsequently zeroed. @chihminchao @aswaterman


chihminchao:
Hi knightsifive,
You are right.  It loads each element from 0 to vlmax-1 and discards the result when it is unexpected load.

chihminchao (approved):
it looks good to me.

PR #353, opened 2019-10-24
Fix vnclip/vnclipu rounding

a0u:
The rounding increment should be derived from the shift amount, not `SEW`.

This also removes some redundant code for capping the shift amount to `(2 * sew) - 1`, which is already accomplished by taking the low `log2(2 * SEW)` bits.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -541,30 +541,27 @@ static inline bool is_overlaped(const int astart, const int asize,
 
 
 #define INT_ROUNDING(result, xrm, gb) \
-  if (gb > 0) { \
-    switch(xrm) {\
+  do { \
+    const uint64_t lsb = 1UL << (gb); \
+    const uint64_t lsb_half = lsb >> 1; \
+    switch (xrm) {\
       case VRM::RNU:\
-        result += ((uint64_t)1 << ((gb) - 1));\
+        result += lsb_half; \
         break;\
       case VRM::RNE:\
-        if ((result & ((uint64_t)0x3 << ((gb) - 1))) == 0x1){\
-            result -= ((uint64_t)1 << ((gb) - 1));\
-            }else if ((result & ((uint64_t)0x3 << ((gb) - 1))) == 0x3){\
-            result += ((uint64_t)1 << ((gb) - 1));\
-        }\
+        if ((result & lsb_half) && ((result & (lsb_half - 1)) || (result & lsb))) \
+          result += lsb; \
         break;\
       case VRM::RDN:\
-        result = (result >> ((gb) - 1)) << ((gb) - 1);\
         break;\
       case VRM::ROD:\
-        result |= ((uint64_t)1ul << (gb)); \
+        if (result & (lsb - 1)) \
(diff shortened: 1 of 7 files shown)


aswaterman:
Can you review, @chihminchao?

riscv/decode.h
```diff
@@ -532,30 +532,27 @@ static inline bool is_overlaped(const int astart, const int asize,
...
+    const uint64_t lsb_half = lsb >> 1; \
+    switch (xrm) {\
       case VRM::RNU:\
-        result += ((uint64_t)1 << ((gb) - 1));\
+        result += lsb_half; \
         break;\
       case VRM::RNE:\
-        if ((result & ((uint64_t)0x3 << ((gb) - 1))) == 0x1){\
-            result -= ((uint64_t)1 << ((gb) - 1));\
-            }else if ((result & ((uint64_t)0x3 << ((gb) - 1))) == 0x3){\
```
a0u, line 552: In the case that `result[gb] == 0`, the result should still be rounded upward if `(result[gb-1] != 0) && (result[gb-2:0] != 0)` (i.e., the usual round-to-nearest behavior outside of the tiebreaker).

riscv/decode.h
```diff
@@ -532,30 +532,27 @@ static inline bool is_overlaped(const int astart, const int asize,
...
-            result += ((uint64_t)1 << ((gb) - 1));\
-        }\
+        if ((result & lsb_half) && ((result & (lsb_half - 1)) || (result & lsb))) \
+          result += lsb; \
         break;\
       case VRM::RDN:\
-        result = (result >> ((gb) - 1)) << ((gb) - 1);\
```
a0u, line 557: Since all uses of `INT_ROUNDING()` are immediately followed with a right shift by `gb`, clearing the lower bits is unnecessary.

a0u (commented):
I've also noticed a few issues with the `INT_ROUNDING` macro with respect to the v spec.  Comments are inline.

riscv/decode.h
chihminchao: Although it is the current usage that low bits are truncated after rounding,  I prefer to keep function doing just what rounding should behave and not to expect what caller will do.

chihminchao (commented):
Hi a0u,

I also have refined the same part to follow v0.8-draft change but your patches are more tidier.  I will test them and gfeedback later.

riscv/decode.h
a0u: This would be reasonable except that none of the other rounding modes clear the lower bits, so treating this one differently is a bit strange.

chihminchao:
@a0u 
The rounding part has passed my local test but nclip part needs some refinement to handle corner case (when shift amount is 63 in rv64).  I will send them in new PR with my other change.  Could you give me your "signed-off-by" ?

chihminchao:
@a0u 
I have sent my PR and included your change here 
https://github.com/riscv/riscv-isa-sim/pull/355

a0u:
Closing this since #355 has been merged.

PR #354, opened 2019-10-24
Fix uninitialized variables in processor_t

scottj97:
These two booleans (see diff) were not initialized by the constructor and were causing bizarre issues in my application. (Install one additional python package, and the variable changes!)


aswaterman:
Thanks.

PR #355, opened 2019-11-05
rvv-0.8-2019-11

chihminchao:
Change summary

1. add completed register usage checking
2. follow new integer rounding definition and fix related saturation instructions
3. enhance vsetvl for rs1 == r0 feature
4. trigger illegal instruction when there is no 'V' extension support
5. make vxrm/vxsat readable from FCSR
6. make tail-zero configurable by command-line option rather than fixed in compilation time and change default to no tail-zero.


aswaterman:
I don't understand why we are making tail zeroing a command-line parameter, since the ISA has a clear specification (don't zero).  Making this part of the command-line interface forces users to specify a parameter that isn't actually supposed to be a parameter.  I am OK with leaving the compile-time switch in for now.

riscv/decode.h
```diff
@@ -390,38 +387,84 @@ static inline bool is_overlaped(const int astart, const int asize,
...
-  require(insn.rd() + P.VU.vlmul * 2 <= 32); \
+  require((insn.rd() & (P.VU.vlmul * 2 - 1)) == 0); \
   if (insn.v_vm() == 0) \
     require(insn.rd() != 0);
 
-#define VI_CHECK_VREG_OVERLAP(v1, v2) \
-  require(!is_overlaped(v1, P.VU.vlmul, v2, P.VU.vlmul));
+#define VI_CHECK_MSS(is_vs1) \
+  if (P.VU.vlmul > 1) { \
+    require(!is_overlaped(insn.rd(), 1, insn.rs2(), P.VU.vlmul)); \
```
aswaterman: "overlapped" is misspelled

riscv/processor.cc
```diff
@@ -735,11 +752,28 @@ reg_t processor_t::get_csr(int which)
       if (!supports_extension('F'))
         break;
       return state.frm;
-    case CSR_FCSR:
-      require_fp;
-      if (!supports_extension('F'))
-        break;
-      return (state.fflags << FSR_AEXC_SHIFT) | (state.frm << FSR_RD_SHIFT);
+    case CSR_FCSR: {
```
aswaterman: According to the current proposal (https://github.com/riscv/riscv-v-spec/pull/135/files), access to FCSR, VXRM, and VXSAT depends only on FS, not VS.  So I think we need to rewrite this block of code as:

      require_fp;
      reg_t val = 0;
      bool is_set = false;

      if (supports_extension('F')) {
        val |= (state.fflags << FSR_AEXC_SHIFT) |
               (state.frm << FSR_RD_SHIFT);

        is_set = true;
      }
      if (supports_extension('V')) {
        val |= (VU.vxrm << FSR_VXRM_SHIFT) |
               (VU.vxsat << FSR_VXSAT_SHIFT);
        is_set = true;
      }

      if (is_set)
        return val;

      break;

aswaterman: Similarly, we should change the checks for VXRM and VXSAT to use `require_fp` instead of `require_vector_vs`.

riscv/decode.h
chihminchao: 1. The tail-zero option is implemented in the transition between tail-zero in spec and non-tail-zero in some implementation.  It is ambiguous to tell whether the tail-zero is on or off with just spike executable.  If the spec has clearly define  "dont' zero", I will move it back to configurable only in compilation phase.
2. mis-spelling will be fixed.

aswaterman: Thanks.  In this section of the spec, the behavior is now defined to be "don't zero": https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#54-prestart-active-inactive-body-and-tail-element-definitions

riscv/processor.cc
chihminchao: Agree with the modification but I will discard this for this PR and update the spike after spec change is merged.
To add "VS" should be important since some developers also question the missing of "VS" for v0.7 spec in Qemu mailing list.

riscv/decode.h
nick-knight: Since the behavior is now defined to be "don't zero", I vote that we also remove the compile-time switch `TAIL_ZEROING` and related logic altogether. In my opinion, this will improve the readability of `decode.h`. On the other hand, if our goal is to support earlier versions of the spec, then we should leave it in.

README.md
```diff
@@ -18,7 +18,7 @@ Spike supports the following RISC-V ISA features:
   - D extension, v2.2
   - Q extension, v2.2
   - C extension, v2.0
-  - V extension, v0.7.1 (_requires a 64-bit host_)
+  - V extension, v0.8-draft-20191004 (_requires a 64-bit host_)
```
nick-knight: I think we should also clarify that we support the `Zvlsseg` but not the `Zvamo` and `Zvediv` sub-extensions.

chihminchao: agree and will update this part.

riscv/decode.h
chihminchao: I have consulted some software users and they could agree with the removal.  I will remove it.

Marc43: Hi all, sorry for commenting in an old issue, have you preserved the `TAIL_ZEROING` switch? I think you don't, but in any case, I am using specification `0.7.1`. Do you know in which **commit** you had zeroing to re-insert it in my fork?

chihminchao: You may try to revert b9d9e1eb to restore the tail-zeroing feature in this PR.

PR #356, opened 2019-11-13
Add --priv command-line option to set which privilege modes are available

aswaterman:
Valid options are `--priv=msu` (the default), `--priv=mu`, and `--priv=m`.


aswaterman:
@chihminchao @khannaudit

khannaudit:
Are we now hardwiring `mstatus.fs` to 0 for systems without S mode and without F/D extension?

aswaterman:
Forgot about that one.  Fixed.

chihminchao:
It looks good to me.

aswaterman:
Thanks for the review.  I'll merge now, with the expectation that @khannaudit might find a bug or two along the way.  (There shouldn't be any regressions when the `--priv` flag isn't used.)

PR #366, opened 2019-12-04
Rvv 0.8 draft 20191118

chihminchao:
follow the change in v-spec-0.8-draft-20191118
1. narrow instructions  use 'w' suffix
2. add whole register load/store and quad mac
3. vsetvl[i] is changed to use r0 to specify vlmax as vl
4. vlx/vsx is changed to refer index as unsigned value
5. add mstatus.vs to control vector unit state.


chihminchao:
@aswaterman just a mention

riscv/processor.cc
```diff
@@ -614,7 +622,7 @@ void processor_t::set_csr(int which, reg_t val)
       break;
     case CSR_SSTATUS: {
       reg_t mask = SSTATUS_SIE | SSTATUS_SPIE | SSTATUS_SPP | SSTATUS_FS
-                 | SSTATUS_XS | SSTATUS_SUM | SSTATUS_MXR;
+                 | SSTATUS_XS | SSTATUS_SUM | SSTATUS_MXR | SSTATUS_VS;
```
aswaterman: I think this should be `(supports_extension('V') ? SSTATUS_VS : 0` -- in other words, the field doesn't exist without V extension.  This is different than FS, but I think that's what we need to do since this is a new extension.

riscv/processor.cc
```diff
@@ -930,6 +941,11 @@ reg_t processor_t::get_csr(int which)
       if (!supports_extension('V'))
         break;
       return VU.vtype;
+    case CSR_VLENB:
+      require_vector_vs;
+      if (!supports_extension('V'))
+        break;
```
aswaterman: If you agree with my comment about SSTATUS_VS, then we can remove this if-statement.

aswaterman (approved):

riscv/processor.cc
chihminchao: It is reasonable and has been changed in new PR

riscv/processor.cc
chihminchao: agree and the 'if' part has been remobed in new PR.

chihminchao:
The update includes the fix of fault-first load.

aswaterman:
Did you incorporate 100% of my PR? There were other bug fixes, too.

aswaterman:
I merged my PR, so please rebase. Sorry for the headache.

chihminchao:
rebase for health :>

PR #367, opened 2019-12-06
Require VM enabled for SFENCE.VMA and S-mode for mstatus.SUM

khannaudit:
Make SFENCE.VMA raise an exception if Virtual Memory is not implemented (i.e. no supervisor mode or `satp.mode == 0`).

Hardwire mstatus.SUM to 0 if no supervisor mode.

Tested by observing illegal instruction exception being raised on `sfence.vma` with `--priv=mu` and successful TLB flush with `--priv=msu`

### DIFF
--- a/riscv/insns/sfence_vma.h
+++ b/riscv/insns/sfence_vma.h
@@ -1,2 +1,3 @@
+require_extension('S');
 require_privilege(get_field(STATE.mstatus, MSTATUS_TVM) ? PRV_M : PRV_S);
 MMU.flush_tlb();
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -541,7 +541,8 @@ void processor_t::set_csr(int which, reg_t val)
                   || supports_extension('V');
 
       reg_t mask = MSTATUS_SIE | MSTATUS_SPIE | MSTATUS_MIE | MSTATUS_MPIE
-                 | MSTATUS_MPRV | MSTATUS_SUM
+                 | MSTATUS_MPRV
+                 | (supports_extension('S') ? MSTATUS_SUM : 0)
                  | MSTATUS_MXR | MSTATUS_TW | MSTATUS_TVM
                  | MSTATUS_TSR | MSTATUS_UXL | MSTATUS_SXL |
                  (has_fs ? MSTATUS_FS : 0) |


gfavor:
On Thu, Dec 5, 2019 at 8:56 PM Udit Khanna <notifications@github.com> wrote:

> Make SFENCE.VMA raise an exception if Virtual Memory is not implemented
> (i.e. no supervisor mode or satp.mode == 0).
>
This first proposed change seems a bit problematic or at least undesirable.

For example, if M-mode changes mstatus.SBE (which alters the interpretation
of page tables) or changes PMP settings (that could affect either the
physical memory that holds the page tables or the physical memory to which
the page tables point), then it needs to use an sfence.vma (unless it knows
that it is safe to not bother).  But this change would require that M-mode
software must check that vsatp.mode != 0 before doing the sfence.vma -
versus being allowed to do it always when changing SBE or PMP settings.  In
essence, the common cases where software needs to perform an sfence.vma
become complicated by having to make sure that translation actually is
active at that moment (versus just in Bare mode).

Secondly, software would never be able to flush translation caches while
vsatp.mode == 0.  While not a common usage scenario, this could prove to be
desirable in certain situations (e.g. debug, initialization, power
management).

Now maybe one could argue that the proposed change should be modified to
only apply outside of M-mode (thus addressing the preceding particular
examples), but it seems a little dangerous to presume that S/HS-mode
software will never ever want to do an sfence.vma while translation is
inactive.  In essence, sfence.vma is a useful tool to flush translation
caches in special cases as well as in the common usage scenario.  And it is
convenient to sometimes be able to use it without having to be certain that
it is needed.

Greg

khannaudit:
@gfavor so you're not opposed to requiring the presence of S mode for `SFENCE.VMA`, only opposed to requiring address translation being active, right?

gfavor:
I would have maybe said Yes, but then I wonder "why should all M-mode code
using sfence.vma's have to be aware of what modes are supported under it
(and currently enabled by the corresponding misa bits)?".

aswaterman:
We shouldn't check that VM is enabled; we should only check that VM is supported.

Since Spike currently only supports S-mode when VM is enabled (and vice-versa), it is correct to trap SFENCE.VMA when S-mode is not supported.  In the future, we might also support the possibility of S-mode without VM, in which case SFENCE.VMA might be illegal even when S-mode is supported.  But this is a different scenario.

khannaudit:
Fixed to only require S-extension for `sfence.vma`. Let me know if there's a better way to implement the suggested change.

PR #370, opened 2019-12-12
extend the memory writes log feature with memory reads

ingallsj:
Follow-up to #324:
This provides a little more information for debugging instruction traces, allowing you to check the memory addresses accessed as the trace is processed.

The following sample trace output illustrates the formatting of the new memory reads.  Memory reads appear very similarly to memory writes, but only print the address, not the value, because the value is already shown in the register write data value.

```
# jump
0 0xPC (0xINSTR)
# arithmetic
0 0xPC (0xINSTR) xRD 0xVAL
# store
0 0xPC (0xINSTR) mem 0xADDR 0xVAL
# load
0 0xPC (0xINSTR) xRD 0xVAL mem 0xADDR
# amo
0 0xPC (0xINSTR) xRD 0xVAL mem 0xADDR mem 0xADDR 0xVAL
```

@dave-estes-syzexion I moved the `WRITE_MEM` call upwards in the `store_##type` function to only print twice for misaligned stores (once for each naturally aligned piece) instead of three times (once for each naturally aligned piece and once for the whole).
@ccelio can you please review this similarly to #324?

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -42,7 +42,8 @@ static void commit_log_print_insn(state_t* state, reg_t pc, insn_t insn)
 {
 #ifdef RISCV_ENABLE_COMMITLOG
   auto& reg = state->log_reg_write;
-  auto& mem = state->log_mem_write;
+  auto& load = state->log_mem_read;
+  auto& store = state->log_mem_write;
   int priv = state->last_inst_priv;
   int xlen = state->last_inst_xlen;
   int flen = state->last_inst_flen;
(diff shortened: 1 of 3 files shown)


aswaterman:
@chihminchao can you CR this and also consider any implications for adding such features for the V extension?

chihminchao:
It looks good to me.  My thought about vector part is to allow multiple register and memory fields in the same line. 
ex:  
\# vfadd.vv   lmul 2
0 0xPC (0xINSTR) xRD0 0xVAL  xRD1  0xVAL
\# vlw.v  vd, (rs1)    lmul1
 0 0xPC (0xINSTR) xRD0 0xVAL  mem  0xADDR
\# vlsw.v  vd, (rs1), rs2    lmul2
0 0xPC (0xINSTR) xRD0 0xVAL  xRD1  0xVAL  mem  0xADDR mem 0xADDR ...  mem 0xADDR

The log structure should be extended to use std::vector or something like.

PR #371, opened 2019-12-16
Vector load/store fixes

aswaterman:
- Terminate first-fault loads on exceptions, not zero data values
- Simplify vleff implementation
- Set vstart correctly for loads & stores

@chihminchao

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -1400,6 +1400,7 @@ for (reg_t i = 0; i < vlmax; ++i) { \
   for (reg_t i = 0; i < vlmax && vl != 0; ++i) { \
     VI_STRIP(i) \
     VI_ELEMENT_SKIP(i); \
+    P.VU.vstart = i; \
     for (reg_t fn = 0; fn < nf; ++fn) { \
       st_width##_t val = 0; \
       switch (P.VU.vsew) { \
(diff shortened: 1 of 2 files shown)


chihminchao:
Take the strlen example from spec. The example  shouldn't work except that the memory exception occurs exactly at the end of the string. 

    strlen example using unit-stride fault-only-first instruction

    # size_t strlen(const char *str)
    # a0 holds *str

    strlen:
        mv a3, a0             # Save start
        li t0, -1             # Infinite AVL
    loop:
        vsetvli a1, t0, e8  # Vector of bytes of maximum length
        vlbff.v v1, (a3)      # Load bytes
        csrr a1, vl           # Get bytes read
        vmseq.vi v0, v1, 0    # Set v0[i] where v1[i] = 0
        vfirst.m a2, v0       # Find first set bit
        add a3, a3, a1        # Bump pointer    
        bltz a2, loop         # Not found?

        add a0, a0, a1        # Sum start + bump
        add a3, a3, a2        # Add index
        sub a0, a3, a0        # Subtract start address+bump

        ret

aswaterman:
It looks right to me. Can you provide a specific counterexample and explain why it fails?

chihminchao:
# input :
    0x80_0000 :  0x41  0x42 0x43 0x00  0x45 0x46 0x47 0x48         // string "abc\0efghijklmnop"
    0x80_0010 :  0x49  0x4a 0x4b 0x4c  0x4d 0x4e 0x4f 0x50
    # assume *str is 0x80_0000, vl = vlmax is 16
To use the previous strlen, there is only iteration for the loop.  The vl should be not be changed to 3 since no exception happens.

aswaterman:
But strlen still returns the correct answer because we use vmfirst/vmseq to search for the zero byte. After we fall out of the loop, we will have a3=0x800010. Then, the next three instructions set a0=0x800010, a3=0x800013, then a0=3.

chihminchao:
I misunderstand the purpose of fault-first load.  I have include this part to this PR https://github.com/riscv/riscv-isa-sim/pull/366.

aswaterman:
Ok, cool.

PR #375, opened 2020-01-03
Make minimum RTI behavior more realistic.

timsifive:
Now DMI will return busy when you'd expect it to, instead of a few scans
later. This only matters when testing OpenOCD. There is no other reason
to use --dmi-rti.


aswaterman (approved):

timsifive:
Thank you!

PR #378, opened 2020-01-09
Rvv 0.8 float64

chihminchao:
1. add float64 support
2. add missing vmvNFr.v instructions for compiler spilling use


aswaterman:
For widening and narrowing ops, don't you need to check for `vsew == e32 && require_extension('D')`?

chihminchao:
> For widening and narrowing ops, don't you need to check for `vsew == e32 && require_extension('D')`?
yes, I miss it and add them to new PR.

PR #383, opened 2020-01-18
Rvv commitlog

chihminchao:
Add commit log to all rvv instruction.  Since vector instructions may have multiple register and memory accesses, the patch sets use containter to keep access record. The format of register and memory record are still the same but may appeal many times in line for multiple access

ex:  vlb.v  instruction log is
v27 0xbe311649e962244962fe457d119bc150b219104d234fc9708d9cc6b2752288ae404f273b8dfd69d2b5a4a75217485349b59de08f775aabbc43b700127b14c17f mem 0x00000000800072db mem 0x00000000800072dc mem 0x00000000800072e1 mem 0x00000000800072e5 mem 0x00000000800072e9

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -538,7 +538,7 @@ static inline bool is_overlapped(const int astart, const int asize,
   for (reg_t i=P.VU.vstart; i<vl; ++i){ \
     VI_LOOP_ELEMENT_SKIP(); \
     uint64_t mmask = (UINT64_MAX << (64 - mlen)) >> (64 - mlen - mpos); \
-    uint64_t &vdi = P.VU.elt<uint64_t>(insn.rd(), midx); \
+    uint64_t &vdi = P.VU.elt<uint64_t>(insn.rd(), midx, true); \
     uint64_t res = 0;
 
 #define VI_LOOP_CMP_END \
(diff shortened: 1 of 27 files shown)


ss2783:
Can we dump out the LMUL and ELEN values along with the register-specifier and the raw bits? That way post-processing of the log has enough information. 

Likewise, adding the element-size for each memory operation could be useful as well. 

For example, can we change the above log to 

`lmul=1 elen=8 v27 0xbe311649e962244962fe457d119bc150b219104d234fc9708d9cc6b2752288ae404f273b8dfd69d2b5a4a75217485349b59de08f775aabbc43b700127b14c17f mem 0x00000000800072db mem 0x00000000800072dc mem 0x00000000800072e1 mem 0x00000000800072e5 mem 0x00000000800072e9`

chihminchao:
If the lmul and elen are used to check number of destination register and accessing element size, the information are doesn't applicable to all instructions.

lmul doesn't mean there are the same number of written registers as lmul.  Some instructions do mask operation and write the value into single destination.  Some instructions, turning on mask feature, may write registers less than lmul.
elen also doesn't mean the access element is is elen.  Widening instructions operate on double size element and vle-like instructions may load data from memory whose size is smaller than elen.

girishpai:
@chihminchao I think at a high level it might not do harm to indicate these values as part of the printed log for each vector register commit log. The post processing code is free to use or discard these and use only the vector register written data. 
Even if due to masking the actual registers written will not be the same as determined by lmul, at least it will tell the end user of how to interpret the printed value of each register.

For example for the above case
When lmul=1 sew=8 (assuming VLEN=512) the user knows that there are 64 elements of 8 bits each within the 512 bit printed value. 

Just to avoid confusion please use sew instead of elen. ELEN is the fixed static parameter for the platform.

ss2783:
Ah yes. I mean't sew and not elen as @girishpai mentions. Regarding the memory operations, adding the size for the transaction along with the memory dump could be useful for any post processing.

chihminchao:
@ss2783
I agree that lmul and sew information is helpful but I don't use the lmul=#, sew=# or elen=#.  I use the vsetvl/vsetvli assembly format, e1/e2/e4/e8 for element size and m# for lmul number.

ss2783:
@chihminchao I see. I think this looks good for now. Long term it would be nice we can figure out a way to dump that information as some downstream parsers can be very simple and might not require it.

ss2783 (commented):
lgtm

ss2783 (approved):

PR #390, opened 2020-01-31
Support loading multiple ELF files via a new payload HTIF option

jrtc27:
Firmware implementations, such as OpenSBI's fw_jump, make use of this feature on other targets to avoid having to be rebuilt every time the payload is updated.

I intend to shortly submit a new spike platform to OpenSBI, which has been used to test that this feature works as intended. The ability to use fw_jump will make development significantly easier.

I have also bundled in +h/+help support as querying the HTIF help from the command line currently seems impossible otherwise? Anything with a `-` prefix is parsed by spike itself and either uses its own option handling (if recognised) or gives an error, so the plusargs are the only way to pass host arguments through to HTIF.

### DIFF
--- a/fesvr/htif.h
+++ b/fesvr/htif.h
@@ -7,6 +7,7 @@
 #include "syscall.h"
 #include "device.h"
 #include <string.h>
+#include <map>
 #include <vector>
 
 class htif_t : public chunked_memif_t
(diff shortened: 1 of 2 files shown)


jrtc27:
OpenSBI PR mentioned is now at https://github.com/riscv/opensbi/pull/156.

aswaterman:
Looks good to me.

fesvr/htif.h
```diff
@@ -93,6 +97,8 @@ class htif_t : public chunked_memif_t
        +signature=FILE\n\
       --chroot=PATH        Use PATH as location of syscall-servicing binaries\n\
        +chroot=PATH\n\
+      --payload=PATH       Load PATH memory as an additional ELF payload\n\
```
jrtc27, line 100: Oops, I realised I missed out a word (*into* memory...). Do you want a trivial PR for that or will you fix it yourself?

PR #391, opened 2020-02-07
Fix some uninitialized variables in state_t caught by Valgrind

scottj97:
Please review carefully as I'm not sure my fix here is really the best way to do it.

Valgrind reported usage of uninitialized variables on `single_step` and `mcontrol` in `execute.cc`.

I'm not sure why I didn't see these previously when I used Valgrind and got clean results.


aswaterman:
Looks fine to me.  @timsifive?

PR #393, opened 2020-02-11
FESVR: ensure dmactive is 1 before reading debug module registers

mwachs5:
Similar to #392 , we need to set dmactive=1 before attempting to read any other debug module registers.

Otherwise, we fail an `assert` that we always get error response of 0 on reads and writes.


timsifive (approved):
Looks right.

PR #394, opened 2020-02-12
Improve --varch error checking.

timsifive:
Print out why an option has problems.
Add check that elen must be >= xlen, flen, per the spec.
Since RV32G includes D by default, bump default elen to 64.


aswaterman:
lgtm

PR #395, opened 2020-02-14
Make spike capable of booting Linux

avpatel:
Latest Linux does not boot Spike mainly because:
1. Spike does not set bootargs in DTS
2. Spike does not provide mechanism to load initrd for Linux

This patch addresses both above issues and we can now
get latest Linux to prompt on Spike.

### DIFF
--- a/riscv/dts.cc
+++ b/riscv/dts.cc
@@ -9,6 +9,7 @@
 #include <sys/types.h>
 
 std::string make_dts(size_t insns_per_rtc_tick, size_t cpu_hz,
+                     reg_t initrd_start, reg_t initrd_end,
                      std::vector<processor_t*> procs,
                      std::vector<std::pair<reg_t, mem_t*>> mems)
 {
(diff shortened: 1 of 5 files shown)


aswaterman:
thanks!

avpatel:
Andrew,

I am seeing some issues with rate of timer interrupts. I will debug and send separate PR. 

Regards,
Anup

NazerkeT:
Hi there and thanks for enabling running Linux on spike! 

I have one concern tho, is there any limit on initrd size according to specifications? I notice that whenever I set more than 2GB of memory via -m instruction, linux does not boot with kernel panic. I assume problem might be coming from here, because of more than 2GB or 32bit addressing not being recognized? https://github.com/riscv-software-src/riscv-isa-sim/blob/master/spike_main/spike.cc#L495

Do you have some recommendations on how to solve this problem by chance? 

Will be pinging @avpatel and @aswaterman just in case :)

Thanks,
Best wishes

aswaterman:
At first glance, I spot no issues of that nature.  But if you debug it and find a fix, we'd be happy to accept a PR.

NazerkeT:
> At first glance, I spot no issues of that nature. But if you debug it and find a fix, we'd be happy to accept a PR.

Hi @aswaterman, thanks for your response! One more question about the debugging part tho. I was trying to change spike's internal's to debug the problem better, however any of my changes for default memory layout or just regular frpintf() are not outputting what I am expecting, particularly there are no changes at output whatever changes I make for spike.cc and after make clean and make again. It is kind of a naive question, but do you have a couple of pointers for how can I debug spike's internal better?

Thanks,
Best wishes,

aswaterman:
`printf`-style debugging should work fine--I do a lot of that, too--so if you aren't seeing your `printf`s, then you're probably on a code path that isn't being invoked.

PR #396, opened 2020-02-14
Rvv fix 2020 02 14

chihminchao:
fix some vector instruction corner cases

### DIFF
--- a/riscv/insns/vmsgtu_vi.h
+++ b/riscv/insns/vmsgtu_vi.h
@@ -1,5 +1,5 @@
 // vsgtu.vi  vd, vd2, simm5
 VI_VI_ULOOP_CMP
 ({
-  res = vs2 > simm5;
+  res = vs2 > (insn.v_simm5() & (UINT64_MAX >> (64 - P.VU.vsew)));
 })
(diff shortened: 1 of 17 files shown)


riscv/insns/vmsgtu_vi.h
```diff
@@ -1,5 +1,6 @@
 // vsgtu.vi  vd, vd2, simm5
 VI_VI_ULOOP_CMP
 ({
-  res = vs2 > simm5;
+  uint64_t imm = (int8_t)(simm5 << 3) >> 3;
```
aswaterman: The `v_simm5()` function already does this sign-extension to uint64_t; why do you need to do it again?

riscv/insns/vmsgtu_vi.h
```diff
@@ -1,5 +1,6 @@
 // vsgtu.vi  vd, vd2, simm5
 VI_VI_ULOOP_CMP
 ({
-  res = vs2 > simm5;
+  uint64_t imm = (int8_t)(zimm5 << 3) >> 3;
```
aswaterman: Just call v_simm5() here instead?

aswaterman (approved):
See my one comment, otherwise looks good, thanks!

riscv/insns/vmsgtu_vi.h
chihminchao: good suggestions.

aswaterman:
Thank you, @chihminchao!

PR #397, opened 2020-02-15
Add optional support for real-time clint

avpatel:
This patch adds optional support clint timer incrementing at
real-time rate. This can be enabled by passing command line
parameter "--real-time-clint".

This feature can be used for:
1. Checking whether any code addition to Spike is slowing down
simulation too much
2. Comparing run-time for software on Spike with other functional
simulators (such as QEMU)

### DIFF
--- a/riscv/clint.cc
+++ b/riscv/clint.cc
@@ -1,9 +1,16 @@
+#include <sys/time.h>
 #include "devices.h"
 #include "processor.h"
 
-clint_t::clint_t(std::vector<processor_t*>& procs)
-  : procs(procs), mtime(0), mtimecmp(procs.size())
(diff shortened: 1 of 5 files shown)


aswaterman:
Thanks, @avpatel.  I can see why this would be useful.

My only objection is that it should be possible to model CLINT frequencies less than 1 MHz (e.g. to model a 32.768 KHz crystal, by setting INSNS_PER_RTC_TICK to 30518).  So I've updated your patch to make the CLINT API use Hz instead of MHz.  Confirm this is OK? a8c83b8190f8ac72b04794c03e68cee873a28f3c

avpatel:
Your change regarding clint frequency looks good to me.

Thanks,
Anup

PR #398, opened 2020-02-17
Add interactive command to show/clear stats

avpatel:
We add an interactive "stats" command to show/clear
stats from different parts of each CPU/HART. As an example,
the patch also implements TLB related stats.

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -73,6 +73,7 @@ void sim_t::interactive()
   funcs["fregd"] = &sim_t::interactive_fregd;
   funcs["pc"] = &sim_t::interactive_pc;
   funcs["mem"] = &sim_t::interactive_mem;
+  funcs["stats"] = &sim_t::interactive_stats;
   funcs["str"] = &sim_t::interactive_str;
   funcs["until"] = &sim_t::interactive_until_silent;
   funcs["untiln"] = &sim_t::interactive_until_noisy;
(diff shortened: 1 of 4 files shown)


avpatel:
This feature allow to us judge whether any software improvements help performance by looking at CPU/HART stats (such as TLB hit/miss etc).

aswaterman:
I understand the goal, but I see a few problems with the approach:

- Spike's TLB is unrealistic for a hardware L1 TLB; it's very large and direct mapped, because that's faster for a software implementation.  So these results could be deceptive.
- Along the same lines, Spike flushes the TLB way more often and more coarsely than an optimized hardware implementation.  It doesn't implement ASIDs or per-line flushes, and it even flushes the TLB on privilege transfers.  Again, this is to keep the hit time low in the context of a software simulator, but could produce misleading results.
- We already have hooks for cache simulations: try, for example, passing the flag `--dc=128:4:64`.  This allows modeling caches that are different sizes than the Spike host caches.
- This approach adds code to the critical execution path for loads and stores.  This can be avoided by registering a `memtracer_t` instead of adding new static code.  (This is how the cache simulator works.)

Maybe it makes sense to build a more realistic TLB model as part of the existing cache simulation framework?  The cache simulator stuff is already most of the way there (just set the line size to 4096 bytes); we'd just need to add the hooks to model SFENCE.VMA.

If that sounds like too much work, and this simple instrumentation is useful for you in the mean time, I recommend maintaining it on a branch for now.

avpatel:
Hi Andrew,

I am already working on a better and more generic TLB implementation for spike where we can change TLB set/way using command-line parameter. In order to do that l needed TLB stats hence this patch.

Are you fine if I send this patch as part of larger TLB series?

Regards,
Anup

avpatel:
Hi Andrew,

I will also implement ASID based tagging in my new TLB implementation.

Regards,
Anup

aswaterman:
I want to clearly separate the Spike host TLB (the one in mmu.h) and any target TLB models.  I don't want to accept a patch that adds ASID support to the host TLB, because it adds complexity, but doesn't make the simulator capable of running more software, and will actually slow down the simulator.

But if we are talking about modeling a TLB in the context of the existing cache-simulation framework, then I am OK with it.  (Of course, if you go that route, the current patch is not necessary, because the cache models do their own stat tracking and printing.)

PR #400, opened 2020-02-19
Vector stores don't care if rd overlaps v0

aswaterman:
Since vector stores read rd, rather than write rd, there is no overlap constraint.

@chihminchao


chihminchao:
It is reasonable that store's destination is not register.

PR #404, opened 2020-02-23
Added #defines for extension N (user mode interrupts).

michael-roe:
This just adds #defines for the CSRs and status bits for extension N.

I also have an implementation of extension N, which isn't included here. In this commit, I'm upstreaming just the changes to encoding.h, as I think this is an obvious, non-controversial, fix.

### DIFF
--- a/riscv/encoding.h
+++ b/riscv/encoding.h
@@ -42,6 +42,9 @@
 #define SSTATUS_UXL         0x0000000300000000
 #define SSTATUS64_SD        0x8000000000000000
 
+#define USTATUS_UIE         0x00000001
+#define USTATUS_UPIE        0x00000010
+
 #define DCSR_XDEBUGVER      (3U<<30)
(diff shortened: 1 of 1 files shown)


aswaterman:
encoding.h is an auto-generated file. Can you add them here?
https://github.com/riscv/riscv-opcodes/blob/master/parse_opcodes

I don’t think we are ready to accept a patch that adds support for the N
extension. We are contemplating deprecating that approach in favor of using
S-mode without virtual memory as the means to provide secure and insecure
interrupts. This is a lot less invasive than the N extension, and it seems
to cover the most interesting embedded use cases. (It doesn’t handle the
case of user-mode interrupts in OSes like Linux, but that’s a more
controversial topic anyway.)

michael-roe:
Thanks - I hadn't realized that encoding.h was auto-generated.  I have now submitted a pull request (with the equivalent fix) against riscv-opcodes.

PR #405, opened 2020-02-27
Check presence of [S|U] extension for mstatus.[sxl|uxl] read/write

khannaudit:
Hardwire `mstatus.[sxl|uxl]` to zero if `[S|U]` mode unspported

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -608,11 +608,10 @@ void processor_t::set_csr(int which, reg_t val)
       reg_t mask = MSTATUS_SIE | MSTATUS_SPIE | MSTATUS_MIE | MSTATUS_MPIE
                  | MSTATUS_MPRV
                  | (supports_extension('S') ? MSTATUS_SUM : 0)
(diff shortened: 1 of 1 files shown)


riscv/processor.cc
```diff
@@ -609,7 +609,9 @@ void processor_t::set_csr(int which, reg_t val)
                  | MSTATUS_MPRV
                  | (supports_extension('S') ? MSTATUS_SUM : 0)
                  | MSTATUS_MXR | MSTATUS_TW | MSTATUS_TVM
-                 | MSTATUS_TSR | MSTATUS_UXL | MSTATUS_SXL |
+                 | MSTATUS_TSR
+                 | (supports_extension('U') ? MSTATUS_UXL : 0)
+                 | (supports_extension('S') ? MSTATUS_SXL : 0) |
```
aswaterman: Since UXL and SXL are not actually writable in Spike (they're either hardwired to 0 or hardwired to MXL), I think we can delete these two lines altogether.

aswaterman (approved):
Feel free to merge after deleting lines 613-614.

PR #406, opened 2020-02-27
Allow overriding CFLAGS and similar when building

rswarbrick:
Before this patch, I don't think it was possible to change (say)
CFLAGS as part of running the make command. Nor did setting them when
running configure do anything. Getting this right is a little fiddly:
for example, see Automake's approach at [1] ("AM_CFLAGS" and friends).

This patch adds an "mcppbs-" prefix, and sets things up properly for
CFLAGS, CPPFLAGS, CXXFLAGS and LDFLAGS. Note that the bulk of the
patch is either the auto-generated configure script or the ax_*.m4
files vendored in from the autoconf archive (needed to handle
--export-dynamic correctly without trashing settings from the user
running configure).

What's supposed to happen is as follows:

  - Base compilation flags that should apply to everything (standard
    optimisation flags, warning flags etc.) are defined in
	Makefile.in.

  - When the user runs configure, they can set compilation flags on
    the command line. These end up as environment variables in the
	shell script.

  - Compilation flags that can only be decided when we run
    configure (this is currently just whether we support
    -Wl,--export-dynamic) are appended to the configure-time LDFLAGS
    environment variable.

  - At the end of the configure script, these environment variables
    are spliced into Makefile.in to fill out the corresponding
	@<varname>@ entries.

  - When running make, the user might again override compilation
    flags. These will get appended to the flags found so far.

As a concrete example:

  mkdir build
  cd build
  ../configure CXXFLAGS='-O3'
  make CXXFLAGS='-O0'

will result in c++ compile commands that look like this:

  g++ -MMD -MP \
      -DPREFIX=\"/usr/local\" -Wall -Wno-unused -g -O2 -std=c++11 \
      -O3 \
      -O0 \
      -I. -I.. -I../fesvr -I../riscv -I../dummy_rocc -I../softfloat \
      -I../spike_main -fPIC -c ../fesvr/elfloader.cc

(I've added some newlines to wrap the long line).

Note that we have the base flags from Makefile.in (called
$(default-CXXFLAGS) there) first. Then we have the -O3 from the
configure command. Finally we have the -O0 from the Make command line.

And I can finally run "make CXXFLAGS='-O0 -g3'". Phew!

[1] https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -77,25 +77,52 @@ VPATH := $(addprefix $(src_dir)/, $(sprojs_enabled))
 # C++ compiler
 #  - CPPFLAGS : flags for the preprocessor (eg. -I,-D)
 #  - CXXFLAGS : flags for C++ compiler (eg. -Wall,-g,-O3)
+#
+# To allow a user to specify CFLAGS or similar as part of the Make
+# command, we also have mcpps-CFLAGS etc. with stuff that shouldn't be
+# lost in such a case.
+#
+# The order of precedence (highest to lowest) is then:
+#
+#    - Specified as part of Make command line
+#    - Specified as part of running configure
+#    - Specified here (default-CFLAGS)
+#
+# These all appear on the command line, from lowest precedence to
+# highest.
+
+default-CFLAGS   := -DPREFIX=\"$(prefix)\" -Wall -Wno-unused -g -O2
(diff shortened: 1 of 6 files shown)


aswaterman:
Looks fine to me, thanks.

PR #407, opened 2020-02-28
Add do-nothing support for mcountinhibit CSR

rswarbrick:
This CSR appeared in version 1.11 of the ISA and is described in the
Volume II: Privileged Architecture manual. It's an optional register
and should read as zero if not implemented, which is what this patch
does.


aswaterman:
Merged via fe24c46547f98576e74d3cabd5548a721cb38b3f

PR #409, opened 2020-03-04
Write execution logs to a named log file

rswarbrick:
This patch adds a --log argument to spike. If not given, the behaviour
is unchanged: messages logging execution of instructions and (if
commit logging is enabled) commits go to stderr.

If --log=P is given, Spike now writes these messages to a log file at
the path P. This is nice, because they are no longer tangled up with
other errors and warnings.

In the spirit of disentangling things further, if there is more than
one processor, each writes its logs to a separate file, with names
P.0, P.1, P.2 etc.

The code is mostly plumbing: passing a FILE* object through to the
functions that were using stderr. I've written a simple "log_file_t"
class, which opens a log file if necessary and yields it or stderr.

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -20,41 +20,47 @@ static void commit_log_stash_privilege(processor_t* p)
   state->last_inst_flen = p->get_flen();
 }
 
-static void commit_log_print_value(int width, const void *data)
+static void commit_log_print_value(FILE *log_file, int width, const void *data)
 {
+  assert(log_file);
   const uint64_t *arr = (const uint64_t *)data;
 
-  fprintf(stderr, "0x");
+  fprintf(log_file, "0x");
   for (int idx = width / 64 - 1; idx >= 0; --idx) {
-    fprintf(stderr, "%016" PRIx64, arr[idx]);
+    fprintf(log_file, "%016" PRIx64, arr[idx]);
   }
 }
 
-static void commit_log_print_value(int width, uint64_t hi, uint64_t lo)
+static void commit_log_print_value(FILE *log_file,
+                                   int width, uint64_t hi, uint64_t lo)
 {
+  assert(log_file);
+
   switch (width) {
     case 8:
-      fprintf(stderr, "0x%01" PRIx8, (uint8_t)lo);
+      fprintf(log_file, "0x%01" PRIx8, (uint8_t)lo);
       break;
     case 16:
-      fprintf(stderr, "0x%04" PRIx16, (uint16_t)lo);
+      fprintf(log_file, "0x%04" PRIx16, (uint16_t)lo);
       break;
     case 32:
-      fprintf(stderr, "0x%08" PRIx32, (uint32_t)lo);
+      fprintf(log_file, "0x%08" PRIx32, (uint32_t)lo);
       break;
     case 64:
-      fprintf(stderr, "0x%016" PRIx64, lo);
+      fprintf(log_file, "0x%016" PRIx64, lo);
       break;
     case 128:
-      fprintf(stderr, "0x%016" PRIx64 "%016" PRIx64, hi, lo);
+      fprintf(log_file, "0x%016" PRIx64 "%016" PRIx64, hi, lo);
       break;
     default:
       abort();
   }
 }
 
-static void commit_log_print_insn(processor_t* p, reg_t pc, insn_t insn)
+static void commit_log_print_insn(processor_t *p, reg_t pc, insn_t insn)
 {
+  FILE *log_file = p->get_log_file();
+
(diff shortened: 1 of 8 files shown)


rswarbrick:
I was a little torn on whether to split outputs per processor (the P.0, P.1, P.2... in the message above) and I don't do multi-processor simulations, so I'm not sure whether it's helpful or a hindrance. Removing that logic would be a simple change: let me know if that would be better.

mwachs5:
Well, it's certainly a change to the current behavior, is there any way to make the --log argument just take the usual trace stuff to the same file, then a different argument that. would split it up (--log_split ?)

rswarbrick:
@mwachs5 Wow, that was quick! There's no change to the current behaviour, because no splitting is done if you don't specify a log file (and everything ends up on stderr). If you *do* specify a path where logs should go, this currently uses it as a sort of base path and appends processor indices.

I think that if we make this configurable, it would probably be nicer to add a boolean option (--log_split or whatever) which says whether to treat the argument to --log as a full path (and mux everything together) or a base path (and append indices).

aswaterman:
The P.0, P.1, etc. thing doesn’t seem ideal to me. First of all, it’s easy
to split up the log file by hart with sed/awk postprocessing. Second,
splitting up the log loses the temporal relationship between the execution
streams. Printing to a single log file and relying on external tools to
split it up seems preferable to me.

rswarbrick:
I've just pushed up an amended version that writes to a single file with the --log argument. It turns out that I'd also messed up the "should I log commits?" logic too, which this also fixes.

rswarbrick:
@aswaterman This PR seems to have fallen through the cracks a bit. I've just rebased over recent edits to `execute.cc` and I think it should be mergeable again. Do you have any remaining changes that you would like made?

aswaterman:
Slipped through the cracks, indeed.  Thanks for rebasing.

rswarbrick:
Brill, thanks!

timsifive:
As of change bf296ca0643fa445b83d8bd45eefa3fca02d9921, spike crashes in some cases in the debug module weeds:
```
0x0000555555689aad in debug_module_t::dmi_read (this=0x7fffffffdd28, address=17, value=0x555555a32698)
    at ./riscv/debug_module.cc:434
434	          result = set_field(result, DMI_DMSTATUS_ALLHAVERESET,
(gdb) bt
#0  0x0000555555689aad in debug_module_t::dmi_read (this=0x7fffffffdd28, address=17, value=0x555555a32698)
    at ./riscv/debug_module.cc:434
#1  0x000055555568da81 in jtag_dtm_t::update_dr (this=0x5555559d1520) at ./riscv/jtag_dtm.cc:183
#2  0x000055555568d7dc in jtag_dtm_t::set_pins (this=0x5555559d1520, tck=true, tms=true, tdi=false)
    at ./riscv/jtag_dtm.cc:110
#3  0x000055555568d153 in remote_bitbang_t::execute_commands (this=0x555555a12470)
    at ./riscv/remote_bitbang.cc:130
#4  0x000055555568cf77 in remote_bitbang_t::tick (this=0x555555a12470) at ./riscv/remote_bitbang.cc:101
#5  0x0000555555609443 in sim_t::main (this=0x7fffffffd9b0) at ./riscv/sim.cc:98
#6  0x000055555560937e in sim_thread_main (arg=0x7fffffffd9b0) at ./riscv/sim.cc:83
#7  0x00005555558e51bb in context_t::wrapper (hi=32767, lo=4294958344) at ./fesvr/context.cc:29
#8  0x00007ffff7be85c0 in ?? () at ../sysdeps/unix/sysv/linux/x86_64/__start_context.S:91
   from /lib/x86_64-linux-gnu/libc.so.6
#9  0x0000555555a224a0 in ?? ()
#10 0x0000000000000000 in ?? ()
```

I'm trying to figure out exactly what the problem is, but `git bisect` points at this change, which looks to be harmless.

timsifive:
I think it has to do with the reorganizing of sim_t(), specifically the processors (procs) needs to be initialized before the constructor of debug_module_t() is called. I'm playing around to make things work again, and be a bit more resilient.

aswaterman:
Oops, thanks, Tim.

rswarbrick:
Oh, drat. Sorry about this. If no-one else has fixed it by tomorrow (Monday) morning, UK time, I'll sort it out and open a PR with a fix.

aswaterman:
I've just reverted the part of the commit uses `push_back` instead of sizing `procs` a priori (which, IMO, was totally fine to begin with).  This has the debug tests passing again for me.

rswarbrick:
Thanks for that.

timsifive:
Thanks, Andrew.

PR #417, opened 2020-03-09
Rvv fix 2020 03 09

chihminchao:
1. fix some corner cases
2. add configurable and parallel vf[w]redsum algorithm
3. update mstatus.vs for v0.9 spec


aswaterman:
I have reservations about 7702e160017029060e53154357bccd27dd4b89ae.  It adds a lot of implementation complexity.  I assume the intent is to support verifying a certain microarchitecture.  But the spec gives _way_ more freedom than we can ever hope to model in Spike.  So I worry that the increase in complexity won't actually help that much.

It's more practical to verify vf[w]redsum by computing the error bounds for a particular summation, or (in the degenerate case) just using inputs that always produce an exact result.

chihminchao:
I have removed the parallel algorithm part and let user to decide what precision they want.

aswaterman:
Thanks for understanding my position (especially since I know it was painful to write that code).

The rest of the PR looks great, so I'm merging it.  Thanks again!

PR #421, opened 2020-03-13
rvv: commitlog: fix vrgather_vv dump

chihminchao:


aswaterman (commented):
Thanks!

PR #430, opened 2020-03-22
Add generated files to .gitignore (Issue 429)

joelsherrill:


aswaterman:
This is a rather heavy-handed way of dealing with the problem. I always
just configure & build inside a build directory, instead of in the root of
the repository, which sidesteps the problem altogether. Some projects (GCC)
don’t even allow configuring and building inside the root of the repository
anymore.

scottj97:
NACK. The build instructions in the README have shown using a separate `build` directory since 2011, neatly avoiding all this mess.

PR #437, opened 2020-03-30
Assert that debug_module is initialized correctly.

timsifive:
This would have prevented the regression in #409.


aswaterman:
Good idea.

PR #448, opened 2020-04-21
Rvv spec 0.9

chihminchao:


riscv/decode.h
```diff
@@ -1510,13 +1519,15 @@ for (reg_t i = 0; i < vlmax && P.VU.vl != 0; ++i) { \
...
-  const reg_t nf = insn.v_nf() + 1; \
+#define VI_LD_COMMON(stride, offset, ld_width, elt_byte, is_seg) \
+  const reg_t nf = is_seg ? insn.v_nf() + 1 : 1; \
   const reg_t vl = P.VU.vl; \
   const reg_t baseAddr = RS1; \
   const reg_t vd = insn.rd(); \
   require((nf * P.VU.vlmul) <= (NVPR / 4) && \
           (vd + nf * P.VU.vlmul) <= NVPR); \
+  if (!is_seg) \
+    require(nf == 1); \
```
aswaterman, line 1552: Isn't this dead code, since on line 1523, you set nf to 1 if `!is_seg`?

riscv/decode.h
```diff
@@ -1542,28 +1553,30 @@ for (reg_t i = 0; i < vlmax && P.VU.vl != 0; ++i) { \
...
   VI_ST_COMMON(stride, offset, st_width, elt_byte) \
 
-#define VI_LDST_FF(itype, tsew) \
+#define VI_LDST_FF(itype, tsew, is_seg) \
   require(p->VU.vsew >= e##tsew && p->VU.vsew <= e64); \
-  const reg_t nf = insn.v_nf() + 1; \
+  const reg_t nf = is_seg ? insn.v_nf() + 1 : 1; \
   require((nf * P.VU.vlmul) <= (NVPR / 4)); \
+  if (!is_seg) \
+    require(nf == 1); \
```
aswaterman, line 1601: Same thing here

riscv/processor.h
```diff
@@ -265,8 +274,10 @@ class processor_t : public abstract_device_t
   }
   extension_t* get_extension() { return ext; }
   bool supports_extension(unsigned char ext) {
-    if (ext >= 'a' && ext <= 'z') ext += 'A' - 'a';
-    return ext >= 'A' && ext <= 'Z' && ((state.misa >> (ext - 'A')) & 1);
+    if (isupper(ext))
+      return ((state.misa >> (ext - 'A')) & 1);
+    else
+      return extension_table[ext];
```
aswaterman, line 280: It would be more robust to do a bounds check here.  You can add an exta entry in the enum, `EXT_MAX`, and check if `ext < EXT_MAX` before indexing the table, and otherwise `abort()`.  The compiler will optimize this, since `supports_extension` is usually called with a constant argument.

chihminchao: The table size is 256 equaling to the size of argument type "unsigned char"

riscv/decode.h
chihminchao: as mentioned above, it is wrong and will be fix in next PR

riscv/processor.h
aswaterman: Oh, I missed that detail. Sounds good.

riscv/decode.h
aswaterman: Ahh, makes sense.

chihminchao:
3rd PR to include some commitlog fix

aswaterman:
Should I wait for v3 or merge this?

chihminchao:
Please merge this and this is the 3rd PR.   Sorry for the confusion.

PR #452, opened 2020-04-23
Check and handle the memory regions when user specified memory regions

davetw:
If user specified the memory regions in the same way before spike aligns them to the page size (https://github.com/riscv/riscv-isa-sim/commit/2091113e041f9d7765990057fe6a6ba17f2fae15).
 
> ./spike -m0x08000000:0x4000,0x08000e00:0x1000,0x08020000:0x4000,0x08020e00:0x10
> 00 program.elf

The above memory regions will cause the following error

> ERROR (duplicate_node_names): Duplicate node name /memory@8000000
> ERROR (duplicate_node_names): Duplicate node name /memory@8020000 

so I added a function to merge the overlapping or remove the contains memory regions before return the regions.

### DIFF
--- a/spike_main/spike.cc
+++ b/spike_main/spike.cc
@@ -93,6 +93,44 @@ static void read_file_bytes(const char *filename,size_t fileoff,
   in.read(read_buf, read_sz);
 }
 
+bool sort_mem_region(const std::pair<reg_t, mem_t*> &a,
+                       const std::pair<reg_t, mem_t*> &b)
+{
+  if (a.first == b.first)
+    return (a.second->size() < b.second->size());
+  else
+    return (a.first < b.first);
+}
+
+void merge_overlapping_memory_regions(std::vector<std::pair<reg_t, mem_t*>>& mems)
+{
+  // check the user specified memory regions and merge the overlapping or
+  // eliminate the containing parts
+  std::sort(mems.begin(), mems.end(), sort_mem_region);
+  reg_t start_page = 0, end_page = 0;
+  std::vector<std::pair<reg_t, mem_t*>>::reverse_iterator it = mems.rbegin();
+  std::vector<std::pair<reg_t, mem_t*>>::reverse_iterator _it = mems.rbegin();
+  for(; it != mems.rend(); ++it) {
+    reg_t _start_page = it->first/PGSIZE;
+    reg_t _end_page = _start_page + it->second->size()/PGSIZE;
+    if (_start_page >= start_page && _end_page <= end_page) {
+      // contains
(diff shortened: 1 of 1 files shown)


aswaterman:
Looks like you have some unrelated commits in there.

davetw:
yap, I forgot to rebase the upstream firstly. Already updated. thanks.

spike_main/spike.cc
```diff
@@ -93,6 +93,44 @@ static void read_file_bytes(const char *filename,size_t fileoff,
...
+bool sort_mem_region(const std::pair<reg_t, mem_t*> &a,
+                       const std::pair<reg_t, mem_t*> &b)
+{
+  if (a.first == b.first)
+    return (a.second->size() < b.second->size());
+  else
+    return (a.first < b.first);
+}
+
+void check_memory_regions(std::vector<std::pair<reg_t, mem_t*>>& mems)
```
aswaterman: This should be called `merge_overlapping_memory_regions` or something.

spike_main/spike.cc
```diff
@@ -93,6 +93,44 @@ static void read_file_bytes(const char *filename,size_t fileoff,
...
+{
+  // check the user specified memory regions and merge the overlapping or
+  // eliminate the containing parts
+  std::sort(mems.begin(), mems.end(), sort_mem_region);
+  int32_t start_page = 0, end_page = 0;
+  std::vector<std::pair<reg_t, mem_t*>>::reverse_iterator it = mems.rbegin();
+  std::vector<std::pair<reg_t, mem_t*>>::reverse_iterator _it = mems.rbegin();
+  for(; it != mems.rend(); ++it) {
+    int32_t _start_page = it->first/PGSIZE;
+    int32_t _end_page = _start_page + it->second->size()/PGSIZE;
```
aswaterman: This will fail for physical addresses greater than 44 bits.  Should use reg_t, not int32_t.

spike_main/spike.cc
```diff
@@ -93,6 +93,44 @@ static void read_file_bytes(const char *filename,size_t fileoff,
...
+  else
+    return (a.first < b.first);
+}
+
+void check_memory_regions(std::vector<std::pair<reg_t, mem_t*>>& mems)
+{
+  // check the user specified memory regions and merge the overlapping or
+  // eliminate the containing parts
+  std::sort(mems.begin(), mems.end(), sort_mem_region);
+  int32_t start_page = 0, end_page = 0;
```
aswaterman: reg_t instead of int32_t here, too

PR #454, opened 2020-04-27
Load dtb

chihminchao:
1. add simple dtb parsing library from OpenSBI
2. add --dtb option to load binary dtb file
3. support configurable CLINT base from dtb

PR #455, opened 2020-04-27
Configurable platform parameters

davetw:
To make the platform-specify parameters configurable, such as the number of PMP regions. Users can choose how many PMP entries are implemented for the particular core. This would be useful to make sure the platform is using the dedicated FW or driver.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -204,7 +204,8 @@ reg_t mmu_t::pmp_ok(reg_t addr, reg_t len, access_type type, reg_t mode)
     return true;
 
   reg_t base = 0;
-  for (size_t i = 0; i < proc->state.n_pmp; i++) {
+  for (size_t i = 0; i < (size_t)proc->state.platform_n_pmp; i++) {
+
     reg_t tor = proc->state.pmpaddr[i] << PMP_SHIFT;
     uint8_t cfg = proc->state.pmpcfg[i];
 
@@ -226,7 +227,6 @@ reg_t mmu_t::pmp_ok(reg_t addr, reg_t len, access_type type, reg_t mode)
         any_match |= match;
         all_match &= match;
       }
-
       if (any_match) {
         // If the PMP matches only a strict subset of the access, fail it
         if (!all_match)
@@ -255,7 +255,7 @@ reg_t mmu_t::pmp_homogeneous(reg_t addr, reg_t len)
     return true;
 
   reg_t base = 0;
-  for (size_t i = 0; i < proc->state.n_pmp; i++) {
+  for (size_t i = 0; i < (size_t)proc->state.platform_n_pmp; i++) {
     reg_t tor = proc->state.pmpaddr[i] << PMP_SHIFT;
     uint8_t cfg = proc->state.pmpcfg[i];
 
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -440,6 +440,35 @@ static int ctz(reg_t val)
   return res;
 }
 
+void processor_t::set_pmp_num(reg_t pmp_num) {
+
+  // check the number of pmp is in a reasonable range
+  if( pmp_num >= 0 && pmp_num <= 16) {
+    state.platform_n_pmp = pmp_num;
+  } else {
+    fprintf(stderr, "error: bad number of pmp regions: '%ld' from the dtb\n", pmp_num);
+    abort();
+  }
+}
+
+void processor_t::set_pmp_granularity(reg_t pmp_granularity) {
+  // check the pmp granularity is set from dtb(!=0) and is power of 2
+  if (ctz(pmp_granularity) > 0 && ((pmp_granularity & (pmp_granularity - 1)) == 0)) {
+    state.pmp_granularity = ctz(pmp_granularity);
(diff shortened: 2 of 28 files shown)


aswaterman:
I note that https://github.com/riscv/riscv-isa-sim/pull/454 is using a very different strategy than this.  For controlling the minor details like the CLINT address and the number of PMPs, doesn't it make more sense to pick one strategy (command-line interface or FDT) and stick with it?  Supporting both seems like unnecessary complexity that will lead to bugs when the strategies are mixed together.

aswaterman:
cc @chihminchao

chihminchao:
#454 is included in this PR.

riscv/mmu.cc
```diff
@@ -204,21 +204,21 @@ reg_t mmu_t::pmp_ok(reg_t addr, reg_t len, access_type type, reg_t mode)
     return true;
 
   reg_t base = 0;
-  for (size_t i = 0; i < proc->state.n_pmp; i++) {
-    reg_t tor = proc->state.pmpaddr[i] << PMP_SHIFT;
+  for (size_t i = 0; i < (size_t)proc->state.platform_n_pmp; i++) {
+    reg_t tor = proc->state.pmpaddr[i] << proc->state.pmp_granularity;
```
aswaterman: I don't believe this new logic is correct.  PMP_SHIFT is a constant no matter what the granularity is.  The granularity just changes the range of values that the register can hold.

riscv/processor.cc
```diff
@@ -107,6 +107,27 @@ static bool check_pow2(int val)
...
+  } else {
+      fprintf(stderr, "error: bad number of pmp regions: '%ld' from the dtb\n", pmp_num);
+      abort();
+  }
+}
+
+void processor_t::set_pmp_granularity(reg_t pmp_granularity) {
+  // check the pmp granularity is set from dtb(!=0) and is power of 2
+  if (pmp_granularity >= 4 && ((pmp_granularity & (pmp_granularity - 1)) == 0)) {
+    state.pmp_granularity = pmp_granularity;
```
aswaterman: Since you are using `state.pmp_granularity` as a shift amount elsewhere, don't you need a `ctz` here?

riscv/processor.cc
```diff
@@ -609,18 +630,23 @@ void processor_t::set_csr(int which, reg_t val)
...
+  size_t i = which - CSR_PMPADDR0;
+  if (which >= CSR_PMPADDR0 && which < CSR_PMPADDR0 + state.platform_n_pmp) {
     bool locked = state.pmpcfg[i] & PMP_L;
-    bool next_locked = i+1 < state.n_pmp && (state.pmpcfg[i+1] & PMP_L);
-    bool next_tor = i+1 < state.n_pmp && (state.pmpcfg[i+1] & PMP_A) == PMP_TOR;
+    bool next_locked = i+1 < (size_t)state.platform_n_pmp && (state.pmpcfg[i+1] & PMP_L);
+    bool next_tor = i+1 < (size_t)state.platform_n_pmp && (state.pmpcfg[i+1] & PMP_A) == PMP_TOR;
     if (!locked && !(next_locked && next_tor))
-      state.pmpaddr[i] = val & ((reg_t(1) << (MAX_PADDR_BITS - PMP_SHIFT)) - 1);
+      state.pmpaddr[i] = val & ((reg_t(1) << (MAX_PADDR_BITS - state.pmp_granularity)) - 1);
```
aswaterman: This doesn't seem correct, either.  You should read the Rocket source code and see how `pmpGranularity` is used.  https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/rocket/PMP.scala

riscv/processor.cc
```diff
@@ -609,18 +630,23 @@ void processor_t::set_csr(int which, reg_t val)
...
 
     mmu->flush_tlb();
+  } else if(which >= CSR_PMPADDR0 + state.platform_n_pmp &&  which < CSR_PMPADDR0 + state.n_pmp) {
+    fprintf(stderr, "error: trying to set PMP (%ld), but the system has %d PMP regions \n",
+            i, state.platform_n_pmp);
+    abort();
```
aswaterman: Following the convention of the rest of this function, you should just delete the `else if` part and let this case fall through.

riscv/mmu.cc
aswaterman: Err, I meant to say, "the granularity just changes how the comparators behave".  See the Rocket source code I referenced in the other comment.  The granularity does not affect how stuff gets stored in the register, how much it's shifted, etc.

riscv/processor.cc
davetw: the minimum granularity is when G=0 in 2^(G+2).

aswaterman: No, but I mean, `pmp_granularity` is `2^(G+2)`, whereas elsewhere in the code you are using `state.pmp_granularity` to mean `G+2`, so you need to do `state.pmp_granularity = ctz(pmp_granularity)` to take the logarithm of the RHS.

khannaudit:
@davetw is this ready for testing and merging?

aswaterman:
I was having trouble following some of the logic in the coarse-grained PMP implementation, so I broke it into smaller commits and refactored it a bit.  It now follows the same logic as the Rocket RTL does.  I'm going to merge now, but please review my changes on master and LMK if I messed anything up.

PR #458, opened 2020-05-04
Rvv fp16

chihminchao:


aswaterman:
I think we might also need vfncvt and vfwcvt to support 8-bit integers when FP16 support is included.  (I think we also need to support 16-bit integers even when FP16 is not supported.)  You could ask for clarification on the riscv-v-spec repo if you aren't sure.

PR #459, opened 2020-05-06
Add missing stdexcept imports (for GCC 10)

seldridge:
[GCC 10 changed some C++ standard library headers to not import `stdexcept`](https://gcc.gnu.org/gcc-10/porting_to.html). This adds those back in so that spike will build on distros with GCC 10 as the default, e.g., Fedora 32.

This should be backwards compatible.

### DIFF
--- a/fesvr/dtm.cc
+++ b/fesvr/dtm.cc
@@ -6,6 +6,7 @@
 #include <string.h>
 #include <assert.h>
 #include <pthread.h>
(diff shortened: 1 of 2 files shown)


aswaterman:
Thanks @seldridge! It's good that they're making the headers less sloppy over time, but I wish they got it right the first time, so that we couldn't accidentally rely on the sloppiness.

PR #461, opened 2020-05-12
Hardwire mstatus.[sie,spie] to zero if 'S' mode absent

khannaudit:
UIE and UPIE fields of mstatus are explicitly marked as 'WPRI' in Privileged ISA 1.12-draft, which implies they are hardwired to zero when N extension is absent. By the same logic, hardwire SIE and SPIE to zero (although this is not explicitly mentioned in the spec).


aswaterman (approved):
LGTM, thanks!

PR #462, opened 2020-05-16
Read interactive debug commands from standard input

michaelmelanson:
I recently wrote a cosimulator for [my own RISC-V emulator](https://github.com/michaelmelanson/riscy), intending to drive Spike by starting it as a child process and interacting with it over standard input and standard error. But I ran into a problem because apparently Spike was, in fact, reading its commands from standard _error_ instead of standard input.

I've always assumed that you could not read from standard output/error so that was a weird discovery. 🤯

This change fixes the interactive debug loop so it reads from file descriptor 0 (standard input) instead of 2 (standard error). It should work as before when used interactively from the terminal, but it should now accept piped commands or (as was my use case) from a parent process.

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -85,7 +85,7 @@ void sim_t::interactive()
   while (!done())
   {
     std::cerr << ": " << std::flush;
-    std::string s = readline(2);
+    std::string s = readline(0);
 
     std::stringstream ss(s);
     std::string cmd, tmp;


aswaterman:
Yeah, I did it this way for a reason... if you read from stdin here, it messes up the debugging of target-machine programs that read from stdin (via system calls proxied through pk).

Obviously my approach assumes we don't need to debug target-machine programs that read from stderr, but as you point out, reading from stderr is pretty rare...

Can you consider working around this in your own way?  You might be able to use a shell script to redirect some file descriptors, or writing a wrapper around spike in C that moves them around before calling exec.

michaelmelanson:
(Sorry for the delay here...)

Thanks @aswaterman, I think I can work around it. I'll close this.

PR #466, opened 2020-05-20
Fix state.misa garbage initialization

khannaudit:
Original bug-report: Simulator is seeing an `flen` of 128 for a `--isa=rv64imafdc` argument

I found that [this](https://github.com/riscv/riscv-isa-sim/blob/2d8a94234ec1b1d57bb81bf5b826cb832cd8a2e1/riscv/processor.cc#L153) call to `get_flen` checks the state of `misa` [here](https://github.com/riscv/riscv-isa-sim/blob/2d8a94234ec1b1d57bb81bf5b826cb832cd8a2e1/riscv/processor.h#L278), and because `processor.reset()` hasn't been called until this point, `misa` is actually holding the value from [this](https://github.com/riscv/riscv-isa-sim/blob/a21011116cdaaa0eea2d4f16175cfc3f267d4765/riscv/processor.cc#L273) code. The initial `state.misa` value could be garbage, leading to this bug.

Since `misa` isn't being initialized anywhere before the line in question (is it?), it should be sufficient to set it directly to the parsed value of `max_isa`.

@aswaterman *please* give this a thorough review if you can, to make sure it doesn't break anything else. I tested it on the original bug report and a few other ISA strings.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -270,7 +270,7 @@ void processor_t::parse_isa_string(const char* str)
     }
   }
 
-  state.misa |= max_isa;
+  state.misa = max_isa;
 
   if (!supports_extension('I'))
     bad_isa_string(str, "'I' extension is required");


BrianGraysonSiV:
Was this a latent bug that has been there for a long time, and only recently exposed in some weird way, or was there a change recently that exposed this? I've been doing waves with verilator for months, and this only cropped up in the last week or two, so something changed in some way....

khannaudit:
Looks to me like it was a latent bug that only recently got exposed in a weird way, so something changed about how we use spike, not necessarily about spike itself. I’m yet to figure out what exactly exposed this bug.

ss2783 (approved):
lgtm

one suggestion would be actually initialize the state.misa value explicitly before the parsing any of the isa/priv/varch strings as that way you wouldn't have to rely on that order.

aswaterman:
Thanks, guys.  It appears this is a month-old regression from fdd146a802385f3cfd96311d65c6a516fd67dff1.  @chihminchao can you comment on the rationale?  Regardless, I'm going to merge this now, just to get the bug fixed.

FWIW, Valgrind found this bug immediately.  It also found others related to recent vector PRs.  I'll fix those separately.

aswaterman:
BTW @ss2783 I agree that zeroing `state.misa` beforehand is the better solution if the `|=` is necessary for some reason I don't understand.

PR #470, opened 2020-05-22
Make gen_icache script use only POSIX commands

anthony-coulter:
The `seq` utility is not available in OpenBSD.


aswaterman:
Thanks!

PR #493, opened 2020-06-22
RISC-V H-Extension v0.6.1 Support

avpatel:
This patch series adds RISC-V H-Extension v0.6.1 Support to Spike.

Current state of H-Extension support:
1. Latest OpenSBI v0.8 with H-Extension v0.6.1 support works fine
2. The nested MMU test-suite in latest Xvisor-next repo works fine
3. We are able to boot Linux on SMP Guest with 2 VCPUs on single/multi-HART Spike using latest Xvisor-next repo
4. We can now pass custom kernel (i.e. Linux/Xvisor) bootargs to Spike using command-line options
5. We can now separately pass kernel flat image (i.e. Linux/Xvisor) to Spike using command-line options

Regards,
Anup


aswaterman:
This is awesome.  Thanks, @avpatel 

@chihminchao can I get your help to code-review this patch set?

riscv/mmu.h
```diff
@@ -134,8 +155,10 @@ class mmu_t
 #endif
 
   // template for functions that store an aligned value to memory
-  #define store_func(type) \
-    void store_##type(reg_t addr, type##_t val) { \
+  #define store_func(type, prefix, xlate_flags) \
+    void prefix##_##type(reg_t addr, type##_t val) { \
+      if (xlate_flags) \
+        flush_tlb(); \
```
aswaterman, line 161: This is heavyweight, but given that guest loads and stores are much rarer than regular loads and stores, it's good for now.

avpatel: I have avoided changing Spike host TLB implementation in this series.

We need at least "V" bit in TLB tags to distinguish between HS/U-mode TLB entries and VS/VU-mode TLB entries. If this is available then we don't need to flush TLB for guest loads and stores.

Updating host TLB for H-extension and removing these TLB flushes can be done separately.

aswaterman: Agreed (if it ever needs to be done at all).

avpatel:
I have rebased this series on latest master branch. Everything works fine on latest master branch.

@aswaterman do you have review comments ??

Regards,
Anup

aswaterman:
I think we should add something like this in decode.h:

    #define PRV_HS (PRV_S + 1)  // internal Spike encoding of S-mode with V=0

rather than using the expression `PRV_S + 1` in several places.

Once you fix that, I think we should merge it.

avpatel:
I have updated as-per your suggestion.

Thanks,
Anup

riscv/decode.h
```diff
@@ -352,10 +353,18 @@ inline freg_t f128_negate(freg_t a)
...
   STATE.serialized = false; \
   unsigned csr_priv = get_field((which), 0x300); \
-  bool mode_unsupported = csr_priv == PRV_S && !P.supports_extension('S'); \
-  unsigned csr_read_only = get_field((which), 0xC00) == 3; \
-  if (((write) && csr_read_only) || STATE.prv < csr_priv || mode_unsupported) \
+  bool mode_unsupported = (csr_priv == PRV_S && !P.supports_extension('S')) || \
+                          (csr_priv == PRV_HS && !P.supports_extension('H')); \
+  if (mode_unsupported) \
     throw trap_illegal_instruction(0); \
+  unsigned state_prv = (STATE.prv == PRV_S && !STATE.v) ? PRV_S + 1: STATE.prv; \
```
aswaterman: Can you also replace this instance of `(PRV_S + 1)` with `PRV_HS`?

riscv/decode.h
```diff
@@ -352,10 +353,18 @@ inline freg_t f128_negate(freg_t a)
...
+  unsigned csr_read_only = get_field((which), 0xC00) == 3; \
+  if (((write) && csr_read_only) || state_prv < csr_priv) { \
+    if (csr_priv == (PRV_S + 1)) \
```
aswaterman: this one, too

avpatel:
My bad for missing these places. I will update soon.

Regards,
Anup

riscv/processor.cc
```diff
@@ -861,11 +1052,118 @@ void processor_t::set_csr(int which, reg_t val)
...
+      mask |= (supports_extension('F') ? SSTATUS_FS : 0);
+      mask |= (supports_extension('V') ? SSTATUS_VS : 0);
+      mask |= (xlen == 64 ? SSTATUS64_SD : SSTATUS32_SD);
+      state.vsstatus = (state.vsstatus & ~mask) | (val & mask);
+      break;
+    }
+    case CSR_VSIE: {
+      reg_t mask = MIP_VS_MASK;
+      state.mie = (state.mie & ~mask) | ((val << 1) & mask);
+      break;
```
aharris-ventana, line 1142: hideleg should be considered here. I've confirmed this behavior with J. Hauser:
```
Second, as you say, hideleg should always be active.  (Although older
versions of the spec did accidentally say otherwise.)
```
Ditto for VSIP.

riscv/processor.cc
```diff
@@ -1153,6 +1521,33 @@ reg_t processor_t::get_csr(int which)
...
+    case CSR_HGATP: return state.hgatp;
+    case CSR_HGEIP: return 0;
+    case CSR_VSSTATUS: {
+      reg_t mask = SSTATUS_VS_MASK;
+      mask |= (supports_extension('F') ? SSTATUS_FS : 0);
+      mask |= (supports_extension('V') ? SSTATUS_VS : 0);
+      mask |= (xlen == 64 ? SSTATUS64_SD : SSTATUS32_SD);
+      return state.vsstatus & mask;
+    }
+    case CSR_VSIE: return (state.mie & MIP_VS_MASK) >> 1;
```
aharris-ventana: hideleg, per 5.2.11:
```
When bit 10 of hideleg is zero, vsip.SEIP and vsie.SEIE are read-only zeros. Else, vsip.SEIP
and vsie.SEIE are aliases of hip.VSEIP and hie.VSEIE.
When bit 6 of hideleg is zero, vsip.STIP and vsie.STIE are read-only zeros. Else, vsip.STIP
and vsie.STIE are aliases of hip.VSTIP and hie.VSTIE.
When bit 2 of hideleg is zero, vsip.SSIP and vsie.SSIE are read-only zeros. Else, vsip.SSIP and
vsie.SSIE are aliases of hip.VSSIP and hie.VSSIE.
```

riscv/processor.cc
avpatel: I had considered HIDELEG for SIP and SIE access from VS-mode but forgot to consider it for VSIP and VSIE.

Thanks for catching. I will update.

riscv/processor.cc
avpatel: Same as above, I should have considered HIDELEG when reading CSR_VSIE and CSR_VSIP.

I will update.

avpatel:
Addressed all comments so far and updated this PR.

Thanks,
Anup

aswaterman:
I'll merge this now, and we can address any other bug reports in follow-on PRs.  Thanks, @avpatel (and @aharris-ventana)

PR #494, opened 2020-06-25
Rvv fix 2020 06 25

chihminchao:


aswaterman:
I don't understand the logic in viota, but as long as the effect is "disallow all vd/vs2 overlap, and disallow vd/v0 overlap if masked", it's OK.

PR #495, opened 2020-06-26
Warn about pitfall involving invoking spike+pk

nick-knight:


aswaterman (commented):
Isn’t there more to it than that? There’s a search path built into spike, I think. But I’m on a cell phone, so I can’t point to the code.

PR #496, opened 2020-06-27
Return non-zero trap value for illegal CSR accessing instructions

IanJiangICT:
The original Spike always returns zero in mtval/stval for illegal instruction traps that accessing CSR.
Here another choice is provided that returns instruction bits rather zero.
A command line parameter --tval-nz-csr is also added to turn this function on.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -353,7 +353,10 @@ inline freg_t f128_negate(freg_t a)
   bool mode_unsupported = csr_priv == PRV_S && !P.supports_extension('S'); \
   unsigned csr_read_only = get_field((which), 0xC00) == 3; \
   if (((write) && csr_read_only) || STATE.prv < csr_priv || mode_unsupported) \
-    throw trap_illegal_instruction((uint32_t)insn_bits); \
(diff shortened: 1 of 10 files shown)


aswaterman (commented):
This patch doesn’t make sense to me. The illegal-instruction reporting feature doesn’t have anything to do with the csrrx instructions.

aswaterman (commented):
This patch doesn’t make sense to me. The illegal-instruction reporting feature doesn’t have anything to do with the csrrx instructions.

IanJiangICT:
> This patch doesn’t make sense to me. The illegal-instruction reporting feature doesn’t have anything to do with the csrrx instructions.

The RISC-V Instruction Set Manual Volume II: Privileged Architecture Document Version 1.12-draft
2.1 CSR Address Mapping Conventions
"Attempts to access a CSR without appropriate privilege level or to write a read-only register also raise illegal instruction
exceptions."

aswaterman:
But you’re only enabling the feature for csrrx, not other instructions. A proper fix would orthogonally implement it for all illegal instructions, not only a tiny subset.

PR #507, opened 2020-07-15
[RFC] Implement snapshot feature for spike

lazyparser:
we implement snapshot feature which can save the runtime states of spike machine.

usage:

save snapshot: in interactive mode， use "snapshot " to save snapshot file to path, e.g. by inputing "snapshot a" will create a snapshot file ./a

load snapshot: use the optional "--snapshot=" command line option.

note: although loaded snapshot file, you still should specify which ELF file to load(to provide boot args, and some symbols in ELF, it won't be loaded actually)

### DIFF
--- a/fesvr/elfloader.cc
+++ b/fesvr/elfloader.cc
@@ -15,7 +15,7 @@
 #include <vector>
 #include <map>
 
-std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry,bool ramdump)
+std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry, bool ramdump)
 {
   int fd = open(fn, O_RDONLY);
   struct stat s;
--- a/fesvr/elfloader.cc
+++ b/fesvr/elfloader.cc
@@ -15,7 +15,7 @@
 #include <vector>
 #include <map>
 
-std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry)
+std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry,bool ramdump)
 {
   int fd = open(fn, O_RDONLY);
   struct stat s;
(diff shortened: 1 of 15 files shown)


fesvr/elfloader.cc
```diff
@@ -82,11 +82,50 @@ std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t*
       } \
     } \
   } while(0)
+  #define LOAD_ELF_NO_MEM(ehdr_t, phdr_t, shdr_t, sym_t, bswap) do { \
```
kito-cheng, line 85: I saw the only difference is _NO_MEM version didn't load anything to memory, just load the symbol table, so add a parameter to LOAD_ELF instead of copy-paste would be better.

```c
    for (unsigned i = 0; i < bswap(eh->e_phnum); i++) {			\
      if(bswap(ph[i].p_type) == PT_LOAD && bswap(ph[i].p_memsz)) {	\
        if (bswap(ph[i].p_filesz)) {					\
          assert(size >= bswap(ph[i].p_offset) + bswap(ph[i].p_filesz)); \
          memif->write(bswap(ph[i].p_paddr), bswap(ph[i].p_filesz), (uint8_t*)buf + bswap(ph[i].p_offset)); \
        } \
        zeros.resize(bswap(ph[i].p_memsz) - bswap(ph[i].p_filesz)); \
        memif->write(bswap(ph[i].p_paddr) + bswap(ph[i].p_filesz), bswap(ph[i].p_memsz) - bswap(ph[i].p_filesz), &zeros[0]); \
      } \
    } \
```

fesvr/elfloader.cc
```diff
@@ -82,11 +82,50 @@ std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t*
...
+        unsigned max_len = bswap(sh[strtabidx].sh_size) - bswap(sym[i].st_name); \
+        assert(bswap(sym[i].st_name) < bswap(sh[strtabidx].sh_size));	\
+        assert(strnlen(strtab + bswap(sym[i].st_name), max_len) < max_len); \
+        symbols[strtab + bswap(sym[i].st_name)] = bswap(sym[i].st_value); \
+      } \
+    } \
+  } while(0)
 
   if (IS_ELF32(*eh64))
-    LOAD_ELF(Elf32_Ehdr, Elf32_Phdr, Elf32_Shdr, Elf32_Sym, from_le);
```
kito-cheng, line 87: Then here just `LOAD_ELF(Elf32_Ehdr, Elf32_Phdr, Elf32_Shdr, Elf32_Sym, from_le, ramdump);` is fine.

fesvr/elfloader.cc
```diff
@@ -15,7 +15,7 @@
 #include <vector>
 #include <map>
 
-std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry)
+std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry,bool ramdump)
```
kito-cheng: space after comma(,) before last argumet.
and maybe snapshot_mode instead of ramdump?

fesvr/elfloader.h
```diff
@@ -8,6 +8,6 @@
 #include <string>
 
 class memif_t;
-std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry);
+std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry, bool ramdump = 0);
```
kito-cheng, line 11: `false` rather than `0` for bool type.

fesvr/htif.h
```diff
@@ -15,7 +15,7 @@ class htif_t : public chunked_memif_t
  public:
   htif_t();
   htif_t(int argc, char** argv);
-  htif_t(const std::vector<std::string>& args);
+  htif_t(const std::vector<std::string>& args, bool ramdump = 0);
```
kito-cheng, line 18: `false` rather than `0` for bool type.

snapshot/snapshot.cc
```diff
@@ -0,0 +1,73 @@
+#include "snapshot.h"
+#include <fstream>
+#define MTIMECMP_BASE	0x4000
+#define MTIME_BASE	0xbff8
```
kito-cheng, line 4: Extract MTIMECMP_BASE and MTIME_BASE into header file and then use that would be better, those two value already defined in clint.cc.

spike_main/spike.cc
```diff
@@ -352,7 +363,20 @@ int main(int argc, char** argv)
 
   sim_t s(isa, priv, varch, nprocs, halted, real_time_clint,
       initrd_start, initrd_end, start_pc, mems, plugin_devices, htif_args,
-      std::move(hartids), dm_config, log_path, dtb_enabled, dtb_file);
+      std::move(hartids), dm_config, log_path, dtb_enabled, dtb_file,ramdump);
+  if(ramdump) {
+    #define MTIMECMP_BASE	0x4000
+    #define MTIME_BASE	0xbff8
```
kito-cheng, line 369: Same issue in snapshot.cc

spike_main/spike.cc
```diff
@@ -352,7 +363,20 @@ int main(int argc, char** argv)
...
+    typedef uint64_t mtime_t;
+    typedef uint64_t mtimecmp_t;
```
kito-cheng, line 371: Maybe make those types to public for clint_t and use that would be better.

kito-cheng:
Another general comment is, maybe you could squash your into 2 commit, 1 for spike and 1 for fesvr.

fesvr/elfloader.cc
liweiwei90: Thanks for the reviews！
I’m the mentor of mengls666. We will merge it into LOAD_ELF as what you suggest.

fesvr/elfloader.cc
liweiwei90: OK. We will fix this as well as check similar format issues.

fesvr/elfloader.h
liweiwei90: we'll fix it.

snapshot/snapshot.cc
liweiwei90: OK. We'll extract these into a header file.

liweiwei90:
> Another general comment is, maybe you could squash your into 2 commit, 1 for spike and 1 for fesvr.

I’m not sure whether is it better to directly squash the commits in the next update or incrementally submit to keep the review record and squash the commits finally？

kito-cheng:
> I’m not sure whether is it better to directly squash the commits in the next update or incrementally submit to keep the review record and squash the commits finally？

I am not sure the policy for spike, @aswaterman any comment?

aswaterman:
Whatever makes it easiest for people reviewing the code to follow the logical progression.

liweiwei90:
OK. We'll squash the commits in the next update. Thanks a lot!

liweiwei90:
Thanks for the reviews again！We create a new PR #510 to squash the commits into two parts, and add a new commit to fix the issues suggested in this review

lazyparser:
> Thanks for the reviews again！We create a new PR #510 to squash the commits into two parts, and add a new commit to fix the issues suggested in this review

Thanks for reviewing. Since squashed commits would break the relationship / map between reviews and codes, a new PR is created for the modified version. PR #510

PR #508, opened 2020-07-17
Fix legalize_privilege for extension H

abhinay-kayastha:
Intorduced by #493

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -592,7 +592,7 @@ reg_t processor_t::legalize_privilege(reg_t prv)
   if (!supports_extension('U'))
     return PRV_M;
 
-  if (prv == PRV_S && !supports_extension('S'))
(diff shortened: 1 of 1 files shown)


aswaterman:
@abhinay-kayastha Turns out ~this is a bug~ this fix is incomplete: see #557.  `legalize_privilege` should disallow selecting MPP=2 even if the hypervisor extension is provided. Fixed in 8966775ab443879d064a6e8419eeea82b901ccca

aswaterman:
I think I understand the intent now - you were trying to prevent it from being possible to set MPP=2 unless the hypervisor extension is present.  So your fix was correct for the no-hypervisor case but insufficient for the hypervisor case.  I'll complete it by making it always impossible.

abhinay-kayastha:
Thanks @aswaterman for fixing the issue.

PR #513, opened 2020-07-23
Rvv fix 2020 07 23

chihminchao:


chihminchao:
will be re-sent with 1.0  patch

PR #515, opened 2020-07-26
bash-style function for interactive mode

13163223717:
this commit add some bash-style function to the interactive mode such as:
1.command completion
2.history list
3.line edit with arrow keys like in a bash environment

in addition to these , some setting funcs in interactive mode has been added which you can use them set the value of reg, pc or memory.


riscv/encoding.h
```diff
@@ -3132,6 +3132,35 @@ DECLARE_CSR(mhpmcounter29h, CSR_MHPMCOUNTER29H)
 DECLARE_CSR(mhpmcounter30h, CSR_MHPMCOUNTER30H)
 DECLARE_CSR(mhpmcounter31h, CSR_MHPMCOUNTER31H)
 #endif
+#ifdef CSR_DECLARE
```
kito-cheng: `encoding.h` is generating by `riscv-opcode`, don't edit it directly, and seems like `CSR_DECLARE` is similar to `DECLARE_CSR`, but a small subset, I guess the reason is want to prevent `sim_t::interactive_csr` dump all CSR registers?

riscv/interactive.cc
```diff
@@ -426,3 +628,248 @@ void sim_t::interactive_until(const std::string& cmd, const std::vector<std::str
...
+  if(args.size() != 1)
+    throw trap_interactive();
+
+    int i = 0;
+    std::map<std::string,reg_t> csr;
+    processor_t *p = get_core(args[0]);
+
+    #define CSR_DECLARE(name, number) csr[#name] = p->get_csr(number);
+    #include "encoding.h"              // generates if's for all csrs
+    #undef CSR_DECLARE
```
kito-cheng: Maybe using `DECLARE_CSR` would be better? and move `csr` to member data and then init once at constructor of `sim_t` or lazy init could prevent construct that again and again.

riscv/encoding.h
13163223717: Exactly, call get_csr() with some uninplemented csr may cause a illegal instrucions trap so i want to skip them.

riscv/interactive.cc
13163223717: thanks for review, i have modified it

hawkinsw:
I was working on making spike use libreadline for its command line input/editing. Should I continue that work? Or will this great patchset make that work obsolete?

PR #527, opened 2020-08-11
Add option to dissable implicit ebreak in program buffer

sobuch:
This can be useful e.g. when testing OpenOCD


timsifive (approved):
This looks like a useful feature (for testing debuggers), and a reasonable implementation.

PR #530, opened 2020-08-18
Fix debug tests failing with impebreak enabled.

timsifive:
Introduced in #527.


timsifive:
@aswaterman, what do you want in terms of reviews for code like this? I could just merge it if you don't want to be bothered reviewing changes in these files. Should I find a reviewer?

aswaterman:
@timsifive For debug changes you consider obvious/uncontroversial, please feel free to merge them yourself.  Add me as a reviewer whenever you'd like my review.

timsifive:
Sounds good. Thanks.

PR #535, opened 2020-08-28
Rvv pre 1.0 2020 08 27

chihminchao:


aswaterman:
Why remove quad-widening instructions?  Even if not in "V", they're still valid, right?

softfloat/fall_reciprocal.c
```diff
@@ -0,0 +1,414 @@
...
+    return (UINT64_MAX >> (64 - len)) << pos;
+}
+
+//user needs to truncate output to required length
+static inline uint64_t rsqrte(uint64_t val, int e, int s, int p, bool sub) {
+  uint64_t exp = extract64(val, s, e);
+  uint64_t sig = extract64(val, 0, s);
+  uint64_t sign = extract64(val, s + e, 1);
+
+  const uint8_t table[] = {
```
aswaterman: You should make this `static const` so it doesn't get generated on the stack every time it's called.

chihminchao:
The [spec update](https://github.com/riscv/riscv-v-spec/commit/72142aa47906ca4f92810d5a073aa046a394991e) has removed the opcode.  I think it should be invalid in 1.0 version

softfloat/fall_reciprocal.c
chihminchao: good suggestion.  It will be fix in updated push.

softfloat/fall_reciprocal.c
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  int idx = ((exp & 1) << (p-1)) | (sig >> (s-p+1));
+  uint64_t out_sig = (uint64_t)(table[idx]) << (s-p);
+  uint64_t out_exp = (3 * make_mask64(0, e - 1) + ~exp) / 2;
+
+  return (sign << (s+e)) | (out_exp << s) | out_sig;
+}
+
+float16_t f16_rsqrte(float16_t in, int precision)
```
aswaterman: I don't think precision should be an argument - it's implicit.

chihminchao: Do you means to use   "float16_t f16_rsqrte7(float16_t in)   and fix the precision in function ?

aswaterman: I think you have two options:
- change name to `f16_rsqrte7` and remove the parameter
- keep current name, and `assert(precision == 7);`

My preference is the first option.

aswaterman: > The [spec update](https://github.com/riscv/riscv-v-spec/commit/72142aa47906ca4f92810d5a073aa046a394991e) has removed the opcode. I think it should be invalid in 1.0 version

OK, I didn't notice that change.

softfloat/fall_reciprocal.c
chihminchao: has been fixed

softfloat/fall_reciprocal.c
chihminchao: I refine it as option 1

chihminchao:
There is a issue to judge sNaN and qNaN case from fclassify and will be fixed in next push

aswaterman:
@chihminchao ready to merge, or waiting for next push?

chihminchao:
ready to merge.  Please use the latest push.

chihminchao:
@aswaterman  
It is ok to merge.

PR #537, opened 2020-09-01
Fix MIDELEG and MEDELEG emulation when H-extension is available

avpatel:


riscv/processor.cc
```diff
@@ -894,11 +894,11 @@ void processor_t::set_csr(int which, reg_t val)
         (1 << CAUSE_MISALIGNED_FETCH) |
         (1 << CAUSE_BREAKPOINT) |
         (1 << CAUSE_USER_ECALL) |
-        (1 << CAUSE_SUPERVISOR_ECALL) |
```
aswaterman: I don't understand this change.  Doesn't it prevent the delegation of supervisor ECALLs?

avpatel:
Ahh yes, I should have kept SUPERVISOR_ECALL line.

I will update in few minutes.

Regards,
Anup

riscv/processor.cc
scottj97: @aswaterman this was in response to my [comment here](https://github.com/riscv/riscv-isa-sim/commit/9af85e39a550ba031e4fe9c1913e275959a9927b#r41925179). After that commit (and 7775c6fb7cc1b29bf529fee9fb900cef513222c1), the maximum value of medeleg (with non-H misa) changed from 0xb109 to 0xb309, newly allowing delegation of supervisor ecall.

@avpatel says that change was inadvertent...so I'm not sure why he removed the fix here.

avpatel:
I got confused by similar names SUPERVISOR_ECALL (0x9) and VIRTUAL_SUPERVISOR_ECALL (0xa) but Andrew is right M-mode can delegate SUPERVISOR_ECALL (0x9) to S-mode.

Regards,
Anup

aswaterman:
Cool, thanks.

PR #548, opened 2020-09-21
Improved log

sthiruva:
Hi,
This pull request adds the following features to improves the trace:

1. Ability to  load multiple elfs. The symbols of all the elfs are collected but only the last elf load takes effect.
2. All trace lines prepended with "core   x: ". This makes it easier to "grep" out per core activity.
3. PC is looked up in the symbol table to indicate the start of a function or any other label making it easier to figure out what a given cpu is upto without a look up in the disassembly file.

I have attached part of a trace file.

[sample_trace.txt](https://github.com/riscv/riscv-isa-sim/files/5254751/sample_trace.txt)

### DIFF
--- a/fesvr/htif.cc
+++ b/fesvr/htif.cc
@@ -121,7 +121,16 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
 
 void htif_t::load_program()
 {
-  std::map<std::string, uint64_t> symbols = load_payload(targs[0], &entry);
+  // Allow load of multiple elfs
+  // Unfortunately it is not additive now as it clears up the previous loads
+  // It is however useful to get symbols
+  // The last elf should be the combined elf
+  // TODO: make elf load additive
+  for(auto const& targ : targs)
+  {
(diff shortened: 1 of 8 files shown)


khannaudit:
Useful and much needed!

aswaterman:
Please stick to one topic per PR.

The multiple ELF loading scheme seems too half-baked to include.  Better to merge the ELFs into a single ELF using the linker or objcopy or whatever.

(Also, the multi-ELF command-line scheme isn't compatible with the current command-line handling.  The args after targs[0] provide the argc/argv given to the target machine, so e.g. you can run `spike pk ./a.out foo bar` and `foo bar` show up as argv[1]/argv[2].)

I'll review the log part of the PR if it's resubmitted as a standalone PR.

sthiruva:
I will submit the log part as a separate PR

Regarding the multiple elf, if I use objcopy won't the symbols be lost?
I will then not be able to look up symbols in a meaningful way.
Any suggestion on how we work around this?

aswaterman:
You should talk to some toolchain experts about that one.

sthiruva:
Breaking up the pull request into multiple request. So closing this.

PR #550, opened 2020-09-23
Rvv pre 1.0 2020 09 22

chihminchao:


riscv/processor.cc
```diff
@@ -767,7 +767,9 @@ void processor_t::set_csr(int which, reg_t val)
 {
 #if defined(RISCV_ENABLE_COMMITLOG)
 #define LOG_CSR(rd) \
-  STATE.log_reg_write[((which) << 4) | 4] = {get_csr(rd), 0};
+  STATE.log_reg_write[((which) << 4) | 4] = { \
+      get_csr(rd, insn_t(0), false, true), \
+      0};
```
aswaterman: I think you should undo this change, and...

riscv/processor.h
```diff
@@ -272,8 +272,8 @@ class processor_t : public abstract_device_t
   void reset();
   void step(size_t n); // run for n cycles
   void set_csr(int which, reg_t val);
-  reg_t get_csr(int which, insn_t insn, bool write);
-  reg_t get_csr(int which) { return get_csr(which, insn_t(0), false); }
+  reg_t get_csr(int which, insn_t insn, bool write, bool peek = 0);
+  reg_t get_csr(int which) { return get_csr(which, insn_t(0), false, false); }
```
aswaterman: change this one to pass `true`.  My intent was that `get_csr(int)` is the "peek" form that skips permission checks.

riscv/processor.cc
chihminchao: has been changed in push

riscv/processor.h
chihminchao: has been modified in push

aswaterman:
Thank you!

PR #552, opened 2020-09-25
pmp: fix local scope issue

chihminchao:
There is an global 'res' in function and local 'res' in pmpcfg loop.
This makes the 'ret' macro use the local res but what we want is to pass
local result to global 'res'


aswaterman (approved):
Oops, thanks for catching this mistake!

PR #556, opened 2020-09-28
Add core id to lines generated by --log-commits

sthiruva:
Add core id lines generated by --log-commit.
This lets us separate log lines based on cpu_id

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -72,7 +72,7 @@ static void commit_log_print_insn(processor_t *p, reg_t pc, insn_t insn)
 
   // print core id on all lines so it is easy to grep
   uint64_t id = p->get_csr(CSR_MHARTID);
-  fprintf(log_file, "core%4ld: ", id);
+  fprintf(log_file, "core%4" PRId64 ": ", id);
 
   fprintf(log_file, "%1d ", priv);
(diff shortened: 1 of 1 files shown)


riscv/execute.cc
```diff
@@ -70,6 +70,10 @@ static void commit_log_print_insn(processor_t *p, reg_t pc, insn_t insn)
   int xlen = p->get_state()->last_inst_xlen;
   int flen = p->get_state()->last_inst_flen;
 
+  // print core id on all lines so it is easy to grep
+  uint64_t id = p->get_csr(CSR_MHARTID);
+  fprintf(log_file, "core%4ld: ", id);
```
aswaterman: You'll want to cast `id` to `long` if you do it this way, because otherwise this is incorrect on 32-bit systems (since there, `long` isn't the same width as `uint64_t`).  The better alternative is to rewrite this as

    fprintf(log_file, "core%4" PRId64 ": ", id);

(See https://en.cppreference.com/w/cpp/types/integer for an explanation of PRId64 if you haven't seen that syntax before.)

sthiruva:
Fixed

abhinay-kayastha:
There is a typo here and missing `d`. It should be
`fprintf(log_file, "core %4d" PRId64 ": ", id);`

PR #558, opened 2020-09-29
Adding symbol lookup when --enable-commitlog is enabled

sthiruva:
The PR add symbol lookup when --enable-commitlog is enabled.
This makes debug easier.

The output looks like this:
core   0: >>>>  at_zero_end
core   0: 0x000000008000000c (0x000060b3) or      ra, zero, zero
3 0x000000008000000c (0x000060b3) x 1 0x0000000000000000
core   0: 0x0000000080000010 (0x00006133) or      sp, zero, zero
3 0x0000000080000010 (0x00006133) x 2 0x0000000000000000

An earlier PR adds core id to all log-commits line

### DIFF
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -294,12 +294,10 @@ char* sim_t::addr_to_mem(reg_t addr) {
   return NULL;
 }
 
-#ifdef RISCV_ENABLE_COMMITLOG
 const char* sim_t::get_symbol(uint64_t addr)
 {
   return htif_t::get_symbol(addr);
 }
-#endif
 
 // htif
 
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -294,6 +294,13 @@ char* sim_t::addr_to_mem(reg_t addr) {
   return NULL;
 }
 
+#ifdef RISCV_ENABLE_COMMITLOG
+const char* sim_t::get_symbol(uint64_t addr)
+{
+  return htif_t::get_symbol(addr);
+}
+#endif
+
 // htif
 
 void sim_t::reset()
--- a/riscv/sim.h
+++ b/riscv/sim.h
@@ -94,9 +94,7 @@ private:
   void make_dtb();
   void set_rom();
 
-#ifdef RISCV_ENABLE_COMMITLOG
   const char* get_symbol(uint64_t addr);
-#endif
 
   // presents a prompt for introspection into the simulation
   void interactive();
--- a/riscv/sim.h
+++ b/riscv/sim.h
@@ -94,6 +94,10 @@ private:
   void make_dtb();
   void set_rom();
 
+#ifdef RISCV_ENABLE_COMMITLOG
+  const char* get_symbol(uint64_t addr);
+#endif
+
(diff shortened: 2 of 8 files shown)


riscv/sim.cc
```diff
@@ -294,6 +294,13 @@ char* sim_t::addr_to_mem(reg_t addr) {
   return NULL;
 }
 
+#ifdef RISCV_ENABLE_COMMITLOG
+const char* sim_t::get_symbol(uint64_t addr)
+{
+  return htif_t::get_symbol(addr);
+}
+#endif
+
```
aswaterman, line 301: I don't think this should be necessary, since `get_symbol` is already defined on `htif_t`, which is a superclass.

riscv/sim.h
```diff
@@ -94,6 +94,10 @@ class sim_t : public htif_t, public simif_t
   void make_dtb();
   void set_rom();
 
+#ifdef RISCV_ENABLE_COMMITLOG
+  const char* get_symbol(uint64_t addr);
+#endif
```
aswaterman: I don't think this should be necessary, since `get_symbol` is already defined on `htif_t`, which is a superclass.

riscv/simif.h
```diff
@@ -16,6 +16,10 @@ class simif_t
   virtual bool mmio_store(reg_t addr, size_t len, const uint8_t* bytes) = 0;
   // Callback for processors to let the simulation know they were reset.
   virtual void proc_reset(unsigned id) = 0;
+
+#ifdef RISCV_ENABLE_COMMITLOG
+  virtual const char* get_symbol(uint64_t addr) = 0;
+#endif
```
aswaterman: I don't think we want to conditionalize the presence of a virtual function on a preprocessor macro; this will cause ABI issues when Spike is used as a library.  Can we just remove this entirely and have sim_t directly use the version defined in its superclass htif_t?

sthiruva: The processor_t class only has access to the interface simif_t* and not to the sim_t object directly.
So I am not sure how to get hold of the sim_t object.

There are ways to do it, but all of them don't feel *clean*.
1. We could type cast the pointer to simif_t* in processor_t to sim_t*
2. Add a get_sim_t_obj()
3. Change virtual get_symbol(uint64_t addr) = 0; in simif.h to  virtual get_symbol(uint64_t addr) {nullptr};
    We have a check for nullptr, which indicates symbols not found.

Is there a cleaner way to get the sim_t* obj?

I have removed the unnecessary get_symbols() defs in sim.h/cpp, I will check it in once we have resolved the above issue.

aswaterman: Ah, I forgot about that.  OK, yeah, it needs to go in simif_t.

Let's just keep the code as-is, but remove the `#ifdef`.  There's no harm in building this additional code when the commit log is disabled.  The extra time it takes to build the map at startup is tiny in the grand scheme of things.

sthiruva:
Fixed. Removed the #ifdef RISCV_ENABLE_COMMITLOG.
Only the call in processor.cc now has the ifdef enabled.

aswaterman:
Thanks!

PR #563, opened 2020-10-06
rvv: sstatus.SD needs to include vs dirty state

chihminchao:


aswaterman:
Thanks for the bug fix.

Longer term, let's refactor the code that generates the SD bit to avoid duplication of logic between mstatus and sstatus and thereby avoid bugs.

chihminchao:
If the position of FS/XS/VS for mstatus and sstatus are always the same.  sstatus.SD could be got from mstatus.SD directly.

PR #564, opened 2020-10-06
Update SATP and HGATP mask usage to make it clearer

abhinay-kayastha:
Updating the mask used by SATP and HGATP registers to make it clearer so that it if rv64_ppn_mask becomes a function of paddr_bits then the final value will be correct.


aswaterman:
thanks!

PR #567, opened 2020-10-10
Fix new ELF checks on big endian hosts

zeldin:
The new macros IS_ELF_... introduced in 80b5b2f5 were not endian safe.


aswaterman:
Indeed, thank you.

PR #568, opened 2020-10-11
Fix VSSTATUS bits updation

avpatel:
This patch fixes VSSTATUS bits updations as follows:
1. set_csr() should not allow writs to VSSTATUS.SD bit instead
   VSSTATUS.SD bit should be derived again after VSSTATUS is updated
2. set_virt() should copy FS, VS and XS bits when transitioning
   from virt=on (VS/VU-mode) to virt=off (HS/M-mode)

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -605,21 +605,22 @@ void processor_t::set_virt(bool virt)
        * we should sync Guest/VM FS, VS, and XS state with Host FS,
        * VS, and XS state.
        */
-       if ((state.mstatus & SSTATUS_FS) == SSTATUS_FS) {
-         state.vsstatus |= SSTATUS_FS;
+      state.vsstatus &= ~SSTATUS_FS;
+      state.vsstatus |= (state.mstatus & SSTATUS_FS);
+      if (supports_extension('V')) {
+        state.vsstatus &= ~SSTATUS_VS;
+        state.vsstatus |= (state.mstatus & SSTATUS_VS);
+      }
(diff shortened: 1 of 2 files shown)


aswaterman:
Thanks @avpatel, I'll review soon.

riscv/processor.cc
```diff
@@ -605,18 +605,22 @@ void processor_t::set_virt(bool virt)
...
-       if ((state.mstatus & SSTATUS_VS) == SSTATUS_VS) {
-         state.vsstatus |= SSTATUS_VS;
-         state.vsstatus |= (xlen == 64 ? SSTATUS64_SD : SSTATUS32_SD);
-       }
-       if ((state.mstatus & SSTATUS_XS) == SSTATUS_XS) {
-         state.vsstatus |= SSTATUS_XS;
+      if (supports_extension('F')) {
+        state.vsstatus &= ~SSTATUS_FS;
+        state.vsstatus |= (state.mstatus & SSTATUS_FS);
+      }
```
aswaterman: The FS field exists even without the F extension (see https://github.com/riscv/riscv-isa-manual/blob/bb9e0561ce5df5e2ed65ebc2f4b253b59f534e82/src/machine.tex#L916) so you should remove the `supports_extension('F')` check.

(By contrast, it's correct to have the `supports_extension('V')` check for VS.)

aswaterman:
LGTM otherwise.

riscv/processor.cc
avpatel: Sure, I will update as-per your suggestion.

PR #569, opened 2020-10-13
reduce sig_len constraint to 4 bytes

neelgala:
Spike currently asserts that the signature length should always be a multiple of 16-bytes. However, the compliance suite has agreed to upon the signature being a multiple ot 4-bytes. This prevents some of the tests to run on spike since it fails the assertion.

The proposed change fixes this issue and reduces the assertion to 4 bytes.

### DIFF
--- a/fesvr/htif.h
+++ b/fesvr/htif.h
@@ -101,8 +101,8 @@ class htif_t : public chunked_memif_t
        +rfb=DISPLAY          to be accessible on 5900 + DISPLAY (default = 0)\n\
       --signature=FILE     Write torture test signature to FILE\n\
        +signature=FILE\n\
-      --granularity=VAL           Size of each line in signature.\n\
-       +granularity=VAL\n\
+      --signature-granularity=VAL           Size of each line in signature.\n\
+       +signature-granularity=VAL\n\
       --chroot=PATH        Use PATH as location of syscall-servicing binaries\n\
        +chroot=PATH\n\
       --payload=PATH       Load PATH memory as an additional ELF payload\n\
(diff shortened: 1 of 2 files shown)


aswaterman:
Is that the right call for 64-bit systems, too?

If so, then doesn't it make more sense to remove the constraint altogether?

aswaterman:
This commit does more than relax a constraint: it also changes the output format.  I think we should preserve compatibility by continuing to print in 16-byte chunks, zero-padding the last line to 16 bytes if necessary.

neelgala:
I am fine with removing the assert altogether - not sure why it was there in the first place. 
The reason compliance chose the signatures to be 4-byte multiples is since the any bas ISA version will have LW/SW and thus soft routines to extract signatures could leverage those. 

Zero padding will change the signature file and spike will fail compliance. 

Does it make sense to make it command line configurable ? default to 16-bytes unless overriden by the command line ?

aswaterman:
I think making it a command-line argument that defaults to 16 is a good compromise.

neelgala:
great.. let me find someone who can help me with this.. and i'll push in the changes.

neelgala:
would you accept if we made this option available during the configuration (build phase) rather than an active cli option?

aswaterman:
I think it should be a command-line option... it's weird if Spike must be statically configured to be able to pass the compliance suite.  I don't think it's hard to do.

neelgala:
could you review the current changes ?

fesvr/htif.h
```diff
@@ -121,6 +124,7 @@ TARGET (RISC-V BINARY) OPTIONS\n\
 {"signature", required_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 2 },     \
 {"chroot",    required_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 3 },     \
 {"payload",   required_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 4 },     \
+{"size",    optional_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 5 },     \
```
aswaterman: I think the name "size" is too generic for this purpose.  Maybe "signature-alignment" or something?

aswaterman:
Looks fine to me, aside from my one comment.

neelgala:
changed it to signature-granularity.  Good to merge?

neelgala:
@aswaterman reminder.

aswaterman:
@neelgala Sorry for letting this slip; reminders are always welcome.

PR #573, opened 2020-10-16
fixed mtval update for breakpoint instructions

neelgala:
This in regards to the recent clarity in spec regarding ebreaks and mtval : https://github.com/riscv/riscv-isa-manual/issues/600

Mtval is now updated with va of the ebreak instruction.


aswaterman:
Thank you, @neelgala.

PR #575, opened 2020-10-18
Big endian target support

zeldin:
This PR adds support for simulating a big endian RISC-V processor.

The last commit in the series does not add or change any functionality, it just tags values which are (expected to be) in target byteorder, thus making it harder to miss when a conversion is needed.  (This could be useful also without the BE RISC-V support since conversions are needed when emulating an LE target on a BE host as well, but as it is presented here it builds upon the other commits.)

Toolchain updates for building big endian target binaries:
https://github.com/zeldin/riscv-binutils-gdb/commits/big-endian
https://github.com/zeldin/riscv-gcc/commits/big-endian
https://github.com/zeldin/riscv-newlib/commits/big-endian

Proxy kernel update:
https://github.com/zeldin/riscv-pk/commits/big-endian

To build a toolchain defaulting to big endian, use `make XLEN=64_be` and/or `make XLEN=32_be` with riscv-gnu-toolchain.

To build a big endian proxy kernel, use `configure --prefix=$RISCV --host=riscv64_be-unknown-elf`.
(Note that you will need to supply the full path to your big endian `pk` to `spike` as the default `TARGET_DIR` it stil `riscv64-unknown-elf/bin`.)

For reference, here is an FPGA RV32 implementation supporting big endian mode:
https://github.com/SpinalHDL/VexRiscv


riscv/mmu.h
```diff
@@ -275,21 +275,21 @@ class mmu_t
   inline icache_entry_t* refill_icache(reg_t addr, icache_entry_t* entry)
   {
     auto tlb_entry = translate_insn_addr(addr);
-    insn_bits_t insn = from_le(*(uint16_t*)(tlb_entry.host_offset + addr));
+    insn_bits_t insn = from_target(*(target_endian<uint16_t>*)(tlb_entry.host_offset + addr));
```
sorear: This seems to contradict https://github.com/riscv/riscv-isa-manual/blob/master/src/intro.tex#L638-L646

zeldin: Interresting.  When was this added?  I was following the RISC-V Instruction Set Manual v2.2 which says that parcels are stored "in a memory halfword according to the implementation's natural endianness".

If the encoding has changed so that the parcels are always little endian internally as well, then I'll need to update the toolchain and the FPGA implementation too.  Should be a pretty small change though.

sorear: git has a nice feature for that: https://github.com/riscv/riscv-isa-manual/blame/master/src/intro.tex#L638-L646

(Jun 21, 2019)

You may also be interested in https://github.com/riscv/riscv-isa-manual/blob/master/src/machine.tex#L722-L808

zeldin: Thanks for the `mstatus` pointer.  I did have bi-endian on my roadmap so it's nice to see that there is an official spec for it.  Right now I need to fix the toolchain and update this PR though.  :smile_cat:

zeldin:
I updated the commits to reflect the new instruction format (only `refill_icache` in `mmu.h` and `set_rom` in `sim.cc` have changed).
The binutils repo has also been updated.  The FPGA implementation has not been updated yet.

zeldin:
Now VexRiscv has also merged the corresponding fix.

riscv/mmu.h
```diff
@@ -424,6 +460,9 @@ class mmu_t
   reg_t pmp_homogeneous(reg_t addr, reg_t len);
   reg_t pmp_ok(reg_t addr, reg_t len, access_type type, reg_t mode);
 
+#ifdef RISCV_ENABLE_DUAL_ENDIAN
+  bool target_big_endian;
+#endif
```
chihminchao: bool target_big_endian;

Suggestion: let compiler optimize path by constant value and remove some ifdef in code

riscv/mmu.h
```diff
@@ -342,6 +342,42 @@ class mmu_t
 #endif
   }
 
+  void set_target_big_endian(bool enable)
+  {
+#ifdef RISCV_ENABLE_DUAL_ENDIAN
+    target_big_endian = enable;
+#else
+    assert(enable == false);
+#endif
```
chihminchao, line 351:
```
#ifdef` RISCV_ENABLE_DUAL_ENDIAN
    target_big_endian = enable;
#endif
```

riscv/mmu.h
```diff
@@ -342,6 +342,42 @@ class mmu_t
...
+  }
+
+  bool is_target_big_endian()
+  {
+#ifdef RISCV_ENABLE_DUAL_ENDIAN
+    return target_big_endian;
+#else
+    return false;
+#endif
```
chihminchao: remove ifdef

riscv/mmu.h
```diff
@@ -342,6 +342,42 @@ class mmu_t
...
+  }
+
+  template<typename T> inline T from_target(T n) const
+  {
+#ifdef RISCV_ENABLE_DUAL_ENDIAN
+    return target_big_endian? from_be(n) : from_le(n);
+#else
+    return from_le(n);
+#endif
```
chihminchao: remove ifdef

riscv/mmu.h
```diff
@@ -342,6 +342,42 @@ class mmu_t
...
+  }
+
+  template<typename T> inline T to_target(T n) const
+  {
+#ifdef RISCV_ENABLE_DUAL_ENDIAN
+    return target_big_endian? to_be(n) : to_le(n);
+#else
+    return to_le(n);
+#endif
```
chihminchao: remove ifdef

riscv/mmu.h
chihminchao:
```
#ifdef RISCV_ENABLE_DUAL_ENDIAN
  bool target_big_endian;
#else
  const bool target_big_endian
#endif
```

fesvr/memif.h
```diff
@@ -19,6 +25,11 @@ class chunked_memif_t
 
   virtual size_t chunk_align() = 0;
   virtual size_t chunk_max_size() = 0;
+
+  virtual void set_target_endianness(memif_endianness_t endianness) {}
+  virtual memif_endianness_t get_target_endianness() const {
+    return memif_endianness_undecided;
+  }
```
chihminchao, line 33: I think the part may be optimized the target_big_endian part to use constant to help compiler and remove ifdef in overloading functions.

chihminchao:
Besides,  the current version conflict with latest master and needs rebase.

riscv/mmu.h
zeldin: You mean change this to
```
#ifdef RISCV_ENABLE_DUAL_ENDIAN
  bool target_big_endian;
#else
  const bool target_big_endian = false;
#endif
```
?  Sure, I can do that.

riscv/mmu.h
zeldin: Should probably keep the assert in the #else branch here, otherwise you end up with a function which exists but silently does nothing when called.

fesvr/memif.h
zeldin: Not sure what you mean here.  `target_big_endian` is in `mmu_t`, which is a concrete class and a completely separate API from `memif_t`.

riscv/mmu.h
chihminchao: What I expect is 
```diff --git a/riscv/mmu.cc b/riscv/mmu.cc
index db90342..ad48f4c 100644
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -6,9 +6,7 @@
 
 mmu_t::mmu_t(simif_t* sim, processor_t* proc)
  : sim(sim), proc(proc),
-#ifdef RISCV_ENABLE_DUAL_ENDIAN
   target_big_endian(false),
-#endif
   check_triggers_fetch(false),
   check_triggers_load(false),
   check_triggers_store(false),
diff --git a/riscv/mmu.h b/riscv/mmu.h
index e42ac82..2670d9b 100644
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -346,36 +346,22 @@ public:
   {
 #ifdef RISCV_ENABLE_DUAL_ENDIAN
     target_big_endian = enable;
-#else
-    assert(enable == false);
 #endif
   }
 
   bool is_target_big_endian()
   {
-#ifdef RISCV_ENABLE_DUAL_ENDIAN
     return target_big_endian;
-#else
-    return false;
-#endif
   }
 
   template<typename T> inline T from_target(target_endian<T> n) const
   {
-#ifdef RISCV_ENABLE_DUAL_ENDIAN
     return target_big_endian? n.from_be() : n.from_le();
-#else
-    return n.from_le();
-#endif
   }
 
   template<typename T> inline target_endian<T> to_target(T n) const
   {
-#ifdef RISCV_ENABLE_DUAL_ENDIAN
     return target_big_endian? target_endian<T>::to_be(n) : target_endian<T>::to_le(n);
-#else
-    return target_endian<T>::to_le(n);
-#endif
   }
 
 private:
@@ -462,6 +448,8 @@ private:
 
 #ifdef RISCV_ENABLE_DUAL_ENDIAN
   bool target_big_endian;
+#else
+  const bool target_big_endian;
 #endif
   bool check_triggers_fetch;
   bool check_triggers_load;
```

zeldin: If you do it like that then the compiler has no way to optimize away the runtime check.  It only knows that `target_big_endian` is constant, not what the actual constant value _is_ (that must be set by the constructor, which lives in a different compilation unit, and it's also possible for different constructors to set different values).

In fact, even using `const bool target_big_endian = false` is not enough to get the optimization (just checked with clang), it needs to be `static const bool target big_endian = false`.

Of course, if you are fine with not optimizing and always having the check done at runtime regardless of `--enable-dual-endian` or not, then I have no problem with that either.  In that case the `#ifdef` around the declation can be dropped and it can always be `bool target_big_endian`.

aswaterman: We definitely want to optimize these checks away for `--disable-dual-endian`.

zeldin:
I rebased on master, and removed some `#ifdef`s as requested.

aswaterman:
Sorry to cause churn, but does removing those `#ifdef`s mean that little-endan-target-on-little-endian-host will be slowed down for `--disable-dual-endian` -- i.e. that checks on the critical code paths will not be optimized away?  If so, I think it was better before.

zeldin:
@aswaterman By using `static const` instead of just `const`, the compiler can optimize the checks away, as a simple test on [godbolt](https://godbolt.org/z/WcWP9E) shows that both gcc and clang do indeed do.  I think using `inline const` would force the compiler to do this, but that is C++17.

But anyway, I made the change to remove the `#ifdef`s as a separate commit, so it's just a matter of dropping that commit if you don't want it anymore.  I also prefer the version with the `#ifdef`s, but I have no strong opinion in the matter.

sorear:
I haven't touched this code in a while so I might be completely off but it ought to be possible to do dual-endian with very little overhead by having two versions of the load and store instruction functions and changing instruction decoding.  It would complicate the code and probably needs to be benchmarked; it may not be worthwhile.

aswaterman:
@zeldin Yep, that checks out.  Thanks.

@sorear Some quick benchmarking suggests `--enable-dual-endian` is only a minor slowdown (when running on a same-endian host, anyway): around 5% on a workload that's >50% loads and stores.  So probably no need for anything more elaborate.

PR #577, opened 2020-10-22
Update htif.cc

pawks:
The existing code defaulted the final byte to 0x00 instead of the value in the memory. This pull request fixes the bug which was caused due to an incorrect bounds check.


neelgala:
@aswaterman missed an "=" condition in the last PR. This one fixes that. Compliance suite also passes with this.

aswaterman:
OK, thanks.

PR #581, opened 2020-10-23
mseccfg spike bug fix and unit test

soberl:
For mseccfg, fix bugs for pmp_ok() and set_csr(). Also create test cases with highly combinations with GNU gengen.


soberl:
Sorry, the pull target is set incorrectly. Thus close it now

PR #582, opened 2020-10-24
s2xlate() trap fix

avpatel:
Fix trap generation in s2xlate() function.


aswaterman:
Looks good on inspection; waiting for confirmation of fix.  Thanks!

PR #592, opened 2020-11-11
Fix LR missing misaligned exception

### LINKED ISSUE #591: Misaligned LR should take exception
The A extension spec says:

> For LR and SC, the A extension requires that the address held in rs1 be naturally aligned to the size of the operand (i.e., eight-byte aligned for 64-bit words and four-byte aligned for 32-bit words). If the address is not naturally aligned, an address-misaligned exception or an access-fault exception will be generated.

When building Spike with `--enable-misaligned`, as of 956ef9ac3a66d2d6cda5df33a0e3f1c7d57ed0e0 I observe this exception occurring for SC ([from check_load_reservation()](https://github.com/riscv/riscv-isa-sim/blob/956ef9ac3a66d2d6cda5df33a0e3f1c7d57ed0e0/riscv/mmu.h#L259) -- added for #431) but the same exception does not occur for the LR.

It's not clear where to add this check while maintaining the proper exception prioritization. It needs to be higher priority than page fault or access fault. (The spec allows it optionally to be lower priority but that wouldn't be consistent with Spike's SC implementation.)

The check could be the first thing in acquire_load_reservation(), but that's called after the load has already occurred, which is too late.

scottj97:
Strawman proposal to fix #591. I'm not exactly fond of this solution but I think it works.

### DIFF
--- a/riscv/insns/lr_d.h
+++ b/riscv/insns/lr_d.h
@@ -1,5 +1,5 @@
 require_extension('A');
 require_rv64;
-auto res = MMU.load_int64(RS1);
+auto res = MMU.load_int64(RS1, true);
 MMU.acquire_load_reservation(RS1);
 WRITE_RD(res);
--- a/riscv/insns/lr_w.h
+++ b/riscv/insns/lr_w.h
@@ -1,4 +1,4 @@
 require_extension('A');
-auto res = MMU.load_int32(RS1);
+auto res = MMU.load_int32(RS1, true);
 MMU.acquire_load_reservation(RS1);
 WRITE_RD(res);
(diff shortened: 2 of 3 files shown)


scottj97:
Except I just noticed that e91092f21948ba2e77bfbc4629b82b6aa14d5ed6 causes a misaligned AMO to generate a load address misaligned exception instead of the correct store address misaligned.

I could revert that change and leave the old behavior, or add another catch() clause to convert that into the proper exception. What's your preference?

aswaterman:
Ah, good point.  Another `catch` sounds good, since it matches the existing pattern.

PR #593, opened 2020-11-12
Selective mmu mode

chihminchao:
Use "mmu-type" in dts to specified implemented  mmu mode 

If there is no dts
   rv32 -> sv32 
   rv64 -> sv39,sv48
else if there is an dts having :mmu-type" field
   "riscv,bare" -> no mmu
   "riscv,sv32" -> sv32
   "riscv,sv39" -> sv39
   "riscv,sv48" -> sv39,sv48

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -236,6 +236,7 @@ private:
 #define require_rv64 require(xlen == 64)
 #define require_rv32 require(xlen == 32)
 #define require_extension(s) require(p->supports_extension(s))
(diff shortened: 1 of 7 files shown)


aswaterman (approved):

khannaudit:
@chihminchao Does this consider that different cores in a heterogeneous system could have different `mmu` attributes?

aswaterman:
@khannaudit scanning the code, that does seem to be the intent, but I did not test that.

PR #596, opened 2020-11-15
Fix byteorder issues with struct riscv_stat

zeldin:
This fixes the `stat` syscall(s) both for little endian target emulation on a big endian host, and for big endian target emulation on a little endian host.


aswaterman:
LGTM, thanks.  struct stat has always been a pain...

PR #598, opened 2020-11-16
Pmp per core

chihminchao:
The original implementation uses the value in the first core to configure all cores. The patch makes pmp configured by each core's setting in DTS


aswaterman:
LGTM, thanks.

PR #602, opened 2020-11-22
Fix misaligned loads and stores for big endian target

zeldin:


zeldin:
See riscv/riscv-pk#224 for other half.  :smile_cat:

PR #603, opened 2020-11-26
Fix missing include in elfloader.cc

db434:
Build wasn't working out-of-the-box for me on Ubuntu 20.10 with GCC 10.2.0. Including `stdexcept` fixed the issue.


aswaterman:
Thanks for the fix!

PR #605, opened 2020-11-27
Fix hstatus.GVA and mstatus.GVA updation

avpatel:
Update hstatus.GVA and mstatus.GVA as-per H-extension specification.


aswaterman:
I just did a quick skim, but it looks good to me.

PR #608, opened 2020-11-27
Fix #607: Add a core parameter to the interactive str command

### LINKED ISSUE #607: Interactive `str` command does not work to print a string at an address returned from `malloc` system call
First, thank you for such a helpful tool! Without spike I would not have been able to write and debug a program that uses the yet-to-be-officially adopted vector extensions!

While debugging the aforementioned program, I was trying to use spike to be able to print a NUL-terminated string from memory and having a problem. The string was stored at an address returned by a call to `malloc`. It was possible to inspect the memory at the `malloc`-returned address (using the interactive `mem` command). However, `str <address>` was not working. It printed no output.  Debugging spike on that command showed that it was causing a fetch error because spike had determined the address was on a ROM device.

I have a patch that will fix the issue and I will be sending it along in a PR shortly. 

Again, thank you for your work on spike and I hope that the forthcoming PR is helpful and ultimately accepted for inclusion!

Will

hawkinsw:
Add a core parameter to the interactive str command. This makes it
possible for the spike user to specify the device whose memory contains
the NUL-terminated string to be printed.

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -125,7 +125,7 @@ void sim_t::interactive_help(const std::string& cmd, const std::vector<std::stri
     "vreg <core> [reg]               # Display vector [reg] (all if omitted) in <core>\n"
     "pc <core>                       # Show current PC in <core>\n"
     "mem <hex addr>                  # Show contents of physical memory\n"
-    "str <hex addr>                  # Show NUL-terminated C string\n"
+    "str <core> <hex addr>           # Show NUL-terminated C string at <hex addr> in core <core>\n"
     "until reg <core> <reg> <val>    # Stop when <reg> in <core> hits <val>\n"
     "until pc <core> <val>           # Stop when PC in <core> hits <val>\n"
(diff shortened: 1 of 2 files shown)


aswaterman:
I probably would've kept the single-argument version that bypasses address translation (analogous to the `mem` command), but we can add it back if anyone gripes.

hawkinsw:
I _believe_ that the code should do that. If there is only one parameter then it assumes that you want to use the debug MMU. That's from memory, so I could be wrong! Thanks for merging! I will look for some other issues to tackle :-)

chihminchao:
@hawkinsw 
there is some issue about this PR.
[This part](https://github.com/riscv/riscv-isa-sim/blob/f4f6e12eeeabccc31853ab4e8570219d1f5cba37/riscv/sim.cc#L358) makes some clint_t access dump lots of duplicated "clint_t" message such as running linux

hawkinsw:
I am so sorry! I left in debugging output. I will take care of it as soon
as I can!

Will

Issue #609, opened 2020-12-01
Debugging output present when running any simulation operation that involves `addr_to_mem`.


hawkinsw:
https://github.com/riscv/riscv-isa-sim/blob/f4f6e12eeeabccc31853ab4e8570219d1f5cba37/riscv/sim.cc#L358

Thank you to @chihminchao for pointing out this issue. I am terribly sorry for introducing this in the first place! Patch coming.

Will

aswaterman:
Resolved by #610

PR #610, opened 2020-12-01
Fix Issue #609

hawkinsw:
Fix Issue #609 where extraneous debugging output was added when the user
invoked any simulation operation that involved `addr_to_mem`.


aswaterman:
Thanks!

PR #618, opened 2020-12-18
Check and use proc variable in MMU emulation

avpatel:
Don't blindly use proc variable in MMU emulation


timsifive:
I've confirmed that this change fixes the debug issues I reported in https://github.com/riscv/riscv-isa-sim/commit/457f628ad78efa305283ae6cc4dc1530fceffbe4#r45151877

aswaterman:
Thanks y'all.

PR #619, opened 2020-12-18
Update config file to support aarch64

mehmetoguzderin:
I am not sure if pulling the latest version of these files can cause any regression but this fixed build issues for me on aarch64 systems.


aswaterman:
Still works on x86-64 Linux.  Thanks.

PR #620, opened 2020-12-18
rvv: make fractional lmul checking simpler and stricter

chihminchao:

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -396,8 +396,7 @@ reg_t processor_t::vectorUnit_t::set_vl(int rd, int rs1, reg_t reqVL, reg_t newT
     vediv = 1 << extract64(newType, 8, 2);
 
     vill = !(vflmul >= 0.125 && vflmul <= 8)
-           || vsew > ELEN
-           || vflmul < ((float)vsew_min / ELEN)
+           || vsew > std::min(vflmul, 1.0f) * ELEN
            || vediv != 1
            || (newType >> 8) != 0;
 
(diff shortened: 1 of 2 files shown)


riscv/processor.cc
```diff
@@ -395,9 +395,8 @@ reg_t processor_t::vectorUnit_t::set_vl(int rd, int rs1, reg_t reqVL, reg_t newT
     vma = extract64(newType, 7, 1);
     vediv = 1 << extract64(newType, 8, 2);
 
-    vill = !(vflmul >= 0.125 && vflmul <= 8)
-           || vsew > ELEN
-           || vflmul < ((float)vsew_min / ELEN)
+    vill = vflmul > 8
+           || vsew > std::min(vflmul, 1.0f) * ELEN
```
aswaterman, line 399: I think you need to restore the `vfmul >= 0.125` clause.  Once Spike supports ELEN=128, this check is not sufficient, because it would permit SEW=8 and LMUL=1/16, which is not a valid setting.

chihminchao: The  PR has been updated.

chihminchao:
@aswaterman  It is ok to merge.

sequencer:
Is `vsew > std::min(vflmul, 1.0f) * ELEN` defined anywhere in the specification?

aswaterman:
It's permitted by the specification, but it isn't mandated.  https://github.com/riscv/riscv-isa-manual/blob/8b9dc50e562af27c9ab22ae0a2d3bf528c6bd41d/src/v-st-ext.adoc#L292-L293

Spike is deliberately strict about this case, to prevent the case that people accidentally develop software that relies on `ELEN < SEW * ceil(LMUL) <= VLEN` being supported.  (We aren't going to change Spike to support the hypothetical extension you were proposing on riscv-isa-manual; for now, you'll need to fork Spike if you want to support this case.)

PR #623, opened 2020-12-28
Move `byteorder` to utils directory (fix #622)

### LINKED ISSUE #622: Big-endian supports breaks `fesvr` library support
When compiling against `libfesvr` the big-endian support (#575) breaks the `fesvr` library functionality. In particular, the header file `memif.h` depends on `byteorder.h`:

https://github.com/riscv/riscv-isa-sim/blob/032a68c3e94b0d315207a5a42251f836185b5d01/fesvr/memif.h#L8

which is not installed by `make install`. Adding `byteorder.h` to the list of installed header files fails because it relies on `autotool` generated `config.h` which should not be distributed/installed.

zarubaf:
`fesvr` and `riscv` both depend on `byteorder`. I've moved byteorder to
its own subproject which gets installed separately.

Unfortunately, there isn't a unified way of detecting the endianess and
we can't rely on autotools as this would require to install a header
file which depends on `config.h`.

As a work-around I have moved the endianess conversion functions to
their own library which gets statically linked to either spike or
libfesvr.

I haven't measured the performance impact yet and whether LTO can make
up for the loss of inlining. There might be a better way to deal with
this - I would be happy to try other proposed fixes.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -108,6 +108,9 @@
 /* Define to 1 if you have the ANSI C header files. */
 #undef STDC_HEADERS
 
+/* Define if subproject MCPPBS_SPROJ_NORM is enabled */
+#undef UTIL_ENABLED
+
 /* Define WORDS_BIGENDIAN to 1 if your processor stores words with the most
    significant byte first (like Motorola and SPARC, unlike Intel). */
 #if defined AC_APPLE_UNIVERSAL_BUILD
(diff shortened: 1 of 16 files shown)


zarubaf:
Nvm, that doesn't work - it was just linking on my end because I had some old libraries which had the proper definitions.

PR #625, opened 2021-01-11
add command of using yum when building

Xinlong-Wu:
sir, I add a command of using yum when building
I built it on my CentOS computer, but we can't find the software _'device-tree-compiler'_ on yum
cuz _'device-tree-compiler'_  was named as _'dtc'_ . 
So I've added a command to use yum, which I hope will help those who compile it in CentOS in the future, as I did


aswaterman:
Thanks, added to instructions in a slightly different way.

PR #630, opened 2021-01-18
adding spike as a target to the arch-test-framework

neelgala:


neelgala:
@allenjbaum please confirm if this is okay ?

aswaterman:
`arch_test_target/spike/` is a little weird since this is the Spike repo and there won't ever be anything in that subdirectory besides `spike`: shouldn't it just be `arch_test_target/`?

Other than that, I'm fine with it if Allen is.

allenjbaum:
If the subdirectory gets copied to the architectural test clone, it will go
into the spike directory, and there may be other model subdirectories
there, so I wanted to keep the names as similar as possible.
There may be other cases where arch-test has multiple subdirectories for
different versions of the model (don't ask me to name one).
But, not a huge deal either way.

aswaterman:
As long as there’s a reason for it, I’m OK with it. Is it good to merge,
Allen?

aswaterman:
ping @allenjbaum - should I merge this?

allenjbaum:
I think you can. It is backwards compatible - we haven't removed anything
at our end, so nothing will break.
The next steps are for us to do: send email out to
<some-set-of-email-lists> indicating what we are doing, then removing the
directories from our repo, and replacing them with a README that points to
yours.

aswaterman:
Cool.  Thanks.

marcfedorow:
Is this still a necessary part of spike after updating the framework (riscof etc)?

allenjbaum:
There are two cases where it becomes necessary:
1. during test development, we need to compare two models, with Sail as a
reference and some other sim as the Device-Under-Test
2. when support has not been added to the Sail model, but has been added to
Spike, we can use Spike as a reference, instead of the DUT.

There are several holes in this of course:
 - if the support isn't in Sail, how do we develop tests? (answer: with
great difficulty)
 - Spike is not as configurable as Sail (will eventually be), so we may not
actually be able to direct comparisons in many cases.

PR #634, opened 2021-01-21
add support to page on demand

chihminchao:
This patch splites the target-requested memory regions into pages and only
allocates host memory when it is accessed to reduce larget memory sceniaro
in 64 bit target system


aswaterman:
👍

PR #635, opened 2021-01-22
scalar-crypto: Initial spike support for v0.8.1

ben-marshall:
**Brief:** (PR details duplicated from commit message)

- This PR adds spike support for [v0.8.1](https://github.com/riscv/riscv-crypto/releases) of the scalar cryptography extension.  See the [riscv-crypto repository](https://github.com/riscv/riscv-crypto)  for more information on this extension.

- It is based on the experimental patch which has so far been kept in the  riscv-crypto repository. Now that scalar crypto is nearly at the "freeze"  stage and entering opcode consistency review, it makes sense to start  up-streaming our experimental version.

- In terms of compiler support for these new instructions - we are using an experimental patch in the riscv-crypto repository at the moment, others are working on an upstream appropriate version. For now, the experimental toolchain is usable by anyone who needs it.

**Details:**

- Add very basic support for the entropy source (entropy_source.h).  Much of the behaviour of this is implementation specific. This model  aims to provide the bare minimum of functionality which can be used to  quickly develop software. It uses `/dev/urandom` as its entropy source for  now.

- Scalar crypto is unique in that it _borrows_ instructions from the  Bit-manipulation extension. This is currently _not_ encoded in the patch,  as I didn't want to damage anything in Bitmanip which is currently under review. However, I've added a macro in riscv/decode.h called  [`require_either_extension(A,B)`](https://github.com/riscv/riscv-isa-sim/compare/master...ben-marshall:scalar-crypto?expand=1#diff-df9aa186345907cc087df6c6c69bc517a077141880d8d110c6202d3b4a86c9f5R242), which will allow instructions to be valid  opcodes iff they are in one or both extensions. Future commits can add  this to the relevant instructions. My rationale was to get the scalar-crypto specific functionality in first, before modifying bitmanip instructions. If maintainers think it makes sense to include those modifications in this PR, I'm happy adding that in, just let me know.

- I've tested everything by running the set of benchmarks we use for development in the riscv-crypto repository. Architectural tests are high on the list of todos for the task group, which are much easier to develop with an up-streamed Spike.

**Additional:**

I know this is an enormous PR, so apologies for that. There didn't seem like a sensible way to break it into pieces. Of course let me know if more information / context is needed, or indeed any changes / clean-ups to the code. In terms of maintenance going forward, I'm happy being point-person for this code. If maintainers would rather keep this on a separate branch in the main Spike repository (as I think happened with Bitmanip?) that's fine, I just don't have permission to create the branch. Otherwise, it can live in master - like Bitmanip now does.

Many thanks,
Ben


aswaterman:
@ben-marshall I only did a quick skim, but this looks great.  Thanks for the contribution!

PR #636, opened 2021-01-28
Increment minstret when WFI completes

scottj97:
See https://github.com/riscv/riscv-isa-manual/issues/189#issuecomment-768525017

I'm not sure I understand this execute loop, so I'm not 100% confident in this change.


aswaterman:
Looks right to me.

PR #645, opened 2021-02-14
Add Multiple Extension Support

zitaofang:
Hi,

I am an undergraduate student at UC Berkeley's ADEPT lab, and I am currently working on simulating with multiple extensions enabled on spike. Spike currently only allows one extension.

This simple PR replaces the `ext` pointer in `processor_t` with an C++ vector, as well as replacing `rocc_t` abstract class with a set of macros (since `rocc_t` cannot embed the extension name needed for lookup in its callback function). 

Any feedback will be appreciated. Thanks!

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -24,7 +24,7 @@
 processor_t::processor_t(const char* isa, const char* priv, const char* varch,
                          simif_t* sim, uint32_t id, bool halt_on_reset,
                          FILE* log_file)
-  : debug(false), halt_request(HR_NONE), sim(sim), ext(NULL), id(id), xlen(0),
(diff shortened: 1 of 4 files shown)


aswaterman:
The basic idea makes sense, but I'm going to make a few tweaks.  In particular, I see no reason to deprecate the existing API for the common case of a single extension.  No reason to spam existing users with warning messages.  (Along the same lines, we don't really need to issue a warning about opcode conflicts when registering multiple custom extensions, since that caveat already exists for a single custom extension vs. the standard custom extensions.)

I also made some stylistic changes you might want to review.  See 3d19864f39ac525162896ae5025f44834f91c83c

zitaofang:
Sorry for the late response, but the changes you made look good! Thanks for integrating this!

PR #646, opened 2021-02-16
fix require fp since spec said <When V=1, both vsstatus.FS and the HS…

francis4096:
…-level sstatus.FS are in effect>

In spec page 91, When V=1, both vsstatus.FS and the HS-level sstatus.FS are in effect. Attempts to execute a floating-point instruction when either field is 0 (Off) raise an illegal-instruction exception.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -241,7 +241,7 @@ private:
 #define require_extension(s) require(p->supports_extension(s))
 #define require_either_extension(A,B) require(p->supports_extension(A) || p->supports_extension(B));
 #define require_impl(s) require(p->supports_impl(s))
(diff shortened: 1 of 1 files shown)


avpatel:
Please squash the two commits into one commit.

Your second commit fixes your first commit. After squashing the commits in your local branch, you can force push to the branch of this PR and it will automatically update the PR.

Meanwhile, I will debug KVM RISC-V so until then please keep PR on hold.

Regards,
Anup

aswaterman:
@avpatel I’ll just squash & merge this one.

avpatel:
@aswaterman I have fixed the KVM RISC-V issue seen on Spike after this PR.

Here's the PR https://github.com/riscv/riscv-isa-sim/pull/647

aswaterman:
Cool, thanks!

PR #649, opened 2021-02-18
Scalar crypto fixes

ben-marshall:
Hi there

This PR contains fixes for two bugs in the scalar-crypto extension. Both are fixes for the implementation of the extension. They don't reflect bugs in the extension itself.

Details of the changes are the the respective commit messages, but briefly:
- The first fixes a sign extension problem with the RV32 SHA512 instructions.
- The second is more involved, and enables spike to decode instructions based on the current XLEN, which is necessary for some scalar crypto instructions, and likely future extensions as well.

The reason these bugs are being caught now is because we've since moved onto longer and much [more extensive tests](https://github.com/rvkrypto/rvkrypto-fips) than our initial [simple benchmarks.](https://github.com/riscv/riscv-crypto/tree/master/benchmarks) They were classic heisenbugs, where changing what was executed (be it input data, or prior unrelated instruction sequences) changed how/if the bugs manifested.

I understand the decoding change is a big one, and that there might have been a better way to implement it which I'm not aware of. I'm very happy iterating on this PR as needed.

Many thanks,
Ben


aswaterman:
encoding.h is auto-generated by https://github.com/riscv/riscv-opcodes, and I'm loathe to make Spike's encoding.h diverge.  Can you figure out how you'd change that repo so it can emit these new macros?

In the mean time, I'll see if I can think of alternative ways of doing this.  So far, I'm struggling to think of anything that's less disruptive.

ben-marshall:
> encoding.h is auto-generated by https://github.com/riscv/riscv-opcodes, and I'm loathe to make Spike's encoding.h diverge. Can you figure out how you'd change that repo so it can emit these new macros?

Sure. I'll put in a PR to riscv-opcodes and refer to this issue in it.

> In the mean time, I'll see if I can think of alternative ways of doing this. So far, I'm struggling to think of anything that's less disruptive.

Yeah, I appreciate it's a thorny one.

Cheers,
Ben

ben-marshall:
Hi, is there any update on this? I don't mean to hurry, just making sure I have something to say at the next TG meeting. :+1:

aswaterman:
You're welcome to give an update to the TG that you're bottlenecked on my reviewing your riscv-opcodes PR :)

I'll get to it in the next couple days.

PR #655, opened 2021-02-24
Rvv v0.10

chihminchao:


aswaterman:
I recommend renaming `normal_fmt` to `is_mask_ldst` to indicate _why_ it's a special case.

chihminchao:
> I recommend renaming `normal_fmt` to `is_mask_ldst` to indicate _why_ it's a special case.

It has been changed to is_mask_ldst in latest push.

aswaterman:
Thanks!

PR #657, opened 2021-02-25
Fix macro for PKxx16 & PKxx32 commands.

marcfedorow:
Now this correctly works for PKBB32, PKBT32, PKTB32 and PKTT32 (rv64-only).

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -2836,13 +2836,12 @@ for (reg_t i = 0; i < P.VU.vlmax && P.VU.vl != 0; ++i) { \
 #define P_PK(BIT, X, Y) \
   require_extension('P'); \
   require(BIT == e16 || BIT == e32); \
-  reg_t size = BIT / 8 * 2; \
-  reg_t rd_tmp = 0; \
-  type_usew_t<BIT>::type pd[2] = { \
-    P_UFIELD(RS1, Y, BIT), \
-    P_UFIELD(RS1, X, BIT) \
-  }; \
-  memcpy(&rd_tmp, pd, size); \
+  reg_t rd_tmp = 0, rs1 = RS1, rs2 = RS2; \
+    rd_tmp = set_field(rd_tmp, make_mask64(i * 2 * BIT, BIT), \
+      P_UFIELD(RS2, i * 2 + Y, BIT)); \
+    rd_tmp = set_field(rd_tmp, make_mask64((i * 2 + 1) * BIT, BIT), \
+      P_UFIELD(RS1, i * 2 + X, BIT)); \
+  } \
   WRITE_RD(rd_tmp);
 
 #define P_64_PROFILE_BASE() \


marcfedorow:
@hope51607, this patch allows to add [this](https://github.com/riscv/riscv-p-spec/blob/master/P-ext-proposal.adoc#pkbb32-pkbt32-pktt32-pktb32) commands from spec.
I would like you to check this pull request.

marcfedorow:
This macro doesn't actually work for PKxx16 instructions either.
There are two problems:
- Typo in source register
- No RV64 support

The first problem is easy to fix, but the second made me rewrite this macro from scratch.
An illustration of both problems:
pkbb16(0x1234567887654321, 0xaabbccddeeff0011) ==  0x5678ccdd43210011
Your value: 0x0000000043214321
We got 0x43214321 instead of 0x43210011 because you use RS1 instead of RS2.
We got 0x00000000 instead of 0x5678ccdd because there is no RV64 support (no cycle through halves).

The same applies to all other instructions using this macro.

marcfedorow:
Hi, @hope51607, you updated isa-sim so I rebased my commit.
I kindly ask you to check bug fix that I made.

hope51607:
Sorry for my mistake, I didn’t see this PR until today.

riscv/decode.h
```diff
@@ -2836,13 +2836,13 @@ for (reg_t i = 0; i < P.VU.vlmax && P.VU.vl != 0; ++i) { \
...
-    P_UFIELD(RS1, Y, BIT), \
-    P_UFIELD(RS1, X, BIT) \
-  }; \
-  memcpy(&rd_tmp, pd, size); \
+  reg_t rd_tmp = 0, rs1 = RS1, rs2 = RS2; \
+  for (sreg_t i = 0; i < xlen / BIT / 2; i++) { \
+    memcpy((type_usew_t<BIT>::type*)&rd_tmp + i * 2, \
+      (type_usew_t<BIT>::type*)&rs2 + i * 2 + Y, BIT / 8); \
+    memcpy((type_usew_t<BIT>::type*)&rd_tmp + i * 2 + 1, \
+      (type_usew_t<BIT>::type*)&rs1 + i * 2 + X, BIT / 8); \
```
hope51607: I think use `set_field` macro to insert bits is better than cast pointer and use `memcpy`
following code is PoC:
```C++
    rd_tmp = set_field(rd_tmp, make_mask64(i * 2 * BIT, BIT), \
      P_UFIELD(RS2, i * 2 + Y, BIT)); \
    rd_tmp = set_field(rd_tmp, make_mask64((i * 2 + 1) * BIT, BIT), \
      P_UFIELD(RS1, i * 2 + X, BIT));
```

marcfedorow: Hi again, @hope51607! I agree so I updated my commit.

hope51607:
@marcfedorow Thanks, LGTM.

riscv/decode.h
```diff
@@ -2836,13 +2836,12 @@ for (reg_t i = 0; i < P.VU.vlmax && P.VU.vl != 0; ++i) { \
...
   require_extension('P'); \
   require(BIT == e16 || BIT == e32); \
-  reg_t size = BIT / 8 * 2; \
-  reg_t rd_tmp = 0; \
-  type_usew_t<BIT>::type pd[2] = { \
-    P_UFIELD(RS1, Y, BIT), \
-    P_UFIELD(RS1, X, BIT) \
-  }; \
-  memcpy(&rd_tmp, pd, size); \
+  reg_t rd_tmp = 0, rs1 = RS1, rs2 = RS2; \
```
hope51607, line 2839: @marcfedorow After I merged your PR, I found that your commit missed the for expression.
Can you solve this and create new PR?

marcfedorow: Solved

PR #658, opened 2021-02-25
Fix disabling hypervisor via misa

scottj97:
Writing `misa.H=0` should disable hypervisor, and the processor should behave as though it's not implemented.

Today, the new hypervisor state in `hstatus` and `mstatus` retains its value and can affect `mret` and `sret`. Specifically, if `mstatus.MPV=1` then `mret` can enter VS-mode (setting `state.v=1`), which is bad because hypervisor's supposed to be disabled.

I'm not sure if this is complete. I'm still digesting the H spec. There might be more bits that need clearing beyond the few I've added here. What about:
* Any of the other new `h*` CSRs?
* The new bits in `mip`/`mie`?

cc @avpatel

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -1130,11 +1130,17 @@ void processor_t::set_csr(int which, reg_t val)
 
       state.misa = (val & mask) | (state.misa & ~mask);
 
-      // update the forced bits in MIDELEG
+      // update the forced bits in MIDELEG and other CSRs
       if (supports_extension('H'))
         state.mideleg |= MIDELEG_FORCED_MASK;
-      else
+      else {
         state.mideleg &= ~MIDELEG_FORCED_MASK;
+        state.medeleg &= ~hypervisor_exceptions;
+        state.mstatus &= ~(MSTATUS_GVA | MSTATUS_MPV);
+        state.mie &= ~MIP_HS_MASK;  // also takes care of hip, sip, hvip
+        state.mip &= ~MIP_HS_MASK;  // also takes care of hie, sie
+        set_csr(CSR_HSTATUS, 0);
+      }
       break;
     }
     case CSR_HSTATUS: {
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -872,6 +872,13 @@ void processor_t::set_csr(int which, reg_t val)
   reg_t coprocessor_ints = (!custom_extensions.empty()) << IRQ_COP;
   reg_t delegable_ints = supervisor_ints | coprocessor_ints;
   reg_t all_ints = delegable_ints | hypervisor_ints | MIP_MSIP | MIP_MTIP | MIP_MEIP;
+  reg_t hypervisor_exceptions = 0
+    | (1 << CAUSE_VIRTUAL_SUPERVISOR_ECALL)
+    | (1 << CAUSE_FETCH_GUEST_PAGE_FAULT)
+    | (1 << CAUSE_LOAD_GUEST_PAGE_FAULT)
+    | (1 << CAUSE_VIRTUAL_INSTRUCTION)
+    | (1 << CAUSE_STORE_GUEST_PAGE_FAULT)
+    ;
 
   if (which >= CSR_PMPADDR0 && which < CSR_PMPADDR0 + state.max_pmp) {
     // If no PMPs are configured, disallow access to all.  Otherwise, allow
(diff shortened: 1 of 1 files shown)


avpatel:
Looks good to me

aswaterman:
The concept is reasonable, as is the implementation thus far.  However:
- Despite what the spec suggests, I'm unconvinced writable-misa.H is a desirable feature.
- If we go forward with this approach, I'd like to keep updating this PR until it's deemed complete, rather than doing the piecemeal thing.

scottj97:
It seems safest to simply zero out the `h*` registers that have control bits:
* `mstatus.MPV` and `mstatus.GVA` (already in this PR)
* `hstatus` (already in this PR)
* `hip` (which also takes care of the hypervisor bits in `mip`)
* `hie` (which also takes care of the hypervisor bits in `mie`)
* `hvip`
* `hgeip` and `hgeie` (not implemented in Spike)
* Bits in `mideleg` that are hardwired to 1 when H=1 (already in the code)
* Extra bits in `medeleg` that are masked when H=0

These seem unnecessary since they don't apply unless V=1:
* `hedeleg`
* `hideleg`
* `hcounteren`
* `htimedelta[h]`
* `htval`
* `htinst`
* `hgatp`
* All the `v*` regs

Or, we take the easy path, and hardwire `misa.H==1`.

gfavor:
Maybe this is being architecturally picky, but the H extensions says (with my emphasis) that "harts that implement the hypervisor extension are encouraged **not** to hardwire misa[7], so that the extension may be disabled."

gfavor:
And the Priv spec says: "When a standard extension is disabled by clearing its bit in misa, the instructions and CSRs defined or modified by the extension revert to their defined or reserved behaviors as if the extension is not implemented."  Which means trapping H-ext instructions and, I believe, accesses to H-ext CSRs as Illegal Instructions.

Overall, it seems like the issue (or an issue) is whether Spike is allowed to play fast and loose wrt fully implementing the arch spec?  Until (or if) Sail eventually models all this fully, Spike is the reference model for ACT's to run against and for arch compliance to be judged.

gfavor:
Although to be most correct, the Priv spec used to say "When misa[7] is clear, the hart behaves as though this extension were not implemented, and attempts to use hypervisor CSRs or instructions raise an illegal instruction exception."  But that was removed an the currnet "Reserved" wording was left.

Although back then the intent was that "Reserved" means (in this context) that implementations may either raise an illegal-instruction exception or execute an instruction from a non-conforming extension.

gfavor:
So, in short, it seems like Spike chooses to either model the intended arch spec (i.e. takes illegal instruction traps when misa.H=0) or to consciously deviate from the spec (and also ignore the H-ext chapter's recommendation to supprt writeable misa.H).

Also note that if one is accurately modeling things, then setting bits to zero that were introduced by H-ext in existing CSRs, is not the same as making them appear to be hardwired to zero.

scottj97:
Related discussion: [Disabling and re-enabling extensions](https://lists.riscv.org/g/tech-privileged/topic/76739961#324)

Though that has more to do with whether state is kept through a disable-then-reenable sequence. I have no opinion on that matter.

From what I can tell, except for the current issue, Spike already properly considers the H bit and behaves as though the H extension was unimplemented if misa.H=0. Specifically, if the ISA string passed into Spike via cmdline does not include H, everything behaves as though it's simply not there. All that behavior is based on looking at misa.H, which is either writable (if H was in the ISA string), or wired to 0 (if not).

So clearing the state at the time of the misa update is sufficient, given the rest of the code already in place to make H optional, to make sure those bits appear to be hardwired to 0.

Another (more complex) option would be to mask those bits on every access when misa.H is disabled. If misa.H is re-enabled, the previous state will be visible again.

A third option, against spec recommendation but still conformant, is to tie misa.H=1.

What's clear to me is that following a CSRW to misa that disables H, everything should behave as if H had never been enabled. That's not the case today, which is what this PR is attempting to rectify.

gfavor:
As is mentioned in the last post's link, the spec now says "When software enables an extension that was previously disabled, then all state uniquely associated with that extension is unspecified, unless otherwise specified by that extension."  And the H-ext chapter doesn't say anything further on this topic.

gfavor:
Lastly, even if H-ext is not being disabled and then re-enabled, the "enabled and then disabled" case is quite real - since the spec says " At reset, the Extensions field shall contain the **maximal** set of supported extensions, and I shall be selected over E if both are available."

scottj97:
I agree with everything you say, but as I don't believe that contradicts anything I've said, I'm not sure what point you're trying to make.

gfavor:
Sorry, I should have stated that I'm just providing information from the Priv spec as to what it explicitly allows and requires.  It wasn't clear that everyone had searched and found these little sentences spread around in the spec (as well as the "Reserved" intent which was from not-so-public past emails).  I'll leave it to you all to decide what does and doesn't happen for this GH issue.

scottj97:
Ah yes, thank you for the references. I wasn't aware of the "unspecified" nature of disabled-then-reenabled state so that's good to see.

IMO the first option of the [three choices](https://github.com/riscv/riscv-isa-sim/pull/658#issuecomment-786966131) is the best one, since it is the simplest to implement that complies with all requirements and recommendations of all the specs. It's what I've implemented here in this PR, though I believe there's [more to do](https://github.com/riscv/riscv-isa-sim/pull/658#issuecomment-786806525).

gfavor:
I would agree with that choice.

scottj97:
Updated this so it will not clobber the fix from #664.

scottj97:
I've added all the state I [identified here](https://github.com/riscv/riscv-isa-sim/pull/658#issuecomment-786806525) as belonging exclusively to hypervisor.

aswaterman:
I'll wait for @avpatel's confirmation that it's OK to merge this one.

scottj97:
Look closely, because this is not well-tested. :( I just found a bug by inspection and patched in a fix.

scottj97:
@avpatel we would appreciate your feedback

avpatel:
My apology, I thought this was already merged.

I have reviewed it from my side. Looks good to me.

@aswaterman please take this PR.

Regards,
Anup

aswaterman:
So it shall be.

scottj97:
@aswaterman are you waiting for something else before merging?

aswaterman:
No, just forgot to click the button.

PR #659, opened 2021-02-28
Correct RV64 vsstatus.UXL field

scottj97:
It was reading as 0, which is not a legal value.

Spike does not support dynamic multi-width; all *XL fields are wired to 64-bit (=2) when running as RV64.

In mstatus, UXL gets initialized by the call to `set_csr(CSR_MSTATUS)` here in `processor_t::reset()`, so I wait until that's complete and then copy all the visible bits over to the vsstatus reset value. (Though only bit 33, part of UXL, is nonzero.)

cc @avpatel


aswaterman:
LGTM.

PR #663, opened 2021-03-02
Fix AMO guest page fault as store guest fault

francis4096:
AMO guest page fault should be reported as store guest page fault. but current code report as load guest page fault.


scottj97 (approved):

aswaterman:
LGTM.

PR #665, opened 2021-03-04
Fix bug where CSRW to vsstatus would not set SD correctly

scottj97:
If you write e.g. 0x6000 (FS=Dirty) via `csrw vsstatus`, this should also set the SD bit.

Unless `mstatus` already had FS=Dirty, this would not occur.

The priv spec says (under vsstatus): "For example, the value of the HS-level sstatus.FS [aka mstatus.FS] does not affect vsstatus.SD."


scottj97:
There is something still incomplete about this fix. I'm investigating.

aswaterman:
cc @avpatel

PR #667, opened 2021-03-05
Fix bug where hstatus.SPVP was being changed when it should not be

scottj97:
[Spec says](https://github.com/riscv/riscv-isa-manual/blob/0453d462a180927169656e6e3f7faf3042b23e5b/src/hypervisor.tex#L302):

> When V=1 and a trap is taken into HS-mode, bit SPVP (Supervisor Previous Virtual Privilege) is set to the privilege mode at the time of the trap, the same as `sstatus.SPP`. But if V=0 before a trap, SPVP is left unchanged on trap entry.

Spike was mistakenly setting SPVP on any trap into HS-mode, in violation of that second sentence.

cc @avpatel

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -763,7 +763,8 @@ void processor_t::take_trap(trap_t& t, reg_t epc)
     s = set_field(s, MSTATUS_SIE, 0);
     set_csr(CSR_MSTATUS, s);
     s = state.hstatus;
-    s = set_field(s, HSTATUS_SPVP, state.prv);
+    if (curr_virt)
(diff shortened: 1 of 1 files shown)


aswaterman:
Can you remove the comment - it’s basically just saying the same thing the code says clearly, and “the spec says so” is implicitly the case for just about all of the Spike code :-)

scottj97:
Agreed, and fixed

aswaterman:
cool, thanks.

avpatel:
Good catch @scottj97 

This fix looks good to me.

PR #670, opened 2021-03-06
Bug fix: vsstatus should also include XS and UBE fields

scottj97:
The [spec says](https://github.com/riscv/riscv-isa-manual/blob/0453d462a180927169656e6e3f7faf3042b23e5b/src/hypervisor.tex#L1247-L1252) the XS and UBE fields are part of `vsstatus`.

### DIFF
--- a/riscv/encoding.h
+++ b/riscv/encoding.h
@@ -54,9 +54,9 @@
 #define SSTATUS_UXL         0x0000000300000000
 #define SSTATUS64_SD        0x8000000000000000
 
-#define SSTATUS_VS_MASK     (SSTATUS_SIE | SSTATUS_SPIE | \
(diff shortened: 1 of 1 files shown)


scottj97:
cc @avpatel

scottj97:
On second thought, I don't think this change is appropriate. XS and UBE fields are not writable in Spike's sstatus, so they should not be in this mask. (Right?)

aswaterman:
Yeah, excluding them from the mask sounds right to me.

avpatel:
I had not included this bits in mask because Spike does not implement it currently.

scottj97:
It looks like Spike supports a fixed UBE=1 if built with `--enable-dual-endian` and run on a big-endian ELF.

I think there's still a bug here, but this change is not the right way to fix it.

PR #671, opened 2021-03-07
Forbid `csrw vsstatus` from modifying the UXL field

scottj97:
Since this is not modifiable in the real `sstatus`, so it should not be in the virtualized version either.

Code is based on similar code for `mstatus`.

I'm working on sharing the sstatus-related code between `mstatus` and `vsstatus` to avoid this duplication, but that's a much bigger change for another PR.


scottj97:
cc @avpatel

Issue #672, opened 2021-03-10
WFI continues without any interrupt


IanJiangICT:
Description: 
1. Core 0 halts at a WFI
1. Core 1 takes turn
1. Core 0 continues without any interrupt

Note: Parameter INTERLEAVE of Spike is changed from 5000 to 16 to make the running shorter.

Run log
```
$ spike -l -p2 test-wfi 2> run-spike.log
$ less run-spike.log
warning: tohost and fromhost symbols not in ELF; can't communicate with target
core   0: 0x0000000000001000 (0x00000297) auipc   t0, 0x0
core   0: 0x0000000000001004 (0x02028593) addi    a1, t0, 32
core   0: 0x0000000000001008 (0xf1402573) csrr    a0, mhartid
core   0: 0x000000000000100c (0x0182b283) ld      t0, 24(t0)
core   0: 0x0000000000001010 (0x00028067) jr      t0
core   0: >>>>  
core   0: 0x0000000080000000 (0x300022f3) csrr    t0, mstatus
core   0: 0x0000000080000004 (0x0082e293) ori     t0, t0, 8
core   0: 0x0000000080000008 (0x30029073) csrw    mstatus, t0
core   0: >>>>  switch_thread
core   0: 0x000000008000000c (0xf14022f3) csrr    t0, mhartid
core   0: 0x0000000080000010 (0x02028763) beqz    t0, pc + 46
core   0: >>>>  thread_master
core   0: 0x000000008000003e (0x00028293) mv      t0, t0
core   0: 0x0000000080000042 (0x10500073) wfi
core   0: 3 0x0000000080000042 (0x10500073)
core   1: 0x0000000000001000 (0x00000297) auipc   t0, 0x0
core   1: 0x0000000000001004 (0x02028593) addi    a1, t0, 32
core   1: 0x0000000000001008 (0xf1402573) csrr    a0, mhartid
core   1: 0x000000000000100c (0x0182b283) ld      t0, 24(t0)
core   1: 0x0000000000001010 (0x00028067) jr      t0
core   1: >>>>  
core   1: 0x0000000080000000 (0x300022f3) csrr    t0, mstatus
core   1: 0x0000000080000004 (0x0082e293) ori     t0, t0, 8
core   1: 0x0000000080000008 (0x30029073) csrw    mstatus, t0
core   1: >>>>  switch_thread
core   1: 0x000000008000000c (0xf14022f3) csrr    t0, mhartid
[...]

PR #674, opened 2021-03-11
Fix WFI: Resume only on pending interrupt (#672)

IanJiangICT:
The origin Spike resumes WFI without checking interrupt state. (See
https://github.com/riscv/riscv-isa-sim/issues/672)
This patch fix the issue.

Besides, a simple test and the test result are provided at
https://github.com/IanJiangICT/instruction-test/blob/master/riscv/test-wfi.S
https://github.com/IanJiangICT/instruction-test/blob/master/riscv/test-wfi-log-pass.txt

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -252,6 +252,14 @@ void processor_t::step(size_t n)
 
     try
     {
+      if (halt_wfi)
+      {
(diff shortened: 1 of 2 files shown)


IanJiangICT:
Closed because
https://github.com/riscv/riscv-isa-sim/issues/672 
is not a bug, but an implementation choice.

PR #678, opened 2021-03-18
rvp: fix macro for PKxx16 & PKxx32 commands. 

marcfedorow:


marcfedorow:
Hi, @hope51607, this is a fix from #657

PR #680, opened 2021-03-24
HS-level interrupts should always be enabled when in VS-mode

scottj97:
See https://github.com/riscv/riscv-isa-manual/issues/633


aswaterman:
Looks right, thanks.

PR #681, opened 2021-03-25
Add `statx` syscall to Spike

huaixv:
The `statx` syscall is required to run some 32bit programs on Spike and `riscv-pk`, so i tried to add it to the simulator. :)

Please let me know if there exist some problems with the code.

### DIFF
--- a/configure
+++ b/configure
@@ -1748,6 +1748,63 @@ fi
   as_fn_set_status $ac_retval
 
 } # ac_fn_cxx_try_link
+
+# ac_fn_cxx_check_member LINENO AGGR MEMBER VAR INCLUDES
+# ------------------------------------------------------
+# Tries to find if the field MEMBER exists in type AGGR, after including
+# INCLUDES, setting cache variable VAR accordingly.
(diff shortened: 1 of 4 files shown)


huaixv:
Just found that `stx_mnt_id` field in `struct statx` was introduced in linux v5.8, but Ubuntu 20.04 is using v5.4.
So I added another `AC_CHECK_MEMBER` call to check its existence, and it builds well.

aswaterman:
LGTM.

aswaterman:
Actually, it does not LGTM :) it fails to build on Ubuntu 16.04.

aswaterman:
I added another configure check for the statx structure overall, which fixes the Ubuntu 16.04 build.  You should check that you're still getting expected results on newer OSes.

huaixv:
Thank you for your response.
But the kernel header defines inode member as `stx_ino` instead of `ino`, so the configure script will not define `HAVE_STATX`. :(

aswaterman:
OK, I pushed yet another fix.

Issue #682, opened 2021-03-25
Sign extension of xperm insns


marcfedorow:
https://github.com/riscv/riscv-isa-sim/blob/634b7eb46e4d1b38f280ffeec9f0ebd8b579c17f/riscv/insns/xperm_b.h#L2

A problem occurs when simulating 32 bit registers:
```asm
li t1 -1;           # 0xffffffffffffffff is stored in t1
li t2 0;            # 0x0000000000000000 is stored in t2
xperm.b t0, t1, t2; # 0x00000000ffffffff is stored in t0
li t1 0xffffffff;   # 0xffffffffffffffff is stored in t1
bne t0, t1, fail;   # jumps to fail
```
Same problem could be observed on some p-extension insns.
This happens because 32-bit registers are stored in 64-bit variables.

I think this problem can be fixed in one of the following ways:
- Every insn's output should be sign-extended (my current way)
- Every insn's output should be zero-extended
- High-order bits should be ignored by ```beq```, ```bne```, etc.
The last option could be implemented easily but such a decision will probably have a bad effect on the performance.

Please clarify which path is preferable -- or what is my mistake.

aswaterman:
This appears to be a bug in the implementation of xperm.n, xperm.b, and xperm.h; they should call `sext_xlen` to sign-extend when in RV32 mode.  I'll push a fix.

cc @cliffordwolf

aswaterman:
Fixed in 4403faee253339c88d077abf6ef7539febad3e30

PR #685, opened 2021-03-26
rvp: sign-extension for p-ext insns

marcfedorow:
Hi @hope51607! Recently I detected some bug in rvb insns (check: https://github.com/riscv/riscv-isa-sim/issues/682).
It seems like bitrev[i] insns and some macros have the same problem.
I kindly ask you to check if there are any other not sign-extended rvp-insns.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -2770,7 +2770,7 @@ for (reg_t i = 0; i < P.VU.vlmax && P.VU.vl != 0; ++i) { \
 
 #define P_LOOP_END() \
   } \
-  WRITE_RD(rd_tmp);
+  WRITE_RD(sext_xlen(rd_tmp));
 
 #define P_PAIR_LOOP_END() \
   } \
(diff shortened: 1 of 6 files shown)


aswaterman:
@marcfedorow can you remove the unrelated whitespace changes, please?

marcfedorow:
Done. Thank you for review, @aswaterman

hope51607:
@marcfedorow thank you for debug. I will check and push a fix.

hope51607:
@marcfedorow thank you for debugging.
Can you fix the sign-extension problem of the following instructions incidentally? 
```
ave
sra.u
srai.u
sraiw.u
```

marcfedorow:
Hi, @hope51607!
Thank you very much for finding other non-sign-extended insns.
I added fixes for `ave` and `sra[i].u` insns.
**I did not add a fix** for `sraiw.u` insn because I think it is unnecessary since it is an rv64-only insn so it doesn't have a problem with comparing 32-bit registers as 64-bit values.
Actually `sraiw.u.h` code contains a call of `sext32` macro which sign-extends 32 bits to 64 due to specification of this insn.
If you are still convinced it is necessary I believe you can fix it by yourself because this does not apply to the issue described in #682.

hope51607:
@marcfedorow you are right.
Thanks again for debugging!

PR #687, opened 2021-03-30
scalar-crypto: Add architectural test suite support.

ben-marshall:
(Duplicated from commit message)

- This commit adds "K" extension support for the architectural test suite.  It indicates to the test framework (riscv/riscv-arch-test) that Spike supports  the K extension.

- All of the tests pass, and agree with the Imperas OVPSim. The test were merged  very recently as part of riscv/riscv-arch-test#177

- There are two caveates:

  - The toolchain tripple is set to "riscv64-*" for both RV32 and RV64. This is    because right now, that's how the experimental toolchain with crypto support    is set to build. It will be update in future as the upstream support being    built by the PLCT lab matures.

  - The Spike ISA string is set to "rv[32|64]ibk". This is overly permissive    and is because correctly enabling certain instructions based on the K    extension is not *yet* implemented in spike. It will be fixed once the    instruction feature groups/names are finalised after opcode and consistency    review.


jjscheel:
@ben-marshall, @aswaterman, is this a stale PR or is there something still yet to be integrated?

aswaterman:
Not sure. @ben-marshall will have to let me know if this is still ready to be merged.

ben-marshall:
Oh goodness. I'm afraid my memory doesn't stretch this far. The purpose was to support the _then_ framework for integrating with the Arch tests. I believe the arch tests have moved on considerably since when this was raised (RISCoF etc) so it's likely this is out of date. I suggest it can be closed without merging.

aswaterman:
OK. Closing.

PR #692, opened 2021-04-11
Add `--dtb` option to help message

Stupremee:
Currently, the `--dtb` option is available as an option, but is not shown inside the help message.


mbgg:
Aready implemented through
[1df65613 ("Add missing description of --dtb in --help message")](https://github.com/riscv-software-src/riscv-isa-sim/commit/1df65613)

mbgg (changes requested):
PR should be dropped

PR #693, opened 2021-04-13
Display 32 bits

emelcher:
This PR addresses the following bugs:

- Even when using configure `--with-isa=rv32i --target=riscv32-unknown-elf`, pc and registers are shown as 64-bit values.
 
This PR shows only the lower 32 bits of pc and registers if max_xlen==32.

- The `until` command uses 64 bits for rv32. To stop correctly, for example, `until pc 0 0xffffffff80000000` must be used.

This PR compares only the lower 32 bits if max_xlen==32, so `until pc 0 0x80000000` works as expected.

- `pk` must be installed manually in `riscv64-unknown-elf/bin/` for `spike pk` to find it.

This PR makes spike look for pk in `riscv32-unknown-elf/bin/`. However, it does so unconditionally. Configure using `--target=riscv64-unknown-elf` will certainly fail to work correctly. If possible, please advise how to adjust the spike search path depending on `--target`.

This PR has only been tested for
`
../configure --prefix=$RISCV --with-isa=rv32i --target=riscv32-unknown-elf --enable-histogram
`

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -184,8 +184,8 @@ void sim_t::interactive_pc(const std::string& cmd, const std::vector<std::string
 
   processor_t *p = get_core(args[0]);
   int max_xlen = p->get_max_xlen();
-  fprintf(stderr, max_xlen==32 ? "0x%08" PRIx32 "\n" :
-                                 "0x%016" PRIx64 "\n", get_pc(args));
+  fprintf(stderr, max_xlen==32 ? "0x%08" PRIx64 "\n" :
+                                 "0x%016" PRIx64 "\n", ERASE_32MSB(max_xlen,get_pc(args)));
 }
 
 reg_t sim_t::get_reg(const std::vector<std::string>& args)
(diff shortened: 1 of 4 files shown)


riscv/interactive.cc
```diff
@@ -179,7 +179,13 @@ reg_t sim_t::get_pc(const std::vector<std::string>& args)
 
 void sim_t::interactive_pc(const std::string& cmd, const std::vector<std::string>& args)
 {
-  fprintf(stderr, "0x%016" PRIx64 "\n", get_pc(args));
+  if(args.size() != 1)
+    throw trap_interactive();
+
+  processor_t *p = get_core(args[0]);
+  int max_xlen = p->get_max_xlen();
+  fprintf(stderr, max_xlen==32 ? "0x%08" PRIx32 "\n" :
```
aswaterman: This should be `PRIx64` even for RV32, since `get_pc` returns a `reg_t` a.k.a. `uint64_t` regardless of `max_xlen`.

emelcher: Changed all `PRIx32` I added to `PRIx64`.

Reverted the definition of `TARGET_ARCH` back to what it was because the problem can be solved by adding `CXXFLAGS=\\\"-DTARGET_ARCH=riscv32-unknown-elf\\\""` to the configure command.

Now I am investigating other bugs that have shown up in spike 32-bit. They are related to memory.

emelcher: All detected bugs corrected.

aswaterman:
OK, thanks.

riscv/interactive.cc
emelcher: There is a mistake in my last comment:.

he correct addition to the configure command is `CXXFLAGS="-DTARGET_ARCH=\\\"riscv32-unknown-elf\\\""`

PR #694, opened 2021-04-13
rvp: insns fixes

marcfedorow:
* Fixed uclipNN insns: sh >> 64 is an UB. Changed loop while P-spec says uclipNN insns work with **signed** packed values.
* Added missing OV to a saturating macro.
* Added missing sext_xlen to insb and bpick.
* Fixed P_REDUCTION_ULOOP_BASE macro so in works on all its insns but not only one.
* Rewrited P_REDUCTION_LOOP_BASE similar to P_REDUCTION_ULOOP_BASE.

Hi, @hope51607! I kindly ask you to check this patch.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -2444,6 +2444,7 @@ for (reg_t i = 0; i < P.VU.vlmax && P.VU.vl != 0; ++i) { \
     R = UINT##BIT##_MAX; \
     P_SET_OV(1); \
   } else if (R < 0) { \
+    P_SET_OV(1); \
     R = 0; \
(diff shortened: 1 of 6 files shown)


hope51607:
Hi again @marcfedorow! I agree with these.
I also found that `P_I_SULOOP` `P_ONE_SULOOP_BODY` macro is no longer used.
Can you remove them incidentally? Thanks.

marcfedorow:
> Can you remove them incidentally?

Done.

hope51607:
Thanks.

PR #697, opened 2021-04-15
Implement JTAG BYPASS register.

timsifive:
This allows spike to be put into a virtual scan chain with other
remote_bitbang JTAG devices.

### DIFF
--- a/riscv/jtag_dtm.cc
+++ b/riscv/jtag_dtm.cc
@@ -44,6 +44,7 @@ jtag_dtm_t::jtag_dtm_t(debug_module_t *dm, unsigned required_rti_cycles) :
   _tck(false), _tms(false), _tdi(false), _tdo(false),
   dtmcontrol((abits << DTM_DTMCS_ABITS_OFFSET) | 1),
   dmi(DMI_OP_STATUS_SUCCESS << DTM_DMI_OP_OFFSET),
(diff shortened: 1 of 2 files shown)


GregSavin (approved):
I see that this change allows the TAP to accept a BYPASS instruction and link it to the single bit bypass register.   Looks right to me. Is this an intermediate step toward allowing multiple jtag_dtm_t instances to be chained at runtime, with one instance shifting the LSB of IR or DR (when leaving or reentering *_SHIFT state) to the TDI of the next jtag_dtm_t instance in a chain (if there is a next instance)?  Or is any inter-TAP data flow routing implemented on the client side of the remote bitbang protocol, rather than in Spike?

timsifive:
Right. For spike this is the only step. I have an external program that links multiple spike instances together. (Doing it inside spike would be faster, but possibly more complex, and then I still can't test heterogeneous cores together because spike doesn't support that.)

GregSavin:
Oh right, I see.  Doing the inter-TAP TDO->TDI routing on the bitbang client side gives more flexibility; and leaves the door open for heterogeneous mix-and-match, even beyond Spike (could even mix-and-match Spike instances with SystemC/Verilator based bitbang nodes too as long as they support bypass instruction too)

PR #699, opened 2021-04-22
Misc fix 2021 04 21

chihminchao:


chihminchao:
@ben-marshall  I have update the crypto opcode from riscv-opcodes and needs your review if it keep all thing sound

ben-marshall:
It _looks_ okay to me, though I might not get a chance to run it for a while. Does your change pass the WiP [scalar crypto compliance tests](https://github.com/riscv/riscv-arch-test/) (See the `K_unratified` directories under the `test-suite`)?

What was the motivation for the change? It doesn't look like anything _functional_ related to the opcodes has changed, it was just an update from the latest riscv-opcodes repo right?

chihminchao:
I dont' run the scalar crypto compliance tests but I will take a try.  The motivation is to make encoding.h generated from and synced with riscv-opcodes.  I just generate from latest riscv-opcodes and that can avoid lots of unexpected difference when we add new extension opcodes.

chihminchao:
@ben-marshall Where could I get the toolchain or binutils for K-extension ?

ben-marshall:
There is an experimental one in the riscv/riscv-crypto repository. Alternatively, there is the "up-streamable" version  (which will eventually end up in the official GCC repo in development by the PLTC group which you can get from [here](https://github.com/plctlab/riscv-gnu-toolchain).

chihminchao:
@ben-marshall
I don't have time to test the k-extension and decide to remove the opcode part change to make it easier.

@aswaterman
Please take a look

riscv/encoding.h
```diff
@@ -227,7 +221,7 @@
 #define PTE_A     0x040 /* Accessed */
 #define PTE_D     0x080 /* Dirty */
 #define PTE_SOFT  0x300 /* Reserved for Software */
-#define PTE_N     0x4000000000000000 /* Svnapot: NAPOT translation contiguity */
+#define PTE_N     0x4000000000000000 /* Zsn: NAPOT translation contiguity */
```
aswaterman: This looks like an unintentional change.  `Svnapot` is correct.

ben-marshall:
It's probably a good thing you reverted the opcodes change. On @aswaterman's advice some of them are about to change anyway. I'll be putting in a separate PR across riscv-opcodes and riscv-isa-sim in the next few days.

aswaterman:
So... OK to merge?

ben-marshall:
If you were asking me then yes sure.

PR #706, opened 2021-05-01
Fix compiler warning

jmonesti:
At compile time, gcc complains with:
../riscv/processor.cc:787:94: warning: format ‘%x’ expects argument of type ‘unsigned int’, but argument 5 has type ‘uint64_t {aka long unsigned int}’ [-Wformat=]

The variable 'bits' is an uint64_t, so that PRIx64 should be used to print it out.

Imho, pull request 'Display 32 bits #693' should have taken advantage of the printf specifier * to format pc, addresses, etc...
instead of this awkward serie of 
    fprintf(stderr, max_xlen==32 ? "fmt32" : "fmt64", ... );

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -782,8 +782,8 @@ void processor_t::disasm(insn_t insn)
       fprintf(log_file, "core %3d: Executed %" PRIx64 " times\n", id, executions);
     }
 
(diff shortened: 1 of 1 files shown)


aswaterman:
@jmonesti I agree, the two-format-string approach is awkward and hard to maintain.  I'll fix that in a follow-up PR.

PR #713, opened 2021-05-25
Add alignment check for lr instruction

zhongchengyong:
Alignment check is privided for sc_w and sc_d, but missing for lr instruction.
The spec says:
> For **LR and SC,** the A extension requires that the address held in {rs1} be naturally aligned to the size of the operand (i.e.,eight-byte aligned for 64-bit words and four-byte aligned for 32-bitwords).  If the address is not naturally aligned, an address-misalignedexception or an access-fault exception will be generated.

### DIFF
--- a/riscv/insns/lr_d.h
+++ b/riscv/insns/lr_d.h
@@ -1,5 +1,5 @@
 require_extension('A');
 require_rv64;
 auto res = MMU.load_int64(RS1, true);
-MMU.acquire_load_reservation(RS1);
+MMU.acquire_load_reservation(RS1, 8);
 WRITE_RD(res);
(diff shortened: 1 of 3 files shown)


aswaterman:
If misaligned accesses are disabled (as is the default), this behavior will occur anyway.  So I guess you are using a nonstandard configuration.

scottj97:
This seems redundant to me. How is this different than #592?

aswaterman:
Hmm. I think you’re right. If you’ve convinced yourself that’s true, can you submit a PR that reverts this PR?

scottj97:
@zhongchengyong can you explain how/why this was not already Spike's behavior (since #592)? Did I miss some corner case? Or perhaps you observed the issue on an older build of Spike that did not have this fix?

zhongchengyong:
@scottj97 Oops! you are right, it is redundant. I have submit a PR #728 , please review the PR, thanks very much.

aswaterman:
I reviewed and merged the revert. Thanks!

PR #721, opened 2021-06-02
Remove Duff's Device in main simulation loop

aswaterman:
With recent compilers on recent computers, this much simpler version of the code is actually slightly faster.  I suspect, but haven't proven, that more accurate indirect jump prediction is the main explanation.  Reduced I$ pressure might be a secondary factor.


riscv/execute.cc
```diff
@@ -280,49 +281,17 @@ void processor_t::step(size_t n)
...
-          if (unlikely(instret+1 == n)) break; \
-          instret++; \
-          state.pc = pc; \
-        }
-
-        // This switch statement implements the modified Duff's device as
-        // explained above.
-        switch (idx) {
-          // "icache.h" is generated by the gen_icache script
-          #include "icache.h"
```
David-Horner, line 325: Even I can understand the simplified code.
The old adage of nothing more optimizing than elimination.
glad it works out well here.
Thanks.

aswaterman: Useful feedback, thanks :)

PR #724, opened 2021-06-07
PTE N bit moved from bit 62 to bit 63

daniellustig:
Updated spec draft available here:

https://github.com/riscv/virtual-memory/blob/main/specs/628-Svnapot.pdf


aswaterman:
Thanks.  I've synced this change with the riscv-opcodes repo.

PR #726, opened 2021-06-09
Apply scounteren only if S-mode is supported

Shai-Aviv:
When U-mode is enabled but S-mode is disabled, `mcounteren` should control the availability of the hardware performance-monitoring counters in U-mode, and `scounteren` should be ignored.
The current implementation (incorrectly) raises an illegal instruction trap when reading performance-monitoring counters in U-mode while S-mode is disabled, even though the counters are enabled in `mcounteren`.


chihminchao (approved):
it matchs what spec says.

aswaterman:
Agreed, thanks.

PR #727, opened 2021-06-12
Implement --dynamic-endian, which makes MBE/SBE/UBE writable

zeldin:


aswaterman:
I’ll let this marinate for a while. I don’t have time to perform a detailed code review, but hopefully others will.

riscv/processor.cc
```diff
@@ -922,11 +942,25 @@ void processor_t::set_csr(int which, reg_t val)
       VU.vxsat = (val & VCSR_VXSAT) >> VCSR_VXSAT_SHIFT;
       VU.vxrm = (val & VCSR_VXRM) >> VCSR_VXRM_SHIFT;
       break;
-    case CSR_MSTATUS: {
+    case CSR_MSTATUSH:
+      {
+	reg_t mask = MSTATUS_SBE | MSTATUS_MBE;
```
scottj97: Indenting doesn't match

riscv/processor.cc
```diff
@@ -922,11 +942,25 @@ void processor_t::set_csr(int which, reg_t val)
...
+	if (xlen != 32)
+	  break;
+	val = (state.mstatus & ~mask) | ((val << 32) & mask);
+      }
+      if (0) {
```
scottj97: I'm not sure I understand the purpose of this, but it's nasty regardless

riscv/processor.cc
```diff
@@ -922,11 +942,25 @@ void processor_t::set_csr(int which, reg_t val)
...
+	/* FALLTHROUGH */
+    case CSR_MSTATUS:
+        if (xlen == 32)
+	  val = ((state.mstatus >> 32) << 32) | (uint32_t)val;
```
scottj97: Indenting

riscv/processor.cc
```diff
@@ -963,10 +998,17 @@ void processor_t::set_csr(int which, reg_t val)
 
       if (supports_extension('U'))
         state.mstatus = set_field(state.mstatus, MSTATUS_UXL, xlen_to_uxl(max_xlen));
+      else
+	state.mstatus = set_field(state.mstatus, MSTATUS_UBE, 0);
```
scottj97: Indenting

riscv/processor.cc
```diff
@@ -963,10 +998,17 @@ void processor_t::set_csr(int which, reg_t val)
...
       if (supports_extension('S'))
         state.mstatus = set_field(state.mstatus, MSTATUS_SXL, xlen_to_uxl(max_xlen));
+      else
+	state.mstatus = set_field(state.mstatus, MSTATUS_SBE, 0);
```
scottj97: Indenting

riscv/processor.cc
```diff
@@ -963,10 +998,17 @@ void processor_t::set_csr(int which, reg_t val)
...
       // U-XLEN == S-XLEN == M-XLEN
       xlen = max_xlen;
+
+      if (dynamic_endian)
+	change_endianness(state.prv);
```
scottj97: Indenting

scottj97 (changes requested):
Nice work. I have not compiled or tested this but it looks generally good, with only a few nits I left comments about.

I would prefer to see this submitted as a series of commits, instead of one monolithic commit, but it's not worth going back to redo it now.

riscv/processor.cc
zeldin: Ok, I'll fix.

riscv/processor.cc
zeldin: The purpose is that in 32-bit mode writes to mstatus/mstatush should take either the upper or the lower 32 bits from the existing value of state.mstatus (which is always 64 bits), and then fall through to the code which handles 64 bit mode.  Thus the `if(0)` is needed to prevent the fallthrough from executing the first bit of the `CSR_MSTATUS` case, which is for writing mstatus in 32-bit mode.

riscv/processor.cc
zeldin: Not sure what you would like here.  The code is inside an `if()` block (even though it is reached without executing the actual `if`statement), so I indented two steps for that, which is what the codebase does in other places with `if`s.  Is it wrong here?

zeldin:
@scottj97 I can split the commit if you like, it's not that much work.  But following the principle of each commit should be buildable on its own I think it's basically the changes to mmu.c and mmu.h that would go in a separate commit, and the rest of the stuff in one other, as it's pretty intermingled.
The change to `sim_t::set_rom` could also be split out, but it doesn't make all that much sense on its own unless it's put after the reset change, but then things would be broken in between those commits.

riscv/processor.cc
scottj97: Line 956 is inside the if() of line 955 so it should be indented more than line 955

riscv/processor.cc
scottj97: OK I understand, but this is still nasty to have if() and case intermixed like this. The language allows it, but I'd say that's a bug in the language.

riscv/processor.cc
zeldin: Ah, it is actually, but there is an illusion in the patch due to line 956 using a TAB character first on the line.  I'll run `M-x untabify` on the file to sort it out.

scottj97:
> @scottj97 I can split the commit if you like, it's not that much work. But following the principle of each commit should be buildable on its own I think it's basically the changes to mmu.c and mmu.h that would go in a separate commit, and the rest of the stuff in one other, as it's pretty intermingled.

Agreed that each commit must be buildable on its own. But I can see this being implemented in many commits. For example:

1. Separate PTE endianness conversion from non-PTE
2. Add --dynamic-endian cmdline arg (that has no effect)
3. Pass in dynamic_endian param to sim_t
4. Compute endianness based on mstatus *BE fields
5. Enable CSRW to mstatus *BE fields

But again it's probably not worth going back to rebase that.

riscv/processor.cc
zeldin: So what would you prefer?   Duplicating the logic from MSTATUS (but removing the handling of changes that can't happen from writing MSTATUSH only)?  Breaking out the code from MSTATUS into a method and calling it from both cases?

zeldin:
New version force-pushed.  Changes from before:
 * Ran `M-x untabify` on all the files :smiley_cat: 
 * Split the commit into several
 * Removed the `if(0)` trick.  I opted for the solution with the mstatus write logic broken out into a separate method as I forsee this being less of a maintainability problem the next time something new is added to `mstatush` / upper 32 bits of `mstatus`.
 * Added an `assert` that you don't try to create a dynamic_endian `processor_t` if dual endian support is not compiled in.

scottj97 (approved):
Excellent; this PR is much better now.

I've reviewed the code but haven't actually tested anything.

aswaterman:
Thanks again for the contribution @zeldin and for the code review, @scottj97.

@chihminchao want to take a look before I merge?

scottj97:
@zeldin I'm afraid that after #793 and family, this PR is going to require significant rework.

zeldin:
@scottj97 Thanks for the heads-up.  I don't have time to work on this right now, maybe after the Holidays.

PR #728, opened 2021-06-13
Revert the redundant check for lr instruction

zhongchengyong:
As @scottj97 says in #713 , it's redundant since alignment check has been done in load_func.
@scottj97 please check it, thanks.


aswaterman:
Thanks!

PR #733, opened 2021-06-28
fix(aes64ks1i) : fix note in riscv/insns/aes64ks1i.h

dingiso:
the note in aes64ks1i should be " rotate right by 8"


aswaterman:
Thanks!

PR #739, opened 2021-07-08
rvk: Sign-extension for sha256 insns

marcfedorow:
See #682 or #685 -- k-ext insns have the same problem.
There are a few fixes -- I kindly ask to check all the not-RV64-only insns.

### DIFF
--- a/riscv/insns/sha256sig0.h
+++ b/riscv/insns/sha256sig0.h
@@ -6,7 +6,7 @@ require_extension('K');
 uint32_t a = RS1;
 
 WRITE_RD(
-    ROR32(a, 7) ^ ROR32(a,18) ^ (a >> 3)
+    sext_xlen(ROR32(a, 7) ^ ROR32(a,18) ^ (a >> 3))
 );
 
 #undef ROR32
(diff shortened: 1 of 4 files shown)


aswaterman:
Has the spec been updated accordingly yet?

marcfedorow:
> Has the spec been updated accordingly yet?

This it not an issue of the spec but of the spike itself.
Kindly ask you to check #682 for the clarification.
TL;DR: on RV32 spike compares the whole 64-bit register with no sign extension so if some instruction's output was not sign extended (i.e. zero extended) but another (let's say `li`'s) was, spike thinks that e.g. `0x00000000FFFFFFFF != 0xFFFFFFFFFFFFFFFF` while actually only 32 LSBs should be checked.
According to #682 we agreed that every insns's output should be sign-extended to avoid such type of errors.

aswaterman:
Indeed. cc @ben-marshall

ben-marshall:
Thanks for the ping about this. Sign extension of the 32-bit SHA/AES/SM3/4 instructions in the spec is on my todo list, pending any more feedback from the architecture review. Thanks @marcfedorow for catching this issue in Spike!

PR #742, opened 2021-07-09
rvk: Sign-extension for aes32 insns

marcfedorow:
This fixes the same problem as #739  but for aes* insns.

### DIFF
--- a/riscv/insns/aes32dsi.h
+++ b/riscv/insns/aes32dsi.h
@@ -12,5 +12,5 @@ uint32_t     u = x;
 
 u = (u << (8*bs)) | (u >> (32-8*bs));
 
-WRITE_RD(u ^ RS1);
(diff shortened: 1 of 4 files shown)


marcfedorow:
In addition, I found that the same error exists in instructions sm3* and sm4*.
**I didn’t fix `sm[34]*` insns**, as `zext_xlen` should just  be replaced with `sext_xlen` but at the moment this is not in accordance with the specification. This may be fixed with `sext_xlen(zext_xlen(...))` but I am not sure if it should be done,

aswaterman:
The SM3/SM4 instructions will be redefined to perform sign- rather than zero-extension (a normative change to the spec).  @ben-marshall will presumably make that change soon.  Thanks for handling these.

ben-marshall:
Yeah I'll do it alongside any other changes/recommendations from the architecture review process.

Issue #743, opened 2021-07-13
slt gives the wrong result in some situations


fabio-malatesta:
It seems that SPIKE gives sometimes the wrong results when handling slt instructions. Here's an example:

> riscv-isa-sim/build//spike -m0x40000000:0x1000000,0x80000000:0x1000,0x20000000:0x1000 --isa=RV32IMAFDHC --priv=msu -l --log-commits -d test.random 2> test.random.spike.log 
> [...]
> : core   0: 0x40004cc6 (0x03400b33) mul     s6, zero, s4
> core   0: 1 0x40004cc6 (0x03400b33) x22 0x00000000
> : core   0: 0x40004cca (0x007524b3) slt     s1, a0, t2
> core   0: 1 0x40004cca (0x007524b3) x 9 0x00000000
> : reg 0 a0
> 0xbfe75c99
> : reg 0 t2
> 0x00000e34

Attached a more complete log:
[test.random.spike.log](https://github.com/riscv/riscv-isa-sim/files/6808539/test.random.spike.log)

aswaterman:
I seriously doubt the implementation of the `slt` instruction is the problem.  More likely, an unrelated bug in the simulator is causing `a0` or `t2` to be incorrectly represented within the register.

I see that `hlvx.wu` is in the executed instruction list.  That is suspicious, since RV32 hypervisor support is poorly tested.  Please try my fix in #744

aswaterman:
Fixed by #744

PR #744, opened 2021-07-13
Fix possible HLVX.WU bug

aswaterman:
Need to sign extend result for RV32.

Possible fix to #743


fabio-malatesta:
Hi @aswaterman,
Thank you very much for the patch, it fixed the issue.

aswaterman:
Cool!

PR #748, opened 2021-07-16
adding riscof-plugin files for spike

neelgala:
This PR adds the plugin files required to run Spike as a test target on the new Architectural Testing RISCOF framework.


neelgala:
Need to make a few more changes based on feedback from @allenjbaum

PR #750, opened 2021-07-18
Priv virtual memory updates

daniellustig:
* Priv 1.12 requires page faults when the address translation process
reaches a PTE with any reserved bit set

* Svpbmt uses two PTE bits, but otherwise has no effect on Spike (since
Spike is sequentially consistent and does not model PMAs)

* Add Svinval instructions

Even though I updated riscv-opcodes separately, I merged the new
defines into riscv/encoding.h manually, because riscv-opcodes seems
to be a step ahead of riscv-isa-sim for a few vector opcodes, causing
conflicts when regenerating encoding.h...

If that gets fixed, and encoding.h gets regenerated automatically, I can
remove it from this PR to avoid conflicts.


aswaterman:
Since the new inval instructions are implemented identically to their fence counterparts, how about we use either symbolic links or e.g. `#include "sfence_vma.h"` to reuse the code that implements the fence instructions?

aswaterman:
The rest LGTM.

daniellustig:
> Since the new inval instructions are implemented identically to their fence counterparts, how about we use either symbolic links or e.g. `#include "sfence_vma.h"` to reuse the code that implements the fence instructions?

Sure, will make that change.  Any preference on handling encoding.h?  (see also https://github.com/riscv/riscv-tests/issues/350)

aswaterman:
@daniellustig I'm OK with either only importing the necessary changes or fully overwriting the file with the newest auto-generated one.  I guess the latter is slightly preferable to make sure we aren't drifting, but if doing it that way breaks the build for some reason, then doing it the way you're currently doing it is fine.

daniellustig:
OK pushed an update with the includes.

There are encoding.h conflicts with some really recent vector updates to riscv-opcodes, so if you're happy with the current approach I'd say let's leave this encoding.h as is and avoid mixing those vector changes into this PR.  Whenever that's resolved, a new autogenerated encoding.h should work too.  Thanks!

aswaterman:
Cool. Yeah, I think some vector instructions got renamed and Spike is on the old names.

PR #751, opened 2021-07-22
Fix HLVX

aswaterman:
HLVX should check only X permission, not (R || X) as previous.

Implement vsstatus.MXR and vsstatus.SUM, which were previously ignored.

(Also, I did a bit of refactoring prior to implementing these fixes.)

See also https://github.com/riscv/riscv-isa-manual/issues/682

cc @avpatel @jameshippisley

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -53,7 +53,6 @@ reg_t mmu_t::translate(reg_t addr, reg_t len, access_type type, uint32_t xlate_f
   if (!proc)
     return addr;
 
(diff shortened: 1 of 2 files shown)


avpatel:
Thanks @aswaterman for addressing this.

I quickly tested Xvisor nested MMU test-suite on this PR and it works fine.

jameshippisley:
I tested this and I agree that it fixes HLVX permissions to match the spec, i.e. HLVX checks only X permission, not (R || X) as previously.

aswaterman:
Thanks, guys!

PR #752, opened 2021-07-22
Non-leaf PTEs with D/A/U==1 are reserved

daniellustig:
...and hence should trigger page faults

Thanks to @pdonahue-ventana for pointing this out:
https://github.com/riscv/riscv-tests/issues/352


aswaterman:
Yep.  We need something analogous in `s2xlate`, but I'll take care of it.

(Actually, I'd like to refactor all of this since there's so much code duplication between the two... but that's a matter for another day.)

PR #754, opened 2021-07-26
decode: op: remove quad related macro and define

chihminchao:
Let git keep the legacy code


aswaterman:
> Let git keep the legacy code

Agreed.

PR #759, opened 2021-07-28
Scalar crypto: post arch review updates

ben-marshall:
This PR includes:

- Updates to the `aes32*` instructions such that their encodings don't overlap with the `aes64` ones. Implements the changes in riscv/riscv-opcodes#77 and originally discussed in riscv/riscv-crypto#101.

- Makes the `sm3*` and `sm4*` instructions sign extend their 32-bit results to `XLEN` bits. This was a specific bit of feedback from arch-review. The `aes32*` and `sha*` instructions already had these changes made.

- Removes the auxiliary code added to Spike to handle when the `aes32/64` instruction encodings overlapped. This is the change I am least confident about, so feedback is very welcome.


aswaterman:
Great, thanks!

PR #760, opened 2021-07-30
scalar-crypto: v0.9.4/arch-review entropy source updates.

ben-marshall:
Post arch review updates to the entropy source:

- New address is 0x546

- Removed old references to mentropy and mnoise.

- Throw invalid opcode exception when reading without writing. I _think_ I've done this in the cleanest place but happy to be advised of a better one.

- Missing: mseccfg register with SKES bit. Waiting for rest of register to be implemented by other task group.


aswaterman:
@ben-marshall generally try to parenthesize references to C macro arguments, in case someone passes in an expression rather than a single token.  (I'll fix this one, actually by moving the check into `get_csr`, which seems like better separation of concerns.)  Thanks for the update!

aswaterman:
Manually merged  via c28c5d09b9a547fd16ef819a5ad9a9509dd84026

ben-marshall:
Great, thanks for this!

PR #761, opened 2021-07-30
rvk: Sign-extension of sha256 insns

marcfedorow:
@aswaterman, K-ext 0.9.4 spec now requires sign-extension of these insns (as rvkintrin.h does).


aswaterman:
👍

Issue #765, opened 2021-08-06
rvp: SWAP16 old encoding


marcfedorow:
Since 0.9.3 p-spec SWAP16 is an alias to PKBT16.
Replacing SWAP16's encoding with PKBT16's one makes spike interpret all the PKBT16 insns as SWAP16 (even if rs1!=rs2).
I solved this by just removing SWAP16 at all but I am not sure this is the right way.

aswaterman:
This sounds right to me on the surface.  PR it?

aswaterman:
Resolved by #766

PR #766, opened 2021-08-07
Removed SWAP16 encoding and implementation header.

marcfedorow:
SWAP16 is an alias to PKBT16 since p-spec v0.9.3.


marcfedorow:
@aswaterman, this PR closes #765

PR #772, opened 2021-08-18
add option --cmd to read commands from file

emelcher:
This PR adds the option `--cmd=<name>` to the spike command line.
This new option is only useful if used together with option -d.
Before accepting interactive debug commands, the option `--cmd` reads and executes commands from the file given by `<name>`. However, if the last line of the command file is `quit`, the simulation ends without accepting any interactive command.

This option is useful, for example, to automatically simulate until the start of `main` before accepting interactive commands. To achieve this, the command file would contain, e. g., `until pc 0 0x00010144`, where `0x00010144` would have been obtained using `objdump`.

This option can also be used for regression tests in a more flexible way than option `-l`.
Such a regression test is not only useful for simulated executable RISC-V code but also for regression tests of spike itself.

This PR was part of #700, but had been removed from that PR because it was not directly related to it.

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -30,6 +30,8 @@ using std::right;
 using std::setw;
 using std::endl;
 
+#define MAX_CMD_STR 40 // maximum possible size of a command line
+
 #define STR_(X) #X      // these definitions allow to use a macro as a string
 #define STR(X) STR_(X)
 
@@ -110,7 +112,8 @@ std::string sim_t::rin(streambuf *bout_ptr) {
 
 // write sout to socket (via bout)
 void sim_t::wout(streambuf *bout_ptr) {
-  if (acceptor_ptr) { // only if a socket has been created
+  if (!cmd_file && acceptor_ptr) { // only if  we are not getting command inputs from a file
+                                   // and if a socket has been created
     try {
       boost::system::error_code ignored_error;
       boost::asio::write(*socket_ptr, *bout_ptr, transfer_all(), ignored_error);
@@ -151,14 +154,27 @@ void sim_t::interactive()
 
   while (!done())
   {
-    std::string s;
 #ifdef HAVE_BOOST_ASIO
     streambuf bout; // socket output
-    s = rin(&bout); // get command string from socket or terminal
-#else
-    cerr << ": " << std::flush;
-    s = readline(2); // 2 is stderr, but when doing reads it reverts to stdin
 #endif
+    std::string s;
+    char cmd_str[MAX_CMD_STR+1]; // only used for following fscanf
+    // first get commands from file, if cmd_file has been set
+    if (cmd_file && !feof(cmd_file) && fscanf(cmd_file,"%" STR(MAX_CMD_STR) "[^\n]\n", cmd_str)==1) {
+                                                      // up to MAX_CMD_STR characters before \n, skipping \n
+       s = cmd_str;
+       // while we get input from file, output goes to stderr
+       sout.rdbuf(cerr.rdbuf());
+    } else {
+       // when there are no commands left from file or if there was no file from the beginning
+       cmd_file = NULL; // mark file pointer as being not valid, so any method can test this easily
+#ifdef HAVE_BOOST_ASIO
+       s = rin(&bout); // get command string from socket or terminal
+#else
+       cerr << ": " << std::flush;
+       s = readline(2); // 2 is stderr, but when doing reads it reverts to stdin
+#endif
+    }
 
     std::stringstream ss(s);
     std::string cmd, tmp;
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -36,11 +36,11 @@ sim_t::sim_t(const char* isa, const char* priv, const char* varch,
              std::vector<int> const hartids,
              const debug_module_config_t &dm_config,
              const char *log_path,
-             bool dtb_enabled, const char *dtb_file
+             bool dtb_enabled, const char *dtb_file,
 #ifdef HAVE_BOOST_ASIO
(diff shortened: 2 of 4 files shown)


riscv/interactive.cc
```diff
@@ -110,7 +112,8 @@ std::string sim_t::rin(streambuf *bout_ptr) {
 
 // write sout to socket (via bout)
 void sim_t::wout(streambuf *bout_ptr) {
-  if (acceptor_ptr) { // only if a socket has been created
+  if (cmd_file==NULL && acceptor_ptr) { // only if  we are not getting command inputs from a file
```
aswaterman: spaces around `==` operator. Also, it seems a little weird to use an explicit comparison to NULL for the first part and an implicit comparison in the second half, i.e., either `!cmd_file && acceptor_ptr` or `cmd_file == NULL && acceptor_ptr != NULL` is fine, but avoid mixing and matching styles.

riscv/sim.cc
```diff
@@ -36,11 +36,11 @@ sim_t::sim_t(const char* isa, const char* priv, const char* varch,
...
              const debug_module_config_t &dm_config,
              const char *log_path,
-             bool dtb_enabled, const char *dtb_file
+             bool dtb_enabled, const char *dtb_file,
 #ifdef HAVE_BOOST_ASIO
-             , io_service *io_service_ptr_ctor, tcp::acceptor *acceptor_ptr_ctor // option -s
+             io_service *io_service_ptr_ctor, tcp::acceptor *acceptor_ptr_ctor, // option -s
 #endif
-             )
+             FILE *cmd_file_ctor) // neded for command line option --cmd
```
aswaterman: spelling of "needed"

aswaterman: Instead of naming this `cmd_file_ctor` and doing a `cmd_file = cmd_file_ctor;` later on, you can simply give this the same name `cmd_file` and add `cmd_file(cmd_file)` to the initializer list.  This is what we did for the `mems` argument.  You could also do this for `acceptor_ptr_ctor`.

aswaterman:
@chihminchao / @scottj97 can y'all also take a look?

emelcher:
Fixed typo and comparison and eliminated `cmd_file_ctor`.
Please confirm if I should eliminate `acceptor_ptr_ctor` and maybe also `io_service_ptr_ctor` in another commit of this same PR or open another PR for this.

spike_main/spike.cc
```diff
@@ -34,6 +34,7 @@ static void help(int exit_code = 1)
 #endif
   fprintf(stderr, "  -h, --help            Print this help message\n");
   fprintf(stderr, "  -H                    Start halted, allowing a debugger to connect\n");
+  fprintf(stderr, "  --cmd=<name>          Read commands from file (use with -d)\n");
```
chihminchao: I see there is an comment in help messag but prefer to adding prefix to option to explicit express it is for interactive mode.  "--debug-cmd", "--interactive-cmd" or something like

emelcher:
Changed option name to `--debug-cmd`, added condition `feof`, and changed `NULL==` to `==NULL`.
PR ready to be reviewed again.

emelcher:
PR ready to be reviewed again.

emelcher:
PR ready to be reviewed again.
This was the first and now is the last PR that was originally part of PR #700.

emelcher:
Hi @chihminchao and @aswaterman,
the last two pushes in this PR were needed only to resolve conflicts from other merges.
Is there anything else You would like me to improve in this PR ?

riscv/interactive.cc
```diff
@@ -151,14 +154,27 @@ void sim_t::interactive()
...
+    char cmd_str[MAX_CMD_STR]; // only used for following fscanf
+    // first get commands from file, if cmd_file has been set
+    if (cmd_file && fscanf(cmd_file,"%" STR(MAX_CMD_STR) "[^\n]\n", cmd_str)==1) {
+                                  // up to MAX_CMD_STR characters before \n, skipping \n
+       s = cmd_str;
+       // while we get input from file, output goes to stderr
+       sout.rdbuf(cerr.rdbuf());
+    } else {
+       // when there are no commands left from file or if there was no file from the beginning
+       cmd_file = NULL; // mark file pointer as being not valid, so any method can test this easily
```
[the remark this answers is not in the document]
aswaterman: was this resolved?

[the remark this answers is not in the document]
emelcher: Yes,
line 163 of riscv/interactive.cc was changed from
`if (cmd_file && ...`
to
`if (cmd_file && !feof(cmd_file) && ...`
so that reaching EOF goes to the `else` where 'cmd_file = NULL` gets executed.

chihminchao:
I think it could be merged.

riscv/interactive.cc
```diff
@@ -151,14 +154,27 @@ void sim_t::interactive()
...
-    std::string s;
 #ifdef HAVE_BOOST_ASIO
     streambuf bout; // socket output
-    s = rin(&bout); // get command string from socket or terminal
+#endif
+    std::string s;
+    char cmd_str[MAX_CMD_STR]; // only used for following fscanf
+    // first get commands from file, if cmd_file has been set
+    if (cmd_file && !feof(cmd_file) && fscanf(cmd_file,"%" STR(MAX_CMD_STR) "[^\n]\n", cmd_str)==1) {
+                                                      // up to MAX_CMD_STR characters before \n, skipping \n
```
aswaterman, line 164: This smells like a buffer overflow, since you can only read n-1 chars into a n-byte null-terminated string.  You should change the above declaration to `char cmd_str[MAX_CMD_STR+1];`

emelcher:
PR ready to be reviewed again.

aswaterman:
Thanks @emelcher, it's good by me.

Issue #589, opened 2020-11-03
Not able to simulate hello program


rahulraveendran15-coder:
Hi, 
I am getting the below error when running the command 
`spike pk hello_world`

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  could not open pk (did you misspell it? If VCS, did you forget +permissive/+permissive-off?)
```
How to overcome this?
I want to run with 32imc

My Tool chain paths are as below:
```

export PATH=${PATH}:/home/rr/.local/opt/riscv32/lowrisc-toolchain-gcc-rv32imc-20200904-1/bin
export PATH=${PATH}:/home/rr/.local/opt/riscv32/lowrisc-toolchain-gcc-rv32imc-20200904-1/riscv32-unknown-elf/bin ##path for pk

export RISCV_TOOLCHAIN=$HOME/.local/opt/riscv32/lowrisc-toolchain-gcc-rv32imc-20200904-1/

export RISCV_GCC="$RISCV_TOOLCHAIN/bin/riscv32-unknown-elf-gcc"
```
The build commands I tried is below

```
./configure --prefix=$RISCV_TOOLCHAIN/riscv32-unknown-elf --enable-commitlog --enable-misaligned 
./configure --prefix=$RISCV_TOOLCHAIN--enable-commitlog --enable-misaligned (tried this command also later)
make
sudo make install
```
```
riscv32-unknown-elf-gcc -o hello hello.c
spike pk hello
```
Can anyone please let me know where I am missing something?

Also when I tried `spike $(which pk) hello` it is running continuosuly without exiting nothing is getting displayed in terminal. For a simple hello program why it is running continuously how can I see the status of the run?

When I tried running `spike --isa=rv32i  $(which pk) hello`, it is throwing the below error
```
bbl loader
z  00000000 ra 8000c1b0 sp 800120b0 gp 00000000
tp 00000000 t0 80000000 t1 8001210b t2 00000000
s0 8000de10 s1 8000e0f5 a0 0000002e a1 00000000
a2 0000000a a3 00000000 a4 00000025 a5 8000c188
a6 00000000 a7 00000000 s2 80012100 s3 8001223c
s4 80010018 s5 00000015 s6 00000009 s7 00000000
s8 0000002e s9 0000000c sA 8000e0f4 sB 00000100
t3 8001226c t4 00000000 t5 00000003 t6 00000000
pc 8000c524 va 00008e2a insn       00008e2a sr 80046100
[...]

PR #776, opened 2021-08-21
configure option --with-target

emelcher:
This PR adds a new configure option `--with-target`.
Using `--with-target=riscv32-unknown-elf` is a cleaner way to solve issue #589
than using `CXXFLAGS="-DTARGET_ARCH=\\\"riscv32-unknown-elf\\\""`.

This PR was part of #700, but had been removed from that PR because it was not related to it.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -126,6 +126,9 @@
 /* Define to 1 if you have the ANSI C header files. */
 #undef STDC_HEADERS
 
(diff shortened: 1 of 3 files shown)


aswaterman:
cc @chihminchao

chihminchao:
I seldom work with pk but the change works fine with my current working environment .

PR #777, opened 2021-08-23
error message for bad command argument

emelcher:
This PR returns an error message to the user when a debug command is given with a wrong number of arguments or with an incorrect argument.

This PR was part of #700, but had been removed from that PR because it was not directly related to it.

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -184,7 +184,9 @@ void sim_t::interactive()
       else
         sout << "Unknown command " << cmd << endl;
     }
-    catch(trap_t& t) {}
+    catch(trap_t& t) {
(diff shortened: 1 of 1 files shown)


riscv/interactive.cc
```diff
@@ -497,11 +502,17 @@ void sim_t::interactive_until(const std::string& cmd, const std::vector<std::str
   bool cmd_until = cmd == "until" || cmd == "untiln";
 
   if (args.size() < 3)
-    return;
+   throw trap_interactive();
```
aswaterman: indentation doesn't match.  Since this is coming up frequently, you probably need to fix your editor to make the tab stop 2 spaces.

emelcher:
Fixed 2 mismatched indentations.
PR ready to be reviewed again.

PR #782, opened 2021-08-26
fesvr: replace use of `std::vector::operator[0]`

compnerd:
This replaces multiple uses of `std::vector::operator[]` where the
parameter is a constant `0` with the use of C++11's `std::vector::data`
method.  This fixes the root cause of invalid memory accesses.
`std::vector::operator[]` is an unchecked memory access, and when the
buffers are zero-sized (that is the buffer container is empty) either
due to a 0 padding in the case of elfloader or NULL parameters to
syscalls where permitted, the unchecked access may cause an invalid
memory access.  The use of `std::vector::data` is permitted even in such
a case, though the returned memory may not be dereferenced.  The general
usage of the returned pointer is to pass to `memif_t`, which is careful
about 0-sized buffer accesses, and so passing the result of
`std::vector::data` is safe.  This is theoretically a better access
pattern as it also avoids having the compiler to re-materialize the
pointer from the de-referenced location.

### DIFF
--- a/fesvr/device.cc
+++ b/fesvr/device.cc
@@ -110,10 +110,10 @@ void disk_t::handle_read(command_t cmd)
   cmd.memif().read(cmd.payload(), sizeof(req), &req);
 
   std::vector<uint8_t> buf(req.size);
-  if ((size_t)::pread(fd, &buf[0], buf.size(), req.offset) != req.size)
+  if ((size_t)::pread(fd, buf.data(), buf.size(), req.offset) != req.size)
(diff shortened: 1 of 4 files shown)


compnerd:
This is meant to be based on #781, clearing up the problem throughout the fesvr layer which is why it appears to have re-uploaded the same set of changes.  This clears up the case of syscalls such as:

```c
write(2, nullptr, 0);
```
which would fail due to the `buf` parameter being allocated and then referenced through `std::vector::operator[](0)`.

PR #785, opened 2021-08-27
disasm: hyp: add hypervisor instructions

chihminchao:


aswaterman:
Thanks!

PR #788, opened 2021-08-31
Fix virtual instruction trap generation for hpmcounters

avpatel:
We should check for virtual instruction traps before illegal
instruction traps upon hpmcounter reads. Also, virtualization
state should always be checked for virtual instruction traps.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -1441,36 +1441,33 @@ void processor_t::set_csr(int which, reg_t val)
 // side effects on reads.
 reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek)
 {
-  uint32_t ctr_en = -1;
-  if (state.prv < PRV_M)
-    ctr_en &= state.mcounteren;
-  if (supports_extension('S') && state.prv < PRV_S)
-    ctr_en &= state.scounteren;
-  bool ctr_ok = (ctr_en >> (which & 31)) & 1;
-  if (state.v)
-    ctr_en &= state.hcounteren;
-  bool ctr_v_ok = (ctr_en >> (which & 31)) & 1;
+#define mcounteren_ok(__which) \
+({ \
+  bool __ctr_ok = true; \
+  if (state.prv < PRV_M) \
+    __ctr_ok = (state.mcounteren >> (__which & 31)) & 1; \
+  __ctr_ok; \
+})
+#define hcounteren_ok(__which) \
+({ \
+  bool __ctr_ok = true; \
+  if (state.v) \
+    __ctr_ok = (state.hcounteren >> (__which & 31)) & 1; \
+  __ctr_ok; \
+})
+#define scounteren_ok(__which) \
+({ \
+  bool __ctr_ok = true; \
+  if (supports_extension('S') && state.prv < PRV_S) \
+    __ctr_ok = (state.scounteren >> (__which & 31)) & 1; \
+  __ctr_ok; \
+})
 
   reg_t res = 0;
 #define ret(n) do { \
     res = (n); \
     goto out; \
   } while (false)
-
-  if ((which >= CSR_HPMCOUNTER3 && which <= CSR_HPMCOUNTER31) ||
-      (xlen == 32 && which >= CSR_HPMCOUNTER3H && which <= CSR_HPMCOUNTER31H)) {
-    if (!ctr_ok)
-      goto throw_illegal;
-    if (!ctr_v_ok)
-      goto throw_virtual;
-    ret(0);
-  }
-  if (which >= CSR_MHPMCOUNTER3 && which <= CSR_MHPMCOUNTER31)
-    ret(0);
-  if (xlen == 32 && which >= CSR_MHPMCOUNTER3H && which <= CSR_MHPMCOUNTER31H)
-    ret(0);
-  if (which >= CSR_MHPMEVENT3 && which <= CSR_MHPMEVENT31)
-    ret(0);
 
   if (which >= CSR_PMPADDR0 && which < CSR_PMPADDR0 + state.max_pmp) {
     // If n_pmp is zero, that means pmp is not implemented hence raise trap if it tries to access the csr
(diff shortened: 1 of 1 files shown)


aswaterman:
@scottj97 mind reviewing?

riscv/processor.cc
```diff
@@ -1444,12 +1444,12 @@ reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek)
   uint32_t ctr_en = -1;
   if (state.prv < PRV_M)
     ctr_en &= state.mcounteren;
+  if (state.v && state.prv <= PRV_S)
```
scottj97: Is it ever possible to have `state.v && state.prv > PRV_S`? i.e. I don't see the need for the check of `state.prv` here.

riscv/processor.cc
```diff
@@ -1523,20 +1523,20 @@ reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek)
...
         goto throw_illegal;
-      if (!ctr_v_ok)
-        goto throw_virtual;
       ret(state.minstret);
     case CSR_MINSTRET:
     case CSR_MCYCLE:
       ret(state.minstret);
     case CSR_INSTRETH:
     case CSR_CYCLEH:
+      if (state.v && !ctr_v_ok)
```
scottj97: What about the check of `xlen`? Shouldn't that be higher priority than the check of `ctr_v_ok`? (It is for `hpmcounterh`)

scottj97 (changes requested):

riscv/processor.cc
aswaterman: Agreed.

riscv/processor.cc
aswaterman: Agreed, since the CSRs shouldn't exist for RV64, hence should raise illegal-instruction traps.

riscv/processor.cc
avpatel: `state.v && state.prv > PRV_S` is not possible.

The additional `state.prv <= PRV_S` check is totally redundant. I will remove it.

riscv/processor.cc
avpatel: Ahh, good catch. I missed this case. I will update this PR.

avpatel:
I have re-implemented `[m|h|s]counteren` CSR checks. Also, we don't need to touch `[m|h|s]counteren` CSRs upon every CSR read.

riscv/processor.cc
```diff
@@ -1523,25 +1513,47 @@ reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek)
...
+        ret(0);
     case CSR_INSTRETH:
     case CSR_CYCLEH:
-      if (!ctr_ok || xlen != 32)
-        goto throw_illegal;
-      if (!ctr_v_ok)
-        goto throw_virtual;
-      ret(state.minstret >> 32);
+    case CSR_HPMCOUNTER3H ... CSR_HPMCOUNTER31H:
+      if (!ctr_ok(which) || xlen != 32) {
```
scottj97: This still appears to throw virtual instead of illegal if `xlen==64 && state.v`. Is that right?

scottj97 (changes requested):

riscv/processor.cc
avpatel: This case is not clearly stated in H-extension spec.

I think it's better to generate virtual instruction trap when RV64 guest access high-half CSRs so that hypervisor directly gets trap without involving M-mode firmware. If we generate illegal instruction trap then M-mode firmware will take the trap and then redirect it to Hypervisor.

@aswaterman what do you think?

Regards,
Anup

aswaterman: The spec does address this case explicitly: "When V=1 and XLEN>32, an attempt to access a high-half supervisor-level CSR, high-half hypervisor CSR, high-half VS CSR, or high-half unprivileged CSR always raises an illegal instruction exception."

It doesn't make sense to send this exception to the hypervisor because there's nothing interesting to emulate: it's actually an illegal instruction.

avpatel: Okay, I will update this PR.

This would mean that M-mode firmware will take illegal instruction trap which will be further redirected to Hypervisor and  Hypervisor will again redirect it to Guest/VM.

Regards,
Anup

avpatel:
I have taken care of xlen != 32 for high-half CSRs in updated PR.

aswaterman:
Thanks.  I'm turning it back over to @scottj97 to review.

scottj97 (approved):

scottj97:
Just to doublecheck -- is this the new behavior described in https://github.com/riscv/riscv-isa-manual/pull/730 ?

aswaterman:
Seems that way.

scottj97:
This change has a significant refactoring mixed with some behavioral changes, which I should have discouraged in my review. For the record, the behavioral changes can be undone by the following diffs:
```
diff --git a/riscv/processor.cc b/riscv/processor.cc
index dc84bf32c..703834e71 100644
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -1425,9 +1425,6 @@ reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek, bool r
       if (!hcounteren_ok(which))
           goto throw_virtual;
       if (!scounteren_ok(which)) {
-        if (state.v)
-          goto throw_virtual;
-        else
           goto throw_illegal;
       }
       if (which == CSR_INSTRET)
@@ -1457,9 +1454,6 @@ reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek, bool r
       if (!hcounteren_ok(which))
           goto throw_virtual;
       if (!scounteren_ok(which)) {
-        if (state.v)
-          goto throw_virtual;
-        else
           goto throw_illegal;
       }
       if (which == CSR_INSTRETH)
```
The above change reverts part of this PR so that Spike's behavior matches the spec prior to [Make virtual instruction exceptions more consistent for VU mode](https://github.com/riscv/riscv-isa-manual/pull/730).

Then a second diff:
```
diff --git a/riscv/processor.cc b/riscv/processor.cc
index 703834e71..2d5954845 100644
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -1422,11 +1422,10 @@ reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek, bool r
     case CSR_HPMCOUNTER3 ... CSR_HPMCOUNTER31:
       if (!mcounteren_ok(which))
           goto throw_illegal;
+      if (!scounteren_ok(which))
+          goto throw_illegal;
       if (!hcounteren_ok(which))
           goto throw_virtual;
-      if (!scounteren_ok(which)) {
-          goto throw_illegal;
-      }
       if (which == CSR_INSTRET)
         ret(state.minstret);
       else if (which == CSR_CYCLE)
@@ -1451,11 +1450,10 @@ reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek, bool r
     case CSR_HPMCOUNTER3H ... CSR_HPMCOUNTER31H:
       if (!mcounteren_ok(which) || xlen != 32)
           goto throw_illegal;
+      if (!scounteren_ok(which))
+          goto throw_illegal;
       if (!hcounteren_ok(which))
           goto throw_virtual;
-      if (!scounteren_ok(which)) {
-          goto throw_illegal;
-      }
       if (which == CSR_INSTRETH)
         ret(state.minstret >> 32);
       else if (which == CSR_CYCLEH)
```
The above reverts all the remaining behavior changes and results in Spike behaving as it did before this PR.

avpatel:
Before this PR, Spike generated illegal instruction trap for counter X accessed from VU-mode when mcounteren[X]=1, hcounteren[X]=0, and scounteren[X]=0 whereas it should have generated virtual instruction trap.

(Refer, https://github.com/riscv/riscv-isa-sim/issues/747)

Also, accessing high-half counter CSR from RV64 VU/VS-mode should always generate illegal instruction trap which was not the case before this PR.

Regards,
Anup

aswaterman:
I might be missing something, but I don't think we want to revert anything here.  And, FWIW, https://github.com/riscv/riscv-isa-manual/pull/730 has been merged.

scottj97:
No I'm not suggesting we revert anything. I wanted to document the behavior changes since they are obscured by all the refactoring in the same commit.

PR #790, opened 2021-09-01
Removed B from misa. B is reserved.

marcfedorow:


aswaterman:
I don't want to delete all of these instructions that haven't yet been included in another extension: it is actually quite likely that many/most of them will be included in future extensions, and it is harmless to continue to include them in Spike.

To that end, I propose one of the following:
- continue to accept 'B' as a noncompliant Spike feature (i.e. simply close this PR);
- instead define a nonstandard extension 'Xbitmanip' that includes all bitmanip instructions that aren't yet in a defined extension.

aswaterman:
Having slept on it, I'm leaning towards the approach of leaving all of these instructions in, but gating them on the presence of the nonstandard `Xbitmanip` extension.

I've also spotted a few bugs.  Take a closer look at the implementations of `grevi`, `gorci`, `pack`, and `packw` and note that subsets of those instructions are required by the Zbb extension.  The implementation pattern is that we check for the specialized form (e.g. rs2 == x0) and permit it under Zbb, but require B in the general case.

marcfedorow:
The main idea of my commit was not about removing instructions.
It was about removing letter "B" from misa while it is no more used for indicating bitmanip (sub)extension(s) -- and almost certainly will not be in the future.
I want this to be done to equalize traces generated by spike and hardware.

aswaterman:
"equalize traces generated by spike and hardware" doesn't make sense to me, since simply removing "b" from the ISA string would suffice to accomplish that.

aswaterman:
Merged, but with my suggestion of moving everything into Xbitmanip.

lico614257:
Like the old version of Spike, I was able to use --isa=rv32imafcb, but how to include all the ZB * in the new version while running?

marcfedorow:
--isa=rv32imafc_zba_zbb_zbc_zbs
In Spike you may add _xbitmanip to get all the (other) b-ext 0.9.3. insn

lico614257 ***@***.***> schrieb am Do., 20. Jan. 2022, 14:09:

PR #792, opened 2021-09-02
Use correct CSR for SUM/MXR for VS-mode load/stores

scottj97:
See [discussion](https://github.com/riscv/riscv-isa-sim/commit/9cfc3e7fef7b29f6b53879d7c91c35459d9b493d#r55593451).

When `state.v`, `state.mstatus` and `state.vsstatus` are swapped, with `state.mstatus` holding the architectural value of `vsstatus`, leading to very confusing code. I plan to change this soon, but that will be a much bigger PR.

# Misbehavior explained

## SUM

Prior to 9cfc3e7, Spike would erroneously use mstatus.SUM instead of vsstatus.SUM when doing HLV/HSV/HLVX from M-mode or HS-mode.

After 9cfc3e7, load/stores from VS-mode will incorrectly use mstatus.SUM instead of vsstatus.SUM.

## MXR

Prior to 9cfc3e7, Spike would erroneously ignore vsstatus.MXR when doing HLV/HSV/HLVX from M-mode or HS-mode, or when doing load/stores from M-mode when mstatus.MPRV=1 and mstatus.MPV=1. Spike would erroneously ignore mstatus.MXR when doing load/stores from VS-mode or VU-mode.

After 9cfc3e7, load/stores from VS-mode or VU-mode would incorrectly apply vsstatus.MXR to both stages of translation, and apply mstatus.MXR to the VS-stage translation only.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -311,7 +311,8 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
   if (vm.levels == 0)
     return gpa;
 
-  bool mxr = proc->state.mstatus & MSTATUS_MXR;
+  reg_t arch_mstatus = proc->state.v ? proc->state.vsstatus : proc->state.mstatus;
+  bool mxr = arch_mstatus & MSTATUS_MXR;
 
   reg_t base = vm.ptbase;
   for (int i = vm.levels - 1; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


aswaterman:
I would have liked to see the solution involve nuking the swapping, because that is the root of the confusion.  But it's good to have this fixed in the short term.  Thank you for also volunteering to take on the bigger project.

PR #793, opened 2021-09-03
Refactor CSR modeling

scottj97:
This is a work-in-progress proposal to convert Spike's CSR modeling, to switch from using simple scalar variables to objects.

# What?

Every CSR will be an instance of a [`csr_t` object](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/csrs.h#L16-L41), and all reads and writes go through this interface:
```
class csr_t {
 public:
  virtual void verify_permissions(insn_t insn, bool write) const;
  virtual reg_t read() const noexcept = 0;
  virtual void write(const reg_t val) noexcept = 0;
};
```

Class `state_t` then has [`unordered_map<reg_t, csr_t*> csrmap`](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/processor.h#L168) which maps each CSR by its address to its implementation. With this, [`get_csr()`](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/processor.cc#L1158-L1163) becomes as simple as:
```
reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek) {
  auto search = state.csrmap.find(which);
  if (search != state.csrmap.end()) {
    if (!peek)
      search->second->verify_permissions(insn, write);
    return search->second->read();
  }
}
```
And [`set_csr()`](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/processor.cc#L920-L924) is even simpler.

# Why?

## Robustness
* Simplifies the hypervisor virtualization of CSRs such as `sscratch`: the [`virtualized_csr_t` class](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/csrs.h#L124-L144) will access either the original (_e.g._ `sscratch`) or the virtualized CSR (_e.g._ `vsscratch`) according to the current privilege mode.
* Removes the confusing [swapping](https://github.com/riscv/riscv-isa-sim/blob/5a36103e41ea43e8ba0505d78bda9a9dcc990e7c/riscv/processor.cc#L677-L679) of `state.mstatus` and `state.vsstatus` when in VS- or VU-mode (and which led to [#792](https://github.com/riscv/riscv-isa-sim/pull/792)) (see [plan](https://github.com/scottj97/riscv-isa-sim/blob/3f0df2352c7f34594cdfd5748d654f1bc3cf4c2e/riscv/csrs.h#L191-L213) and the many commits that follow to implement that plan)
* Centralizes [CSR logging](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/csrs.cc#L55-L59), which today is mostly in this [duplicate switch statement](https://github.com/riscv/riscv-isa-sim/blob/20efc8288159e437928c249fcea41e77d2b254e4/riscv/processor.cc#L1353-L1435) and has been a source of [bugs](https://github.com/riscv/riscv-isa-sim/pull/639) when it gets out of sync from the actual CSR writes.
* Describes each CSR's behavior in one place
* Simplifies platform customization (which is ~50% CSR changes)
* Simplifies configurability -- optional CSRs are either created or not during init
* Side benefit: [`get_const_xlen()`](https://github.com/scottj97/riscv-isa-sim/commit/b2f58c8da07d37be5f0f89704a5e866392349a43) documents places in the code where XLEN is assumed to be constant throughout the simulation; if Spike ever grows to support variable XLEN, it will be easy to spot the places that need work
* Side benefit: [`extension_enabled_const()`](https://github.com/scottj97/riscv-isa-sim/commit/625b4a619506c7db6e7762699dd6314b5fcf8587) documents places in the code where a given `misa` extension is assumed to be constant throughout the simulation; if the requested bit is ever modified to become writable, it will be easy to spot the places that need work

## Bug fixes

* Fixes logging of hypervisor virtualized CSRs: writing _e.g._ `sscratch` from VS-mode will now correctly log that `vsscratch` was written, not `sscratch` as before
* `vsstatus` now has an [XS field](https://github.com/scottj97/riscv-isa-sim/commit/f79e0bd3615f73b3b67bbf8d2446613df38afda3) when appropriate
* `csrw` to `hie` or `vsie` will now properly log changes to `mie`, which were previously missing
* `csrw` to `hip`, `hvip`, and `vsip` will now properly log changes to `mip`, which were previously missing
* `mstatus.FS` will always [be writable](https://github.com/scottj97/riscv-isa-sim/commit/a5d14803f3dbc468554fec5970784f15f70ef790) even after `misa.F=0`; see [riscv-isa-manual#707](https://github.com/riscv/riscv-isa-manual/pull/707)
* FPU ops that write `mstatus.FS` to Dirty will now be properly logged

## Code cleanup

* Get rid of this 400+ line [switch statement](https://github.com/riscv/riscv-isa-sim/blob/20efc8288159e437928c249fcea41e77d2b254e4/riscv/processor.cc#L941-L1351) in `set_csr()`
* Get rid of this 300 line [switch statement](https://github.com/riscv/riscv-isa-sim/blob/20efc8288159e437928c249fcea41e77d2b254e4/riscv/processor.cc#L1495-L1782) in `get_csr()`
* Move functionality into localized classes instead of in the `processor_t` God object. Examples:
  * pmpaddr: compare the [original `pmp_ok()` and `pmp_homogenous()`](https://github.com/scottj97/riscv-isa-sim/blob/ea5d76121bef5e3d54bbaba21cce75d0299c461e/riscv/mmu.cc#L218-L303) with the [new implementation](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/mmu.cc#L218-L259)
  * satp: moved these [methods specific to `satp`](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/csrs.cc#L744-L770) (and `vsatp`) out of the global `processor_t`
  * `mip` and `mie`: all of the myriad CSRs that manipulate these two underlying registers are now expressed [relatively succinctly](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/processor.cc#L356-L386)

## Drawbacks
* Quite a bit of boilerplate required for many CSRs
* Incremental recompile is slow whenever `csrs.h` changes
* Performance: informal (and inadequate) benchmarking shows approx 10% slowdown; though if `-l --log-commits` is used, there is practically no difference

# Current status

Only about half of the CSRs are converted. The remaining ones still use the old system.

# Next steps

There are many more CSRs yet to be converted. If the general approach is agreeable, I would rather not wait to merge before all are converted -- merge conflicts have been numerous and cumbersome since I first started this project. For example #788 is going to be a mess, even though I haven't yet converted the performance counters to this system.

I believe this PR is production-ready. I humbly await your review.

### DIFF
--- a/riscv/clint.cc
+++ b/riscv/clint.cc
@@ -33,7 +33,7 @@ bool clint_t::load(reg_t addr, size_t len, uint8_t* bytes)
   if (addr >= MSIP_BASE && addr + len <= MSIP_BASE + procs.size()*sizeof(msip_t)) {
     std::vector<msip_t> msip(procs.size());
     for (size_t i = 0; i < procs.size(); ++i)
-      msip[i] = !!(procs[i]->state.mip & MIP_MSIP);
+      msip[i] = !!(procs[i]->state.mip->read() & MIP_MSIP);
     memcpy(bytes, (uint8_t*)&msip[0] + addr - MSIP_BASE, len);
   } else if (addr >= MTIMECMP_BASE && addr + len <= MTIMECMP_BASE + procs.size()*sizeof(mtimecmp_t)) {
     memcpy(bytes, (uint8_t*)&mtimecmp[0] + addr - MTIMECMP_BASE, len);
@@ -54,9 +54,9 @@ bool clint_t::store(reg_t addr, size_t len, const uint8_t* bytes)
     memset((uint8_t*)&mask[0] + addr - MSIP_BASE, 0xff, len);
     for (size_t i = 0; i < procs.size(); ++i) {
       if (!(mask[i] & 0xFF)) continue;
-      procs[i]->state.mip &= ~MIP_MSIP;
+      procs[i]->state.mip->backdoor_write_with_mask(MIP_MSIP, 0);
       if (!!(msip[i] & 1))
-        procs[i]->state.mip |= MIP_MSIP;
+        procs[i]->state.mip->backdoor_write_with_mask(MIP_MSIP, MIP_MSIP);
     }
   } else if (addr >= MTIMECMP_BASE && addr + len <= MTIMECMP_BASE + procs.size()*sizeof(mtimecmp_t)) {
     memcpy((uint8_t*)&mtimecmp[0] + addr - MTIMECMP_BASE, bytes, len);
@@ -82,8 +82,8 @@ void clint_t::increment(reg_t inc)
     mtime += inc;
   }
   for (size_t i = 0; i < procs.size(); i++) {
-    procs[i]->state.mip &= ~MIP_MTIP;
+    procs[i]->state.mip->backdoor_write_with_mask(MIP_MTIP, 0);
     if (mtime >= mtimecmp[i])
-      procs[i]->state.mip |= MIP_MTIP;
+      procs[i]->state.mip->backdoor_write_with_mask(MIP_MTIP, MIP_MTIP);
   }
 }
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -617,7 +617,7 @@ reg_t generic_int_accessor_t::deleg_mask() const {
 
 // implement class mip_proxy_csr_t
 mip_proxy_csr_t::mip_proxy_csr_t(processor_t* const proc, const reg_t addr, generic_int_accessor_t_p accr):
-  csr_t(proc, addr),
+  logged_csr_t(proc, addr),
   accr(accr) {
 }
 
@@ -625,8 +625,9 @@ reg_t mip_proxy_csr_t::read() const noexcept {
   return accr->ip_read();
 }
 
-void mip_proxy_csr_t::write(const reg_t val) noexcept {
+bool mip_proxy_csr_t::unlogged_write(const reg_t val) noexcept {
   accr->ip_write(val);
+  return false;  // accr has already logged
 }
 
 // implement class mie_proxy_csr_t
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -484,9 +484,7 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   const reg_t new_misa = (adjusted_val & write_mask) | (old_misa & ~write_mask);
   const bool new_h = new_misa & (1L << ('H' - 'A'));
 
-  // update the forced bits in MIDELEG and other CSRs
-  if (new_h && !prev_h)
-    state->mideleg |= MIDELEG_FORCED_MASK;
+  // update the hypervisor-only bits in MEDELEG and other CSRs
   if (!new_h && prev_h) {
     reg_t hypervisor_exceptions = 0
       | (1 << CAUSE_VIRTUAL_SUPERVISOR_ECALL)
@@ -495,7 +493,6 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
       | (1 << CAUSE_VIRTUAL_INSTRUCTION)
       | (1 << CAUSE_STORE_GUEST_PAGE_FAULT)
       ;
-    state->mideleg &= ~MIDELEG_FORCED_MASK;
     state->medeleg &= ~hypervisor_exceptions;
     state->mstatus->write(state->mstatus->read() & ~(MSTATUS_GVA | MSTATUS_MPV));
     state->mie->write_with_mask(MIP_HS_MASK, 0);  // also takes care of hie, sie
@@ -613,7 +610,7 @@ void generic_int_accessor_t::ie_write(const reg_t val) noexcept {
 
 reg_t generic_int_accessor_t::deleg_mask() const {
   const reg_t hideleg_mask = mask_hideleg ? state->hideleg : (reg_t)~0;
-  const reg_t mideleg_mask = mask_mideleg ? state->mideleg : (reg_t)~0;
+  const reg_t mideleg_mask = mask_mideleg ? state->mideleg->read() : (reg_t)~0;
   return hideleg_mask & mideleg_mask;
 }
 
@@ -644,4 +641,32 @@ reg_t mie_proxy_csr_t::read() const noexcept {
 
 void mie_proxy_csr_t::write(const reg_t val) noexcept {
   accr->ie_write(val);
+}
+
+
+// implement class mideleg_csr_t
+mideleg_csr_t::mideleg_csr_t(processor_t* const proc, const reg_t addr):
+  basic_csr_t(proc, addr, 0) {
+}
+
+reg_t mideleg_csr_t::read() const noexcept {
+  reg_t val = basic_csr_t::read();
+  if (proc->extension_enabled('H')) return val | MIDELEG_FORCED_MASK;
+  // No need to clear MIDELEG_FORCED_MASK because those bits can never
+  // get set in val.
+  return val;
+}
+
+void mideleg_csr_t::verify_permissions(insn_t insn, bool write) const {
+  basic_csr_t::verify_permissions(insn, write);
+  if (!proc->extension_enabled('S'))
+    throw trap_illegal_instruction(insn.bits());
+}
+
+bool mideleg_csr_t::unlogged_write(const reg_t val) noexcept {
+  const reg_t supervisor_ints = proc->extension_enabled('S') ? MIP_SSIP | MIP_STIP | MIP_SEIP : 0;
+  const reg_t coprocessor_ints = (reg_t)proc->any_custom_extensions() << IRQ_COP;
+  const reg_t delegable_ints = supervisor_ints | coprocessor_ints;
+
+  return basic_csr_t::unlogged_write(val & delegable_ints);
 }
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -40,11 +40,6 @@ void csr_t::verify_permissions(insn_t insn, bool write) const {
 csr_t::~csr_t() {
 }
 
-// implement class logged_csr_t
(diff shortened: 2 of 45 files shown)


aswaterman:
Thanks, @scottj97.  I wouldn't have taken this tack: I view the extra indirection as an onus.  I don't view the large switch statements as an impediment, especially since they so closely mirror HW implementations in practice.

My inclination is to say "no" to this PR.  But, if the community speaks up in favor of this approach, I won't get in the way.

One concern I have is relying too much on indirection.  It'll be hard to discern the specification of a specific CSR if you have to dig through several parent classes to learn all the behavior.  Can we please endeavor to keep the hierarchy as flat as possible?

scottj97:
> One concern I have is relying too much on indirection.

Most of the bug fixes in this PR came about by accident. With the approach I've taken, it's harder to create bugs in the first place, and the bugs already present became obvious during this refactoring, or disappeared on their own sometimes without my even noticing at first.

My team requires reliable CSR logging from Spike. The current system is not reliable, and I don't see how to achieve that (and keep it working as Spike evolves) without a layer of indirection like this.

"We can solve any problem by introducing an extra level of indirection." :stuck_out_tongue: 

I do agree that too-deep hierarchy can burden the reader. But what I've found from working with code like this is that once I finally do identify the place that needs to change, I can make that change with high confidence that I'm not breaking anything else. I am optimizing for long-term maintainability more than readability. I've added hierarchy only where it enables sharing code that should be shared (keeping it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)).

scottj97:
@avpatel I would appreciate your input. The original impetus for this change was to repair the logging of all the hypervisor CSRs, but I find the benefits extend well beyond that.

scottj97:
@chihminchao as you are the original developer of the [CSR commit logging](https://github.com/riscv/riscv-isa-sim/commit/580ef6a6a519ea40d2680eed26d3c6e75bf4f763), you might be interested.

aswaterman:
I am warming to the idea but would like to hear others' thoughts.

aswaterman:
Pardon the merge conflicts caused by #790.  Don't forget to remove the `(1L << ('B' - 'A'))` from `misa_csr_t`.

avpatel:
Overall this looks good, I suggest to add check()/predicate() in csr_t along with read() and write() which will allow us to implement common trap checking. For example, we can create common CSR class for hpmcounter and [m|s|h]counteren checks can be part of this class.

Regards,
Anup

chihminchao:
1. For logging part, I agree that removing the duplicated switch needs some level abstraction since it needs some side effect to do  writing and logging at same time.  it can't be achieved when register value is represented in simple integer.
2. I don't prefer too deep inheritance. It make debugging and traversing difficult.  Is it possible to make it more component base (ex: remove the level of logged_csr_t)
3. @scottj97,  You mentions the PR is  "production-ready".  How is it tested ?
4.  I personally suggest this change but these commits need some tidying and merging for review process

scottj97:
> Overall this looks good, I suggest to add check()/predicate() in csr_t along with read() and write() which will allow us to implement common trap checking. For example, we can create common CSR class for hpmcounter and [m|s|h]counteren checks can be part of this class.

If I understand you correctly, that is exactly what [`verify_permissions()`](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/csrs.h#L20-L21) is for. The [base class implementation](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/csrs.cc#L22-L37) checks permissions based on the CSR's address. Specific CSRs with additional restrictions will [override](https://github.com/scottj97/riscv-isa-sim/blob/370facfab00e556982bc48c5239169f259adba29/riscv/csrs.cc#L776-L780).

I haven't yet converted the counters to this system, but that's where all the checks from #788 will go.

Thank you for the review.

scottj97:
> * I don't prefer too deep inheritance. It make debugging and traversing difficult.  Is it possible to make it more component base (ex: remove the level of logged_csr_t)

The design emerged from the requirements of the code. The `logged_csr_t` exists because some CSRs are implemented in terms of others. For example, `sip/sie` are implemented through masked access to `mip/mie`, which already log. Each virtualized register is implemented by two underlying registers (_e.g._ `sscratch/vsscratch`) which already log. To avoid double logging, those do not inherit from `logged_csr_t`.

> * @scottj97,  You mentions the PR is  "production-ready".  How is it tested ?

Excellent question. I built a regression system that compiles Spike using a variety of compile-time options, then runs a number of the [riscv-tests](https://github.com/riscv-software-src/riscv-tests) using a variety of `--isa` options, and records signatures and commit logs (where enabled) in git. I can run it on each riscv-isa-sim commit in a series and then by looking through the git log, see how the behavior changes. It caught several problems (which I've already patched).

> * I personally suggest this change but these commits need some tidying and merging for review process

Do you think there's too many commits? I personally like lots of tiny commits (makes `git bisect` much more useful), and I think this is considered best practice in the industry, but I can squash it down to fewer if you think it would help review.

scottj97:
> * I don't prefer too deep inheritance. It make debugging and traversing difficult.  Is it possible to make it more component base (ex: remove the level of logged_csr_t)

Upon further reflection, since `unlogged_write()` can return `false` to block the logging, perhaps I don't need this level of hierarchy at all. Let me experiment.

scottj97:
I pushed a few new commits to remove `logged_csr_t`. Great suggestion.

aswaterman:
I'm OK with merging once the merge conflicts are fixed.  (Note the comment I left at https://github.com/riscv-software-src/riscv-isa-sim/pull/793#issuecomment-913929765)

scottj97:
Resolved the merge conflict. I'll get to work on converting the remaining CSRs for a follow-on PR.

aswaterman:
I'll merge this now with fingers crossed.

neelgala:
@scottj97 I had a few questions about the changes in set_virt function in processor.cc. Eariler we had [this](https://github.com/riscv-software-src/riscv-isa-sim/blob/686674dc15c1734940614c8f59419e9f94d51a52/riscv/processor.cc#L668-L679) swapping scheme between mstatus and vsstatus happening. Which I see is removed in your PR. Is this swapping done else where in the new code or was the swapping wrong in the first place ? I am hoping for the latter, as I couldn't see where in the spec the need for swapping is  mentioned.

scottj97:
> @scottj97 I had a few questions about the changes in set_virt function in processor.cc. Eariler we had [this](https://github.com/riscv-software-src/riscv-isa-sim/blob/686674dc15c1734940614c8f59419e9f94d51a52/riscv/processor.cc#L668-L679) swapping scheme between mstatus and vsstatus happening. Which I see is removed in your PR. Is this swapping done else where in the new code or was the swapping wrong in the first place ? I am hoping for the latter, as I couldn't see where in the spec the need for swapping is mentioned.

The architectural requirement is that when V=1, vsstatus substitutes for most uses of sstatus. The swapping was a clever trick to get the existing code (which used the `mstatus` variable, a superset of sstatus) to see bits from vsstatus instead. The swapping was not "wrong" but the resulting code was very confusing because the variable named `mstatus` would sometimes contain bits from the architectural vsstatus, and vice versa, which led to [bugs](https://github.com/riscv-software-src/riscv-isa-sim/pull/792) and [confusion](https://github.com/riscv-software-src/riscv-isa-sim/pull/793/commits/9752f09f2e81a875396910a1abc9f96883a9b6c1).

This PR changes the implementation so that all appropriate code refers to the object `sstatus`, which inherits from type `virtualized_csr_t`. That class maintains [two underlying CSRs](https://github.com/scottj97/riscv-isa-sim/blob/457be55e86fadc63e01127bfe94fd0b86b475a8f/riscv/csrs.h#L131-L132), in this case sstatus and vsstatus, and [chooses between them](https://github.com/scottj97/riscv-isa-sim/blob/457be55e86fadc63e01127bfe94fd0b86b475a8f/riscv/csrs.cc#L227-L241) based on V.

See the [implementation plan](https://github.com/scottj97/riscv-isa-sim/blob/b4d598caa9a73148cec2823f354d382dc5da230c/riscv/csrs.h#L191-L213) and the commits that follow that implement that plan.

neelgala:
@scottj97 thanks for the info. 

Also while using the latest spike (after your PR merge) spike seems to be seg-faulting for some of my applications. Using gdb I was able get the following : 

```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[Detaching after fork from child process 1031758]
[Detaching after fork from child process 1031759]
[Detaching after fork from child process 1031760]
[Detaching after fork from child process 1031761]

0x000055555583292f in pmpaddr_csr_t::next_locked_and_tor (this=0x555555c169e0) at /usr/include/c++/9/bits/shared_ptr_base.h:1020
1020	      _M_get() const noexcept

```

This wasn't happening on spike before this PR. I haven't debugged further, but any quick takes on why this happens or how I could figure out why its seg-faulting ?

scottj97:
> This wasn't happening on spike before this PR. I haven't debugged further, but any quick takes on why this happens or how I could figure out why its seg-faulting ?

I have not seen that, and it's disturbing. Please email me and let's get to the bottom of this.

scottj97:
Opened #798 to fix.

PR #795, opened 2021-09-06
add option --log=<name> to command line help

emelcher:
This PR adds the option `--log=<name>` to the spike command line help listing.

The option `--log` had been implemented in March 2020 but had not been included in the listing of options which is provided by spike command line help.

As far as I can see, the same effect as `--log=<name>` can be achieved by adding `2><name>` to the spike command line. So, alternatively to this PR, another PR could be proposed to remove the implementation of the `--log` option. This would have the benefit of eliminating the function `processor_t::debug_output_log` and simplifying the functions `processor_t::take_trap` and `processor_t::disasm`.
However, any script out there using `--log` would become broken.

This PR was originally included in PR #700, but has been removed from that PR because it was not related to it.

### DIFF
--- a/spike_main/spike.cc
+++ b/spike_main/spike.cc
@@ -34,6 +34,7 @@ static void help(int exit_code = 1)
 #endif
   fprintf(stderr, "  -h, --help            Print this help message\n");
   fprintf(stderr, "  -H                    Start halted, allowing a debugger to connect\n");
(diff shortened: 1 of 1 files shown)


aswaterman:
Let's merge this for now.  I agree that redirecting `stderr` is a more elegant solution and we might consider doing that later, at which point we'd deprecate `--log`.

PR #796, opened 2021-09-07
Splitted K-ext to zeds

marcfedorow:


aswaterman:
@ben-marshall please review

ben-marshall (approved):

ben-marshall:
This was on my to-do list for next week so thank you very much @marcfedorow for getting in there first.

Everything looks as I'd expect. It means we can specify all of the individual extensions in the scalar crypto spec, rather than, as previously, just adding `K` to the ISA string.

aswaterman:
Thanks.  I'm going to wait until #793 is merged before handling this one, because I've already caused @scottj97 enough pain.

aswaterman:
OK, #793 has been merged; please rebase.

marcfedorow:
@aswaterman Done

marcfedorow:
@aswaterman, Should I do something else?

aswaterman:
Nope, looks good now.  Thanks!

PR #798, opened 2021-09-09
Fix segfault when executing 'csrw pmpaddr15' when n_pmp==16

scottj97:
Off-by-one error introduced in 69008075a1f3f9d7ed6fec3489c50e5d7701d366


neelgala:
@scottj97 yes.. this fixes it.. thanks

aswaterman:
Thanks!

PR #802, opened 2021-09-15
Parse FDQ extensions in isa string as in spec

marcfedorow:
See: 27.11 Subset Naming Convention (Table 27.1)

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -217,144 +217,120 @@ void processor_t::parse_priv_string(const char* str)
 
 void processor_t::parse_isa_string(const char* str)
 {
-  std::string lowercase = strtolower(str), tmp;
-
-  char error_msg[256];
-  const char* p = lowercase.c_str();
+  isa_string = strtolower(str);
   const char* all_subsets = "imafdqchp"
 #ifdef __SIZEOF_INT128__
     "v"
 #endif
     "";
 
-  max_xlen = 64;
   max_isa = reg_t(2) << 62;
-
-  if (strncmp(p, "rv32", 4) == 0)
-    max_xlen = 32, max_isa = reg_t(1) << 30, p += 4;
-  else if (strncmp(p, "rv64", 4) == 0)
-    p += 4;
-  else if (strncmp(p, "rv", 2) == 0)
-    p += 2;
-
-  if (!*p) {
-    p = "imafdc";
-  } else if (*p == 'g') { // treat "G" as "IMAFD"
-    tmp = std::string("imafd") + (p+1);
-    p = &tmp[0];
-  }
-
-  isa_string = "rv" + std::to_string(max_xlen) + p;
-
-  while (*p) {
-    if (islower(*p)) {
-      max_isa |= 1L << (*p - 'a');
-      extension_table[toupper(*p)] = true;
-
-      if (strchr(all_subsets, *p)) {
-        p++;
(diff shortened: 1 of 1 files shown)


marcfedorow:
@aswaterman, actually spike parses isa improperly:
- it allows wrong letter order (e.g. imcaqdf) 
- it does not allow no underscore before first zshx subext (except bitmanip)
- it does not allow underscore before bitmanip

Are you or if I make this parsing consistent with the latest riscv-isa-spec?

aswaterman:
Sure, feel free to append to this PR.

riscv/processor.cc
```diff
@@ -248,6 +248,11 @@ void processor_t::parse_isa_string(const char* str)
 
   while (*p) {
     if (islower(*p)) {
+      switch (*p) {
+        case 'q': max_isa |= 1L << ('d' - 'a');
+        case 'd': max_isa |= 1L << ('f' - 'a');
+        // case 'f': extension_table[EXT_ZICSR] = true;
```
aswaterman: Please don't include the commented-out code.

Note that Spike _necessarily_ has Zicsr, because Zicsr is implied by the privileged architecture.  We don't need to call it out specially.  We don't even need to define the EXT_ZICSR enum field.  I suppose we could accept the "Zicsr" string and simply ignore it, but we shouldn't require the string.

marcfedorow:
@aswaterman I tried to achieve some equilibrium between not making this parser as complicated as in gcc but setting required misa bits on the other hand.
Not sure if I missed something spike-specific.

riscv/processor.cc
```diff
@@ -217,144 +217,120 @@ void processor_t::parse_priv_string(const char* str)
...
-    tmp = std::string("imafd") + (p+1);
-    p = &tmp[0];
+  auto p = isa_string.begin();
+  for (p += 4; islower(*p) && !strchr("zshx", *p); ++p) {
+    while (*all_subsets && (*p != *all_subsets))
+      ++all_subsets;
+    if (!*all_subsets)
+      bad_isa_string(str, "Wrong order");
+    switch (*p) {
+      case 'q': max_isa |= 1L << ('d' - 'a');
```
aswaterman, line 246: This needs to be `(1L << ('d' - 'a')) | (1L << ('f' - 'a'))`, because Q implies D then D implies F.  I'll fix it as part of testing it out, though.

aswaterman: Nevermind, missed that this is a fallthrough...  good ol' C.

aswaterman:
Looks like this still has some bugs: e.g., `--isa=rv64gc` doesn't work.

aswaterman:
Also, custom extensions are partially broken.  `--isa=rv64gc_xfoobar` should print `couldn't find extension 'foobar' in shared library 'libcustomext.so'`

`xbitmanip` still seems to be working correctly.

As a special case `xdummy` is simply supposed to be ignored.  This seems to be working correctly, but make sure it still works after fixing the above.

marcfedorow:
@aswaterman, I think it's ready for review
- fixed index for 'g' letter
- reverted remove of `register_extension` call

Actually `rv64gc_xfoobar` never worked (while `rv64gcxfoobar` was). Currently both of them do the same thing.

aswaterman:
Looks good.  If any other issues arise, we can deal with them separately.

PR #804, opened 2021-09-16
Convert minstret and friends to new CSR format

scottj97:
@aswaterman I'm not sure if you want to take these changes in small batches, or if you'd prefer to have all remaining CSRs converted in a single PR?

If the former, this PR is ready for review. If the latter, I'll leave this open as WIP and keep updating it as I make progress.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -48,8 +48,12 @@ void csr_t::write(const reg_t val) noexcept {
 }
 
 void csr_t::log_write() const noexcept {
+  log_special_write(address, read());
+}
+
+void csr_t::log_special_write(const reg_t address, const reg_t val) const noexcept {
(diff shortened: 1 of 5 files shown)


aswaterman:
@scottj97 small PRs preferable.

neelgala:
@scottj97 there is an issue in printing the h*deleg csrs in the log output ( using the `--log-commits`). Let me know if I should raise that as a separate issue. I couldn't figure out why this is happening.

scottj97:
> @scottj97 there is an issue in printing the h*deleg csrs in the log output ( using the `--log-commits`). Let me know if I should raise that as a separate issue. I couldn't figure out why this is happening.

Yes, logging of hideleg/hedeleg has never worked properly. They will start working when I convert them to this new system. I will prioritize those two for my next PR. Expect that within a week.

PR #810, opened 2021-09-26
Convert remaining hypervisor CSRs to csr_t

scottj97:
See #793 for rationale.

* mtval2
* mtinst
* hgeie
* htval
* htinst
* hedeleg
* hideleg
* hgeip
* hgatp

The only functional change intended is in the commit log. These registers were not being logged at all previously.

cc @neelgala


neelgala:
seems about right.

aswaterman:
Looks good from a skim.  Thanks!

PR #813, opened 2021-09-27
Convert debug csrs to csr_t

scottj97:
See #793 for rationale.

* tselect
* tdata1
* tdata2
* dscratch0
* dscratch1
* dpc
* dcsr
* tdata3

I ran the debug tests in riscv-tests/debug; three of them (Sv32Test, Sv39Test, Sv48Test) were already failing on master, and all the rest passed both before and after this PR.

### DIFF
--- a/riscv/csrs.h
+++ b/riscv/csrs.h
@@ -508,4 +508,13 @@ class tselect_csr_t: public basic_csr_t {
 };
 
 
+class tdata1_csr_t: public csr_t {
+ public:
+  tdata1_csr_t(processor_t* const proc, const reg_t addr);
+  virtual reg_t read() const noexcept override;
+ protected:
+  virtual bool unlogged_write(const reg_t val) noexcept override;
+};
+
+
 #endif
--- a/riscv/csrs.h
+++ b/riscv/csrs.h
@@ -543,4 +543,17 @@ class dpc_csr_t: public epc_csr_t {
   virtual void verify_permissions(insn_t insn, bool write) const override;
 };
 
+typedef struct
+{
+  uint8_t prv;
+  bool step;
+  bool ebreakm;
+  bool ebreakh;
+  bool ebreaks;
+  bool ebreaku;
+  bool halt;
+  uint8_t cause;
+} dcsr_t;
+
+
 #endif
--- a/riscv/csrs.h
+++ b/riscv/csrs.h
@@ -500,4 +500,12 @@ class hgatp_csr_t: public basic_csr_t {
 };
 
 
+class tselect_csr_t: public basic_csr_t {
+ public:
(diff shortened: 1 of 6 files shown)


riscv/csrs.h
```diff
@@ -500,4 +500,68 @@ class hgatp_csr_t: public basic_csr_t {
...
+typedef std::shared_ptr<tdata2_csr_t> tdata2_csr_t_p;
+
+
+class dpc_csr_t: public epc_csr_t {
+ public:
+  dpc_csr_t(processor_t* const proc, const reg_t addr);
+  virtual void verify_permissions(insn_t insn, bool write) const override;
+};
+
+class dcsr_csr_t: public csr_t {
```
timsifive, line 546: Shouldn't this inherit debug_mode_csr_t?

riscv/csrs.h
```diff
@@ -500,4 +500,68 @@ class hgatp_csr_t: public basic_csr_t {
...
+class debug_mode_csr_t: public basic_csr_t {
+ public:
+  debug_mode_csr_t(processor_t* const proc, const reg_t addr);
+  virtual void verify_permissions(insn_t insn, bool write) const override;
+};
+
+typedef std::shared_ptr<tdata2_csr_t> tdata2_csr_t_p;
+
+
+class dpc_csr_t: public epc_csr_t {
```
timsifive, line 540: I assume this does not inherit debug_mode_csr_t to avoid multiple inheritance.

timsifive (approved):
This looks right to me, although without any of the surrounding context in my head, I found it hard to review.

riscv/csrs.h
scottj97: I wish C++ had mixins or something equivalent. Class `debug_mode_csr_t` has a 64-bit `val` to record the CSR's current value (inherited from `basic_csr_t`). For `dcsr` I kept the implementation close to the original, which records the bits in a series of bools (the old `dcsr_t`). Which means I couldn't inherit from `debug_mode_csr_t`.

riscv/csrs.h
scottj97: Yes. The disappointing thing is that without using multiple inheritance (and diamond-pattern multiple inheritance) I have to have three separate implementations of `verify_permissions()` that are all identical (for `debug_mode_csr_t`, `dpc_csr_t`, and `dcsr_csr_t`).

Do you think there's any good way to avoid that? I haven't researched C++ multiple inheritance in decades.

timsifive: I do not. Personally I don't understand the details of multiple inheritance well enough so I stay away from it.

riscv/csrs.h
timsifive: Thanks for the explanation.

PR #815, opened 2021-09-27
Convert mstatush to csr_t

### LINKED ISSUE #812: mstatush not writable
It seems there is no code in [set_csr()](https://github.com/riscv-software-src/riscv-isa-sim/blob/82c5d885d14fbbb81a856752aa4668333ce8094c/riscv/processor.cc#L979-L1071) to write to `mstatush` in RV32, which is necessary to set MPV=1 before `mret` to enter VS-mode or VU-mode.

@avpatel does your hypervisor software rely on `mstatush.MPV`? It might instead use `hstatus.SPV` and `sret` to enter a virtualized mode; perhaps that's why this issue has not been seen.

scottj97:
Also fixes #812 by making bits `MPV` and `GVA` writable when Hypervisor is enabled.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -442,6 +442,21 @@ bool mstatus_csr_t::unlogged_write(const reg_t val) noexcept {
   return true;
 }
 
+// implement class mstatush_csr_t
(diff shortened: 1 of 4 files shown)


aswaterman:
LGTM.

PR #816, opened 2021-09-28
Convert floating-point CSRs to csr_t

scottj97:
See #793 for rationale.

* frm
* fflags
* fcsr

The only functional change is in the commit logging, where I fixed some [obvious bugs](https://github.com/riscv-software-src/riscv-isa-sim/commit/ede97384f6f760a1a2c8b15f329e3e7c574a0301), and removed logging of fcsr entirely, since that is already covered by the logs of frm and fflags.


aswaterman (approved):
LGTM, thank you!

PR #818, opened 2021-09-29
Convert sentropy and minor CSRs to csr_t

scottj97:
See #793 for rationale.

* sentropy (from the Zkr extension)
* marchid
* mimpid
* mvendorid
* mhartid

The write-only nature of sentropy is unique, and showed a problem when enabling logging, as explained in [this fix](https://github.com/riscv-software-src/riscv-isa-sim/commit/3083b125d69fec064a29f467274a9af13b79698f).

Perhaps it's not useful to log writes to sentropy at all?


aswaterman:
LGTM.

PR #821, opened 2021-09-30
Convert vector CSRs to csr_t -- completes refactoring of CSRs

scottj97:
See #793 for rationale.

This covers the following CSRs, all from the Vector extension:

* vxsat
* vstart
* vxrm
* vl
* vtype
* vlenb
* vcsr

The only functional change is in the commit log, where the writable CSRs above will now appear as expected.

I also cleaned up the remaining code around reading/writing CSRs.

This completes the CSR refactoring begun in #793 and continued in #804, #810, #813, #815, #816, and #818.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1188,8 +1188,8 @@ bool sentropy_csr_t::unlogged_write(const reg_t val) noexcept {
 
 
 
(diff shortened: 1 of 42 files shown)


aswaterman:
Khalas `switch`!  Thanks, Scott!

PR #822, opened 2021-10-04
Fix VSIE CSR write emulation

avpatel:
The VSIE CSR write emulation is broken in latest Spike because
it is allowing only VSSIE bit to be update. This patches fixes
ie_write_mask for VSIE CSR to allow VSTIE, VSSIE, and VSEIE
bits to be updated.

With this patch, we have both Xvisor RISC-V and KVM RISC-V
working again on Spike.

Fixes: 5c1d635c6e56 ("Convert sie/hie/vsie to csr_t family")

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -413,7 +413,7 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   auto vsip_vsie_accr = std::make_shared<generic_int_accessor_t>(this,
                                                                  MIP_VS_MASK,   // read_mask
                                                                  MIP_VSSIP,     // ip_write_mask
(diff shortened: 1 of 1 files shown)


scottj97 (approved):
I expanded my tests to cover this case. I observed the inadvertent behavior change in 5c1d635c and I confirm that this PR fixes it back to the previous, correct behavior.

@neelgala might this explain the behavior change you were seeing a few weeks ago?

aswaterman:
Looks good, thanks for tracking it down.

neelgala:
@scottj97 Yup. I was narrowing it down slowly. Anup got to it early.
Thanks for this patch.

PR #826, opened 2021-10-06
Fix sharing of vxsat between V and P extensions

### LINKED ISSUE #823: vxsat CSR doesn't work with P-ext
`vxsat` requires V-ext enabled. Also it sets some vector states.
Thus kadd16 is illegal if -march=rv64imp.
@scottj97

scottj97:
Fixes #823; see discussion there for context.

Behavior changes:

1. Don't write `vxsat` (or set `mstatus.VS=Dirty` if that field exists) unless the instruction actually saturates
2. If `mstatus.VS` field exists (which today in Spike means `misa.V==1`), P-extension instructions that may write `vxsat` will take an illegal instruction exception if `mstatus.VS==Off`
3. When the `--isa` string contains P but not V, doing `csrrw vxsat` will no longer take an illegal instruction exception

cc @marcfedorow @chuanhua


aswaterman:
Thank you for tackling this, @scottj97.  I'd like a P-extension expert to perform a sanity check before I merge, but the template sounds right to me.

marcfedorow:
LGTM

PR #830, opened 2021-10-13
Split 'P' to EXT_ZPN and friends

marcfedorow:


marcfedorow:
@chuanhua I did not add a permission for FSRIW insns for P-ext but I still think that it should be added
https://github.com/riscv/riscv-p-spec/issues/103

scottj97:
The ZMMUL change means that writing `misa.M=0` no longer disables the multiply instructions. Was that change intentional?

marcfedorow:
> The ZMMUL change means that writing `misa.M=0` no longer disables the multiply instructions. Was that change intentional?

It seems that ZMMUL can be enabled without the whole M-extension.
https://lists.riscv.org/g/tech-unprivileged/topic/81580347#199

marcfedorow:
It is also described in the latest unpriv spec draft
`8.3 Zmmul Extension, Version 0.1`

scottj97:
If I'm reading the ISA string parsing code correctly, setting `--isa=rv64imac` will set both misa.M and EXT_ZMMUL. Is that right?

Is there any way to enable EXT_ZMMUL without also enabling the M extension? I don't see that code.

Is there any way to enable the full M extension and have it controlled by misa.M? I don't see that code.

marcfedorow:
> Is there any way to enable EXT_ZMMUL without also enabling the M extension? I don't see that code.

--isa=rv64ip enables ZMMUL and does not enable the other M-ext insns.
I will add a possibility to enable it explicitly now.



> Is there any way to enable the full M extension and have it controlled by misa.M? I don't see that code.

I do not think that ZMMUL insns should be controlled by misa.M while they should be presented e.g. if misa.P == 1 and misa.M == 0

scottj97:
> I do not think that ZMMUL insns should be controlled by misa.M while they should be presented e.g. if misa.P == 1 and misa.M == 0

That makes this a backward-incompatible change. Today the M extension instructions are all controlled by misa.M. With this PR, there's no way to get that behavior anymore. With this PR, if the M extension is in the ISA string, then misa.M only controls divide and not multiply. That doesn't sound right.

marcfedorow:
> Today the M extension

I may enable the whole 'M' extension is 'P' is enabled -- if this is acceptible.
This is not consistent with [this](https://github.com/riscv/riscv-p-spec/issues/98) tho.

marcfedorow:
Otherwise I may `require_either_extension('M', EXT_ZMMUL)`.
So 'M' will switch off 'M'-ext insns if they are not enabled via `EXT_ZMMUL`

scottj97:
I'm not concerned about the P extension; that behavior is deliberately changing.

I'm concerned about the apparently accidental behavior change of `--isa=rv64imac`. This PR changes the behavior of misa.M for that configuration.

> Otherwise I may require_either_extension('M', EXT_ZMMUL).

Yes, plus make `--isa=rv64imac` not enable EXT_ZMMUL anymore.

marcfedorow:
@scottj97 Are current changes acceptable?

scottj97 (approved):
Yes, this looks good to me now. Thank you!

PR #831, opened 2021-10-14
add support for zfinx v1.0

liweiwei90:
Add support for Zfinx v1.0:
-   Add Zfinx/Zdinx/Zhinx{min} flags
-   Hardwire mstatus_fs to zero
-   Modify extensions check for F/D/ZFH{min}  instructions
-   Modify  F/D/ZFH{min}  instructions to read X regs intead of F regs when Zfinx enabled
-   Add interactive support for Zfinx


riscv/csrs.cc
```diff
@@ -1145,6 +1145,8 @@ float_csr_t::float_csr_t(processor_t* const proc, const reg_t addr, const reg_t
 }
 
 void float_csr_t::verify_permissions(insn_t insn, bool write) const {
+  if (proc->extension_enabled(EXT_ZFINX))
+    return;
```
scottj97: I think this works, but it would be better to add the new extension to the existing check a few lines below:

`if (!proc->extension_enabled('F') && !proc->extension_enabled(EXT_ZFINX))`

The `require_fp` should not hurt anything (after 66c4853bdc5b22bf4c4b364218c713e3f1e487f3).

[I also realize that this method doesn't call `masked_csr_t::verify_permissions()` like it should; I'll file a new PR for that.]

riscv/processor.cc
```diff
@@ -304,6 +310,24 @@ void processor_t::parse_isa_string(const char* str)
       extension_table[EXT_SVPBMT] = true;
     } else if (ext_str == "svinval") {
       extension_table[EXT_SVINVAL] = true;
+    } else if (ext_str == "zfhmin") {
+      if (extension_table[EXT_ZFINX]) {  //z[fdq]inx_zfhmin
```
scottj97: This seems to be dependent on the order. This will work with `--isa=rv64gc_zfinx_zfhmin` but will it work with `--isa=rv64gc_zfhmin_zfinx`?

riscv/decode.h
```diff
@@ -2048,6 +2072,8 @@ reg_t index[P.VU.vlmax]; \
   VI_VFP_COMMON \
   require((P.VU.vsew == e16 && p->extension_enabled('F')) || \
           (P.VU.vsew == e32 && p->extension_enabled('D'))); \
```
scottj97: Shouldn't these lines be deleted?

riscv/insns/fadd_h.h
```diff
@@ -1,5 +1,10 @@
...
+require_either_extension(EXT_ZFH, EXT_ZHINX);
+require_fp_or_zfinx;
 softfloat_roundingMode = RM;
-WRITE_FRD(f16_add(f16(FRS1), f16(FRS2)));
+if(p->extension_enabled(EXT_ZFINX)) {
+  WRITE_FRD_H(f16_add(f16(FRS1_H), f16(FRS2_H)).v);
+} else {
+  WRITE_FRD(f16_add(f16(FRS1), f16(FRS2)));
+}
+
```
scottj97: This is really unfortunate, having to duplicate logic in each and every FP instruction. There's got to be a better way.

Couldn't the `FRS1`, `WRITE_FRD`, etc macros consider Zfinx and do the right thing?

spike_dasm/spike-dasm.cc
```diff
@@ -41,7 +41,11 @@ int main(int argc, char** argv)
     return 1;
   }
 
-  disassembler_t* disassembler = new disassembler_t(xlen);
+  bool zfinx = lowercase.find(std::string("zfinx")) != std::string::npos;
+  bool zdinx = lowercase.find(std::string("zdinx")) != std::string::npos;
+  bool zqinx = lowercase.find(std::string("zqinx")) != std::string::npos;
+
+  disassembler_t* disassembler = new disassembler_t(xlen, zfinx || zdinx || zqinx);
```
scottj97: Can this use the same ISA string processing as `spike_main`?

disasm/disasm.cc
```diff
@@ -507,34 +507,52 @@ static void NOINLINE add_b1type_insn(disassembler_t* d, const char* name, uint32
...
 }
 
-static void NOINLINE add_frtype_insn(disassembler_t* d, const char* name, uint32_t match, uint32_t mask)
+static void NOINLINE add_frtype_insn(disassembler_t* d, const char* name, uint32_t match, uint32_t mask, bool zfinx)
 {
-  d->add_insn(new disasm_insn_t(name, match, mask, {&frd, &frs1, &frs2}));
+  if (zfinx)
+    d->add_insn(new disasm_insn_t(name, match, mask, {&xrd, &xrs1, &xrs2}));
+  else
+    d->add_insn(new disasm_insn_t(name, match, mask, {&frd, &frs1, &frs2}));
```
scottj97: Would it be possible to have Zfinx redefine frd, frs1, frs2 variables, so we could avoid all these `if (zfinx)` with duplicated calls to `d->add_insn()`?

scottj97 (changes requested):

riscv/decode.h
```diff
@@ -250,6 +273,7 @@ class regfile_t
 #define require_either_extension(A,B) require(p->extension_enabled(A) || p->extension_enabled(B));
 #define require_impl(s) require(p->supports_impl(s))
 #define require_fp          require(STATE.sstatus->enabled(SSTATUS_FS))
+#define require_fp_or_zfinx require(p->extension_enabled(EXT_ZFINX) || STATE.sstatus->enabled(SSTATUS_FS))
```
scottj97: This should not be necessary since 66c4853bdc5b22bf4c4b364218c713e3f1e487f3; if `mstatus.FS` does not exist, `require_fp` will do nothing.

riscv/csrs.cc
liweiwei90: Ok, I'll change it later.

riscv/processor.cc
liweiwei90: In the zfinx spec(https://github.com/riscv/riscv-zfinx/blob/main/Zfinx_spec.adoc#configurations), the zfinx version for zfh is Z[FDQ]inx_Zfh{,min}. So I think rv64gc_zfhmin_zfinx is illegal. the same to 'V'(Z[FDQ]inx_V).

riscv/decode.h
liweiwei90: Yes. This may be added when rebased to upstream. I'll delete it later.

riscv/decode.h
liweiwei90: Ok. I'll change this later.

disasm/disasm.cc
liweiwei90: I didn't find a good way to get the zfinx parameter in frd definition. So I pass it from the constructor.

riscv/insns/fadd_h.h
liweiwei90: The type (or size) of the register is important for zfinx when reading the source: Firstly. the NaN-Boxing of Zfinx is different from F/D. Secondly, if we read double data in RV32, register-pair is used.
WRITE_FRD probably can be reused, since the type of the written data passed to it is different. However, I think reusing it is not necessary when the others cannot be reused.

spike_dasm/spike-dasm.cc
liweiwei90: Yes. However, the ISA parser in spike-dasm is very simple: it only parse the xlen info before. So I don't know whether is it necessary to use the same ISA string processing as spike_main.

aswaterman: I think we should move to having unified ISA-string processing, even if it results in minor backwards incompatibilities for spike-dasm.

liweiwei90: Do we need to add instructions to disassembler according to the ISA-string?  Or just move the current code to unified ISA-string processing?

riscv/insns/fadd_h.h
liweiwei90: Maybe we can reuse FRS1/2/3_H/S/D and WRITE_FRD_H/S/D in normal case. Change the above code into :
 `WRITE_FRD_H(f16_add(FRS1_H, FRS2_H));`
 is this OK?

disasm/disasm.cc
liweiwei90: Or we pass the zfinx as global variable？

riscv/insns/fadd_h.h
aswaterman: I don't think that change is necessary either.  We are already relying on overloading and conversions to perform NaN boxing and unboxing.  It seems conceivable that we can use the same mechanisms to avoid boxing and unboxing, i.e., only `decode.h` need be modified.

disasm/disasm.cc
aswaterman: Globals are not an acceptable alternative.

spike_dasm/spike-dasm.cc
aswaterman: The latter, not the former.  Seeing "unknown" all over the place is not helpful; it's better to disassemble as much as possible.

riscv/insns/fadd_h.h
liweiwei90: The problem is how to determine the size of data will be read for FRS1:
```
#define FRS1_H freg(f16(STATE.XPR[insn.rs1()] & (uint16_t)-1))
#define FRS1_F freg(f32(STATE.XPR[insn.rs1()] & (uint32_t)-1))
#define FRS1_D READ_ZDINX_REG(insn.rs1())

#define READ_ZDINX_REG(reg) (xlen == 32 ? freg(f64(READ_REG_PAIR(reg))) : freg(f64(STATE.XPR[reg] & (uint64_t)-1)))
```
OR we should need the outside f16/f32 (f16(FRS1)) overloading and conversions to get the right data and FRS1 only return the  STATE.XPR[insn.rs1()] ?
However, this will be not suitable for FRS1_D, since it should read register pair in RV32 (this cannot be used in H/F).

aswaterman: Good point, `FRS1_D` for RV32_Zdinx will require an alternate approach regardless.

I think you should proceed with your proposal `WRITE_FRD_H(f16_add(FRS1_H, FRS2_H))`.

disasm/disasm.cc
liweiwei90: Or we change the code into:
```
  d->add_insn(zfinx? new disasm_insn_t(name, match, mask, {&xrd, &xrs1, &xrs2}) : 
                     new disasm_insn_t(name, match, mask, {&frd, &frs1, &frs2}));
```

spike_dasm/spike-dasm.cc
liweiwei90: Ok, I'll try to move them.

disasm/disasm.cc
aswaterman: Or make e.g. an `FRD` macro that expands to something like `(zfinx ? &xrd : &frd)`?  I am just spitballing here; not sure this is the best alternative.

riscv/insns/fadd_h.h
liweiwei90: OK.

disasm/disasm.cc
liweiwei90: Unfortunately, this cannot pass the compile:
```
../disasm/disasm.cc:518:76: error: conditional expression between distinct pointer types ‘<unnamed struct>*’ and ‘<unnamed struct>*’ lacks a cast
 d_insn(zfinx ? new disasm_insn_t(name, match, mask, {zfinx? &xrd : &frd, &xrs1}) :
```

aswaterman: I'm speaking abstractly, of course.  Think about other approaches along similar lines.

liweiwei90: I find it's possible to add zfinx viriable in frd/frs1/2/3, just as global variable:

```
struct : public arg_t {
  bool zfinx;
  std::string to_string(insn_t insn) const {
    return zfinx ? xpr_name[insn.rd()] : fpr_name[insn.rd()];
  }
} frd;
```

riscv/processor.cc
scottj97: If it's illegal will the parsing here generate an error message? Or will it silently do something weird?

spike_dasm/spike-dasm.cc
scottj97: If/when Spike supports conflicting extensions (with overlapping opcodes) then we'd need to have the disassembler consider more of the ISA string. But for now it can simply always support all extensions.

riscv/decode.h
```diff
@@ -2083,23 +2125,23 @@ reg_t index[P.VU.vlmax]; \
   switch(P.VU.vsew) { \
     case e16: {\
       float16_t &vd = P.VU.elt<float16_t>(rd_num, i, true); \
-      float16_t rs1 = f16(READ_FREG(rs1_num)); \
+      float16_t rs1 = p->extension_enabled(EXT_ZFINX) ? f16(READ_REG(rs1_num) & (uint16_t)-1) : f16(READ_FREG(rs1_num)); \
```
scottj97: Can this be FRS1_H now? And same comment for a few more places below in these vector macros.

disasm/disasm.cc
```diff
@@ -643,6 +647,11 @@ disassembler_t::disassembler_t(int xlen)
   const uint32_t mask_amoop = 0x1fUl << 27;
   const uint32_t mask_width = 0x7Ul << 12;
 
+  frd.zfinx = zfinx;
+  frs1.zfinx = zfinx;
+  frs2.zfinx = zfinx;
+  frs3.zfinx = zfinx;
```
scottj97: I know @aswaterman doesn't want a global for `zfinx` but as `frd`, `frs1`, etc are already globals, this `zfinx` is effectively a global too. Is there really any benefit of putting it into each of these 4 classes?

spike_dasm/spike-dasm.cc
```diff
@@ -14,6 +14,32 @@
 #include <fesvr/option_parser.h>
 using namespace std;
 
+struct isa_config {
+  int xlen;
+  bool zfinx;
+};
+
+bool parse_isa_string(const char* str, struct isa_config* isa) {
```
scottj97: I don't see how this is "unified" ISA string processing. What I meant was to call `processor_t::parse_isa_string()` from here, or factor out that code to some common routine that both `processor_t` and `spike_disasm` can each call.

Though maybe that should be a separate PR for later.

scottj97 (changes requested):
This PR is much improved, thank you.

Please squash the fixes back into the originals; specifically:
* Squash d12e4b2eecb94d87a3c95d16881f199e10025884 into 9c70bc3722d6a245a4204c33628e8d388724e239
* Squash e4db39bce7421556f5fed4fad7982274b9e9215c into 955bc01d669b195ffc335a790565bb1af8789b4d
* Squash ef82cc023d73d1742e289e5f74c2a81f691af8c8 into 9c70bc3722d6a245a4204c33628e8d388724e239

riscv/decode.h
liweiwei90: No, the reg index of FRS1_H is insn.rs1().

disasm/disasm.cc
liweiwei90: > This PR is much improved, thank you.
> 
> Please squash the fixes back into the originals; specifically:
> 
>     * Squash [d12e4b2](https://github.com/riscv-software-src/riscv-isa-sim/commit/d12e4b2eecb94d87a3c95d16881f199e10025884) into [9c70bc3](https://github.com/riscv-software-src/riscv-isa-sim/commit/9c70bc3722d6a245a4204c33628e8d388724e239)
> 
>     * Squash [e4db39b](https://github.com/riscv-software-src/riscv-isa-sim/commit/e4db39bce7421556f5fed4fad7982274b9e9215c) into [955bc01](https://github.com/riscv-software-src/riscv-isa-sim/commit/955bc01d669b195ffc335a790565bb1af8789b4d)
> 
>     * Squash [ef82cc0](https://github.com/riscv-software-src/riscv-isa-sim/commit/ef82cc023d73d1742e289e5f74c2a81f691af8c8) into [9c70bc3](https://github.com/riscv-software-src/riscv-isa-sim/commit/9c70bc3722d6a245a4204c33628e8d388724e239)

OK. I'll do this later.

liweiwei90: > I know @aswaterman doesn't want a global for `zfinx` but as `frd`, `frs1`, etc are already globals, this `zfinx` is effectively a global too. Is there really any benefit of putting it into each of these 4 classes?

We will not add new global variable  apparently if add  'zfinx' in frd, frs1, etc each.

spike_dasm/spike-dasm.cc
liweiwei90: Sorry, I misunderstood the idea. I'll try to fix it later.

riscv/processor.cc
liweiwei90: It will generate error message, such as rv32ima_zfh_zfinx will generate:
`error: bad --isa option 'rv32ima_zfh_zfinx'. 'Zfh' extension requires 'F'`

rv32imaf_zfh_zfinx will generate:
`error: bad --isa option 'rv32imaf_zfh_zfinx'. zfinx extension conflict with F/D/Q/ZFH{min} extension`

disasm/disasm.cc
liweiwei90: If frd, frs1 is class, zfinx can be a private member of them, and passed from the construction

scottj97: > If frd, frs1 is class, zfinx can be a private member of them, and passed from the construction

Yes, I looked into that possibility, but these are global variables that are referenced throughout the file, so they need file scope. We could change them to type `arg_t*` and then construct them in `disassembler_t()` constructor, but then those four variables would be different than all the rest.

Maybe we could put all these arg_t subclasses into class disassembler_t so they aren't globals? And make `add_*_insn()` all into methods of `disassembler_t`?

But that's a big change worthy of its own PR. So probably best to just leave it like this.

riscv/decode.h
scottj97: That should still work, if I'm reading this correctly:

https://github.com/riscv-software-src/riscv-isa-sim/blob/d12e4b2eecb94d87a3c95d16881f199e10025884/riscv/decode.h#L657-L659

riscv/processor.cc
scottj97: > It will generate error message, such as rv32ima_zfh_zfinx will generate: `error: bad --isa option 'rv32ima_zfh_zfinx'. 'Zfh' extension requires 'F'`

But the Zfh extension does not actually require F, apparently? The code supports rv32ima_zfinx_zfh. The error message is very misleading.

> rv32imaf_zfh_zfinx will generate: `error: bad --isa option 'rv32imaf_zfh_zfinx'. zfinx extension conflict with F/D/Q/ZFH{min} extension`

This also sounds misleading, since it looks like the code supports both rv32imac_zfinx_zfh and rv32imac_zfinx_zfhmin.

liweiwei90: zfh actually requires F, the original code is:
```
    if (ext_str == "zfh") {
      if (!((max_isa >> ('f' - 'a')) & 1))
        bad_isa_string(str, "'Zfh' extension requires 'F'");
```
I think zfh_zfinx means support original zfh and zfinx, however zfinx_zfh means support zfinx and zhinx(zfinx version of zfh)

the code do support both  rv32imac_zfinx_zfh and rv32imac_zfinx_zfhmin.

riscv/decode.h
liweiwei90: Ok, I'll try to modify it.

riscv/processor.cc
scottj97: > zfh actually requires F

> the code do support both rv32imac_zfinx_zfh

These two statements are mutually exclusive, since the second has zfh but not F, and is supported.

There are a large number of extensions and combinations of those extensions, and I think users will be very confused if the error messages are unclear about what is and isn't supported.

liweiwei90: Ok. The error message may be "Base 'Zfh' extension requires 'F'. Or Zfinx version of 'Zfh' should be 'Z[FDQ]inx_zfh'"

riscv/processor.h
```diff
@@ -276,8 +276,37 @@ static int cto(reg_t val)
   return res;
 }
 
+class isa_parser_t {
+public:
+  isa_parser_t(): xlen(0), extension_table(256, false){};
```
scottj97: This is great.

Is there any reason the constructor can't take the ISA string as a parameter? Instead of having to call `parse_isa_string()` later to initialize it?

liweiwei90: Taking the ISA string as a parameter of the constructor is Ok. The reason why I call parse_isa_string() later to initialize it is just to preserve the original code structure. I'll fix it later.

riscv/processor.cc
```diff
@@ -310,6 +315,26 @@ void processor_t::parse_isa_string(const char* str)
...
+          bad_isa_string(str, "Base 'Zfhmin' extension requires 'F'. Or Zfinx version of 'Zfhmin' should be 'Z[FDQ]inx_zfhmin'");
+        extension_table[EXT_ZFHMIN] = true;
+      }
+    } else if (ext_str == "zfinx") {
+      extension_table[EXT_ZFINX] = true;
+    } else if (ext_str == "zdinx") {
+      extension_table[EXT_ZDINX] = true;
+      extension_table[EXT_ZFINX] = true;
+    } else if (ext_str == "zqinx") {
+      extension_table[EXT_ZQINX] = true; //not work currently
```
scottj97: One last nit: if Zqinx doesn't work, this should give a reasonable error message, instead of silently doing the wrong thing.

scottj97 (changes requested):

riscv/processor.cc
liweiwei90: Ok, Zqinx flag is defined at the beginning of the v0.41 spec, However, there is no further description about it. Maybe we can just delete this flag or report message "Zqinx is not supported currently." here.

aswaterman: Just delete this case.  No need to have a special error message Zqinx--it should issue the same error message as any other extension Spike doesn't implement.

liweiwei90: OK.

riscv/processor.cc
```diff
@@ -330,9 +330,10 @@ isa_parser_t::isa_parser_t(const char* str)
       extension_table[EXT_ZDINX] = true;
       extension_table[EXT_ZFINX] = true;
     } else if (ext_str == "zqinx") {
-      extension_table[EXT_ZQINX] = true; //not work currently
+      extension_table[EXT_ZQINX] = true;
       extension_table[EXT_ZDINX] = true;
       extension_table[EXT_ZFINX] = true;
+      bad_isa_string(str, "'ZQinx' is not supported currently");
```
aswaterman: Just delete all of this code.  The standard error handling is fine; no need for a Zqinx-specific message.

riscv/processor.cc
aswaterman: BTW, Zqinx isn't part of the standard, anyway.  See this note: https://github.com/riscv/riscv-isa-manual/blob/399c9a759eb4540a65c60e2cc236164821ff2346/src/zfinx.tex#L140

riscv/processor.cc
liweiwei90: OK.

riscv/processor.cc
liweiwei90: OK. Thanks.

scottj97:
It looks like @lazyparser reverted something... 17402ce looked good to me, but now it's back to f8b0065 from 2 days ago.

liweiwei90:
> It looks like @lazyparser reverted something... [17402ce](https://github.com/riscv-software-src/riscv-isa-sim/commit/17402cee701e0f62fb232a3589ac9174529b5040) looked good to me, but now it's back to [f8b0065](https://github.com/riscv-software-src/riscv-isa-sim/commit/f8b0065ca32f148c23b358c1a23a68d1e7f50130) from 2 days ago.

Sorry. Something went wrong with our code sync mechanism.

scottj97 (approved):

disasm/disasm.cc
aswaterman: I still don't think this is acceptable, because it is functionally incorrect in the presence of multiple disassemblers with different zfinx settings.  Maybe that's contrived, but since this is the only situation that breaks that abstraction, I don't want to merge it until it's somehow fixed.

liweiwei90: Zfinx version float point extension reuse the same opcode as original float point extension. The only difference is that zfinx version will reuse integer register instead of float point register. So their disassemble code are only different in the registers and will not be distinguished by their opcode without additional zfinx settings. 
I didn't find another similar situation. And representing its disassemble code the same as original float point code will only have minor effect on our understanding of code logic. Maybe we can revert these changes and remain the original disasm code unchanged if this is not acceptable by current abstraction.

aswaterman: In that case, my preference is to revert the changes to the disassembler, so that we can merge this PR.  Later, when we figure out how we want to make the disassembler changes, we can do so as a follow-up.

riscv/processor.h
```diff
@@ -300,12 +325,10 @@ class processor_t : public abstract_device_t
...
     return xlen;
   }
-  unsigned get_max_xlen() { return max_xlen; }
-  std::string get_isa_string() { return isa_string; }
   unsigned get_flen() {
     return extension_enabled('Q') ? 128 :
-           extension_enabled('D') ? 64 :
-           extension_enabled('F') ? 32 : 0;
+           extension_enabled('D') || extension_enabled(EXT_ZDINX) ? 64 :
+           extension_enabled('F') || extension_enabled(EXT_ZFINX) ? 32 : 0;
```
aswaterman: This doesn't make sense.  FLEN refers to the width of an F-register.  For Zfinx and Zdinx, there are no F-registers, and therefore it doesn't make sense to talk about their length.

disasm/disasm.cc
liweiwei90: Ok.

riscv/processor.h
liweiwei90: The FLEN is also used in Zfinx. It refers to the width of virual 'F-register'(which store the value in integer register or registers, such as https://github.com/riscv/riscv-zfinx/blob/main/Zfinx_spec.adoc#rv32fd-zfh-ZFinx-register-pair-handling).

aswaterman: That's not actually the Zfinx spec; that is an out-of-date proposal.  The actual spec is here https://github.com/riscv/riscv-zfinx/blob/main/zfinx-1.0.0-rc.pdf and makes no mention of the term FLEN.  FLEN still means "the width of an F register".

liweiwei90: Ok. I'll try to change this.

tariqkurd-repo: Hi @aswaterman there's no defined discovery mechanism yet, and certainly nothing in the Z[fd]inx spec about detecting which of these two is implemented. I agree that FLEN is the incorrect term as there are no F-registers, but the model does need to know which of these two are implemented, so how should it distinguish between them? I suppose setting a Zfinx_valid / Zdinx_valid flag would do it.

aswaterman: The discovery mechanism is irrelevant here, because we're talking about a data structure internal to Spike.  `processor_t::extension_enabled(EXT_ZFINX)` suffices - that's equivalent to your "zfinx_valid" flag.

liweiwei90: @aswaterman Do I need to revert the unified ISA-string processing for spike-dasm and spike?
> In that case, my preference is to revert the changes to the disassembler, so that we can merge this PR. Later, when we figure out how we want to make the disassembler changes, we can do so as a follow-up.

liweiwei90:
Stateen related check for float point instructions have been added.

riscv/v_ext_macros.h
```diff
@@ -1802,15 +1802,15 @@ reg_t index[P.VU.vlmax]; \
...
+      float32_t rs1 = f16_to_f32(FRS1_H); \
       BODY16; \
       set_fp_exceptions; \
       break; \
     } \
     case e32: { \
       float64_t &vd = P.VU.elt<float64_t>(rd_num, i, true); \
       float64_t vs2 = f32_to_f64(P.VU.elt<float32_t>(rs2_num, i)); \
-      float64_t rs1 = f32_to_f64(FRS1_F)); \
+      float64_t rs1 = f32_to_f64(FRS1_F); \
```
scottj97: I don't understand the changes to this file. This doesn't seem to have anything to do with the rest of this commit.

riscv/decode.h
```diff
@@ -330,6 +330,23 @@ do { \
                         ((STATE.prv == PRV_U) && (s##field == 0)))) \
       throw trap_virtual_instruction(insn.bits()); \
   } while (0);
+#define require_stateen_fcsr \
```
scottj97: This is an unfortunate copy/paste from `fcsr_csr_t::verify_permissions()`, which is a public member. Can you simply call `STATE.fflags->verify_permissions()` here?

On a related note, shouldn't the code in `fcsr_csr_t::verify_permissions()` belong in `float_csr_t::verify_permissions()`? It seems that the sstaten bit should apply to both `frm` and `fflags` and not only the composite `fcsr`, right?

riscv/insns/fadd_d.h
```diff
@@ -1,5 +1,6 @@
 require_either_extension('D', EXT_ZDINX);
 require_fp;
+require_stateen_fcsr;
```
scottj97: Can this check be part of `require_fp`? The Smstateen spec says:

> For convenience, when the stateen CSRs are implemented and misa.F = 0, then if bit 1 of a controlling stateen0 CSR is zero, all floating-point instructions cause an illegal instruction trap (or virtual instruction trap, if relevant), as though they all access fcsr, regardless of whether they really do.

riscv/v_ext_macros.h
liweiwei90: Sorry, I fixed this mistake, which was introduced when I tried to rebase to upstream, in the wrong commit. I'll update this later.

riscv/decode.h
liweiwei90: I'm not sure about whether stateen.FCSR bit should apply to frm and fflags which is not stated in the spec. They doesn't currently. And fcsr is not an element of state. So I copied the check logic here or Maybe I need add fcsr to state or add verification_csr function to search CSR_FCSR from csrmap and do the verify_permissions

riscv/insns/fadd_d.h
liweiwei90: It's OK to take it into require_fp from the perspective of function realization. However, I'm not sure whether the logic of stateen check which is to control the access of FCSR is suitable for the original logic of require_fp.
.

riscv/decode.h
aswaterman: fflags and frm are part of fcsr, so anything that controls access to fcsr must also control access to fflags and frm.

liweiwei90: OK. I agree with this. I'll send a fix to this.

scottj97: PR #1059 fixes this

scottj97 (approved):

aswaterman (approved):

PR #832, opened 2021-10-14
Properly check CSR permissions for floating-point CSRs

scottj97:
We were not properly checking for Machine-level or Supervisor-level CSR accessibility on `frm`, `fflags`, or `fcsr`.

Since those three are all User-accessible, this has no functional effect, but it's the safe thing to do.


aswaterman:
Yeah, it's seemingly a no-op, but this will reduce the risk that copying-and-pasting of code loses the permissions check elsewhere.

PR #801, opened 2021-09-13
Added insn overlapping (bitmanip / krypto)

marcfedorow:

PR #838, opened 2021-10-18
Delete SHAMT check in pack

liweiwei90:
I didn't find any description about SHAMT check in pack instruction from the spec. 
This check will make the ACT test of pack failed


scottj97:
Please explain in the commit message the _why_ of this change. That's much more important for the historical record than the _what_.

aswaterman:
Yeah, this just looks like a bug introduced in #801.  SHAMT is not an argument to this instruction.

Scott's of course right, but I'll squash and merge this, so that the commit message will point back to this PR.

marcfedorow:
It seems like packw needs the same fixup.

aswaterman:
@marcfedorow can you prepare a PR? Thanks.

PR #842, opened 2021-10-26
Fixed a segmentation fault bug

eric-xtang1008:
After executing vector load/store whole register instructions,
            spike would be crashed when executed the next vector instruction.

Signed-off-by: Eric Tang <tangxingxin1008@gmail.com>


aswaterman:
thanks!

scottj97:
I broke this in 72a6f9fac8f3c058d72af54a8554ee6fab2627ec -- thank you for the fix.

PR #844, opened 2021-10-28
Fix a link error when compiled without optimization option "-O2"

eric-xtang1008:
Error log:
  libriscv.a(processor.o): In function `state_t::reset(processor_t*, unsigned long)':
  .../_build/../riscv/processor.cc:509: undefined reference to `state_t::num_triggers'
  collect2: error: ld returned 1 exit status
  Makefile:349: recipe for target 'spike-log-parser' failed

Signed-off-by: Eric Tang <tangxingxin1008@gmail.com>

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -339,6 +339,8 @@ static int xlen_to_uxl(int xlen)
   abort();
 }
 
+const int state_t::num_triggers;
+
 void state_t::reset(processor_t* const proc, reg_t max_isa)
 {
   pc = DEFAULT_RSTVEC;


chihminchao (commented):
Please also handle this 
https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/processor.h#L199

eric-xtang1008:
> Please also handle this https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/processor.h#L199


ord-used of a variable need a definition like as 
csrmap[CSR_TDATA2] = tdata2 = std::make_shared<tdata2_csr_t>(proc, CSR_TDATA2, num_triggers); //ord-used
but there is no ord-used of this variable [static const int max_pmp = 16;] in code,  do we need handle this?

scottj97 (approved):
TIL what _odr-used_ means.

Looks like I broke this in 428c2aa1f074e8280fb9fc799e38c7289abc7c74.

The fix looks correct and I concur with @eric-xtang1008 that the same fix is unnecessary for `max_pmp` since it is not _odr-used_. (Should not hurt though, and might fix a similar problem in the future.)

aswaterman:
How 'bout that.

PR #845, opened 2021-10-29
Added instruction and data trace CSV output capability.

andyduller:
Hi Andrew/Scott,
As promised a pull request for the instruction and data trace patch. I'm afraid it is not using a C++ class, which would probably be best, our only excuse is that the initial patch was simple and was done many years ago now ;-)
Any questions please let myself and Gadge know.
Thanks,
Andy Duller

### DIFF
--- a/riscv/cachesim.h
+++ b/riscv/cachesim.h
@@ -49,7 +49,7 @@ class cache_sim_t
   size_t idx_shift;
 
   uint64_t* tags;
-  
+
   uint64_t read_accesses;
   uint64_t read_misses;
   uint64_t bytes_read;
@@ -107,9 +107,9 @@ class icache_sim_t : public cache_memtracer_t
   {
     return type == FETCH;
   }
-  void trace(uint64_t addr, size_t bytes, access_type type)
+  void trace(uint64_t addr, size_t len_bytes, access_type type, const uint8_t* bytes)
   {
-    if (type == FETCH) cache->access(addr, bytes, false);
+    if (type == FETCH) cache->access(addr, len_bytes, false);
   }
 };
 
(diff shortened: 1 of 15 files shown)


aswaterman:
@andyduller Thank you for the patch! But since I hear you won’t be working with us to iterate on it and revise it, I don’t think it’s actionable.  We’ll leave it open for a while in case someone else is willing to run with it.

scottj97:
Where is the output CSV format described?

Does this work when there's multiple processors?

As Andy hinted, this ought to be a C++ class instantiated inside each `processor_t` instead of a bunch of global variables.

This really needs to be a series of Git commits, not just one giant commit.

aswaterman:
I think the next logical step is to address all of @scottj97's questions and concerns, then I'll help with the re-review.

scottj97:
I also wonder if this could somehow be integrated into the commit log facility, instead of having a completely separate mechanism for approximately the same information.

gajinderpanesar:
Scott,

  Good point. The CSV format needs to be specified somewhere - we had it as
a separate internal document for instruction trace about 4 years ago when
we did the original patch. The format now needs to cover all the new stuff
added. Let me look at that. It does not make sense to put it in the E-trace
specification. Let me first bring the specification upto date and then we
can discuss where it is best to put it.
  The current patch is for single processors - you would typically have one
trace encoder per processor.
  Agreed on the C++ class. We will have a look at that.
  I will think about how best to split this up into smaller commits once we
have considered the C++ route properly.

  Thanks,

    Gadge

andyduller:
Hi Andrew/Scott,
Apologies for the delay in updating this. I have now started to develop the C++ version of the changes and so I will close this request and open another one. Hopefully that will be in the next few days.
Thanks,
Andy

PR #846, opened 2021-11-01
Krypto 1.0 changes.

mjosaarinen:
Related to https://github.com/riscv/riscv-opcodes/pull/84
- Arch review naming changes: `xperm.n` and `xperm.b` are now `xperm4` and `xperm8`, `rev.b` is `brev8`. Note that `xperm4`/`8` is a Krypto instruction (Xbkx) as it was not included in Bitmanip 1.0
- The entropy source CSR  is now in user mode (0x015), the name is `seed` (not `sentropy`/`mentropy`) and the pseudo instructions are gone.
- Missing from K implementation: `mseccfg.sseed` and `mseccfg.useed` (since the rest of `mseccfg` is also missing.) These bits are already defined in the privileged spec.
- Also includes a diff to encodings.h (auto-generated from opcodes) to allow building spike.


aswaterman:
The "K extension" is no longer a thing, right?  i.e. should we update the readme to say `Zk` instead?

riscv/csrs.cc
```diff
@@ -1181,23 +1181,24 @@ bool composite_csr_t::unlogged_write(const reg_t val) noexcept {
...
 
-sentropy_csr_t::sentropy_csr_t(processor_t* const proc, const reg_t addr):
+seed_csr_t::seed_csr_t(processor_t* const proc, const reg_t addr):
   csr_t(proc, addr) {
 }
 
-void sentropy_csr_t::verify_permissions(insn_t insn, bool write) const {
+void seed_csr_t::verify_permissions(insn_t insn, bool write) const {
   /* Read-only access disallowed due to wipe-on-read side effect */
+  /* XXX mseccfg.sseed and mseccfg.useed should be verified. */
```
aswaterman, line 1190: Just to confirm: the reason you think it's OK to commit this `XXX` is that `mseccfg` hasn't been implemented yet, and so this is legitimately future work.  I'm OK with this.

disasm/disasm.cc
```diff
@@ -855,7 +855,7 @@ disassembler_t::disassembler_t(int xlen)
 
   DEFINE_NOARG(ecall);
   DEFINE_NOARG(ebreak);
-  DEFINE_NOARG(uret);
+//  DEFINE_NOARG(uret);
```
aswaterman: We shouldn't have commented-out code.  (I'm OK with removing `uret` as a separate PR, though.)

riscv/entropy_source.h
```diff
@@ -53,8 +53,8 @@ class entropy_source {
     if(return_status == OPST_ES16) {
 
         // Add some sampled entropy into the low 16 bits
-        uint16_t seed  = this -> get_two_random_bytes();
-        result        |= seed;
+        uint16_t entropy  = this -> get_two_random_bytes();
+        result        |= entropy;
```
aswaterman: The formatting here is weird; just use this:

```
        uint16_t entropy = this -> get_two_random_bytes();
        result |= entropy;
```

(As a separate matter, this entire file uses a different formatting discipline from the rest of Spike, but fixing that should be done in another PR.)

riscv/csrs.cc
mjosaarinen: Yes once mseccfg is there, we will add those access control features in a consistent way.

mjosaarinen: > Just to confirm: the reason you think it's OK to commit this `XXX` is that `mseccfg` hasn't been implemented yet, and so this is legitimately future work. I'm OK with this.

Yes once mseccfg is there, we will add those access control features in a consistent way.

mjosaarinen:
> The "K extension" is no longer a thing, right? i.e. should we update the readme to say `Zk` instead?

While `Zk` would be a useful shorthand for all features in the crypto specification (e.g. for Spike command-line invocation) I don't think it's officially defined any more. So for the purposes of the README, perhaps saying that "RISC-V Cryptography Extensions" version 1.0 (which only has "Scalar & Entropy Source Instructions") seems accurate.  Version 2.0 support would then imply scalar cryptography etc. The desire is of course to keep Spike support as complete as possible in this respect.

aswaterman:
The items in the README list are generally meant to be passed in as ISA strings; you'll see that the bitmanip extensions are listed out individually (Zba, Zbb, Zbb, Zbs).  It would be more consistent to list out the individual extensions that are supported, rather than using an umbrella term that can't be used as an ISA string.

mjosaarinen:
Looking at the spec again, Zk actually means a part of the extension that excludes Zksed and Zksh (the Chinese ciphers). So that's why saying Zk is not entirely accurate. The subtitle of the document says _"Scalar & Entropy Source Instructions"_ and entropy source of course applies to all kinds of cryptography and is functionally independent of those. Hence a different bullet point for Zkr.

scottj97:
Please rebase and force-push onto your branch, instead of merging master into this branch like a3cb337

riscv/encoding.h
```diff
@@ -38,8 +38,6 @@
 
 #define MSTATUSH_SBE        0x00000010
 #define MSTATUSH_MBE        0x00000020
-#define MSTATUSH_GVA        0x00000040
-#define MSTATUSH_MPV        0x00000080
```
scottj97: It seems this file has been hand-edited recently, instead of generating it from riscv-opcodes. (I know I am guilty of this.) There are several changes in this commit that seem to be reverting recent hand changes to this file.

It would be best to update riscv-opcodes as needed first, auto-generate a clean version of this encoding.h that matches what had been hand changes before, then make any changes needed for Krypto as a separate commit.

Especially the removal of `uret` and related CSRs should be a separate commit or even separate PR.

riscv/encoding.h
```diff
@@ -954,9 +954,9 @@
 #define MATCH_SINVAL_VMA 0x16000073
 #define MASK_SINVAL_VMA  0xfe007fff
 #define MATCH_SFENCE_W_INVAL 0x18000073
-#define MASK_SFENCE_W_INVAL  0xfff07fff
+#define MASK_SFENCE_W_INVAL  0xffffffff
 #define MATCH_SFENCE_INVAL_IR 0x18100073
-#define MASK_SFENCE_INVAL_IR  0xfff07fff
+#define MASK_SFENCE_INVAL_IR  0xffffffff
```
scottj97: This looks suspicious...was it incorrect before?

riscv/encoding.h
```diff
@@ -1033,6 +1033,10 @@
 #define MASK_FCVT_H_L  0xfff0007f
 #define MATCH_FCVT_H_LU 0xd4300053
 #define MASK_FCVT_H_LU  0xfff0007f
+#define MATCH_POLLENTROPY 0xf1500073
+#define MASK_POLLENTROPY  0xfffff07f
+#define MATCH_GETNOISE 0x7a900073
+#define MASK_GETNOISE  0xfffff07f
```
scottj97: This commit adds these defines, then the following commit removes them. If they shouldn't be there then we should fix that here in the first commit, instead of adding a second commit to correct this one.

riscv/encoding.h
```diff
@@ -3005,7 +3001,8 @@
 #define CSR_MARCHID 0xf12
 #define CSR_MIMPID 0xf13
 #define CSR_MHARTID 0xf14
-#define CSR_SENTROPY 0x546
+#define CSR_MENTROPY 0xf15
+#define CSR_MNOISE 0x7a9
```
scottj97: Same comment as above with MATCH_POLLENTROPY et al.

scottj97 (changes requested):
Please clean up the commit messages; "commenting out is bad" and "white space formatting" don't seem to make sense anymore.

riscv/encoding.h
mjosaarinen: Yeah they were squashed together and this commit was gone.

riscv/encoding.h
mjosaarinen: It was already a separate PR and was merged before this one.

scottj97: I guess you're talking about #847? That removed one line in disasm.cc.

If it had already been removed in a separate PR then it wouldn't be showing up here in this diff.

riscv/encoding.h
scottj97: How is this resolved? It's still there in 236fd642967bddeece5312b6a8fa52c5ab693b38

riscv/encoding.h
mjosaarinen: Andrew requested me to file a separate PR while he was reviewing this, so that's how it was gone. You're currently commenting on a squashed (outdated) commit of an autogenerated header file.

mjosaarinen: .. which I probably shouldn't have included in the original commit. I wasn't expecting to have to split and squash it into things that go into separate repos *sigh*

scottj97: I'm commenting on 236fd642967bddeece5312b6a8fa52c5ab693b38 which is in your latest mjosaarinen:master branch as of this second.

mjosaarinen: Ok.. the PR is for these bits: https://github.com/riscv/riscv-opcodes/pull/86

riscv/encoding.h
mjosaarinen: Outdated.

riscv/encoding.h
mjosaarinen: Via riscv-opcodes that generates this file.

scottj97: @daniellustig in cb8f09a4d6fed30527fdd832b885c898b4591a5f you added these mask values by hand. For some reason the values here don't match with riscv-opcodes, resulting in this diff when we recreate the encoding.h from riscv-opcodes.

Can you confirm which repo has the correct mask value?

daniellustig: `0xffffffff` is correct.  I originally added the values manually into spike because at the time, encoding.h in riscv-isa-sim and encoding.h in riscv-opcodes had diverged.  I then later fixed a bug in the version in riscv-opcodes (https://github.com/riscv/riscv-opcodes/pull/78), but then apparently I never propagated that fix over to spike.  My mistake, sorry.

If you don't mind these two lines of diff being mixed into your unrelated PR, please feel free to just include them into your merge.  Otherwise, I can push a separate fix for those two lines.  Let me know which you prefer.

(I assume there's still not a more automated process of keeping encoding.h in sync across the various repos that have a copy?)

scottj97: Fixed by #848; @mjosaarinen please rebase, regenerating encoding.h from the latest riscv-opcodes, and it should remove all these non-Krypto changes from encoding.h.

riscv/encoding.h
```diff
@@ -56,6 +56,10 @@
 #define SSTATUS_UXL         0x0000000300000000
 #define SSTATUS64_SD        0x8000000000000000
 
+#define SSTATUS_VS_MASK     (SSTATUS_SIE | SSTATUS_SPIE | \
+                             SSTATUS_SPP | SSTATUS_FS | SSTATUS_SUM | \
+                             SSTATUS_MXR | SSTATUS_UXL)
+
```
scottj97: This was not generated from the latest riscv-opcodes. If it was, the hash at the top of this file would have changed, and this SSTATUS_VS_MASK would not be here.

scottj97 (changes requested):

riscv/encoding.h
mjosaarinen: Regenerated from latest.

scottj97 (approved):
Looks good, thank you for your patience.

aswaterman (approved):
LGTM, too; thanks, @mjosaarinen!

PR #850, opened 2021-11-04
Set GVA properly on breakpoint traps

scottj97:
Privileged spec says about `mstatus.GVA`:

> Field GVA (Guest Virtual Address) is written by the implementation whenever a trap is taken into M-mode. For any trap (breakpoint, address misaligned, access fault, page fault, or guest-page fault) that writes a guest virtual address to mtval, GVA is set to 1. For any other trap into M-mode, GVA is set to 0.

Spike was always writing GVA=0 on breakpoint traps.


aswaterman (approved):
LGTM.

PR #856, opened 2021-11-10
fixed fdt #include to search in local diretory first

emelcher:
Replacing `<>` by `""` fixed the problem.
Even if there is another file called `libfdt.a` in the system search path, the linker uses the file `libfdt.a` from `riscv-isa-sim/build`.

### DIFF
--- a/fdt/fdt.c
+++ b/fdt/fdt.c
@@ -5,8 +5,8 @@
  */
 #include "libfdt_env.h"
 
-#include <fdt.h>
-#include <libfdt.h>
+#include "fdt.h"
+#include "libfdt.h"
 
 #include "libfdt_internal.h"
 


scottj97:
Really? I thought this would only affect the search path for the preprocessor to find the include files. It affects the linker search path too?

emelcher:
No, I did not express my self clearly: my first suspicion, which I mentioned in an earlier email, was a conflict of libraries. Later I discovered the problem with the `#include`. Now I found out that in fact there is (neither was) no conflict of libraries, because only fixing the `#include`s resolved the problem.

scottj97:
Ah. Your initial comment above made me think this change modified the linker behavior. You were saying that the linker was doing the right thing all along.

Can you please fix the typo in the commit message? Directory is misspelled.

PR #854, opened 2021-11-09
Fix #820 by relocating definition of P and require macros

aswaterman:
Fixes a namespace conflict with Boost, allow us to use precompiled headers with Clang again.

Issue #834, opened 2021-10-15
Undefined reference error in boost when compiling the latest


sammy17:
I am getting the following error when I try to compile latest 

```
/share/pkg.7/gcc/9.3.0/install/bin/g++ -L. -Wl,--export-dynamic -L/usr/lib64  -Wl,-rpath,/projectnb2/risc-v/riscv-isa-sim/build/lib  -o spike-dasm spike-dasm.o  libspike_dasm.a  libriscv.a  libdisasm.a -lpthread -ldl -lboost_regex -lboost_system -lpthread  -lboost_system-mt -lboost_regex-mt
libriscv.a(interactive.o): In function `boost::cpp_regex_traits<char>::transform[abi:cxx11](char const*, char const*) const':
/usr/include/boost/regex/v4/cpp_regex_traits.hpp:919: undefined reference to `boost::re_detail::cpp_regex_traits_implementation<char>::transform[abi:cxx11](char const*, char const*) const'
libriscv.a(interactive.o): In function `boost::cpp_regex_traits<char>::transform_primary[abi:cxx11](char const*, char const*) const':
/usr/include/boost/regex/v4/cpp_regex_traits.hpp:923: undefined reference to `boost::re_detail::cpp_regex_traits_implementation<char>::transform_primary[abi:cxx11](char const*, char const*) const'
```

As shown, the error is coming when spike-dasm is being linked. Even if I spike spike-dasm, same error comes when linking spike-main.

Please let me know if there is any way to identify the root cause and resolve the issue.

aswaterman:
As a workaround, try using the --without-boost argument to configure.

aswaterman:
Actually, --without-boost --without-boost-asio --without-boost-regex

aswaterman:
cc @emelcher — it appears the autoconf checks for Boost aren’t sufficient. (This is part of the reason I was leery of incorporating Boost into Spike to begin with…)

sammy17:
It compiled successfully when --without-boost --without-boost-asio --without-boost-regex options were used to configure. Thanks!

scottj97:
Please leave this bug open until the issue is fixed without any workaround needed. This has come up many times before ([here](https://github.com/riscv-software-src/riscv-isa-sim/issues/829#issuecomment-939035540)
and [here](https://github.com/riscv-software-src/homebrew-riscv/pull/71#issuecomment-930809706) and #820).

scottj97:
[...]

PR #858, opened 2021-11-12
fix code style of pointer to processor_t

emelcher:
This PR improves coding style.

In the files `decode.h` in the `riscv/` directory
and in the files `v*.h` in the `riscv/insns/` directory
there was an inconsistent use of `p->` and `(*p).`, being that
`(*p).` was used by means of a macro definition as `P.`.
For example, one could find the expression `P.get_max_xlen()`
as well as the expression `p->get_max_xlen()`.

This PR proposes to eliminate all ocurrences of `P.`,
replacing them by `p->`.

As a side effect, the macro definition of `P` becomes
unnecessary and is also eliminated.
Eliminating the macro named `P` has the beneficial
side effect that a foreign library header file
which happens to use `P` as a template definition type
placeholder (for example boost) can be used without problem.

To guarantee that funcionality is preserved, a static
verification approach was used. This approach is based
on the assumption that all C source code, without exception,
gets compiled to an object file in the `build/` directory.

Using the command `objdump -d *.o` before any modification
and comparing its output with the output of the same
command after `make` with the modifications proposed in this PR,
no difference is reported. This proves that the modifications
do not affect the funtion of the code in any way.

The verification was done on CentOS7 (gcc 4.8.5),
Ubuntu 21.10 (gcc 11.2.0), and MacOS Monterey (clang 13.0.0).

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -254,9 +254,9 @@ private:
   do { \
     require_vector_vs; \
     require_extension('V'); \
-    require(!P.VU.vill); \
-    if (alu && !P.VU.vstart_alu) \
-      require(P.VU.vstart->read() == 0);        \
+    require(!p->VU.vill); \
+    if (alu && !p->VU.vstart_alu) \
+      require(p->VU.vstart->read() == 0);        \
     WRITE_VSTATUS; \
     dirty_vs_state; \
   } while (0);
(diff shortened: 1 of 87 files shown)


aswaterman:
As I mentioned on https://github.com/riscv-software-src/riscv-isa-sim/issues/834#issuecomment-963655347, I wanted to find a less disruptive way of fixing this problem, which I did in #854.  PCH now works with Clang, without the need to change all the uses of P.

PR #859, opened 2021-11-13
Mask hideleg/hip/hie according to recent spec clarification

scottj97:
This [spec clarification](https://github.com/riscv/riscv-isa-manual/pull/771) says that these three CSRs should be masked by `mideleg`:
* `hideleg`
* `hip`
* `hie`

This has no functional effect today, since the only active bits in these three CSRs are all [tied to 1 in `mideleg`](https://github.com/riscv-software-src/riscv-isa-sim/blob/f1bc56264ed214ee941b3a601a693ab655724f52/riscv/csrs.cc#L694) when hypervisor is enabled. But it [might affect](https://github.com/riscv/riscv-isa-manual/pull/771#issuecomment-967724784) future features, so let's explicitly do what the spec requires so there's no subtle bugs in the future.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -956,6 +956,16 @@ void hypervisor_csr_t::verify_permissions(insn_t insn, bool write) const {
 }
 
 
+hideleg_csr_t::hideleg_csr_t(processor_t* const proc, const reg_t addr, csr_t_p mideleg):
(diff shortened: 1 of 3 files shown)


aswaterman:
Agreed it's prudent to get this out of the way now.  LGTM.

PR #861, opened 2021-11-16
Simplify vwmulsu_vv and vwmulsu_vx

eopXD:


aswaterman:
cc @chihminchao

scottj97:
I have some basic test coverage of `vwmulsu.vv` (but none for `vwmulsu.vx`); I see no difference in behavior from this PR.

aswaterman:
Makes sense, since this is just using the MACC implementation with a zero addend.

PR #866, opened 2021-11-27
[RFC] Accept dummy extensions ("Zifencei" and "ZiHintPause")

a4lg:
# Patch 1, 2: Accept Dummy Extensions

To prevent "unsupported extension" error, this pull request adds two valid extensions (valid but does nothing in the simulator) to `processor_t::parse_isa_string`.

*   "Zifencei" (`fence.i` instruction)
*   "ZiHintPause" (`pause` instruction - a special encoding of `fence` instruction)

# RFC Patch 3: Expand "G" to full extension name

In addition, third commit expands "G" extension to "IMAFDZicsr_Zifencei" (because "Zifencei" is now a valid extension name).
This commit would not meet your requirements because...

*   It may confuse some operating systems such as Linux, which does not correctly parse (or ignore) multi-letter extensions for now  
    Recently, I submitted RFC PATCH to linux-riscv ([v1 with background](http://lists.infradead.org/pipermail/linux-riscv/2021-November/010252.html) and [v2](http://lists.infradead.org/pipermail/linux-riscv/2021-November/010350.html)) to fix this
*   Extension name ordering is not respected  
    Putting some extensions with `--isa` argument (e.g. `Xhwacha`) may make full set of extensions invalid (because `Z` multi-letter extensions cannot come after `X` ones).
    Current `processor_t::parse_isa_string` does not respect multi-letter ordering but still, applying patch 3 may not be a good idea as first two patches.

So, third commit is not a requirement (just a proposal). I just want to hear your thoughts.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -215,7 +215,7 @@ void processor_t::parse_isa_string(const char* str)
   else
     bad_isa_string(str, "Spike supports either RV32I or RV64I");
   if (isa_string[4] == 'g')
-    isa_string = isa_string.substr(0, 4) + "imafd" + isa_string.substr(5);
+    isa_string = isa_string.substr(0, 4) + "imafd" + isa_string.substr(5) + "_zicsr_zifencei";
   if (isa_string[4] != 'i')
     bad_isa_string(str, "'I' extension is required");
 
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -256,6 +256,8 @@ void processor_t::parse_isa_string(const char* str)
       // Zicsr is implied by the privileged architecture
     } else if (ext_str == "zifencei") {
       // Zifencei is ignored and implied
+    } else if (ext_str == "zihintpause") {
+      // ZiHintPause is ignored (only a hint)
     } else if (ext_str == "zmmul") {
       extension_table[EXT_ZMMUL] = true;
     } else if (ext_str == "zba") {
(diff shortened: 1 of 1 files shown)


a4lg:
## Another extension which is valid but not in the pull request:
"Zam" (misaligned atomics) requires special handling so this extension is not in the pull request (even if it's valid).

aswaterman:
Part 3 has no functional effect, but I suppose it is harmless, and it might improve the readability of the code for people who are unfamiliar with the history of `Zifencei` and `Zicsr` previously being part of the base ISA.

I'm OK with merging this as-is.  @scottj97, any thoughts?

riscv/processor.cc
```diff
@@ -215,7 +215,7 @@ void processor_t::parse_isa_string(const char* str)
   else
     bad_isa_string(str, "Spike supports either RV32I or RV64I");
   if (isa_string[4] == 'g')
-    isa_string = isa_string.substr(0, 4) + "imafd" + isa_string.substr(5);
+    isa_string = isa_string.substr(0, 4) + "imafd" + isa_string.substr(5) + "_zicsr_zifencei";
```
scottj97, line 218: If the ISA string already has Zicsr or Zifencei, this would add it again...is that okay?

aswaterman: It's not a problem now, because we don't enforce canonical order of multi-letter extension names, and we allow repetition.  This is arguably a feature, because it makes it easier to assemble ISA strings piecemeal in scripts that invoke Spike.

However, I can see it becoming a problem eventually, and since this has no functional benefit, I think I've changed my mind.  Instead, I'll leave a comment saying that G also implies Zicsr + Zifencei, but since we include those implicitly for backwards compatibility, we don't need to explicitly list them here.

aswaterman:
I've manually merged the first two commits and replaced the third with a comment explaining the logic.  Thanks!

a4lg:
Thanks for better comments for dummy extensions and I support your decision not merging Patch 3.

PR #867, opened 2021-11-27
Simplify single-width averaging add and subtract

eopXD:
Leverage existing macros.


aswaterman:
These look good.

aswaterman:
@scottj97 can you verify these, too?

scottj97:
I've seen no differences here, though I'm not sure how thorough my test coverage is.

PR #868, opened 2021-11-28
Simplify narrowing instruction

eopXD:
Eliminate redundant parameters and reuse parameter macros.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -769,19 +769,19 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
   auto &vd = P.VU.elt<type_sew_t<x>::type>(rd_num, i, true); \
   auto vs2 = P.VU.elt<type_sew_t<x>::type>(rs2_num, i - offset);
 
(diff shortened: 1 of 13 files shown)


eopXD:
May I ask why does narrowing right-shift instructions need check this extra condition for `VI_LOOP_ELEMENT_SKIP`?
Why does a narrowing right-shift instructions requires either `insn.rd() != 0` or `P.VU.vflmul <= 1`?

```cpp
#define VI_LOOP_NSHIFT_BASE \
  VI_GENERAL_LOOP_BASE; \
  VI_LOOP_ELEMENT_SKIP({\
    require(!(insn.rd() == 0 && P.VU.vflmul > 1));\
  });
```

aswaterman:
Since I'm no longer on vacation, I will defer to @chihminchao and others.

(As a general principle, I support simplifying the code, but I don't have time to verify the correctness of the changes.)

scottj97:
I have some basic coverage of these instructions. I can't see any difference caused by this PR.

PR #869, opened 2021-11-29
Take guest page fault if guest PA out of bounds

scottj97:
Hypervisor spec section 5.5.1 says about Sv39x4:

> Address bits 63:41 must all be zeros, or else a guest-page-fault exception occurs.

It seems that `s2xlate()` was not checking for this, instead ignoring any upper bits that were set.

cc @daniellustig


aswaterman:
Nice catch.

PR #870, opened 2021-11-30
Simplify mulhsu

eopXD:
The last PR is bugged, as Andrew had pointed out.

> the vs1 argument is being sign-extended before it's being zero-extended. I'll revert this right now. Not sure if there's a more elegant way to handle this that's also correct.

This PR will have `vs1` not being sign-extended before an unsigned-extend.


aswaterman:
Thanks.  Code looks good by inspection, but I'll let @scottj97 test and then merge.

eopXD:
@scottj97 Thank you for verifying this and catching bug from the last PR.

PR #877, opened 2021-12-06
Simplify vmadc and vmsbc

eopXD:
Extract prefix and suffix as macro, leaving only main logic in header.
```cpp
#define VI_LOOP_CARRY_BASE \
  VI_GENERAL_LOOP_BASE \
  VI_MASK_VARS \
  auto v0 = P.VU.elt<uint64_t>(0, midx); \
  const uint64_t mmask = UINT64_C(1) << mpos; \
  const uint128_t op_mask = (UINT64_MAX >> (64 - sew)); \
  uint64_t carry = insn.v_vm() == 0 ? (v0 >> mpos) & 0x1 : 0; \
  uint128_t res = 0; \
  auto &vd = P.VU.elt<uint64_t>(rd_num, midx, true);

#define VI_LOOP_CARRY_END \
    vd = (vd & ~mmask) | (((res) << mpos) & mmask); \
  } \
  P.VU.vstart->write(0);
```

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -630,6 +630,21 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
   } \
   P.VU.vstart->write(0);
 
+#define VI_LOOP_CARRY_BASE \
+  VI_GENERAL_LOOP_BASE \
+  VI_MASK_VARS \
+  auto v0 = P.VU.elt<uint64_t>(0, midx); \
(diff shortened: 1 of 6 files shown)


aswaterman:
This seems like a good cleanup, but as usual I’ll wait for @scottj97 to validate it with his tests.

scottj97:
I have only minimal test coverage for this, but I don't see any change in behavior.

PR #879, opened 2021-12-07
Simply floating point parameters and merge operations

eopXD:
Organizes macros in decode.h and cleanup headers.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -1882,7 +1882,7 @@ reg_t index[P.VU.vlmax]; \
   for (reg_t i = P.VU.vstart->read(); i < vl; ++i) { \
     VI_LOOP_ELEMENT_SKIP(); \
     uint64_t mmask = UINT64_C(1) << mpos; \
-    uint64_t &vdi = P.VU.elt<uint64_t>(rd_num, midx, true); \
+    uint64_t &vd = P.VU.elt<uint64_t>(rd_num, midx, true); \
     uint64_t res = 0;
 
 #define VI_VFP_LOOP_REDUCTION_BASE(width) \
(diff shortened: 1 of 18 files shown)


scottj97:
I have minimal test coverage for `vmerge` and none at all for `vfmerge` but as of 1f06e0ecaafb94c595805ad036ccb07affb479f5 this looks good to me.

I've never been excited about the use of CPP macros. C++ has better ways of handling abstractions and I'd like to move away from macros where feasible.

aswaterman:
Although I agree, the V extension implementation has already gone so far down the path of using C macros that I don't view this as problematic.

eopXD:
@scottj97 
I agree that macro may not be the best way.
The least I'm trying to do is to let headers under `riscv/insns/` contains straight forward logic of the instructions with no details exposed. In my opinion I think having direct calls to `P.VU` are too detailed, and my recent PR-s on simplifications are trying to eliminate these. If the headers are clean enough, refactoring them to adapt modern cpp features won't be too hard.

```cpp
// impl
template<typename Instruction>
int compute(reg_t vd, reg_t vs1, reg_t vs2, Instruction inst) { 
  return inst(vd, vs1, vs2);
}
void vadd_vv(reg_t &vd, reg_t vs1, reg_t vs2) { 
  /* ... */
}

// usage
type_sew_t<x>::type &vd = P.VU.elt<type_sew_t<x>::type>(rd_num, i, true);
type_sew_t<x>::type vs1 = P.VU.elt<type_sew_t<x>::type>(rs1_num, i);
type_sew_t<x>::type vs2 = P.VU.elt<type_sew_t<x>::type>(rs2_num, i);
compute(vd, vs1, vs2, vadd_vv);
```

However as Andrew has mentioned, the implementation has gone deep. It may take time to do this refactoring.

PR #881, opened 2021-12-09
Simplify float convert instructions

eopXD:
Organizes macros in decode.h and cleanup headers for `vfcvt`, `vfncvt`, `vfwcvt`.
Also have vd.v unexposed for `vclass`.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -852,6 +852,18 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
   float##width##_t rs1 = f##width(READ_FREG(rs1_num)); \
   float##width##_t vs2 = P.VU.elt<float##width##_t>(rs2_num, i);
 
+#define CVT_FP_TO_FP_PARAMS(from_width, to_width) \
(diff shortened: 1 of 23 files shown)


aswaterman:
The code doesn’t really seem to have gotten simpler/cleaner in this case.

eopXD:
Apart from lines saved on format-ing for the instruction headers, I want to claim that it does simplify the header in the following points:

- In the instruction headers, the original last bracket `eew_check` is redundant because if the switch is set properly for each instruction, eew that is not covered shall fall into the default case with `assert(0)`. This PR have the switches set properly and hence `eew_check` can be eliminated.
- The instruction headers originally contain variable (`vd, vs2`) fetching code. Conventionally for most of the headers it is hidden under the loop macro. This PR hides them to and so the header is contains only what functions (e.g. softfloat convertion functions) that are called upon `vs2` to write into `vd`.

scottj97 (approved):
I have no test coverage for these instructions but I see no issues with this change.

PR #882, opened 2021-12-09
The pmpcfg csr should be WARL other than WIRI

zhongchengyong:
The riscv-privileged version1.11 has revised the pmpcfg csr fields: "Made the unused pmpaddr and pmpcfg fields WARL, rather than WIRI."


scottj97:
I don’t understand this. The behavior already seems to be WARL to me. If I read it correctly, this change allows reserved bits of pmpcfg to be read and written. That’s not WARL!

Also you pushed these changes to the same branch as #840 so that PR now includes these changes. Please make a different branch for each PR.

zhongchengyong:
> 

Sorry, I misunderstood the meaning of WARL.
The both PR will be closed, thanks.

PR #883, opened 2021-12-09
P-ext v.0.9.11. update

marcfedorow:
[u]k(add|sub)h insns now operate half-word data (not word as 0.9.10 and earlier).


aswaterman:
As usual, I’m not able to verify any P-extension PRs, but by inspection it looks fine.

PR #886, opened 2021-12-14
Add disassembler support for `unimp'

### LINKED ISSUE #49: Wrong disassembly of 0x00000000
Hi,

`spike` and `spike-dasm` e977e229 disassemble 0x00000000 as `addi    s0, sp, 0`, which is wrong in two ways:
- the two least significant bits are `00`, so it would be an RVC instruction. One can see in spike's instruction trace mode that it is decoded as 32 bits.
- `0000` is defined as an invalid instruction in the RVC spec 1.9.

Interestingly, when executing a `00000000` word in spike, an illegal instruction trap is generated.

a4lg:
Now it disassembles 0x0000 (invalid encoding of `c.addi4spn`) as `c.unimp` (RVC).
Non-RVC variant of `unimp` pseudoinstruction (0xc0001073) is also implemented.

It resolves #49 but... is there any good reason *not* to do that? I think `c.unimp` (RVC variant) should be handled (at least).
cf. https://github.com/riscv-non-isa/riscv-asm-manual/blob/b01f83328a8176a7351402414b14a0ac9bd1802f/riscv-asm.md

### DIFF
--- a/disasm/disasm.cc
+++ b/disasm/disasm.cc
@@ -676,6 +676,9 @@ disassembler_t::disassembler_t(int xlen)
   #define DEFINE_XFTYPE(code) add_xftype_insn(this, #code, match_##code, mask_##code);
   #define DEFINE_SFENCE_TYPE(code) add_sfence_insn(this, #code, match_##code, mask_##code);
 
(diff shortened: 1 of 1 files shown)


a4lg:
Considering how long that issue is open, I assume that there might be some good reason/context I missed.
If there is, I'm happy to hear that :)

scottj97 (approved):

PR #889, opened 2021-12-15
PLIC and NS16550 UART Support

avpatel:
Having PLIC and UART devices in Spike provides many benefits:
1. The existing HTIF interface causes OSes (Linux) to use a poll based console driver over SBI calls. The OS console polling threads keeps eating CPU in background causing instruction stream to be polluted with extra instructions. This issue will go away if we have UART device with RX and TX interrupts in Spike.
2.  Real world distro images are large and require a block device (such as VirtIO block, AHCI, etc) for easy use. Currently, there is no interrupt controller in Spike so we can't emulate memory-mapped block devices. We need at least PLIC to emulate sophisticated memory-mapped devices.
3. The UART device in Spike will also help us do some testing of new AIA interrupt controllers (IMSIC and APLIC) on Spike
4. With a PLIC and UART in Spike, we can eventually target OS-A platform compliance in future

### DIFF
--- a/riscv/plic.cc
+++ b/riscv/plic.cc
@@ -1,0 +1,393 @@
+#include <sys/time.h>
+#include "devices.h"
+#include "processor.h"
+
+#define PLIC_MAX_CONTEXTS 15872
+
+/*
+ * The PLIC consists of memory-mapped control registers, with a memory map
+ * as follows:
+ *
+ * base + 0x000000: Reserved (interrupt source 0 does not exist)
+ * base + 0x000004: Interrupt source 1 priority
+ * base + 0x000008: Interrupt source 2 priority
+ * ...
+ * base + 0x000FFC: Interrupt source 1023 priority
+ * base + 0x001000: Pending 0
+ * base + 0x001FFF: Pending
+ * base + 0x002000: Enable bits for sources 0-31 on context 0
+ * base + 0x002004: Enable bits for sources 32-63 on context 0
+ * ...
+ * base + 0x0020FC: Enable bits for sources 992-1023 on context 0
+ * base + 0x002080: Enable bits for sources 0-31 on context 1
+ * ...
+ * base + 0x002100: Enable bits for sources 0-31 on context 2
+ * ...
+ * base + 0x1F1F80: Enable bits for sources 992-1023 on context 15871
+ * base + 0x1F1F84: Reserved
+ * ...		    (higher context IDs would fit here, but wouldn't fit
+ *		     inside the per-context priority vector)
+ * base + 0x1FFFFC: Reserved
+ * base + 0x200000: Priority threshold for context 0
+ * base + 0x200004: Claim/complete for context 0
+ * base + 0x200008: Reserved
+ * ...
+ * base + 0x200FFC: Reserved
+ * base + 0x201000: Priority threshold for context 1
+ * base + 0x201004: Claim/complete for context 1
+ * ...
+ * base + 0xFFE000: Priority threshold for context 15871
+ * base + 0xFFE004: Claim/complete for context 15871
+ * base + 0xFFE008: Reserved
+ * ...
+ * base + 0xFFFFFC: Reserved
+ */
+
+/* Each interrupt source has a priority register associated with it. */
+#define PRIORITY_BASE           0
+#define PRIORITY_PER_ID         4
+
+/*
+ * Each hart context has a vector of interupt enable bits associated with it.
+ * There's one bit for each interrupt source.
+ */
+#define ENABLE_BASE             0x2000
+#define ENABLE_PER_HART         0x80
+
+/*
+ * Each hart context has a set of control registers associated with it.  Right
+ * now there's only two: a source priority threshold over which the hart will
+ * take an interrupt, and a register to claim interrupts.
+ */
+#define CONTEXT_BASE            0x200000
+#define CONTEXT_PER_HART        0x1000
+#define CONTEXT_THRESHOLD       0
+#define CONTEXT_CLAIM           4
+
+#define REG_SIZE                0x1000000
+
+plic_t::plic_t(std::vector<processor_t*>& procs, bool smode, uint32_t ndev)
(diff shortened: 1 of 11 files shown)


avpatel:
@aswaterman The SBI v0.1 is disabled by default in latest Linux-5.18 kernel so this series also helps us boot unmodified Linux-5.18 on Spike. There is also bug fix for hgatp CSR in this series which gets KVM RISC-V working again on latest Spike. The series is primarily foundation work for adding AIA support in Spike.

aswaterman:
@avpatel Sorry for letting this slip through the cracks.  In future, please submit bug fixes as separate PRs, so that they get the higher-priority treatment that they deserve.

I will review the PR later this week.

avpatel:
Sure, I will keep bug fix PRs separate in future.

scottj97:
> Sure, I will keep bug fix PRs separate in future.

Not just in the future...I'd say this is a high-priority fix and should be its own PR today.

Update: filed as #972

aswaterman:
I'm supportive of this effort, but since this is going to be a long-lived interface, we need to get the naming right.  In Spike, we should stick to standard RISC-V terminology when possible.

"irqchip" is an unfortunate Linuxism (what is the relevance of chips in this context?) for something more conventionally called an "interrupt controller".  I think it would be a substantial improvement to `s/irqchip/interrupt_controller/g`.

Along the same lines. "IRQ" is an x86ism; the term doesn't show up in the RISC-V specs.  `irq_level` is probably best replaced by `interrupt_id`, which matches the terminology in the PLIC spec.

avpatel:
@aswaterman Okay, I will do the renaming like you suggested.

riscv/devices.cc
```diff
@@ -12,6 +12,22 @@ void bus_t::add_device(reg_t addr, abstract_device_t* dev)
...
+}
+
+abstract_interrupt_controller_t* bus_t::get_interrupt_controller(void)
+{
+  return intctrl;
+}
+
+void bus_t::set_interrupt_level(uint32_t interrupt_id, int level)
+{
+  if (intctrl)
```
aswaterman: Is it a bug to call `set_interrupt_level` if `set_interrupt_controller` has not been called?  If so, should this be `assert(intctrl);` instead?

riscv/devices.cc
```diff
@@ -12,6 +12,22 @@ void bus_t::add_device(reg_t addr, abstract_device_t* dev)
   devices[addr] = dev;
 }
 
+void bus_t::set_interrupt_controller(abstract_interrupt_controller_t* intdev)
```
aswaterman: Why is a bus associated with an interrupt controller?  Is this just a software-engineering simplification so that all devices on a bus are associated with the same interrupt controller?

riscv/sim.cc
```diff
@@ -109,6 +109,21 @@ sim_t::sim_t(const cfg_t *cfg, const char* varch, bool halted, bool real_time_cl
     bus.add_device(clint_base, clint.get());
   }
 
+  // create plic
```
aswaterman, line 104: I think this should be conditionalized similar to the CLINT so that users who pass in their own DeviceTree can choose not to instantiate the PLIC.

https://github.com/riscv-software-src/riscv-isa-sim/blob/fc88987f4587556a8dbf184825263263b19c01e7/riscv/sim.cc#L107

riscv/sim.cc
```diff
@@ -109,6 +109,21 @@ sim_t::sim_t(const cfg_t *cfg, const char* varch, bool halted, bool real_time_cl
...
+  reg_t plic_base = PLIC_BASE;
+  uint32_t plic_ndev = PLIC_NDEV;
+  fdt_parse_plic(fdt, &plic_base, &plic_ndev, "riscv,plic0");
+  plic.reset(new plic_t(procs, true, plic_ndev));
+  bus.add_device(plic_base, plic.get());
+  bus.set_interrupt_controller(plic.get());
+
+  // create ns16550
```
aswaterman, line 113: Ditto conditionalization here.

aswaterman:
@scottj97 You want to take a look now?  The guts of the PLIC and the interrupt controller might be outside the domain that you're interested in, but I'd like your input on the software-architectural decisions at least.

riscv/devices.cc
avpatel: It's just a software-engineering simplification.

I guess a more flexible way would be to have interrupt controller associated with a device. This will allow multiple interrupt controllers on the same bus where different sets of devices are associated with interrupt controller.

riscv/sim.cc
avpatel: okay, I will update.

riscv/sim.cc
avpatel: Okay, I will update.

riscv/plic.cc
```diff
@@ -0,0 +1,393 @@
...
+      c->mmode = true;
+    }
+    memset(&c->enable, 0, sizeof(c->enable));
+    memset(&c->pending, 0, sizeof(c->pending));
+    memset(&c->pending_priority, 0, sizeof(c->pending_priority));
+    memset(&c->claimed, 0, sizeof(c->claimed));
+  }
+}
+
+uint32_t plic_t::context_best_pending(plic_context_t *c)
```
scottj97, line 100: It looks like you are not changing `*c` so could this formal parameter be `const plic_context& c`? (Same question for other methods here)

scottj97 (changes requested):
I don't have the time or motivation to review this thoroughly, sorry. I do have one nitpick.

aswaterman:
@scottj97 No worries, I'll take it from here.

riscv/plic.cc
aswaterman: I've resolved this on my branch.

aswaterman:
Manually merged.  Sorry for the delay.

jerryz123:
How should Spike support selective enable/disable of devices as more device models are implemented?

aswaterman:
Passing in DeviceTrees seems to be a scalable option, but it does admittedly require a fair amount of tooling.

PR #892, opened 2021-12-20
Update instruction vmandnot.mm, vmornot.mm -> vmandn.mm, vmorn.mm

eopXD:
According to v-spec v1.0-rc2:

> The previous assembler mnemonics vmandnot and vmornot have been changed to vmandn and vmorn to be consistent with the equivalent scalar instructions. The old vmandnot and vmornot mnemonics will be retained as aliases in assembler.

This PR addresses it.


scottj97:
Why closed?

eopXD:
I must have accidentally deleted the branch when I was cleaning up my forked-repo. Reopened in #896.

PR #893, opened 2021-12-22
Add missing check for floating-point instructions

eopXD:
#879 missed out this check when refactoring, it should be added so exception can be thrown and handled.


aswaterman:
If you haven't already done so, please go check over any other recent refactorings to make sure they don't suffer from the same bug.

aswaterman:
BTW, the PR appears to have failed to compile in the CI flow.

eopXD:
Ok I will check closely. The CI shall pass now.

PR #894, opened 2021-12-22
Fix check for fcvt

eopXD:
Error is introduced in #881.


eopXD:
Credits to `Jerry Joe Thomas <jthomas[at]valtrix[dot]in>` for providing test case.

PR #897, opened 2021-12-27
Fix check for fcvt

eopXD:
Fix check for fcvt when sew = e64.


eopXD:
Thanks @chunshoulin for verifying with some internal test generator.

PR #898, opened 2022-01-06
Don't say "master"

timsifive:
Requested by "LfX Security - Non Inclusive Language Alerts"


aswaterman:
I've been disregarding those alerts, because they're flagging some silly stuff that's impractical to change (like the C standard library function `abort`).  But this is OK.

PR #899, opened 2022-01-07
Add RV32E/RV64E base ISA support

aswaterman:
The basic approach is to hook into the RS1/RS2/WRITE_RD accessors to check for valid register specifiers and possibly throw illegal-instruction exceptions as a side effect.


scottj97 (approved):
Source code looks clean. I did some sanity regression testing and saw no issues with RV32I/RV64I. I didn't try RV32E/RV64E.

aswaterman:
I've done some basic sanity checking for RV32E.  The rest will be crowdsourced.

PR #901, opened 2022-01-09
Dev/joy/fix wextra build warnings

jerinjoy:
These commits fix issues that are flagged by -Wextra. Wextra is not enabled in the build yet.

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -87,20 +87,20 @@ void debug_module_t::reset()
       proc->halt_request = proc->HR_NONE;
   }
 
-  dmcontrol = {0};
+  memset(&dmcontrol, 0, sizeof(dmcontrol));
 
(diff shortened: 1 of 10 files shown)


aswaterman:
We prefer not to change the SoftFloat source code (more than we prefer to be `-Wextra`-clean), and so the approach to the fallthrough warnings will be to add `-Wno-implicit-fallthrough`, rather than to accept that aspect of the patch.  And I think that'll be the common theme going forward: we'll accept these patches when convenient, but we won't pledge to be `-Wextra`-clean when it isn't particularly convenient.

aswaterman:
I've manually merged the non-softfloat changes in d1a3a4255295b742728b8c851a222cfbf06e8116

jerinjoy:
Thanks for pulling in the partial changes. I'll close this PR.

PR #905, opened 2022-01-17
Allow to enable p-ext via its zeds

marcfedorow:
@chuanhua @Kevin-Andes
Also I want to remind that P-ext spec does not specify what means 'p' in isa string for rv32 (i.e. if rv32gp enables zpsfoperand).

### DIFF
--- a/riscv/insns/vfrsqrt7_v.h
+++ b/riscv/insns/vfrsqrt7_v.h
@@ -1,4 +1,4 @@
-// vfclass.v vd, vs2, vm
+// vfrsqrt7.v vd, vs2, vm
 VI_VFP_V_LOOP
 ({
   vd = f16_rsqrte7(vs2);


riscv/processor.cc
```diff
@@ -351,6 +357,15 @@ void processor_t::parse_isa_string(const char* str)
     }
     p = end;
   }
+  if (extension_table[EXT_ZBPBO]
+   || extension_table[EXT_ZPN]
+   || extension_table[EXT_ZPSFOPERAND]) {
```
aswaterman: We prefer the following style for multi-line conditionals; would you mind fixing up these statements accordingly?  https://github.com/riscv-software-src/riscv-isa-sim/blob/ba9f5da8c5f269adf95e34b3c8f434d2084608da/riscv/csrs.cc#L988-L991

marcfedorow: @aswaterman, is this suitable?

aswaterman: Yeah, thanks.

PR #909, opened 2022-01-23
Improve boost detect

jolivain:
The current boost detection logic search for the base first, then for asio and regex libraries. This sequence works well when the boost library is not installed at all and not detected, or when it's fully installed and detected.

In some specific cases (when a boost version is available, but asio or regex libraries are not present, or a full boost version is available but riscv-isa-sim needs to be built without boost support), configuration or build fails.

Attempting to disable boost on a system with boost with:

    ./configure --without-boost

will fail with:

    checking whether the Boost::ASIO library is available... yes
    configure: error: Could not find a version of the Boost::Asio library!

Attempting to disable boost with:

    ./configure --without-boost --without-boost-regex --without-boost-asio

will result to a binary linked to boost.

Those cases happened when riscv-isa-sim was added into Buildroot. See:
https://git.buildroot.org/buildroot/commit/?id=ea033cecf990ed3f86619647523a9ad2ccb09adb
https://git.buildroot.org/buildroot/commit/?id=24d07fdc1404901fb1872ac55f875fe1d555cbb3

This commit change the logic of the boost library detection: boost-asio and boost-regex will be searched only if the boost base is requested or detected. This fixes the case when riscv-isa-sim is configured with the --without-boost option.

Tested build on several configurations:
- Host system without boost,
  `./configure` (no option)
- Host system with full boost installation,
  `./configure` (no option)
- Host system with full boost installation,
  `./configure --with-boost=/usr`
- Host system with incomplete boost (no regex),
  `./configure --without-boost`
  Final spike binary does not link libboost.so
- Host system with full boost installation,
  `./configure --without-boost`
  Final spike binary does not link libboost.so


aswaterman:
@jolivain can you provide a description of what you’ve done that has improved Boost detection at configure time? We’d like to improve this story, but we’d also like to be able to explain what has changed :)

scottj97:
@aswaterman the commit messages explain it well.

Also see #834 -- @jolivain do you think this would solve that issue?

scottj97 (changes requested):
Please fix typo in `autoreconf` (in the commit message)

jolivain:
Hi!  Sorry, I did not filled the pull request description.  I'll copy the main commit content in it, for the record.
I also fixed the "autoreconf" typo.
I also added a description of the various build cases I tested.
Thanks!

scottj97 (approved):

jolivain:
> Also see #834 -- @jolivain do you think this would solve that issue?

Hi @scottj97

Looking deeper at #834, it seems some issues are created from multiple installations of boost, in system directories (e.g. one with regex in `/usr/local`, and another without regex in `/usr`).  Link errors are likely from mixing two different boost versions.

Those cases are hard to always auto-detect correctly (e.g. `--with-boost=yes`), because it really depends on the search order and where is the "good" version.
It could be somewhat controlled by giving the path to the good version (e.g. `--with-boost=/usr/local`), but the search order could conflict with other libraries or features.

Regarding the search order, I noticed that:
https://github.com/riscv-software-src/riscv-isa-sim/blob/v1.1.0/Makefile.in#L127
adds a rpath library search on the install lib dir.  When doing a simple `./configure`, this rpath end up to `/usr/local/lib`. This can get tricky in system with multiple boost versions if spike links against `/usr/lib/libboost*.so` but then runtime-link to `/usr/local/lib`.

When working with multiple versions, I think the best practice would be to keep only one version in system directories, and other versions in specific user directory.  This gives an precise control of the search order, for all libraries (e.g. `--with-boost="$HOME/my-boost-x.y.z"`).

My patches are not specifically improving in those specific points.  This could be managed with a small word of warning in the readme.

My patches are more improving:
1. the logic of what is optional/mandatory. (i.e. asio and regex are mandatory boost libraries, only IF the optional boost support is enabled).
2. in case boost is disabled (`--without-boost`), the final binary is not linked against libboost (if a library was present).

Looking at that, I realize the lines `AC_CHECK_LIB` for `boost_system` and `boost_regex` might be superfluous:
https://github.com/riscv-software-src/riscv-isa-sim/blob/v1.1.0/riscv/riscv.ac#L7

They are normally part of the AX_BOOST* macros, when needed:
https://github.com/riscv-software-src/riscv-isa-sim/blob/v1.1.0/ax_boost_regex.m4#L80
https://github.com/riscv-software-src/riscv-isa-sim/blob/v1.1.0/ax_boost_asio.m4#L88

Looking at the Autoconf Boost macro usage examples at:
https://www.randspringer.de/boost/examples.html
those `AC_CHECK_LIB()` calls are not present either.

I quickly tested on a Fedora 35 and a Ubuntu 20.04.3, configuration and builds are working fine (with and without boost support) when I remove those `AC_CHECK_LIB()` lines.

Any idea why they were added in the first place?

Best regards,

Julien.

scottj97:
> Any idea why they were added in the first place?

No, I don't understand much about this. I know @emelcher has done some work here.

emelcher:
Hi @jolivain, the problems You point out are due to lack of experience with autoconf on my part. @scottj97 had mentioned some strange behavior https://github.com/riscv-software-src/riscv-isa-sim/issues/834#issuecomment-944451079 but I am unsure how to resolve it and hesitated to make a pull request to improve, following the saying "don't fix what is not broken". My suggestion would be as: https://github.com/emelcher/riscv-isa-sim/tree/boost_config

jolivain:
Hi @emelcher

From your Boost detection improvement proposal:

- I'm not sure to understand what exact problem your changes in ax_boost_base.m4 are fixing.  This ax_boost_base.m4 is coming from the autoconf-archive package, I would rather not change it specifically for this project, unless there is bug in it.
See: https://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob;f=m4/ax_boost_base.m4

- The AX_BOOST_ASIO and AX_BOOST_REGEX could indeed be moved into the [action-if-found] block of AX_BOOST_BASE().  I updated my changes for doing that.

- I still don't see the reason for having those AC_CHECK_LIB() for boost_system and boost_regex. Are those needed for some specific systems, toolchain or boost version?  I removed those in my updated pull request.

For info, I tested this new proposal on Fedora 35 and Ubuntu 20.04.3, in the configurations described in the commit log.

Best regards,

Julien.

emelcher:
Hi @jolivain,

- I agree that ax_boost_base.m4 should not be changed in this project. If a change is needed, which I am not sure of, it should be done by the maintainers of that file, or at least be approved by the maintainers. The modification that I proposed avoids that the paths, where configure looked for Boost, still get included, even if Boost has not been detected, see https://github.com/riscv-software-src/riscv-isa-sim/issues/834#issuecomment-944451079.

- ok

- As far as I remember, the library link path was not included without AC_CHECK_LIB() in my production system CentOS7. Please give me some days to find time to compile Your pull request on that system.

Regards,
Elmar

jolivain:
For info, I quickly tested by branch in a CentOS 7 Docker container with the following file:
```
# To build this image:
#   docker build -t spike:$(TZ=UTC date +%Y%m%d.%H%M) .
#   docker run -ti spike

FROM centos:7

RUN yum -y upgrade && \
    \
    yum -y install \
        git make gcc autoconf automake dtc gcc-c++ boost-devel \
        && \
    yum clean all

RUN useradd -m -s /bin/bash build && \
    chown -R build:build /home/build

USER build
WORKDIR /home/build
ENV HOME /home/build
ENV LC_ALL en_US.UTF-8

RUN git clone -b improve_boost_detect https://github.com/jolivain/riscv-isa-sim.git
RUN cd riscv-isa-sim && ./configure && make -j$(nproc)
```

Boost libraries (`-mt` versions) seems correctly detected.  The generated Makefile contain:
```
LIBS          := -lpthread -ldl -lpthread  -lboost_system-mt -lboost_regex-mt
```

Build fails to an apparently unrelated issue with:
```
In file included from ./riscv/mmu.h:6:0,
                 from ./riscv/insn_template.h:4:
./riscv/insns/sunpkd810.h: In function ‘reg_t rv32i_sunpkd810(processor_t*, insn_t, reg_t)’:
./riscv/decode.h:242:60: error: expression cannot be used as a function
 #define get_field(reg, mask) (((reg) & (decltype(reg))(mask)) / ((mask) & ~((mask) << 1)))
                                                            ^
```
I'll have a look if I can fix this issue later...

Thanks!

jolivain:
Hi,

After reviewing this after some time, I think:

1. this pull request is still valid, so I rebased, refreshed it (regenerated generated files with autoconf 2.71) and retested it.
This pull request also adds additional autoconf fixes needed to regenerate the configure script. For info, fixes were initially found while proposing PR #1262 but not kept as patches were reworked without needing to update autoconf files.

2. The CentOS 7 issue mentioned in comment
https://github.com/riscv-software-src/riscv-isa-sim/pull/909#issuecomment-1039446207
no longer applies:
Commit 55f664b6b7 changed the language standard from C++11 to C++17.
Full C++17 support is included to GCC >= 5. See:
https://gcc.gnu.org/projects/cxx-status.html#cxx17

CentOS 7 gcc version is currently at 4.8.5:
https://git.centos.org/rpms/gcc/blob/8859c01b82b1ce88742646d92916ca8f2d0b92be/f/SPECS/gcc.spec#_3578

For info, CentOS 7 full updates ended on 2020-08-06. Maintenance
updates (end-of-life) will end 2024-06-30:
https://wiki.centos.org/About/Product

Best regards,

Julien.

scottj97:
I ~~don't~~ have Boost installed in a place where the Spike build can find it. Yet my compiles all have `-I/include/boost-0`, and this PR doesn't seem to change that. :(

Edit: On closer look, I do have `$BOOST_ROOT` set to a filesystem path containing Boost 1.67.0; if that is what's causing Spike to detect Boost, it's certainly not setting the right include path, because there is no `/include` on this system.

I do not need or intend to compile Spike with Boost.

scottj97:
I think it would be best if the `autoconf`-generated files were kept in sync with the source files at every commit point. That is, anybody running `autoconf` at any commit point should not see any changes made.

Otherwise we'd have a commit that has no effects yet, until the following commit updates the generated files.

PR #912, opened 2022-01-27
bind disas support with isa

liweiwei90:
There are many encoding overlaps for different extensions currently, such as Zfinx/Zdinx/Zhinx{min} with 'F'/'D'/'Zfh{min}', Zce(Zcmb, Zcmp, Zcmt) with 'D', cmo(Zicboz, Zicbom) with RV128(lq instruction). 
However, In current disas support, only xlen is passed to disassembler  which make it difficult to distinguish the overlaped instructions.
In this PR, we try to bind the disas support for instructions with  extension to solve this problem:
- Firstly, we use an unified ISA-string processing in spike-dasm and spike (partitioned from https://github.com/riscv-software-src/riscv-isa-sim/pull/831). 
- Then, we try to only add the disas support for instructions within the enabled extension in the ISA string.
- Finally, we add the disas support for Zfinx as a example.

### DIFF
--- a/disasm/disasm.cc
+++ b/disasm/disasm.cc
@@ -931,6 +931,38 @@ disassembler_t::disassembler_t(isa_parser_t* isa)
     DEFINE_FX2TYPE(fle_s);
   }
 
+  if (isa->extension_enabled(EXT_ZFINX)) {
+    DEFINE_RTYPE(fadd_s);
+    DEFINE_RTYPE(fsub_s);
+    DEFINE_RTYPE(fmul_s);
+    DEFINE_RTYPE(fdiv_s);
+    DEFINE_R1TYPE(fsqrt_s);
+    DEFINE_RTYPE(fmin_s);
+    DEFINE_RTYPE(fmax_s);
+    DEFINE_R3TYPE(fmadd_s);
+    DEFINE_R3TYPE(fmsub_s);
+    DEFINE_R3TYPE(fnmadd_s);
+    DEFINE_R3TYPE(fnmsub_s);
+    DEFINE_RTYPE(fsgnj_s);
+    DEFINE_RTYPE(fsgnjn_s);
+    DEFINE_RTYPE(fsgnjx_s);
+    DEFINE_R1TYPE(fcvt_s_d);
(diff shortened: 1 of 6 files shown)


aswaterman:
I agree with the need, but this is an extremely large diff (even with whitespace changes hidden).  Have you tested this on a large corpus (e.g. booting Linux and running a vector workload)?

liweiwei90:
Sorry. I didn't test this on a large corpus(only tested this on ACT). This PR mainly affect the disasembler and I don't known how to test this effectively. Booting Linux with -l option and check the disasembly code? If so, it may be a big workload.

> I agree with the need, but this is an extremely large diff (even with whitespace changes hidden). Have you tested this on a large corpus (e.g. booting Linux and running a vector workload)?

aswaterman:
Tested on Linux boot and looks OK.

scottj97:
Interestingly, the P-extension instruction `kadd16`, when P is not enabled, now disassembles as:
```
core   0: 0x0000000080002206 (0x10940577) kadd16 (args unknown)
```

Why does Spike understand the opcode but not the operands? I would have expected it to know both (if P is in the ISA string) or neither (if P is not).

(I assume other P-extension opcodes would behave similarly; `kadd16` is the only one in my little test suite.)

liweiwei90:
Yes, this is true. This is the function of "add_unknown_insns(this);" at the end of constructor for disassembler.  It's the default disas support for all instructions. Direct disas support will insert before them. So it's only work for instructions which have implemented their function but no disas support before this PR.  In this PR, we bind the disas support with the isa string. When P is not enabled, direct disas support for instrucitons in P extension will not be inserted. so the default case works.

aswaterman:
@liweiwei90 this is a usability problem. For cases where there is no opcode aliasing (including P, V, etc.), we should disassemble correctly even if unspecified. It’s only for the aliasing cases (Zfinx, etc.) that we should get argument-dependent results.

aswaterman:
I can work on this one if you prefer.

liweiwei90:
It's OK for me.

Issue #873, opened 2021-12-05
AMO with store page fault and load guest-page fault


scottj97:
Consider a two-stage (hypervisor) translation where:
* Final VS-stage PTE has W=0
* Final G-stage PTE has U=0

A load to this page will take a guest-page fault.
A store to this page will take a page fault.

Which one should an AMO take?

The spec says:

> Attempting to execute a store, store-conditional, or AMO instruction whose effective address lies within a page without write permissions raises a store page-fault exception.

And accompanying commentary:
> AMOs never raise load page-fault exceptions. Since any unreadable page is also unwritable, attempting to perform an AMO on an unreadable page always raises a store page-fault exception.

This suggests that checking the page for store permissions (only) is sufficient, and therefore the AMO should take a page fault.

Spike, however, is taking a guest-page fault, because it executes AMOs as a [load followed by a store](https://github.com/riscv-software-src/riscv-isa-sim/blob/a9b3babb8f27c063213d6189f4862269756a1a0c/riscv/mmu.h#L192-L193).

Is that okay?

scottj97:
This might be the first case in the ISA where a load can trigger a completely separate exception cause than a store (beyond simply load-_X_-fault versus store-_X_-fault), so this is the first time this implementation detail has become architecturally visible.

scottj97:
On second thought, this might be visible even without hypervisor.
* Sv39 leaf PTE has W=0
* PMP has R=0

A load to this page would take an access fault.
A store to this page would take a page fault.

Which one should an AMO take?

gfavor:
For both cases one can think in terms of there being an inherent natural order of checks.  First a VA is translated by the MMU.  Either that results in a Page Fault or in a valid PA to pass on to PMAs and PMPs to check before they then pass the PA along to do an actual memory access.  Hence Page Faults are generally detected and reported before Access Faults.

Although, to be more accurate, each PA access during a tablewalk also gets checked by PMAs and PMPs.  So Access Faults can be detected during a tablewalk.  In general, there is a long sequence of actions to perform a translation, and Page Faults and Access Faults are detected and reported in that order.

Similarly, in the case of two-stage translation, first a VA is translated by VS-stage.  Either that results in a Page Fault or in a GPA to pass on to G-stage to translate.  Then that results either in a Guest Page Fault or in a PA to pass on to PMAs and PMPs to check before they then pass the PA along to do an actual memory access.  Hence Page Faults are generally detected and reported before Guest Page Faults.

[...]

PR #913, opened 2022-01-27
Instruction and data trace (refactor to use C++ classes) 

andyduller:
Hi Andrew/Scott,
I've rewritten the original Siemens instruction and data trace patch so that it uses a C++ class. I have split it into 5 separate commits which hopefully makes things easier to review.
I think it is much simpler, and hopefully less invasive than it was before as it doesn't touch the sim_t class anymore, as suggested by Scott I think.

I believe there is still the question of the documentation of the CSV formats produced by the instruction and data trace; Gadge is looking at how best to provide this information to you.

Any questions, please let me know. Thanks,
Andy Duller


scottj97:
Please clean up the git history: it should not have the original commit + revert, or the merge commit from upstream/master. (See any good blog on `git rebase --interactive`.) There's also a merge conflict with the latest master that will need to be resolved.

scottj97:
I think this PR should explain a little more of the motivation here, especially regarding if/when this CSV format is intended to become standardized.

scottj97:
Looking at the commits a little closer, this isn't really what I meant when I recommended breaking it up into pieces. It looks like you simply took the completed implementation and made a separate commit out of each final file.

Commit 0e93f2cdf7188bd66a36ffd3ee6ca259b6fd758c adds a complete new class that's unused and not even compiled. That's not useful.

Commit 0ad499a08c4b8d3f27fa6a330e85fe9e8ca31ce4 calls method `proc->set_i_trace()` that doesn't exist yet. Compilation fails.

What I meant is something like:
1. Add empty trace class and instantiate it inside `proc_t`
2. Add cmdline args and pass them into the tracer
3. Add basic trace of PC
4. Add tracing of load/stores
5. ...etc...

The commit history should be an archaeological record of how and (more importantly) why the code came to be the way it is. The commit messages are a place to record design decisions and rationale. That's not possible when an entire 326-line file suddenly appears in complete form.

I have lots of nits to pick on the code itself but we'll get to that in due time. If you don't want to go back and split this up properly it would be better to simply squash it all into one commit.

andyduller:
Sorry, just noticed all of these useful comments about yesterday's pull request so I'll remove today's attempt as it's clearly not what you wanted.
Thanks,
Andy

andyduller:
I'll discuss with Gadge what he wants to do next. Be in touch next week.

andyduller:
Hi Scott,
Regarding the archaeology of the code development, I'm afraid that just doesn't exist. It started as a simple patch against the isa-sim back in 2017 (I think) and has gone through a number of different revision control regimes since then.

I have now reverted to a single commit for the main code portions and rebased on the latest version of master. I've also added a README.processor_trace in that commit, to describe the CSV formats and to provide a reference to the source of most of the information regarding this work, namely:
https://github.com/riscv-non-isa/riscv-trace-spec

I've then added another commit with a set of tests against expected files. I'm afraid that these are the only type of tests that we have ever done to check that the spike output is as expected. What I have added is a small subset of the tests that are available in the riscv-trace-spec repository and more of them can be added if necessary. I have tried to keep the footprint of these small (I have compressed the expected files) and if you would like more added then perhaps we ought to use git-lfs?
Thanks,
Andy

README.processor_trace
```diff
@@ -0,0 +1,100 @@
+Processor trace background information
+--------------------------------------
+
+Processor trace allows instruction and/or data information to be
+traced. Further details of how this is achieved can be found in the
+"Efficient Trace for RISC-V" document which is available from here:
+
+https://github.com/riscv-non-isa/riscv-trace-spec/blob/main/riscv-trace-spec.pdf
+
+In addition python refence models of the trace encoder and trace
```
scottj97: reference is misspelled

scottj97:
I've only skimmed the trace spec. I understand there is Branch Trace and Data Trace, and those correspond to the two files output by this change.

Are these two trace streams truly independent, though? A real system could presumably correlate the two streams by the timing of the messages (though I couldn't find anything useful in the trace spec when I searched for "correlate").

I'm asking because I wonder if it would make more sense to output them all in one file, where each line was either a data or a branch trace.

andyduller:
Hi Scott,
Interesting thoughts. Since the two streams come from the same executing
program they are not independent in that sense. However, the instruction
trace will produce a trace for every executed instruction and the data
trace only outputs for a subset of instructions.
I think that when this "flavour" of instruction and data trace was
developed, by UltraSoC initially, it was envisaged that the two types of
trace were independent, since adding each type of trace would cost silicon
area. I believe that the expectation was that instruction trace would be
far more common than data trace.
Thanks,
Andy

README.processor_trace
andyduller: Thanks

README.processor_trace
```diff
@@ -0,0 +1,100 @@
...
+
+ADDRESS: Address of the instruction (%lx format, reg_t)
+
+INSN: Instruction (%lx format, uint64_t)
+
+PRIVILEGE: Privilege level (%x format, uint8_t)
+
+EXCEPTION: Whether an exception occurred (%x format, uint8_t)
+
+ECAUSE: The exception value when an exception occurred (%lx format, reg_t)
```
scottj97: I assume you mean "exception **cause** value" here?

README.processor_trace
```diff
@@ -0,0 +1,100 @@
...
+
+These can be run by using the script
+  tests/processor_trace/run_trace_tests.sh path-to-spike-binary
+
+The binaries and expected files are stored in tests/processor_trace/test_files.
+Note that the expected files have been compressed using gzip and that the test script
+automatically uncompresses these before comparing with the generated traces.
+
+The output is placed in a sub-directory of the current working directory, called trace_regression.
+Each run of the script will remove any previously run tests.
```
scottj97: Missing final newline

riscv/data_tracer.h
```diff
@@ -0,0 +1,25 @@
...
+class datatracer_t : public memtracer_t
+{
+ public:
+    datatracer_t() {}
+    bool interested_in_range(uint64_t begin, uint64_t end, access_type type) {
+        return type == LOAD || type == STORE;
+    }
+    void trace(uint64_t addr, size_t bytes, access_type type) {
+        (void)addr;
+        (void)type;
```
scottj97: Do these lines serve any purpose? If they are only for quieting compiler warnings about unused arguments, prefer this:
```
 void trace(uint64_t /*addr*/, size_t bytes, access_type /*type*/) {
```

riscv/data_tracer.h
```diff
@@ -0,0 +1,25 @@
...
+#define _RISCV_DATA_TRACER_H
+
+#include "proc_trace.h"
+#include "memtracer.h"
+
+class datatracer_t : public memtracer_t
+{
+ public:
+    datatracer_t() {}
+    bool interested_in_range(uint64_t begin, uint64_t end, access_type type) {
```
scottj97: Both methods here should be explicitly `virtual` and `override`.

riscv/mmu.h
```diff
@@ -106,7 +107,9 @@ class mmu_t
       size_t size = sizeof(type##_t); \
       if ((xlate_flags) == 0 && likely(tlb_load_tag[vpn % TLB_ENTRIES] == vpn)) { \
         if (proc) READ_MEM(addr, size); \
-        return from_target(*(target_endian<type##_t>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr)); \
+        type##_t data = from_target(*(target_endian<type##_t>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr)); \
+        if (proc) proc->get_proc_trace()->record_load(addr, data, size, DATA_SRC_LOAD_TLB_1); \
```
scottj97: The `READ_MEM()` macro call above already does tracing, in the form of the commit log. It would be preferable if there weren't two completely independent tracing mechanisms here. Can you add the necessary parameters to READ_MEM() and call `record_load()` from within that macro?

riscv/mmu.h
```diff
@@ -164,6 +169,7 @@ class mmu_t
         return misaligned_store(addr, val, sizeof(type##_t), xlate_flags); \
       reg_t vpn = addr >> PGSHIFT; \
       size_t size = sizeof(type##_t); \
+      if (proc) proc->get_proc_trace()->record_store(addr, val, size, DATA_SRC_STORE); \
```
scottj97: Could `record_store()` be called from `WRITE_MEM()` macro?

riscv/processor.cc
```diff
@@ -999,6 +1041,18 @@ void processor_t::disasm(insn_t insn)
 
     debug_output_log(&s);
 
+    if (i_trace || d_trace) {
+      proc_trace->step();
+      // Need to translate addresses in a trap so do an mmu translate in that case.
+      // However, in general we want virtual addresses for the instruction trace.
+      if (state.pc & 0x7000000000000000ULL)
```
scottj97: What is this magic number?

riscv/processor.cc
```diff
@@ -1032,7 +1088,10 @@ reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek)
   if (search != state.csrmap.end()) {
     if (!peek)
       search->second->verify_permissions(insn, write);
-    return search->second->read();
+    reg_t old = search->second->read();
```
scottj97: Variable name is misleading...there's nothing "old" about this value. It's the current value of the CSR.

riscv/proc_trace.h
```diff
@@ -0,0 +1,99 @@
...
+    void set_exception(reg_t cause);
+    void set_tval(reg_t tval);
+    void set_interrupt(uint8_t interrupt);
+    void set_is_32bit_isa(bool is_32bit_isa);
+
+    void record_load(reg_t addr, uint64_t res, size_t size, data_src_t src);
+    void record_store(reg_t addr, uint64_t val, size_t size, data_src_t src);
+    void record_amo(reg_t addr, uint64_t load_data, uint64_t store_data, size_t size);
+    void record_csr_get_trace(int which, insn_t insn, bool write, reg_t load_data, size_t size);
+    void record_csr_set_trace(int which, reg_t store_data);
```
scottj97: Method names are inconsistent. The others in this group don't end in `_trace` even though they also record traces.

riscv/proc_trace.cc
```diff
@@ -0,0 +1,325 @@
...
+      m_tval(0), m_interrupt(0), m_is_32bit_isa(false), m_is_first_step(true),
+      m_dtrace_fd(NULL), m_data_trace_debug(false), m_has_insn(false), m_insn(insn_t()),
+      m_is_load(false), m_is_store(false), m_is_amo(false), m_dtype(DTYPE_INVALID),
+      m_recorded_dtype(DTYPE_INVALID), m_csr_which(CSR_INVALID), m_csr_write(false),
+      m_csr_load_data(0), m_csr_size(0)
+{
+}
+
+void proc_trace_t::open_i_trace(const char *instruction_trace_file_name) {
+    m_itrace_fd = fopen(instruction_trace_file_name, "w");
```
scottj97: We have tried to move away from C stdio towards C++ streams. I'd hate to add a bunch of new `fprintf()` calls here.

riscv/proc_trace.cc
```diff
@@ -0,0 +1,325 @@
...
+        return;
+    }
+
+    reg_t addr = m_addr;
+    reg_t tval = m_tval;
+    if (m_is_32bit_isa) {
+        addr &= 0xffffffff;
+        tval &= 0xffffffff;
+    }
+    fprintf(m_itrace_fd, "1,%lx,%lx,%x,%x,%lx,%lx,%x\n",
```
scottj97: This is not portable. If you must use C stdio, use the defines from `inttypes.h` such as PRIx64. You can see a few examples of this elsewhere in Spike.

riscv/proc_trace.cc
```diff
@@ -0,0 +1,325 @@
...
+        m_csr_size = 0;
+
+        bool is_compact = (binary & 0x3) != 0x3;
+        m_is_amo = false;
+        if (is_compact) {
+            // Not interested in load immediate
+            uint64_t inst_15_13 = (binary >> 13) & ((insn_bits_t(1) << 3) - 1);
+            uint64_t inst_1_0 = binary & 0x3;
+            m_is_load = (((inst_1_0 == 0) || (inst_1_0 == 2)) && (inst_15_13 > 0) && (inst_15_13 < 4));
+            m_is_store = ((inst_1_0 == 0) || (inst_1_0 == 2)) & (inst_15_13 > 4);
```
scottj97: This is not robust, as new load & store opcodes might be added to the ISA at any time, or in a custom fork of Spike, and they would not know to update this cryptic decoder here.

Spike already has some way of decoding whether an opcode is a load/store, though it's not clear to me why this is needed here, since the act of loading or storing will call other methods in this class. That would be a much more reliable way of determining that an instruction did a load or store.

riscv/proc_trace.cc
```diff
@@ -0,0 +1,325 @@
...
+
+        if (!m_is_load) {
+            // It's not clear why these are being triggered.
+            // They seem to come from DATA_SRC_LOAD_TLB_1
+            if (m_data_trace_debug)
+                fprintf(m_dtrace_fd, "NOT LOAD SRC %d %lx ld %d st %d amo %d %lx\n",
+                        src, addr, m_is_load, m_is_store, m_is_amo, m_insn_binary);
+            else
+                fprintf(stderr, "NOT LOAD SRC %d %lx ld %d st %d amo %d %lx\n",
+                        src, addr, m_is_load, m_is_store, m_is_amo, m_insn_binary);
```
scottj97: No need to duplicate the entire` fprintf()` when the only difference is the output filehandle

riscv/proc_trace.cc
```diff
@@ -0,0 +1,325 @@
...
+    if (!m_csr_write) {
+        assert(m_csr_which != CSR_INVALID);
+        uint64_t pair[2] = {load_data, 0};
+        record_data(m_csr_which, (uint8_t*)pair, 2 * size);
+    }
+}
+
+void proc_trace_t::record_csr_set_trace(int which, reg_t store_data) {
+    // mret is a special case where there is no call to csr_get to go with this.
+    // In this case the csr must be MSTATUS and it does not get recorded.
```
scottj97: Why doesn't it get recorded? Is that specified in the trace spec?

spike_main/spike.cc
```diff
@@ -55,6 +55,9 @@ static void help(int exit_code = 1)
   fprintf(stderr, "                        This flag can be used multiple times.\n");
   fprintf(stderr, "  --rbb-port=<port>     Listen on <port> for remote bitbang connection\n");
   fprintf(stderr, "  --dump-dts            Print device tree string and exit\n");
+  fprintf(stderr, "  --ust-trace=<file>    Write Instruction trace file\n");
```
scottj97: What does "UST" mean? Why not `--inst-trace`?

README.processor_trace
```diff
@@ -0,0 +1,100 @@
...
+DSIZE: The data transfer is 2^DSIZE bytes. (%ld format, size_t)
+
+DATA: The data value or values transferred. (hex format, length is
+dependent on DSIZE)
+
+Testing
+-------
+
+A small collection of tests are included which run spike to generate instruction and data trace
+file and then checks the output against expected files. Both 64-bit and 32-bit binaries are
```
scottj97, line 92: Where did the expected files originate? Has there been any independent validation of their contents?

scottj97 (changes requested):
My comments are mostly small things that could be improved, but I'm a little skeptical of this entire PR. It feels very invasive. I also don't like having a long list of class attributes that get reset and reused for each line of trace output. It would be better to somehow accumulate the info as it becomes available into a new object for each line of trace. (Though it's not obvious to me how best to do that.) That would make it impossible to accidentally mix up two instructions' fields in one line of trace.

This PR includes several binary test programs. Where did these binaries come from? How can we recreate them? The ISA changes over time and Spike changes with it, and sometimes that can necessitate updating these kind of test programs. (To initialize a new CSR, for example.) Would it be better if their source code were included here, and your script ran the compiler/assembler first?

spike_main/spike.cc
andyduller: UST is UltraSoC Trace, it's historic as the work on instruction was initially done by UltraSoC in 2017.

README.processor_trace
andyduller: Corrected.

README.processor_trace
andyduller: Corrected.

README.processor_trace
andyduller: Corrected.

riscv/data_tracer.h
andyduller: Corrected.

riscv/data_tracer.h
andyduller: Corrected.

riscv/processor.cc
andyduller: Corrected.

riscv/proc_trace.h
andyduller: Corrected.

riscv/proc_trace.cc
andyduller: Changed to using C++ streams.

riscv/proc_trace.cc
andyduller: Changed to using C++ streams.

riscv/proc_trace.cc
andyduller: Changed to using C++ streams.

riscv/proc_trace.cc
andyduller: Could you please point me at the way of deciding whether an opcode is a load or store? I had looked in decode.h, but couldn't see any obvious way of doing it. I agree that I really don't want this partial disassembly in this code.
Thanks.

riscv/mmu.h
andyduller: Yes, good idea. Changed.

riscv/mmu.h
andyduller: Also changed, as suggested, thanks. Data trace now requires --enable-commitlog to function correctly.

spike_main/spike.cc
scottj97: So the switch is "UltraSoC Trace trace"? Why not switch this to `--inst-trace`? There's no other references to UltraSoC that I can see in this PR, so no reason to keep this historical name.

riscv/processor.cc
andyduller: As you say, this is a strange magic number. To be honest, I can't remember where it came from.
The addresses recorded in the instruction trace need to be used by the trace decoder to look-up the instruction in the original elf file. I've made a change to this check, but I'm not sure it's the best way of expressing what is required. What do you think?

riscv/mmu.h
scottj97: Is there [code like this](https://github.com/riscv-software-src/riscv-isa-sim/blob/5ba63a44706173b556b8d5632872b39a09d7f16d/riscv/sim.cc#L250-L256) that will give a reasonable error message if someone forgets the configure option and asks for a data trace?

spike_main/spike.cc
andyduller: Corrected.

riscv/mmu.h
scottj97: It looks like you could also easily make `READ_MEM()` do the right thing here even if `RISCV_ENABLE_COMMITLOG` is disabled. The only reason to make it conditional on that compilation flag is if you can provide a performance improvement by disabling it.

riscv/processor.cc
scottj97: You changed this to: `if (state.pc & (1ULL << MAX_PADDR_BITS))`

Which is different, but also unexplained. It seems the goal here is to record the physical address? And this is a (broken) attempt to determine if the PC value is a physical address? But the old comment says the trace wants virtual addresses. So I don't understand why you would ever call `mmu->translate()` here.

The new comment says "the decoder must be able to relate the pc addresses back to the elf file" -- but ELF files include both virtual addresses and physical addresses, and Spike will initialize memory based on the physical address. Some bare-metal test cases will lie to the assembler/linker and use a physical address in place of the virtual, so I understand why sometimes the decoder will not be able to find a given virtual address in the ELF. But that's a problem better solved either upstream (when the ELF is created) or downstream (in the decoder). I don't see any reason to ever translate the address here.

riscv/proc_trace.cc
scottj97: What I meant is, if you see `record_load()`, `record_store()`, or `record_amo()` called, then you know it's a load/store. If not, then it's not.

riscv/proc_trace.cc
andyduller: Since the csr for the mret instruction is implicitly known to be MSTATUS there is no need to record that in the data trace stream (for efficiency reasons). I don't think this is explicitly stated in the trace spec, but a clarification will be added in the next revision of the spec.

README.processor_trace
andyduller: They were produced by a run through a version of spike. The validation that has been done is twofold:
- the traces have been run through various models of the trace encoder and decoder within Siemens and by their customers, and it has been checked that the original trace was reconstructable.
- for some of the smaller files the data has been scrutinized within Siemens and by their customers to check that it agrees with hardware models. In particular, Seagate has contributed a "torture" test example as part of their RISCV development work (https://github.com/riscv-non-isa/riscv-trace-spec/blob/main/referenceFlow/tests/examples/br_j_asm_test/test.S).

riscv/proc_trace.cc
andyduller: I have removed the disassembly code. The check for whether it's an atomic instruction is required because otherwise you get spurious load/stores being recorded. I didn't have a test for this in the subset of tests I have imported, so I've added a test which requires this check.
I also need to be able to determine what type of instruction it is for the data trace of CSR and atomics so I've now done that using the constants defined in encoding.h.

riscv/proc_trace.cc
andyduller: I have removed the magic number in this function and replaced with a constant defined in encoding.h

riscv/mmu.h
andyduller: Yes, I did wonder about whether to put the record_load/record_store in the #ifndef RISCV_ENABLE_COMMITLOG, but as you say, it gives a performance gain if data trace isn't being used.
I've added this check, a good idea, thanks.

riscv/processor.cc
andyduller: Hmmm, ok, I don't know how to change this in that case.

scottj97: IIUC the instruction trace is supposed to record only virtual addresses, yes? The address should never be translated here.

If that causes problems elsewhere in the flow, those problems should be solved outside of Spike.

riscv/proc_trace.cc
scottj97: Can you explain what you mean by "spurious load/stores"? An AMO is both a load and a store. What's supposed to get recorded in the trace in that case? I wonder if a solution for #873 would be able to help here.

README.processor_trace
scottj97: My concern is how we keep these traces up to date over time. As Spike changes, these traces might need to change too.

I guess it wouldn't be too hard to eyeball any differences and then manually copy the new traces over the golden traces (assuming the changes are intentional)? Could/should we document that process?

Should we run these tests as part of the GitHub CI? Otherwise people will submit PRs and not even realize they've broken the tracing.

andyduller: I share your concern over using expected files for CI.
It's hard to quantify, but our experience with the instruction trace files, over a number of years (data trace was much more recent), is that the trace output has not changed as we have updated the patch file to keep it up to date with your upstream changes. Given the fixed elf file input, I would expect that there should be no reason for the trace to change.

Looking at  some of the core instruction trace code
```
      proc_trace->set_addr(state.pc);
      proc_trace->set_insn(insn, bits);
      proc_trace->set_priv(state.prv);
```
e.g. if the encoding or data type of state.prv were to change, then I would expect that there would be large numbers of differences against the expected files (or it might not even compile?).

Regarding documenting how to deal with CI failures:
- the traces include the start-up code that is part of spike; if that changes then in terms of content or location, then the expected files will need to be updated
- if the data types of e.g. internal state of processors was to be changed, then the C++ formatted output of the trace data may well need to be changed
- given that the elf file input to the tests is fixed (and is not subject to changes in the toolchain), I would expect that the instructions addresses will not change (since proxy kernel operation is not supported for trace)

riscv/proc_trace.cc
andyduller: The data trace spec differentiates AMO from other load/store operations. The problem I am trying to solve here is, to make sure that the load/store operations are not recorded in the trace, if the instruction is actually an AMO.
Regarding #873, it may be that a solution to that helps here.

README.processor_trace
scottj97: The elf files may need to change though...it's possible the ISA could change in the future to require e.g. a new CSR that needs to be initialized during startup.

scottj97:
When you rebase the commits, you want to squash any updates needed back into the original commits. For example [13da500](https://github.com/riscv-software-src/riscv-isa-sim/pull/913/commits/13da5006a3561b8683238549118a1c8d1fa378aa) and [d4b24a8](https://github.com/riscv-software-src/riscv-isa-sim/pull/913/commits/d4b24a8184f03b396ddcea67e2f37004a47f4f86) should be squashed. Otherwise some of the earlier commits will have compile errors.

Also please squash [8923e2e](https://github.com/riscv-software-src/riscv-isa-sim/pull/913/commits/8923e2e867d37ca21d2e1727b5333c33c618d3b2) (and [74e11c3](https://github.com/riscv-software-src/riscv-isa-sim/pull/913/commits/74e11c3f20195de105283ad86f96f2f2d6b1dfba)) so this inappropriate address translation is erased from the history completely.

riscv/proc_trace.cc
andyduller: Table 4.12 in the trace spec (https://github.com/riscv-non-isa/riscv-trace-spec/blob/main/riscv-trace-spec.pdf) gives a list of the types of data access that are to be recorded as part of the data trace.

README.processor_trace
andyduller: I suppose in that case there would probably need to be a re-compilation of the source and the new output from running the instruction and data trace taken as the expected output. Are you saying that, in the case that you have cited, the old elf files would not run using an updated spike?

scottj97: In the case I'm imagining the old elf files would not run the same way anymore, and would not give good trace coverage anymore. Maybe this is far-fetched.

There is also the case where we discover a problem with the tracing, and we want to expand coverage to make this problem visible, which means modifying one of the elfs, or adding a new elf.

I want to avoid the provenance of these elf files being a mystery to anyone new coming into this project. "I made these by hand once and I'm assuming they will never need to be changed" is not generally a good plan.

andyduller: I agree that lack of provenance is not a good plan to build in. I'll create something to provide some provenance.

andyduller: I've added extra information about how to rebuild the various test binaries. In particular, I've added the script and C entry point code required to build the embench code for use with spike.

riscv/execute.cc
```diff
@@ -215,7 +215,7 @@ static inline reg_t execute_insn(processor_t* p, reg_t pc, insn_fetch_t fetch)
 
 bool processor_t::slow_path()
 {
-  return debug || state.single_step != state.STEP_NONE || state.debug_mode;
+  return i_trace | d_trace | debug || state.single_step != state.STEP_NONE || state.debug_mode;
```
sifiverobert: This '|' instead of '||' seems like logical bug (but it will work). Below it is correct.

README.processor_trace
sifiverobert: I was expecting, that SPIKE output itself can be used to check if encoder+decoder will produce desired PC sequence.
Spike patch is producing PC sequence (+ other details). If we extract 'PC-only flow', we can check if encoder+decoder are is producing same PC flow. Using such a technique ANY file/program which can be run by spike can be used to test trace. Similar technique can be used to validate data trace - trace should reconstruct all memory transactions (same addresses and data).

riscv/processor.h
```diff
@@ -331,6 +342,11 @@ class processor_t : public abstract_device_t
 
   simif_t* sim;
   mmu_t* mmu; // main memory is always accessed via the mmu
+
+  // Instruction and data trace
+  proc_trace_t* proc_trace;
+  datatracer_t* d_tracer;
```
sifiverobert, line 346: It would be good to have these classes named in similar way. We have 'set_i_trace/set_d_trace/i_trace/d_trace', but these are very different. Maybe 'data_trace_t' instead of 'datatracer_t'? I also see 'data_tracer.h', which is not in sync even with 'datatracer_t'.

spike_main/spike.cc
```diff
@@ -512,6 +518,25 @@ int main(int argc, char** argv)
     for (auto e : extensions)
       s.get_core(i)->register_extension(e());
     s.get_core(i)->get_mmu()->set_cache_blocksz(blocksz);
+
+    if (i_trace_file != NULL) {
+        std::string i_trace_filename(i_trace_file);
+        if (cfg.nprocs() != 1) {
+            std::stringstream ss;
```
sifiverobert, line 526: Only here (end below ...) I see that it actually supports more than one core.
It should be mentioned in 'help' (and maybe full/modified name of file printed ...).

tests/processor_trace/build_files/README.embench
```diff
@@ -0,0 +1,18 @@
...
+Only two additional files are required to build these tests for use with spike:
+- a build script
+- a crt0 source code file
+
+The crt0-spike.c file needs to be put into the directory:
+  config/riscv32/boards/ri5cyverilator/
+within the embench-iot repository.
+
+Note that the build script requires access to a 32-bit RISCV compiler. The embench_build.sh needs to be modified to be able
+to locate this compiler.
```
sifiverobert, line 16: What about 64-bit testing? Here it says '32-bit' only ...

README.processor_trace
```diff
@@ -0,0 +1,126 @@
...
+embench-statemate source is from:
+  https://github.com/embench/embench-iot/tree/master/src
+
+  In order to build these tests for RISCV a bash script was written to setup the compile options, prior to calling
+  the standard build_all.py script from the embench-iot repository. In addition, a spike entry point was written to work with
+  embench. These files, and information on building embench, can be found in
+    tests/processor_trace/build_files
+
+br_j_asm is from:
+  https://github.com/riscv-non-isa/riscv-trace-spec/tree/main/referenceFlow/tests/examples/br_j_asm (written by Bill McSpadden, Seagate Technology)
```
sifiverobert, line 119: Maybe in order to validate if SPIKE output is NOT broken, it would be enough to use just this test?
As far as I remember, this file is generating all possible (hard and easy ...) PC sequences.
Full testing of 'model of encoder' + 'model of decoder' chain will require to run 'any program' (including programs with interrupts and exceptions, which are key to test of trace). Plain C-programs (running on fixed address space) cannot test trace well.

sifiverobert:
Hi Andy & Scott,
I accidentally stepped into this pull-request and commented a bit. I hope you do not mind.
I am interested in this as I am a chair of Nexus Trace TG and I want to connect to Spike output as well. Some initial verification infrastructure (for Nexus Trace ...) is available here: https://github.com/riscv-non-isa/tg-nexus-trace/blob/master/refcode/c/NexusTraceTG-RefCode.pdf. I was using something what I called 'PCSEQ' file, but I want to use Spike output instead.

PR #917, opened 2022-02-16
Fix libdir and includedir in pkg-config files

rswarbrick:
The autoconf `@libdir@` and `@includedir@` variables include the
prefix part so we were generating things like "`/usr//usr/lib`".

### DIFF
--- a/riscv-disasm.pc.in
+++ b/riscv-disasm.pc.in
@@ -1,7 +1,7 @@
 prefix=@prefix@
 exec_prefix=@prefix@
-libdir=${prefix}/@libdir@
-includedir=${prefix}/@includedir@
+libdir=@libdir@
+includedir=@includedir@
 
 Name: riscv-disasm
 Description: RISC-V disassembler
--- a/riscv-fesvr.pc.in
+++ b/riscv-fesvr.pc.in
@@ -1,7 +1,7 @@
 prefix=@prefix@
 exec_prefix=@prefix@
-libdir=${prefix}/@libdir@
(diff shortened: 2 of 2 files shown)


aswaterman:
I don't completely understand what this does, and everything works fine on my side (on both Linux and Mac), so I'm not going to merge this.  Perhaps someone more knowledgable will chime in.

rswarbrick:
The pkg-config files here are used to get CFLAGS/LDFLAGS if you want to compile software against (in this case) spike. For example, you could run something like this:
```
$ pkg-config --libs riscv-disasm
-L/opt/spike/ibex-cosim/lib -Wl,-rpath,/opt/spike/ibex-cosim/lib -ldisasm
```
Here, I've fixed the paths! Without the change in this PR, it would come out with paths like `/opt/spike/ibex-cosim//opt/spike/ibex-cosim/lib`.

The reason I'm interested in this is because we've got a locally modified spike that we're using for cosimulation work for the Ibex processor. Of course, to do cosim stuff you need to compile against spike headers and link against spike libraries, which is why I noticed this at all. We plan to try to upstream the changes needed to make everything work once it's stable, but thought it made more sense to keep things separate for now.

But this is a straight-up bug, so I thought it was worth pushing the fix immediately :-)

ramonaerne:
Bringing this up again, using spike as a dependency to a cmake project. I see the same behaviour of the unnecessary prefix that should be removed. Tested on Linux

aswaterman (approved):
OK.

PR #918, opened 2022-02-17
Split Xbitmanip into its proposed component extensions

rswarbrick:
Before this patch, spike just had an `Xbitmanip` extension which
covered everything in the proposed bitmanip extension that hadn't been
ratified. The problem is that if you want to model (or verify) a
processor that targetted just some of the proposed bitmanip extension,
you couldn't configure Spike to do that.

For example, the lowRISC Ibex processor has several different
configurations. The "balanced" configuration targetted Zba, Zbb, Zbs,
Zbf and Zbt of the 0.92 spec. With the Zba, Zbb and Zbs ratified,
we'll now be able to use an ISA string like

    rv32imc_Zba_Zbb_Zbs_XZbf_XZbt

and Spike will correctly fail to decode instructions like `bcompress`,
which would have been decoded with `Xbitmanip`.

This patch adds a new custom extension name for each part of the
extension that wasn't fully ratified. These have an 'X' prefix so, for
example, the bit permutation instructions that were proposed as Zbp
can be found under `XZbp`.

Specifying `Xbitmanip` gets all of these extensions, so its behaviour
should be unchanged.

Note that the `slo(i)` / `sro(i)` instructions have been moved from the
proposed Zbb to `XZbp`. This matches a comment in the Change History
section of v0.93 of the bitmanip spec: it seems that the authors
forgot to also move them in Table 2.1 (which gives the lists of
instructions for each extension). This change won't break anything
that currently exists, but it took quite a while to figure out what
was going on so I thought I'd leave a breadcrumb trail for the next
engineer!

The bulk of the patch is just defining some more entries in the
`isa_extension_t` enum and rewriting each of the instructions to depend
on the relevant entry. This is mostly a straight textual replacement
but it's slightly more complicated for things like the "pack"
instruction that are defined by several different proposed extensions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,14 @@ is provided under the Spike-custom extension name _Xbitmanip_.
 These instructions (and, of course, the extension name) are not RISC-V
 standards.
 
+These proposed bit-manipulation extensions can be split into further
+groups: Zbp, Zbs, Zbe, Zbf, Zbc, Zbm, Zbr, Zbt. Note that Zbc is
+ratified, but the original proposal contained some extra instructions
+(64-bit carryless multiplies) which are captured here.
+
+To enable these extensions individually, use the Spike-custom
+extension names _XZbp_, _XZbs_, _XZbc_, and so on.
+
 Versioning and APIs
 -------------------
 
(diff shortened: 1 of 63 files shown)


aswaterman:
Thank you for preserving backwards compatibility with the `Xbitmanip` option.  LGTM.

PR #920, opened 2022-02-18
Add simple error checking to DTB parsing code

rswarbrick:
The first commit removes some duplicated code which doesn't do anything (I think the rom loading stuff predated the code that's now in the `sim_t` constructor?). The second commit is why I wanted to make the change, after spending 10 minutes very confused about why I couldn't pass a device tree...

Second commit message below:

This catches silly mistakes like accidentally passing a DTS file when
it should have been a DTB.

Now, you get something like this:

    $ /opt/spike/latest/bin/spike --dtb=bogus.dtb -l obj.o
    Failed to read DTB from `bogus.dtb': FDT_ERR_BADMAGIC.

### DIFF
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -6,6 +6,7 @@
 #include "remote_bitbang.h"
 #include "byteorder.h"
 #include "platform.h"
+#include "libfdt.h"
 #include <fstream>
 #include <map>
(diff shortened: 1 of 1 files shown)


aswaterman:
This looks good, too.  Thanks.

PR #921, opened 2022-02-18
Don't instantiate a CLINT if there is none in the device config

rswarbrick:
This change makes it possible to faithfully simulate systems which
don't have a CLINT (without adding yet another command line argument
to pass through!).

Without a change like this, lowRISC has been using a local hack in its
Spike fork, where we've just commented out the internals of
`clint_t::increment()`. This approach is rather cleaner and is hopefully
general enough to use upstream.

Of course, it's possible I've completely misunderstood something and that there's a much easier way to do this! If so, I'd love to hear it :-)

### DIFF
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -98,12 +98,17 @@ sim_t::sim_t(const char* isa, const char* priv, const char* varch,
   make_dtb();
 
   void *fdt = (void *)dtb.c_str();
-  //handle clic
-  clint.reset(new clint_t(procs, CPU_HZ / INSNS_PER_RTC_TICK, real_time_clint));
+
+  // Only make a CLINT (Core-Local INTerrupt controller) if one is specified in
+  // the device tree configuration.
+  //
+  // This isn't *quite* as general as we could get (because you might have one
+  // that's not bus-accessible), but it should handle the normal use cases. In
+  // particular, the default device tree configuration that you get without
+  // setting the dtb_file argument has one.
   reg_t clint_base;
-  if (fdt_parse_clint(fdt, &clint_base, "riscv,clint0")) {
-    bus.add_device(CLINT_BASE, clint.get());
-  } else {
+  if (fdt_parse_clint(fdt, &clint_base, "riscv,clint0") == 0) {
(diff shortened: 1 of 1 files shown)


aswaterman:
This is probably the best way to do this at this time.  No objections here.

rswarbrick:
Ah, brill, thanks!

Slightly longer-term, I wonder whether it would make sense to allow other core attributes to be passed as part of the device tree? I'm imagining a setup where an explicit command line argument overrides the default, but that default comes from a DTB if provided (and the device tree that comes out of `make_dts()` if not).

That would also allow us to expose more config knobs without needing to add new command line arguments (and arguments to `sim_t::sim_t`) each time.

aswaterman:
I think some folks at SiFive were thinking along similar lines (but I don't remember if there was any follow-through).

rswarbrick:
Cool! There are some more customisations (hacks!) we've got locally that I'll try to upstream nicely over the next few days. I'll try to get things in roughly that shape.

scottj97:
I had a discussion with Andrew from a few months back about the configuration system. I don't have the bandwidth to work on this but I had some ideas.

My quote first, followed by Andrew's response:

> > Today the processor_t parses the ISA string and configures itself, and then sim_t parses the DTB and modifies a few things in the config (PMP, MMU). This means that some config options may change after processor_t::reset() runs, and that makes some things unnecessarily difficult. Such as having to consider n_pmp on every PMP CSR access, instead of only once at construction time.

> >I'd like to refactor the configuration so it all happens before processor_t is created, perhaps into a new config_t class, then pass that into the constructor of processor_t as const so it's sure not to change after reset.

> I agree that configuring processor_t at two different times seems improperly factored.  I don't yet have an opinion on how best to rectify it.  Your proposal seems like one reasonable approach.  The other is to pass the DTB to the processor.  But I think (as you mentioned below) that your approach is more flexible with respect to configuration options coming from multiple places.  The processor_config_t class essentially acts an adapter from {all sources of configuration} to the concrete configuration parameters that processor_t understands.

> > This also provides a good place to integrate the Configuration Task Group's future work.

rswarbrick:
Sounds like a good plan. I've got some half-finished stuff that I wrote over the weekend that goes in that direction. I should be able to push it up in the next day or so.

PR #922, opened 2022-02-18
Split out MINSTRET and MCYCLE

rswarbrick:
Before this change, the MCYCLE CSR was just a proxy for MINSTRET.
Similarly, CYCLE was a proxy for INSTRET. This models a machine where
every instruction takes exactly one cycle to execute.

That's not quite precise enough if you want to do cosimulation: there,
you're going to want to MCYCLE to actually match the behaviour of your
processor (because you need reads from the relevant CSRs to give the
expected result).

This commit splits the two CSRs, leaving the other proxy relationships
unchanged. The code in `processor_t::step()` which bumps MINSTRET now
bumps MCYCLE by the same amount, maintaining the previous behaviour.

Of course, now a cosimulation environment can update the value of
MCYCLE to fix things up for multi-cycle instructions after they run.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -833,21 +833,21 @@ bool virtualized_satp_csr_t::unlogged_write(const reg_t val) noexcept {
 }
 
 
-// implement class minstret_csr_t
-minstret_csr_t::minstret_csr_t(processor_t* const proc, const reg_t addr):
+// implement class wide_counter_csr_t
+wide_counter_csr_t::wide_counter_csr_t(processor_t* const proc, const reg_t addr):
   csr_t(proc, addr),
   val(0) {
 }
 
-reg_t minstret_csr_t::read() const noexcept {
+reg_t wide_counter_csr_t::read() const noexcept {
   return val;
 }
 
-void minstret_csr_t::bump(const reg_t howmuch) noexcept {
+void wide_counter_csr_t::bump(const reg_t howmuch) noexcept {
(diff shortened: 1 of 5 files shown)


scottj97 (approved):
This almost exactly matches the same change I'd made in my local fork of Spike.

The one commit was a bit much to review at once though. The renaming of the two classes should have been done as two refactoring commits before a third commit splits the registers.

rswarbrick:
Good point. I've just split things so that the rename goes in a separate commit. I didn't split into two rename commits because that didn't really feel like it made much sense (the first would be e.g. renaming just the top half class, which looked a bit random!). The only difference between the end results is that I've put back a newline that I'd accidentally deleted.

scottj97 (approved):
Thank you, this is improved.

I don't think that each commit needs to make total sense in isolation. It might require looking at a series of commits to understand the goal and purpose. But it's still useful to break down each change into its own commit to ease review, and simplify archaeology in case of a regression discovered later.

aswaterman:
As usual, I endorse Scott's advice going forward.

I'm going to amend this PR to alter the comment to simply say that this scheme models a CPI of 1.  The cosim remark seems extraneous, given that cosim isn't a standard Spike feature.  That use case should be documented separately.

On a historical note, we had originally kept the counters the same to improve simulator performance.  After we moved the `minstret` update off the critical path, we probably should've done so for `mcycle`, too.  Thanks.

PR #923, opened 2022-02-18
Documentation: Cache block size for CMO operations default 64

ingallsj:
https://github.com/riscv-software-src/riscv-isa-sim/pull/891#discussion_r810250913

match https://github.com/riscv-software-src/riscv-isa-sim/blob/42cbf10dc1f6c6ce0a9591a53be2555f75da6c87/spike_main/spike.cc#L245


aswaterman:
thx

Issue #872, opened 2021-12-04
Various traps fail to properly set mstatus.GVA=1


scottj97:
There are several special cases where on a trap, Spike will erroneously set `mstatus.GVA=0` when the value written into `mtval` is a guest virtual address. It should set `mstatus.GVA=1` according to this requirement from the spec section 5.4.1:

> Field GVA (Guest Virtual Address) is written by the implementation whenever a trap is taken into M-mode. For any trap (breakpoint, address misaligned, access fault, page fault, or guest-page fault) that writes a guest virtual address to mtval, GVA is set to 1. For any other trap into M-mode, GVA is set to 0.

1. [Debug trigger traps](https://github.com/riscv-software-src/riscv-isa-sim/blob/c4fdc8fe95e497ab6e107117a7de47ec40e7de7d/riscv/execute.cc#L332) on HLV/HLVX/HSV or load/store with MPRV=1 and MPV=1 (in #850 I fixed only normal loads & stores)
2. Misaligned [LR](https://github.com/riscv-software-src/riscv-isa-sim/blob/c4fdc8fe95e497ab6e107117a7de47ec40e7de7d/riscv/mmu.h#L262-L266) or [SC](https://github.com/riscv-software-src/riscv-isa-sim/blob/c4fdc8fe95e497ab6e107117a7de47ec40e7de7d/riscv/mmu.h#L272-L276) access fault when MPRV=1 and MPV=1
3. MMIO [load](https://github.com/riscv-software-src/riscv-isa-sim/blob/c4fdc8fe95e497ab6e107117a7de47ec40e7de7d/riscv/mmu.cc#L153) or [store](https://github.com/riscv-software-src/riscv-isa-sim/blob/c4fdc8fe95e497ab6e107117a7de47ec40e7de7d/riscv/mmu.cc#L182) where the device detects an access fault, on HLV/HLVX/HSV or load/store with MPRV=1 and MPV=1
4. [LR to I/O space](https://github.com/riscv-software-src/riscv-isa-sim/blob/c4fdc8fe95e497ab6e107117a7de47ec40e7de7d/riscv/mmu.h#L256), when MPRV=1 and MPV=1

scottj97:
To solve this robustly, I am thinking of a data structure that holds info about the desired access:
* virtual address
* effective mode (including 2 bits of privilege and 1 bit of V)
* special cases that alter permissions like HLVX

Then we pass around this data structure instead of a simple virtual address.

[...]

PR #926, opened 2022-02-22
Misc 2022 02 21

chihminchao:


aswaterman:
I don't think the commit about U-mode counters is correct.  The counters are _unprivileged_, not _U-mode_, so it is OK to provide them in M-mode, even if U-mode does not exist.  (To be clear, it is also legal for implementations to skip providing these CSRs, as long as they don't claim to implement Zicntr/Zihpm.)

riscv/insns/mret.h
```diff
@@ -7,8 +7,12 @@ if (prev_prv != PRV_M)
   s = set_field(s, MSTATUS_MPRV, 0);
 s = set_field(s, MSTATUS_MIE, get_field(s, MSTATUS_MPIE));
 s = set_field(s, MSTATUS_MPIE, 1);
-s = set_field(s, MSTATUS_MPP, PRV_U);
 s = set_field(s, MSTATUS_MPV, 0);
+if (p->extension_enabled('H')) {
+  s = set_field(s, MSTATUS_MPP, 0);
```
aswaterman, line 12: PRV_U instead of 0?

riscv/insns/mret.h
```diff
@@ -7,8 +7,12 @@ if (prev_prv != PRV_M)
...
+} else {
+  s = set_field(s, MSTATUS_MPP, p->extension_enabled('U') ? PRV_U : PRV_M);
+}
```
aswaterman: If `p->extension_enabled('H')` is true, then `p->extension_enabled('U')` is also necessarily true.  So you should be able to delete lines 11-13 and 15, and keep only line 14.

chihminchao:
I am  a little confused about the "unprivileged".  Do it means
the existence is decided by zicntr and zihpm in unprivileged spec but the accessibility is  still controlled by [msh]counteren in privilege spec ?

riscv/insns/mret.h
chihminchao: agree  with you and it will be fix in next update

aswaterman:
@chihminchao I think the Intro chapter of the unpriv spec says it well: "Unprivileged instructions are those that are generally usable in all privilege modes in all privileged architectures, though behavior might vary depending on privilege mode and privilege architecture."

Yes, in general the existence of `cycle`/`instret`/`time` is governed by Zicntr, and the existence of `hpmcounter*` is governed by Zihpm.  This is an unprivileged concept, meaning it is true regardless of which privilege modes are implemented.

For M-mode-only implementations, `mcounteren` does not exist.  But even if `mcounteren` did exist, its definition says it affects only modes less than M.  So, `cycle` is always readable in M-mode, provided Zicntr is implemented.

riscv/mmu.cc
```diff
@@ -150,7 +150,8 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
     else if (xlate_flags == 0)
       refill_tlb(addr, paddr, host_addr, LOAD);
   } else if (!mmio_load(paddr, len, bytes)) {
-    throw trap_load_access_fault((proc) ? proc->state.v : false, addr, 0, 0);
+    bool gva = ((proc) ? proc->state.v : false) || (RISCV_XLATE_VIRT & xlate_flags);
```
scottj97: There are many more cases of broken GVA bit. See #872 where I proposed a more general solution to the problem.

riscv/csrs.cc
```diff
@@ -918,17 +918,24 @@ bool const_csr_t::unlogged_write(const reg_t val) noexcept {
...
 }
 
 void counter_proxy_csr_t::verify_permissions(insn_t insn, bool write) const {
-  proxy_csr_t::verify_permissions(insn, write);
+  //check implementation status before privileage
+  unsigned priv = state->prv == PRV_S && !state->v ? PRV_HS : state->prv;
+
+  if ((csr_priv == PRV_S && !proc->extension_enabled('S')) ||
+      (csr_priv == PRV_HS && !proc->extension_enabled('H')))
+    throw trap_illegal_instruction(insn.bits());
```
scottj97: This is nasty copypasta of the permission-checking code that already exists in class `csr_t`.

If this change is really correct (sounds like Andrew might disagree?) I see two alternatives.

In `processor_t::reset()`, check for the existence of these CSRs, and don't instantiate them if they shouldn't exist.

If for some reason that can't work, split up the code in `csr_t::verify_permissions()` into two, so that we have:
```
void csr_t::verify_permissions(insn_t insn, bool write) const {
  verify_existence(insn);
  verify_accessibility(insn, write);
}
```
And the additional checks can be added in an override `counter_proxy_csr_t::verify_existence()`.

scottj97 (changes requested):

riscv/csrs.cc
chihminchao: If I understand the spec correctly,  verify_permission in base class is unnecessary.

chihminchao: I choose the first suggestion is don't initialized them in processor_t::reset_t by isa status.

riscv/insns/mret.h
chihminchao: done

chihminchao:
Q: Is it good to make zicntr default to spike ? Just like zifenci and zicsr ?

riscv/mmu.cc
scottj97: I guess you decided to punt on this fix?

scottj97 (approved):
I'm not sure I understand all of it, but this is much improved.

aswaterman:
@chihminchao For backwards compatibility, I think it would be better if `Zicntr` and `Zihpm` were enabled by default and ignored by the parser, like `Zifencei` and `Zicsr`.  This will avoid breaking people's existing use cases.

chihminchao:
The zicntr and zihpm have been enabled in the beginning of isa_parser_t constructor.

aswaterman:
Works for me.  Thank you, @chihminchao.

PR #928, opened 2022-02-24
mmu: support asid/vmid

chihminchao:
The change makes [v]satp.asid and hgatp.vmid writtable and supports
maximum length for rv32 and rv64. Software could write and read the
satp.asid to get the valid length or check if the core supports
asid/vmid or not.  However, there is no official way to describe this hardware
capability (device tree or something else).  Two implementation flags
are also added for future use and enabled by default.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -782,8 +782,10 @@ reg_t base_atp_csr_t::compute_new_satp(reg_t val) const noexcept {
   reg_t rv64_ppn_mask = (reg_t(1) << (MAX_PADDR_BITS - PGSHIFT)) - 1;
 
   reg_t mode_mask = proc->get_xlen() == 32 ? SATP32_MODE : SATP64_MODE;
+  reg_t asid_mask_if_enabled = proc->get_xlen() == 32 ? SATP32_ASID : SATP64_ASID;
+  reg_t asid_mask = proc->supports_impl(IMPL_MMU_ASID) ? asid_mask_if_enabled : 0;
   reg_t ppn_mask = proc->get_xlen() == 32 ? SATP32_PPN : SATP64_PPN & rv64_ppn_mask;
-  reg_t new_mask = (satp_valid(val) ? mode_mask : 0) | ppn_mask;
+  reg_t new_mask = (satp_valid(val) ? mode_mask : 0) | asid_mask | ppn_mask;
   reg_t old_mask = satp_valid(val) ? 0 : mode_mask;
 
   return (new_mask & val) | (old_mask & read());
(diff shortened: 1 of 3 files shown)


riscv/csrs.cc
```diff
@@ -790,8 +790,9 @@ reg_t base_atp_csr_t::compute_new_satp(reg_t val) const noexcept {
   reg_t rv64_ppn_mask = (reg_t(1) << (MAX_PADDR_BITS - PGSHIFT)) - 1;
 
   reg_t mode_mask = proc->get_xlen() == 32 ? SATP32_MODE : SATP64_MODE;
+  reg_t asid_mask = proc->supports_impl(IMPL_MMU_ASID) ? proc->get_xlen() == 32 ? SATP32_ASID : SATP64_ASID : 0;
```
aswaterman: It's a minor stylistic point, but I'd suggest breaking up this statement like the following (or whatever you prefer):

    reg_t asid_mask_if_mmu_enabled = proc->get_xlen() == 32 ? SATP32_ASID : SATP64_ASID;
    reg_t asid_mask = proc->supports_impl(IMPL_MMU_ASID) ? asid_mask_if_mmu_enabled : 0;

aswaterman:
LGTM, but I'd like review from a non-SiFive person.

riscv/csrs.cc
```diff
@@ -981,9 +982,12 @@ bool hgatp_csr_t::unlogged_write(const reg_t val) noexcept {
...
   reg_t mask;
   if (proc->get_const_xlen() == 32) {
-    mask = HGATP32_PPN | HGATP32_MODE;
+    mask = HGATP32_PPN |
+        HGATP32_MODE |
+        proc->supports_impl(IMPL_MMU_VMID) ? HGATP32_VMID : 0;
   } else {
-    mask = HGATP64_PPN & ((reg_t(1) << (MAX_PADDR_BITS - PGSHIFT)) - 1);
+    mask = HGATP64_PPN  |
+        proc->supports_impl(IMPL_MMU_VMID) ? HGATP64_VMID : 0;
```
scottj97: This seems to break the `MAX_PADDR_BITS` masking for `hgatp`. (It's still there in `rv64_ppn_mask` for `satp`.)

scottj97 (changes requested):

riscv/csrs.cc
chihminchao: @scottj97 
I don't understand where hgatp is related to MAX_PADDR_BITS and rv64_ppn_mask. Could you point out that ?

scottj97: The PPN field in `hgatp` is WARL, same as `satp`, so the spec allows but does not require this masking. I see no reason to allow a bigger PPN than the system supports though.

In any case this commit says it changes ASID/VMID; it says nothing about changing the PPN field too. If you want to change that, it should be a separate commit with an explanation.

chihminchao: Thanks for your clarification.  I didn't notice the removal of MAX_PADDR_BITS in my change.

scottj97 (approved):

aswaterman:
I suspect we might eventually want the option to curtail the ASID length, but we can cross that bridge when we come to it.

PR #929, opened 2022-02-24
adding ports to run  F, D and E extension architectural tests on spike

neelgala:
- updated the arch_test_target/device directory to include F, D and rv32e_unratified extensions
- using register t1 instead of t5 in RVMODEL_HALT since its supported in E as well 
- Related issue https://github.com/riscv-non-isa/riscv-arch-test/issues/227

### DIFF
--- a/arch_test_target/spike/device/rv32e_unratified/C/Makefile.include
+++ b/arch_test_target/spike/device/rv32e_unratified/C/Makefile.include
@@ -1,0 +1,40 @@
+TARGET_SIM   ?= spike
+TARGET_FLAGS ?= $(RISCV_TARGET_FLAGS)
+ifeq ($(shell command -v $(TARGET_SIM) 2> /dev/null),)
+    $(error Target simulator executable '$(TARGET_SIM)` not found)
+endif
+
+RISCV_PREFIX   ?= riscv32-unknown-elf-
+RISCV_GCC      ?= $(RISCV_PREFIX)gcc
+RISCV_OBJDUMP  ?= $(RISCV_PREFIX)objdump
+RISCV_GCC_OPTS ?= -g -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles $(RVTEST_DEFINES)
+
+COMPILE_CMD = $$(RISCV_GCC) $(1) $$(RISCV_GCC_OPTS) \
+							-I$(ROOTDIR)/riscv-test-suite/env/ \
+							-I$(TARGETDIR)/$(RISCV_TARGET)/ \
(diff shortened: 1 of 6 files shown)


aswaterman (commented):
I have been treating the `arch_test_target` directory as a black box, and I'll keep doing so if that's what's easiest for everyone.  But I'm concerned by this ABI breakage (`t5` vs. `t1`).  The implication is that only some versions of Spike work with some versions of the tests.  How is this interface being versioned?

neelgala:
The peice of code where I did the change is part of spike's exit routine (RVMODEL_HALT macro is target/device specifi) and is independent of the tests. Keeping this snippet compatible for rv32e and rv32i made more sense - hence the change. I don't think there are any implications of the sorts you are suggesting.

aswaterman:
If that is the case, where is the value feeding `t1` coming from?  I can't figure out why it's OK to simply change `t5` to `t1`.  If any value is OK, why not `x0`?

neelgala:
I have changed `sw x1, tohost, t5;` to `sw x1, tohost, t1;`. So t5 was being used as a temporary base register. so it can't be x0 but can be anything between x1-x15 to work on an rv32e device. I chose t5 because someone else got spike working with this change for the rv32e tests.

aswaterman:
Yeah, I now see that it's a temp argument.  (I should've figured that out, since I'm the one who created that assembly syntax to begin with :-\ )

scottj97:
@neelgala changing t1 to t5 is a separate change and therefore should be a separate commit.

There's an awful lot of copypasta in all the Makefile.include files. It would be better to put all the common stuff in one file and specify only the differences in all these directories' individual Makefiles.

neelgala:
@scottj97 the initial reason for copypasta was to keep the whole thing extremely simple for absolute beginners. But I guess now I can push in another PR to reduce the duplicate stuff.

Skepfyr:
@aswaterman I've just been a a whole arc trying to work out what `sw x1, tohost, t1;` does and ended up here. Where is this asm syntax documented?

For anyone else who ends up here I believe what's going on is that this 3 argument syntax gets translated to an `auipc` instruction that stores the address in t1 (or whatever register you supply) and then executes a normal `sw x1, 0(t1)`. I think this is to allow stores to labels that are up to 32 bits away rather than 12 bits away.

aswaterman:
@Skepfyr you're exactly right.  This behavior is documented in the RISC-V assembly manual: https://github.com/riscv-non-isa/riscv-asm-manual/blob/3b06fa93fd52f277e8e0dc180bbb2ceec1ff5957/src/asm-manual.adoc#load-and-store-global

Issue #932, opened 2022-02-26
avoid printing hstatus when hypervisor not enabled


neelgala:
I am using the following command:
```
spike --log ref.dump --log-commits --priv=msu --isa=rv64ima rv64im-Illegal-p.elf
```
What I observe is that when sret is executed the following entry is is made in ref.dump

```
core   0: 1 0x0000000080000264 (0x10200073) c1536_hstatus 0x0000000200000000 c768_mstatus 0x0000000a001000a8
```

Is it possible to print hstatus in the above line only when hypervisor is enabled via the cli --isa arg ? Where should I start looking to find a fix for this ?

scottj97:
Fixed in #934

PR #934, opened 2022-02-28
perform hstatus probe/update for sret only when H extension enabled.

neelgala:
motivation for this is #932 . This fix is meant to disable logging of hstatus on srets when H extension is not enabled.

### DIFF
--- a/riscv/insns/sret.h
+++ b/riscv/insns/sret.h
@@ -16,11 +16,12 @@ s = set_field(s, MSTATUS_SPP, PRV_U);
 STATE.sstatus->write(s);
 p->set_privilege(prev_prv);
 if (!STATE.v) {
-  reg_t prev_virt = get_field(prev_hstatus, HSTATUS_SPV);
-  p->set_virt(prev_virt);
-
-  reg_t new_hstatus = set_field(prev_hstatus, HSTATUS_SPV, 0);
-  STATE.hstatus->write(new_hstatus);
+  if p->extension_enabled('H') {
+    reg_t prev_virt = get_field(prev_hstatus, HSTATUS_SPV);
+    p->set_virt(prev_virt);
+    reg_t new_hstatus = set_field(prev_hstatus, HSTATUS_SPV, 0);
+    STATE.hstatus->write(new_hstatus);
+  }
 
   STATE.mstatus->write(set_field(STATE.mstatus->read(), MSTATUS_MPRV, 0));
 }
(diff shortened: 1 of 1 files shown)


neelgala:
apologies for the typo in the branch branch name - it should have been hstatus. 
@aswaterman @scottj97 let me know if any more changes required related to this ?

riscv/insns/sret.h
```diff
@@ -16,11 +16,12 @@ s = set_field(s, MSTATUS_SPP, PRV_U);
 STATE.sstatus->write(s);
 p->set_privilege(prev_prv);
 if (!STATE.v) {
-  reg_t prev_virt = get_field(prev_hstatus, HSTATUS_SPV);
-  p->set_virt(prev_virt);
-
-  reg_t new_hstatus = set_field(prev_hstatus, HSTATUS_SPV, 0);
-  STATE.hstatus->write(new_hstatus);
+  if p->extension_enabled('H') {
```
aswaterman: missing parens around predicate?

neelgala: fixed

riscv/insns/sret.h
```diff
@@ -16,11 +16,12 @@ s = set_field(s, MSTATUS_SPP, PRV_U);
...
+    reg_t prev_virt = get_field(prev_hstatus, HSTATUS_SPV);
+    p->set_virt(prev_virt);
+    reg_t new_hstatus = set_field(prev_hstatus, HSTATUS_SPV, 0);
+    STATE.hstatus->write(new_hstatus);
+  }
```
aswaterman, line 24: I was initially skeptical that `sret` would be the only place you'd need to make this change.  But based on some cursory grepping, it seems to be the case that this change will suffice.

neelgala: same here.. grep resulted only in this one instruction requiring this change. Everywhere else its just a read operation.

scottj97 (approved):
Change looks safe and innocuous.

Note that when `misa.H=0`, [this line](https://github.com/riscv-software-src/riscv-isa-sim/blob/8ca012a6cc7e5b9edfaacca2d04b5a0f267377cf/riscv/csrs.cc#L541) makes sure that `hstatus` is always 0. So the only line that really needs to be qualified here is the call to `write()`, and then only to avoid logging.

aswaterman:
Indeed.  To my eyes, at least, this version reads better.

PR #938, opened 2022-03-03
Move towards configuring Spike through a device tree

rswarbrick:
*This is a long series of commits (sorry!) and is reasonably invasive. I'm very happy to split the PR into several pieces to be merged separately if that's more helpful, but I thought it might make sense to show an "end goal" for the staging commits that appear near the start. I'm also happy to rewrite/refactor things as desired, but it probably only makes sense to do that if we agree that this is going in a sensible direction overall.*

The basic idea is to represent more and more of Spike configuration by a "DTB" file (a binary representation of a devicetree).

Before these changes, we could dump a devicetree in textual form (with the `--dump-dts` argument) or consume a binary device tree representation (with the `--dtb` argument). If a user passed `--dtb` and also some other configuration arguments (`--isa` or `-p`, say) we'd do *something* based on the two. In particular, we ignored the ISA implied by the DTB file, always taking the value of `--isa` from the command line. But we spat out an error if `-p` from the command line didn't match the devicetree. 

With these changes, the configuration goes like this if you have no DTB:
```
    cmd line -> DTS -> DTB -- (dtb parser) --> config
```
and this if you have a DTB:
```
    DTB -- (dtb parser) -->
                     /----> config -\
                     |              v
    cmd line --------+----> consistency check
```
(pardon the rubbish ASCII art!)

The consistency check is intended to be a bit more principled than what we had before. At the moment, it fails with an error if a command line argument has been explicitly given and doesn't match the DTB. We could equally well change that behaviour to e.g. override the DTB: I went with the semantics that more closely matched what I think is there at the moment, but I'm happy to go the other way.

At the moment, the configuration information that we get from the DTB is used for exactly the arguments that were needed by the `make_dts` function. This is the bare minimum needed to be able to pull this configuration machinery out of `sim_t`. Clearly, there are lots more configuration arguments that can follow suit if we decide to take this further, but this PR was intended as the "MVP" version to avoid wasting lots of effort if people hated the general approach!

### DIFF
--- a/riscv/cfg.cc
+++ b/riscv/cfg.cc
@@ -1,0 +1,25 @@
+// See LICENSE for license details.
+
+#include "cfg.h"
+#include "isa_parser.h"
+
+cfg_t::~cfg_t() {}
+
+const cfg_t &cfg_t::freeze()
+{
+  assert(!frozen);
+
+  assert(!isa_parser);
+  isa_parser.reset(new isa_parser_t(isa(), priv()));
+
+  frozen = true;
+  return *this;
+}
+
+cfg_arg_t<const isa_parser_t *> cfg_t::get_isa_parser() const
+{
+  assert(frozen && isa_parser);
+  cfg_arg_t<const isa_parser_t *> ret(isa_parser.get());
+  if (isa.overridden() || priv.overridden()) ret.set_overridden();
+  return ret;
+}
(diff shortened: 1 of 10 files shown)


rswarbrick:
Force-push rebases over the stuff that's been merged so far (not trivial: the design changed rather dramatically!). I've already split the first commit in the queue out (#955). I'll spin out the next couple of commits into their own PRs in a minute.

PR #939, opened 2022-03-03
Change some methods to take a const isa_parser_t

rswarbrick:
*This is part of the chain of commits in #938 but seemed easy to review separately. I'll rebase the larger PR if this one is merged*

No functional change, but this is needed for a following refactor
where we're passing it around `const`.


aswaterman:
Yes, please only make small PRs when feasible.

PR #940, opened 2022-03-03
Incorporate supported privilege levels into isa_parser_t

rswarbrick:
*This is another commit plucked from #938. It probably makes sense no matter what we think about the general direction of that PR*

These affect the "max_isa" value (now exposed as get_max_isa()) and
feel like they're similar to the other computations done in
isa_parser_t.

### DIFF
--- a/disasm/disasm.cc
+++ b/disasm/disasm.cc
@@ -2032,7 +2032,7 @@ disassembler_t::disassembler_t(const isa_parser_t *isa)
   // next-highest priority: other instructions in same base ISA
   std::string fallback_isa_string = std::string("rv") + std::to_string(isa->get_max_xlen()) +
     "gcv_zfh_zba_zbb_zbc_zbs_zkn_zkr_zks_xbitmanip";
-  isa_parser_t fallback_isa(fallback_isa_string.c_str());
+  isa_parser_t fallback_isa(fallback_isa_string.c_str(), DEFAULT_PRIV);
   add_instructions(&fallback_isa);
 
   // finally: instructions with known opcodes but unknown arguments
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -26,7 +26,7 @@
 processor_t::processor_t(const char* isa, const char* priv, const char* varch,
                          simif_t* sim, uint32_t id, bool halt_on_reset,
                          FILE* log_file, std::ostream& sout_)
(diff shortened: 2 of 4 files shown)


aswaterman:
I'm not sure about this one, for two reasons:
- the priv string, as it currently exists, has nothing to do with ISA strings, so passing it to the isa_parser doesn't seem like proper factoring.
- this probably creates additional technical debt, because `--priv` is a bit of a hack that I imagine we will deprecate and remove once RVI defines ISA strings that have the capability of expressing what we're currently expressing with the combination `--isa` and `--priv`.

rswarbrick:
Thanks for the reply. The reason I did it like this is that the `max_isa` property on the ISA parser is affected by the priv string. At the moment, the processor subclass updates the parent class `max_isa` to make that work, which rather blurs the boundaries between the two classes.

I *think* I still believe this refactoring is a good idea. After all, you say above that we're eventually going to fold the `--priv` functionality into the ISA itself (once RVI defines appropriate ISA strings): isn't that an argument that this is the right place for the temporary hack to live?

If you feel strongly that we shouldn't do it like this, another approach would be to use a getter interface on `processor_t` that does the override for us. (Something like a `processor_t::get_max_isa()`, which will have pre-computed the value based on the underlying ISA parser plus the priv string). Do you think that would be better?

aswaterman:
Maybe we can solicit another opinion from @scottj97.

scottj97:
> The reason I did it like this is that the `max_isa` property on the ISA parser is affected by the priv string. At the moment, the processor subclass updates the parent class `max_isa` to make that work, which rather blurs the boundaries between the two classes.

I like this PR for the above reason.

(Also I just noticed that `processor_t` inherits from `isa_parser_t`, which is rather smelly. It feels like `processor_t` should contain an instance of `isa_parser_t` instead. Prefer composition over inheritance.)

rswarbrick:
>  (Also I just noticed that processor_t inherits from isa_parser_t, which is rather smelly. It feels like processor_t should contain an instance of isa_parser_t instead. Prefer composition over inheritance.)

See the next commit in #938 (which is what prompted this change in the first place) :-D

rswarbrick:
@aswaterman: Are you happy to proceed with this? Or are there changes that you'd like?

aswaterman:
It's fine with me.

PR #944, opened 2022-03-09
Refactor trigger code

timsifive:
In preparation for adding more trigger types, refactor the code a bit.

Almost all trigger logic is now in its own class.
Triggers are stored as internal state, getting rid of the awkward replication inside the tdata2 CSR implementation.
Added support for mcontrol6 triggers, but haven't tested it yet.


riscv/triggers.h
```diff
@@ -0,0 +1,136 @@
...
+
+    trigger_t *trigger(unsigned index) {
+      return triggers[index];
+    }
+
+    unsigned count() const {
+      return triggers.size();
+    }
+
+  //private:
```
aswaterman: Was this commented out during debugging?

scottj97:
Based only on the commit messages, it sounds like some of the commits here are known to be broken? Please rebase/squash so that everything is working at every commit. (Though all new features don't have to be complete of course.)

Also the PR title mentions mcontrol6, so I would expect one of the commit messages to mention its addition.

riscv/triggers.h
timsifive: It's really because I didn't quite finish refactoring all the way. Currently processor_t::trigger_updated() still needs direct access to the member. I'll add a comment.

timsifive:
@scottj97 I squashed the work.

scottj97:
Squashing it all into one commit wasn't really what I meant; it's better to have refactorings separated from the functional changes like adding mcontrol6. I can easily see this being like 25+ commits.

One of my duties as a maintainer of Spike is to make sure that changes are clearly communicated to all affected users (and there will be many here). This PR currently does not achieve that. For example, my (proprietary, natch) tests show a change in the behavior of mcontrol's `tdata1.maskmax` field. That is not mentioned anywhere in this PR or the commit messages. I don't know if that change is a bug or a feature.

I'd use `git bisect` to narrow it down, but there's only one giant commit, so that's not possible.

Another test on the current master branch exits prematurely with code 255; it continues past that point and exits normally on this PR, so I assume there's some kind of bug fix somewhere in here?

timsifive:
It's a pleasant surprise to hear that there are tests (besides the debug tests) that cover the trigger registers. If you can point me to how to run them, or explain in more detail what changed, I'd be happy to take a look at that.

By code inspection, it looks like maskmax was probably reported as 0 before, while now it is set to 0x3f. That's more likely to be right than 0, but I'm not aware of anything that actually tries to use mcontrol.match=1, where this is relevant.

Spending a bunch of time to break this change down into a bunch of tiny ones just in case that helps somebody in the future doesn't seem like a worthwhile use of time. If this change turns out to cause a bug, let me know and I'll help figure it out.

scottj97 (changes requested):
It's not "just in case"...it will definitely help at least one person (me). If this PR (or any PR) causes a change in Spike's behavior, that change needs to be intentional, deliberate, and understood. Doesn't it bother you that this "refactoring" caused a visible change in behavior? It should -- because that makes it not just a refactoring anymore. But because it's one giant commit, that behavioral change is obscured by all the actual refactoring. Now everybody who cares about Spike's triggers will have to study this entire 900-line diff to try to decipher how the behavior changed -- not a worthwhile use of their time.

[Small commits are beautiful.](https://madewithlove.com/blog/software-engineering/small-commits-how-to-make-a-difference-when-reviewing-code/)

NACK on this PR as of c12c3f179cd512dc17595a4d8b6c3cd18a011e15, because:
* It is unreviewable as a single 900-line commit
* It causes unexplained behavior changes in the triggers
* It commingles new functionality (mcontrol6) with large-scale refactoring

timsifive:
> Doesn't it bother you that this "refactoring" caused a visible change in behavior?

No. The change in question changes spike behavior from clearly incorrect to probably correct. It was a one-line change. I should have mentioned it in the description, though.

I will redo this work using small commits, but I'm not convinced that for this simple change such effort is worth the time.

> Another test on the current master branch exits prematurely with code 255; it continues past that point and exits normally on this PR, so I assume there's some kind of bug fix somewhere in here?

Can you point me at this test/how to run the test suite/whatever link? I did reimplement how `chain` works and I think I got it better than the previous implementation. But again to my knowledge nobody has ever used this behavior. I don't mind changing behavior that nobody uses.

scottj97:
> No. The change in question changes spike behavior from clearly incorrect to probably correct. It was a one-line change.

No it was not. It was a 900-line change. It _should_ be a one-line change, with a commit message explaining the change and its rationale.

> I will redo this work using small commits, but I'm not convinced that for this simple change such effort is worth the time.

It is always worth the time if you care about software quality, because change management is a critical component of QC.

> Can you point me at this test/how to run the test suite/whatever link? I did reimplement how `chain` works and I think I got it better than the previous implementation. But again to my knowledge nobody has ever used this behavior. I don't mind changing behavior that nobody uses.

I've been using chained triggers in Spike for 1.5 years, though I had to do quite a bit of work in my local fork to get them to work correctly. Generally I contribute bug fixes back to the open-source repo but in this case I'd made lots of other changes for platform customization and the chain fixes don't apply cleanly to this upstream repo.

I'm afraid my tests are proprietary, but once this is broken up into small changes I can bisect and pinpoint exactly where the behavior changed. If the commits are small enough, it will be obvious at that point.

When it comes to implementing new behaviors in Spike, I think our policy should be to disallow merging untested changes. If the PR author has not proven the implementation with at least some sanity (perhaps proprietary) tests then the feature is not ready for merging. Therefore I recommend you submit mcontrol6 as a separate PR after it's had some basic testing done.

And a sincere thank you for your help.

scottj97:
This is interesting, because as of 521506ac000a66c9eb730fe9c9e07d95a4acd356 I no longer see any behavior changes versus master (in maskmax or anywhere else), except for one assertion starting with 28940caea640b007be11114f35d4229a84dbaf86:
```
spike: riscv-isa-sim/riscv/triggers.cc:119: virtual bool triggers::mcontrol_t::memory_access_match(processor_t*, triggers::operation_t, reg_t, reg_t): Assertion `0' failed.
```

This is triggered by the fallthrough case in `memory_access_match` where there's an unknown `match` value. 

This is the same test that was exiting prematurely before, then succeeding as of c12c3f179cd512dc17595a4d8b6c3cd18a011e15. It writes `tdata1=0x2000000000013e30` which has `match=6` which is undefined. It seems some WARL-ness is missing in `mcontrol_t::tdata1_write()`?

Though perhaps we could say this is out of scope for this PR.

(I still plan to look through all these commits in more detail.)

timsifive:
> This is interesting, because as of https://github.com/riscv-software-src/riscv-isa-sim/commit/521506ac000a66c9eb730fe9c9e07d95a4acd356 I no longer see any behavior changes versus master (in maskmax or anywhere else)

Right. This work is not identical to the previous PR. It goes a little further in the refactoring, and doesn't incorporate any trivial fixes that I made in the original one.

I'll add WARL to fix the assert. spike should only assert if there's an internal error. Not due to some code it's executing.

scottj97:
Excellent, the test is completing as expected now.

Unfortunately that test is the best coverage I have of these debug CSRs, and it's broken all the way up until 0f066bf, so I cannot thoroughly test all the rest of this PR. Would it be a big mess to rebase that change back to the beginning of the series? Then I can get much better test coverage of the rest of the commits to ensure there's no unexpected behavior changes.

riscv/triggers.cc
```diff
@@ -0,0 +1,200 @@
...
+      return MATCH_FIRE_AFTER;
+    else
+      return MATCH_FIRE_BEFORE;
+  }
+  return MATCH_NONE;
+}
+
+module_t::module_t(unsigned count) : triggers(count) {
+  for (unsigned i = 0; i < count; i++) {
+    triggers[i] = new mcontrol_t();
```
scottj97, line 146: Shouldn't this have a matching destructor to `delete`?

riscv/triggers.cc
```diff
@@ -0,0 +1,200 @@
...
+namespace triggers {
+
+mcontrol_t::mcontrol_t() :
+  type(2), maskmax(0), select(false), timing(false), chain_bit(false),
+  match(MATCH_EQUAL), m(false), h(false), s(false), u(false),
+  execute_bit(false), store_bit(false), load_bit(false)
+{
+}
+
+reg_t mcontrol_t::tdata1_read(const processor_t *proc) const noexcept {
```
scottj97: Could this be `const processor_t * const proc`? (Same question for many other `*proc` through this code)

riscv/mmu.h
```diff
@@ -481,19 +470,20 @@ class mmu_t
...
     }
-    int match = proc->trigger_match(operation, address, data);
-    if (match == -1)
+    triggers::action_t action;
+    auto match = proc->TM.memory_access_match(&action, operation, address, data);
+    if (match == triggers::MATCH_NONE)
       return NULL;
-    if (proc->state.mcontrol[match].timing == 0) {
-      throw trigger_matched_t(match, operation, address, data);
+    if (triggers::MATCH_FIRE_BEFORE) {
```
scottj97: Shouldn't this compare vs `match` i.e. `if (match == triggers::MATCH_FIRE_BEFORE)`? Otherwise it seems like this will always be taken, since `triggers::MATCH_FIRE_BEFORE` is nonzero.

riscv/processor.h
```diff
@@ -354,88 +314,7 @@ class processor_t : public abstract_device_t
...
-      if (!state.mcontrol[i].chain) {
-        return i;
-      }
-      chain_ok = true;
-    }
-    return -1;
-  }
-
-  void trigger_updated();
+  void trigger_updated(const std::vector<triggers::trigger_t *> *triggers);
```
scottj97: I think the preferred C++ way here is to use const ref instead of passing a pointer to vector.

scottj97 (changes requested):
I like this refactoring a lot. It's going to cause me some pain because I'd made a very similar set of changes in my local fork, but that's what I get for not contributing back to upstream.

There are still some very large commits in here that could easily be split up (e.g. 9ec51562f0daef3a7de87b41c6bb3753c4f467b9, 519915a064d0fc3ad979a40b3e16b5447f40a36a, 803373ce892208de4af4f868f71025c95b55f971) but I'll stop harping on that.

timsifive:
> Would it be a big mess to rebase that change back to the beginning of the series? Then I can get much better test coverage of the rest of the commits to ensure there's no unexpected behavior changes.

I did the easy rebase, which moves it quite a lot earlier but not quite to the start.

riscv/triggers.cc
timsifive: I'll write one.

riscv/triggers.cc
timsifive: I find the notation for that pretty cumbersome, and the benefit negligible, so I generally don't bother.
But I'll add it.

riscv/mmu.h
timsifive: Yes. Good catch.

riscv/processor.h
timsifive: I'll change it.

riscv/triggers.cc
```diff
@@ -0,0 +1,206 @@
...
+}
+
+module_t::module_t(unsigned count) : triggers(count) {
+  for (unsigned i = 0; i < count; i++) {
+    triggers[i] = new mcontrol_t();
+  }
+}
+
+module_t::~module_t() {
+  for (unsigned i = 0; i < triggers.size(); i++) {
```
scottj97: Any reason we couldn't use a modern loop e.g. `for (auto trigger : triggers)`?

scottj97 (approved):
Excellent. One nit and I think this is ready to merge.

riscv/triggers.cc
timsifive: Done.

scottj97 (approved):
Excellent, thank you for all the work.

timsifive:
What's the process for actually merging this? It looks like I have permission to merge the PR. Is it on me to push the green button?

scottj97:
I was giving Andrew some time to review, but I'll go ahead and merge if you're waiting on this.

timsifive:
Thanks. I'd be fine waiting for Andrew, but was just curious what the process was. I'm going on vacation for a week starting tomorrow so I'm not doing anything immediately in any case.

PR #945, opened 2022-03-12
Construct an isa_parser_t and pass it to processor_t constructor

rswarbrick:
*This commit comes from the series at #938*

This is a minor change, turning `processor_t` from a child class of
`isa_parser_t` into a class that contains an `isa_parser_t` as a field.
The point is that it is a step toward separating out
"configuration" (and ISA string parsing) from processor state. This
should be helpful for rejigging things so that we construct more from
a supplied device tree.


aswaterman (approved):
Agreed that we should favor composition over inheritance.

scottj97:
This is pushing quite large for one commit. Things like renaming variables (`isa`->`isa_string`, `isa_extensions`->`extensions`) should each be their own commit, to avoid cluttering up and obscuring the substantive changes going on.

With that caution, I will go ahead and merge.

PR #948, opened 2022-03-14
Compile using C++17 standard

scottj97:
It would [be nice](https://github.com/riscv-software-src/riscv-isa-sim/pull/946#discussion_r826335981) to be able to use modern C++ features like `std::optional`.

Are all users of Spike using modern C++ compilers by now?


aswaterman:
Speaking for me and my employer, we're OK with requiring C++17 (and we like `optional`).

PR #949, opened 2022-03-15
Fix perf regression from CSR refactoring

aswaterman:
Since many instructions are only conditionally legal, their implementations need to query misa (or isa).  Since reading misa is therefore on the critical path, make sure it's inlined.

Making misa_csr_t a final class removes the need for a vtable lookup for `read()`.

On my simple test case (`for (volatile int i = 0; i < 200000000; i++) {}`), this increases perf by around 45%.

### DIFF
--- a/riscv/csrs.h
+++ b/riscv/csrs.h
@@ -8,6 +8,7 @@
 #include <memory>
 // For access_type:
 #include "memtracer.h"
+#include <cassert>
 
 class processor_t;
 struct state_t;
@@ -61,7 +62,11 @@ typedef std::shared_ptr<csr_t> csr_t_p;
 class basic_csr_t: public csr_t {
  public:
   basic_csr_t(processor_t* const proc, const reg_t addr, const reg_t init);
-  virtual reg_t read() const noexcept override;
+
+  virtual reg_t read() const noexcept override {
+    return val;
+  }
+
  protected:
   virtual bool unlogged_write(const reg_t val) noexcept override;
  private:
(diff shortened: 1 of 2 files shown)


riscv/csrs.h
```diff
@@ -61,7 +62,11 @@ typedef std::shared_ptr<csr_t> csr_t_p;
 class basic_csr_t: public csr_t {
  public:
   basic_csr_t(processor_t* const proc, const reg_t addr, const reg_t init);
-  virtual reg_t read() const noexcept override;
+
+  inline virtual reg_t read() const noexcept override {
```
scottj97: The `inline` keyword doesn't have any effect here, according to [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Inline.html): "As required by ISO C++, GCC considers member functions defined within the body of a class to be marked inline even if they are not explicitly declared with the inline keyword."

So is there any reason to specify it explicitly?

scottj97 (changes requested):

riscv/csrs.h
aswaterman: No, you're right, I forgot about that.

scottj97 (approved):

scottj97:
I'm curious why your simple test case saw such a big difference. Why would that loop need to query any CSR?

Maybe instruction fetch reading `mstatus`? There might be some more low-hanging fruit here.

aswaterman:
Most of the instructions in the loop are compressed and so must query the `misa.C` bit.  There is one other place I observed some low-hanging fruit (checking sstatus.FS for FP instructions) that I'll PR in a moment.  I'm guessing the rest is a long tail and not worth thinking about at the moment.

PR #950, opened 2022-03-16
Fix perf regressions from CSR refactoring in FP-heavy code 

aswaterman:
Since all FP instructions implicitly read `sstatus.FS` (and most implicitly write it), we are sensitive to the performance of those accesses.  Restructure accordingly.

For my simple test case (`for (volatile double f = 0; f < 200000000; f += 1.00000000001) {}`), these changes give a 3.4x speedup, which is roughly evenly divided between the changes to the read path (the first seven commits) and the changes to the write path (the final two).

This is a follow-on to #949.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -421,11 +421,6 @@ mstatus_csr_t::mstatus_csr_t(processor_t* const proc, const reg_t addr):
 }
 
 
-reg_t mstatus_csr_t::read() const noexcept {
-  return val;
-}
-
-
 bool mstatus_csr_t::unlogged_write(const reg_t val) noexcept {
   const bool has_mpv = proc->extension_enabled('S') && proc->extension_enabled('H');
   const bool has_gva = has_mpv;
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -464,9 +464,9 @@ void sstatus_csr_t::dirty(const reg_t dirties) {
   // checking for mstatus.VS!=Off:
   if (!enabled(dirties)) abort();
 
-  orig_csr->write(orig_csr->read() | dirties);
+  orig_sstatus->write(orig_sstatus->read() | dirties);
   if (state->v) {
-    virt_csr->write(virt_csr->read() | dirties);
+    virt_sstatus->write(virt_sstatus->read() | dirties);
   }
 }
 
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -394,10 +394,6 @@ sstatus_proxy_csr_t::sstatus_proxy_csr_t(processor_t* const proc, const reg_t ad
   mstatus(mstatus) {
 }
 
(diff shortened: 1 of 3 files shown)


riscv/csrs.cc
```diff
@@ -454,13 +454,19 @@ sstatus_csr_t::sstatus_csr_t(processor_t* const proc, sstatus_proxy_csr_t_p orig
...
+
   // Catch problems like #823 where P-extension instructions were not
   // checking for mstatus.VS!=Off:
   if (!enabled(dirties)) abort();
 
-  orig_csr->write(orig_csr->read() | dirties);
+  orig_sstatus->write(orig_sstatus->read() | dirties);
   if (state->v) {
-    virt_csr->write(virt_csr->read() | dirties);
+    virt_sstatus->write(virt_sstatus->read() | dirties);
```
scottj97, line 469: I guess this change avoids a vtable lookup? Since the `*_sstatus` pointers are to `final` classes?

Because the first half of 8974912c6237e072284942087fbc1ceefb36e29a modifies the logging (many unnecessary `mstatus` writes disappear from the commit log) I would like that commit split up so it's clear exactly what caused that change. Also the commit message should mention the logging change, which surprised me at first.

riscv/csrs.h
```diff
@@ -621,14 +621,21 @@ class dcsr_csr_t: public csr_t {
...
-class float_csr_t: public masked_csr_t {
+class float_csr_t final: public masked_csr_t {
  public:
   float_csr_t(processor_t* const proc, const reg_t addr, const reg_t mask, const reg_t init);
   virtual void verify_permissions(insn_t insn, bool write) const override;
+
+  reg_t read() const noexcept override {
+    return masked_csr_t::read();
+  }
+
```
scottj97: Does this really add anything? Seems like since class `float_csr_t` is now `final` the compiler should be able to inline the parent `read()` without this.

scottj97 (changes requested):

riscv/csrs.cc
aswaterman: Correct re: vtable.  I actually could've left these alone since the first part of the commit makes this code rarely executed.  But since the first part of the commit needs to access `orig_sstatus` and `virt_sstatus` for perf reasons, I thought it looked confusing to use `orig_csr` and `virt_csr` later on in the same method.

I didn't realize I was removing side effects by removing the writes.  Thanks for pointing that out; I'll split out that commit.

riscv/csrs.h
aswaterman: I would've thought that, too, but it removed a vtable lookup, so I left it in.  Your comment made me drill down a bit further; I discovered moving from gcc 8 to gcc 11 or clang 12 makes this useless, as we'd hope.  So I'll remove it.

aswaterman:
Thanks for the careful review!

scottj97 (approved):

PR #951, opened 2022-03-16
Raise illegal (not virtual) instruction exception on counter writes

aswaterman:
Looks like a regression introduced in 53a3002e8cdbf94016d36c6071945bd663e826d5

Writes to read-only CSRs are not HS-qualified so should not raise virtual instruction exceptions.

cc @ingallsj @chihminchao


ingallsj (approved):

scottj97 (approved):
I'm embarrassed that I didn't have better test coverage for this issue, especially after we discussed it in #926.

I've added more test coverage for this and I confirm that the behavior after this patch matches the behavior prior to 53a3002e8cdbf94016d36c6071945bd663e826d5

aswaterman:
Cool. Thanks for validating the fix, Scott!

PR #954, opened 2022-03-24
Move a few more configuration arguments into cfg_t

rswarbrick:
This is a continuation of #946 and is sort of orthogonal to #938 (which sparked the work in the first place). The idea here is that we move more of the configuration that describes the simulated device into `cfg_t` (which will also eventually be used to represent the stuff that we parse from a device tree). I *think* this is the rest of the command line arguments that are easy to move from `spike.cc`.

Eventually, we might also want to move things like `mems`, parsing the command line arguments to some configuration object in `spike.cc` and then actually instantiating memory objects in the `sim_t` constructor. But that's a bit more difficult and invasive! So I've gone for the low hanging fruit here.

### DIFF
--- a/riscv/cfg.h
+++ b/riscv/cfg.h
@@ -56,12 +56,14 @@ public:
   cfg_t(std::pair<reg_t, reg_t> default_initrd_bounds,
         const char *default_bootargs,
         const char *default_isa, const char *default_priv,
+        const char *default_varch,
         const std::vector<mem_cfg_t> &default_mem_layout,
         const std::vector<int> default_hartids)
     : initrd_bounds(default_initrd_bounds),
       bootargs(default_bootargs),
       isa(default_isa),
       priv(default_priv),
+      varch(default_varch),
       mem_layout(default_mem_layout),
       hartids(default_hartids),
       explicit_hartids(false)
@@ -71,6 +73,7 @@ public:
   cfg_arg_t<const char *>            bootargs;
   cfg_arg_t<const char *>            isa;
   cfg_arg_t<const char *>            priv;
+  cfg_arg_t<const char *>            varch;
   cfg_arg_t<std::vector<mem_cfg_t>>  mem_layout;
   std::optional<reg_t>               start_pc;
   cfg_arg_t<std::vector<int>>        hartids;
--- a/riscv/cfg.h
+++ b/riscv/cfg.h
@@ -54,13 +54,12 @@ class cfg_t
 {
 public:
   cfg_t(std::pair<reg_t, reg_t> default_initrd_bounds,
-        const char *default_bootargs, size_t default_nprocs,
+        const char *default_bootargs,
         const char *default_isa, const char *default_priv,
         const std::vector<mem_cfg_t> &default_mem_layout,
         const std::vector<int> default_hartids)
     : initrd_bounds(default_initrd_bounds),
       bootargs(default_bootargs),
-      nprocs(default_nprocs),
       isa(default_isa),
       priv(default_priv),
       mem_layout(default_mem_layout),
@@ -70,13 +69,14 @@ public:
 
   cfg_arg_t<std::pair<reg_t, reg_t>> initrd_bounds;
   cfg_arg_t<const char *>            bootargs;
-  cfg_arg_t<size_t>                  nprocs;
   cfg_arg_t<const char *>            isa;
   cfg_arg_t<const char *>            priv;
   cfg_arg_t<std::vector<mem_cfg_t>>  mem_layout;
   std::optional<reg_t>               start_pc;
   cfg_arg_t<std::vector<int>>        hartids;
   bool                               explicit_hartids;
+
+  size_t nprocs() const { return hartids().size(); }
 };
 
 #endif
--- a/riscv/cfg.h
+++ b/riscv/cfg.h
@@ -2,6 +2,7 @@
 #ifndef _RISCV_CFG_H
 #define _RISCV_CFG_H
 
+#include <optional>
 #include "decode.h"
 #include "mmu.h"
 #include <cassert>
@@ -70,6 +71,7 @@ public:
   cfg_arg_t<const char *>            isa;
   cfg_arg_t<const char *>            priv;
   cfg_arg_t<std::vector<mem_cfg_t>>  mem_layout;
+  std::optional<reg_t>               start_pc;
 };
 
 #endif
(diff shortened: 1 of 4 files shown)


spike_main/spike.cc
```diff
@@ -253,7 +252,8 @@ int main(int argc, char** argv)
             /*default_bootargs=*/nullptr,
             /*default_nprocs=*/1,
             /*default_isa=*/DEFAULT_ISA,
-            /*default_priv=*/DEFAULT_PRIV);
+            /*default_priv=*/DEFAULT_PRIV,
+            /*default_start_pc=*/reg_t(-1));
```
scottj97: It seems this default value is unused now...is there a good way to get rid of it completely?

spike_main/spike.cc
```diff
@@ -445,9 +452,28 @@ int main(int argc, char** argv)
   }
 #endif
 
+  if (cfg.explicit_hartids) {
```
scottj97, line 465: It seems that nobody outside of this code in `spike.cc` cares about `explicit_hartids`? Can this be a local variable here in `main()` instead of in `cfg_t`?

riscv/sim.cc
```diff
@@ -128,7 +118,7 @@ sim_t::sim_t(const cfg_t *cfg, const char* varch, bool halted, bool real_time_cl
         procs[cpu_idx]->set_pmp_num(pmp_num);
       } else {
         std::cerr << "core ("
-                  << hartids.size()
+                  << cfg->hartids().size()
```
scottj97: How does this work? (How did it ever work?) If I have hartids={0,1} this would report that core 2 has invalid pmpregions?

riscv/sim.cc
```diff
@@ -45,7 +44,7 @@ sim_t::sim_t(const cfg_t *cfg, const char* varch, bool halted, bool real_time_cl
     cfg(cfg),
     mems(mems),
     plugin_devices(plugin_devices),
-    procs(std::max(cfg->nprocs(), size_t(1))),
+    procs(std::max(cfg->hartids().size(), size_t(1))),
```
scottj97: Perhaps there should be a method `cfg_t::nprocs()` that returns `hartids().size()`?

riscv/cfg.h
```diff
@@ -29,23 +29,25 @@ class cfg_t
 {
 public:
   cfg_t(std::pair<reg_t, reg_t> default_initrd_bounds,
-        const char *default_bootargs, size_t default_nprocs,
```
scottj97, line 56: Commit 72c119f03a148fc3512e85547deb531004f8f6e1 is awfully large. Things like moving the nprocs/hartid sanity checking, removing nprocs in lieu of hartids list, could/should be separate commits.

scottj97 (changes requested):

spike_main/spike.cc
rswarbrick: This is analogous to the `overridden()` method on `cfg_arg_t<T>`. Once we are merging with a parsed DTB file, I think we'll want to know whether the IDs have been explicitly or whether they are just inferred from the number of processors.

riscv/sim.cc
rswarbrick: I just replaced it blindly, but you're completely right! I'll push a commit to come before that sets this to something sensible.

riscv/cfg.h
rswarbrick: Hmm, it's only +59/-41 (and the version I'm about to push up is +57/-37).

I'm not entirely sure how to split things up: as long as we're passing nprocs and hartids separately to `sim.cc`, I think the sanity check belongs there. But once we've merged the two fields, we (obviously) have to do the sanity check where we do the merge (in `spike.cc`). So I think that part of the commit is reasonably atomic. We *could* split up the "merge nprocs and hartids" and "move the result into cfg_t" change, but I'm not sure how much that buys: the second commit will just be a couple of lines.

Have I missed something?

spike_main/spike.cc
rswarbrick: Good point, I'll switch this across to using an option type.

riscv/sim.cc
rswarbrick: Good idea. Done.

spike_main/spike.cc
scottj97: Yes...it's really unfortunate that this is a special case though. :(

riscv/cfg.h
scottj97: I don't think each commit needs to make complete sense in isolation. Moving the sanity check first might not make sense in isolation, but if it splits up a large commit, it can help with pinpointing future-discovered regressions. Even commits that are only a couple of lines are useful to separate.

spike_main/spike.cc
rswarbrick: Yeah, I guess the problem is that there are two different "knobs" that can be driven on the command line argument which both affect the same thing. I suspect there might be a few more similar special cases popping up before this has all settled down.

riscv/cfg.h
rswarbrick: I wasn't trying hard enough :-) I've just done a bit of surgery and have unpicked things: that commit is now in 4 parts!

scottj97 (changes requested):
I'm getting a lot of compile errors as of ea11e80b448018be4e95cb0c49b6254b5b835b31, starting with:
```
In file included from /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/sim.h:14,
                 from /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/interactive.cc:3:
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/cfg.h:36:20: error: ‘vector’ in namespace ‘std’ does not name a template type
         const std::vector<int> default_hartids,
                    ^~~~~~
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/cfg.h:36:15: note: ‘std::vector’ is defined in header ‘<vector>’; did you forget to ‘#include <vector>’?
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/cfg.h:7:1:                                                                                   
+#include <vector>
```

rswarbrick:
Huh! Looks like I forgot a header but apparently my toolchain has different transitive includes. Will fix tomorrow.

scottj97:
Please rebase on the latest master

rswarbrick:
Done, thanks.

scottj97 (approved):

PR #957, opened 2022-03-25
Split mem layout computation in spike.cc

rswarbrick:
*Split out from #938*

The motivation here is mostly to enable a refactoring where the memory
layout (sans allocated memory) gets passed to DTS/DTB code before we
ever allocate anything.

But it turns out to make merge_overlapping_memory_regions a bit
simpler, which is an added bonus.

### DIFF
--- a/riscv/cfg.h
+++ b/riscv/cfg.h
@@ -3,6 +3,8 @@
 #define _RISCV_CFG_H
 
 #include "decode.h"
+#include "mmu.h"
+#include <cassert>
 
 template <typename T>
 class cfg_arg_t {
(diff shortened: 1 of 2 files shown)


riscv/cfg.h
```diff
@@ -25,24 +26,46 @@ class cfg_arg_t {
...
+public:
+  mem_cfg_t(reg_t base, reg_t size)
+    : base(base), size(size)
+  {
+    // The truth of these assertions should be ensured by whatever is creating
+    // the regions in the first place, but we have them here to make sure that
+    // we can't end up describing memory regions that don't make sense. They
+    // ask that size is a multiple of 32 bits, that it is positive and that the
+    // top address is still representable in a reg_t.
+    assert((size % 4 == 0) && (base + size > base));
```
aswaterman: We actually need to enforce 4 KiB size and 4 KiB alignment because of the implementation of `mmu_t`.  I agree with your comment that other code should be responsible for that sort of input validation, but if we're going to have an assertion here, we should assert the same thing.

rswarbrick: You're completely right: I have no idea where that 4 came from! (Maybe I mis-read the code in `mem_t::mem_t`?) Fixing now.

riscv/cfg.h
```diff
@@ -25,24 +27,46 @@ class cfg_arg_t {
...
+public:
+  mem_cfg_t(reg_t base, reg_t size)
+    : base(base), size(size)
+  {
+    // The truth of these assertions should be ensured by whatever is creating
+    // the regions in the first place, but we have them here to make sure that
+    // we can't end up describing memory regions that don't make sense. They
+    // ask that size is a multiple of the minimum page size, that it is
+    // positive and that the top address is still representable in a reg_t.
+    assert((size % PGSIZE == 0) && (base + size > base));
```
aswaterman: Can we also add a `(base % PGSIZE == 0)` here?

rswarbrick: Good point! Done.

riscv/cfg.h
```diff
@@ -25,24 +27,49 @@ class cfg_arg_t {
...
+    // ask that the page size is a multiple of the minimum page size, that the
+    // page is aligned to the minimum page size, that the page is non-empty and
+    // that the top address is still representable in a reg_t.
+    assert((size % PGSIZE == 0) &&
+           (base % PGSIZE == 0) &&
+           (base + size > base));
+  }
+
+  reg_t base;
+  reg_t size;
```
scottj97, line 49: Can these be `const`?

scottj97 (approved):
This is quite large for a single commit, but I'll get off my horse for a minute.

riscv/cfg.h
rswarbrick: Yep, they can be: I'll do that now.

rswarbrick: Oh! Actually, that doesn't work. The problem is that then you can't pass it to `std::sort`, which expects an `operator=` (and, of course, you can't implement `operator=` if your fields are const)[1].

Is this actually needed?

[1] The C++ "chapter and verse" seems to be that the type must be "ValueSwappable", but this boils down to roughly the same thing.

scottj97: ok nbd

aswaterman (approved):

Issue #956, opened 2022-03-25
Support for sv57 paging mode


Alenkruth:
Hello,

I wanted to know if support for sv57 paging mode is in the works and when will it be available?

Thanks,
Alenkruth

aswaterman:
I've made a prototype PR (which I haven't been able to meaningfully test yet).  https://github.com/riscv-software-src/riscv-isa-sim/pull/958

scottj97:
@Alenkruth do you have a good way to test #958?

Alenkruth:
@scottj97 I do have a way to generate tests with sv57 paging. I will be testing #958 using the tests I am able to generate.

@aswaterman I tried running one test with the sv57 paging. I find the SATP.mode field is not updated with 0xA.
I am adding a snippet from spike's log. Additionally, I am also zipping and attaching the ELF as well as the dump from SPIKE.

> core   0: 3 0x00000000800000e4 (0x01e2b023) mem 0x0000000080008000 0x0000000020002401
> core   0: 0x00000000800000e8 (0x00a00293) li      t0, 10
> core   0: 3 0x00000000800000e8 (0x00a00293) x 5 0x000000000000000a
> core   0: 0x00000000800000ec (0x03c29313) slli    t1, t0, 60
> core   0: 3 0x00000000800000ec (0x03c29313) x 6 0xa000000000000000
> core   0: 0x00000000800000f0 (0x18032073) csrs    satp, t1
> core   0: 3 0x00000000800000f0 (0x18032073) c384_satp 0x0000000000000000
> core   0: 0x00000000800000f4 (0x00003e17) auipc   t3, 0x3
> core   0: 3 0x00000000800000f4 (0x00003e17) x28 0x00000000800030f4
> core   0: 0x00000000800000f8 (0xf0ce0e13) addi    t3, t3, -244
> core   0: 3 0x00000000800000f8 (0xf0ce0e13) x28 0x0000000080003000
> core   0: 0x00000000800000fc (0x00ce5e93) srli    t4, t3, 12
> core   0: 3 0x00000000800000fc (0x00ce5e93) x29 0x0000000000080003
> core   0: 0x0000000080000100 (0x01d00f33) add     t5, zero, t4
> core   0: 3 0x0000000080000100 (0x01d00f33) x30 0x0000000000080003
> core   0: 0x0000000080000104 (0x180f2073) csrs    satp, t5
[...]

PR #958, opened 2022-03-28
Implement Sv57 and Sv57x4 translation modes

aswaterman:
To do so cleanly, I also did some light refactoring (in the first two commits).

Unfortunately, I don't have a good way of testing this, beyond testing I didn't break Sv39/Sv48.


scottj97:
I don't see any issues here, but I don't think we should merge without at least some sanity testing.

aswaterman:
@scottj97 I agree.  I'm going to take that as approval to push the non-Sv57 commits, then we'll leave this open until someone demonstrates sanity.

scottj97 (approved):

scottj97:
As discussed in #956 this has been sanity tested.

Issue #857, opened 2021-11-11
Vector vmv1r.v erroneously ignores SEW


scottj97:
The spec for vmv1r.v says:

> The instructions operate as if EEW=SEW, EMUL = NREG, effective length evl= EMUL * VLEN/SEW.

But the [vmv1r.v source](https://github.com/riscv-software-src/riscv-isa-sim/blob/f1bc56264ed214ee941b3a601a693ab655724f52/riscv/insns/vmvnfr_v.h) does not seem to take SEW into account at all.

When vstart=1, it should start the copy from the second element according to SEW (right?), but seems to always start from the second byte, as if SEW=8.

cc @chihminchao

aswaterman:
I agree that the whole-register moves should report vstart in terms of SEW-sized elements, not bytes.  I think the reason I haven't observed this is that the implementation I had been working on doesn't support interruption/restart of arithmetic, only of loads & stores.

scottj97:
Fixed by #959

PR #959, opened 2022-03-30
Fix start byte for vmv<nf>r.v

liweiwei90:
As spec for vmv\<nf\>r,v said:

'The instructions operate as if EEW=SEW.EMUL = NREG, effective length evl= EMUL * VLEN/SEW.'

So the start byte should take sew into acount when vstart != 0.

### DIFF
--- a/riscv/insns/vmvnfr_v.h
+++ b/riscv/insns/vmvnfr_v.h
@@ -7,11 +7,12 @@ const reg_t len = insn.rs1() + 1;
 require_align(vd, len);
 require_align(vs2, len);
 const reg_t size = len * P.VU.vlenb;
+const reg_t start = P.VU.vstart->read() * (P.VU.vsew >> 3);
 
 //register needs one-by-one copy to keep commitlog correct
-if (vd != vs2 && P.VU.vstart->read() < size) {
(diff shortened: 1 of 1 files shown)


scottj97:
Does this fix #857?

liweiwei90:
Oh, Yes. Sorry, I didn't observe this issue before. I just find this problem when I try to review a patchset for QEMU.

liweiwei90:
So. Should I close this PR?

scottj97:
> So. Should I close this PR?

No! I didn't have any fix. I just reported the problem in #857.

aswaterman:
This looks correct to me.

I think I've spotted a second bug in this instruction: it appears we are not raising an illegal-instruction exception for non-power-of-2 `len`.  Since by construction we know that `len >= 1`, it should suffice to add a `require((len & (len - 1)) == 0);`.

liweiwei90:
The code in vmvnfr_v.h is the common function for vmv1r.v, vmv2r.v, vmv4r.v and vmv8r.v instructions. I think the encoding for these instructions have ensured  the 'len' to be power-of-2.

aswaterman:
Yeah, you're right.  I thought for a moment this was the direct implementation.

PR #960, opened 2022-03-30
Change misa CSR in associate with priv spec.

marcfedorow:
These patches implement appropriate misa manipulations for `Q` and `V` extensions.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -499,14 +499,17 @@ misa_csr_t::misa_csr_t(processor_t* const proc, const reg_t addr, const reg_t ma
              ) {
 }
 
+const reg_t misa_csr_t::dependency(const reg_t val, const char feature, const char depends_on) const noexcept {
+  return (val & (1L << (depends_on - 'A'))) ? val : (val & ~(1L << (feature - 'A')));
+}
+
 bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   // the write is ignored if increasing IALIGN would misalign the PC
   if (!(val & (1L << ('C' - 'A'))) && (state->pc & 2))
     return false;
 
-  const bool val_supports_f = val & (1L << ('F' - 'A'));
-  const reg_t val_without_d = val & ~(1L << ('D' - 'A'));
-  const reg_t adjusted_val = val_supports_f ? val : val_without_d;
+  reg_t adjusted_val = val;
+  adjusted_val = dependency(adjusted_val, 'D', 'F');
 
   const reg_t old_misa = read();
   const bool prev_h = old_misa & (1L << ('H' - 'A'));
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -488,7 +488,7 @@ bool sstatus_csr_t::enabled(const reg_t which) {
 misa_csr_t::misa_csr_t(processor_t* const proc, const reg_t addr, const reg_t max_isa):
   basic_csr_t(proc, addr, max_isa),
   max_isa(max_isa),
-  write_mask(max_isa & (0  // allow MAFDQCH bits in MISA to be modified
+  write_mask(max_isa & (0  // allow MAFDQCHV bits in MISA to be modified
                         | (1L << ('M' - 'A'))
                         | (1L << ('A' - 'A'))
                         | (1L << ('F' - 'A'))
@@ -496,6 +496,7 @@ misa_csr_t::misa_csr_t(processor_t* const proc, const reg_t addr, const reg_t ma
                         | (1L << ('Q' - 'A'))
                         | (1L << ('C' - 'A'))
                         | (1L << ('H' - 'A'))
+                        | (1L << ('V' - 'A'))
                         )
              ) {
 }
@@ -512,6 +513,7 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   reg_t adjusted_val = val;
   adjusted_val = dependency(adjusted_val, 'D', 'F');
   adjusted_val = dependency(adjusted_val, 'Q', 'D');
+  adjusted_val = dependency(adjusted_val, 'V', 'D');
 
   const reg_t old_misa = read();
   const bool prev_h = old_misa & (1L << ('H' - 'A'));
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -320,8 +320,8 @@ reg_t base_status_csr_t::compute_sstatus_write_mask() const noexcept {
   // If a configuration has FS bits, they will always be accessible no
   // matter the state of misa.
   const bool has_fs = proc->extension_enabled('S') || proc->extension_enabled('F')
-              || proc->extension_enabled_const('V');
-  const bool has_vs = proc->extension_enabled_const('V');
+              || proc->extension_enabled('V');
+  const bool has_vs = proc->extension_enabled('V');
   return 0
     | (proc->extension_enabled('S') ? (SSTATUS_SIE | SSTATUS_SPIE | SSTATUS_SPP) : 0)
     | (has_page ? (SSTATUS_SUM | SSTATUS_MXR) : 0)
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -488,11 +488,12 @@ bool sstatus_csr_t::enabled(const reg_t which) {
 misa_csr_t::misa_csr_t(processor_t* const proc, const reg_t addr, const reg_t max_isa):
   basic_csr_t(proc, addr, max_isa),
   max_isa(max_isa),
-  write_mask(max_isa & (0  // allow MAFDCH bits in MISA to be modified
+  write_mask(max_isa & (0  // allow MAFDQCH bits in MISA to be modified
                         | (1L << ('M' - 'A'))
                         | (1L << ('A' - 'A'))
                         | (1L << ('F' - 'A'))
                         | (1L << ('D' - 'A'))
+                        | (1L << ('Q' - 'A'))
                         | (1L << ('C' - 'A'))
                         | (1L << ('H' - 'A'))
                         )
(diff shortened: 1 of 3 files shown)


marcfedorow:
riscv-privileged spec says:
```
3.1.1. Machine ISA Register misa
<...>
If an ISA feature x depends on an ISA feature y, then attempting to enable feature x but disable
feature y results in both features being disabled. For example, setting “F”=0 and “D”=1 results
in both “F” and “D” being cleared.
```

marcfedorow:
@scottj97, I ask you to check this if you don't mind.

NOTE: this logic actually ignores `zv*`. `zv*` is not implemented in spike.

riscv/csrs.cc
```diff
@@ -493,7 +493,10 @@ misa_csr_t::misa_csr_t(processor_t* const proc, const reg_t addr, const reg_t ma
                         | (1L << ('A' - 'A'))
                         | (1L << ('F' - 'A'))
                         | (1L << ('D' - 'A'))
+                        | (1L << ('Q' - 'A'))
```
scottj97, line 496: The comment above is incorrect now

scottj97, line 496: Why is Q writable now? Is Quad FP implemented?

riscv/csrs.cc
```diff
@@ -504,9 +507,14 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
...
     return false;
 
-  const bool val_supports_f = val & (1L << ('F' - 'A'));
-  const reg_t val_without_d = val & ~(1L << ('D' - 'A'));
-  const reg_t adjusted_val = val_supports_f ? val : val_without_d;
+  reg_t adjusted_val = val;
+  if (!(adjusted_val & (1L << ('F' - 'A')))) adjusted_val &= ~(1L << ('D' - 'A'));
+  if (!(adjusted_val & (1L << ('D' - 'A')))) adjusted_val &= ~((1L << ('V' - 'A')) | (1L << ('Q' - 'A')));
+  if (adjusted_val & ((1L << ('V' - 'A')) | (1L << ('Q' - 'A')))) adjusted_val |= (1L << ('D' - 'A'));
+  if (adjusted_val & (1L << ('D' - 'A'))) adjusted_val |= (1L << ('F' - 'A'));
```
scottj97: This logic doesn't make sense. If D=0, line 512 clears V & Q, so line 513 will never be true. Same for line 514 and 517.

marcfedorow:
WIP

scottj97 (changes requested):

riscv/csrs.cc
scottj97: Yes I see that Quad FP is implemented. Enabling each bit in misa should be a separate commit.

riscv/csrs.cc
marcfedorow: Sorry for tagging.
I already realized my stupidity

marcfedorow: I re-read spec and realised that there is no logic of re-enabling implied extensions.

riscv/csrs.cc
marcfedorow: Done for Q and V via separate commits

marcfedorow: Fixed

riscv/csrs.cc
```diff
@@ -504,9 +505,9 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   if (!(val & (1L << ('C' - 'A'))) && (state->pc & 2))
     return false;
 
-  const bool val_supports_f = val & (1L << ('F' - 'A'));
-  const reg_t val_without_d = val & ~(1L << ('D' - 'A'));
-  const reg_t adjusted_val = val_supports_f ? val : val_without_d;
+  reg_t adjusted_val = val;
+  adjusted_val = (val & (1L << ('F' - 'A'))) ? adjusted_val : (val & ~(1L << ('D' - 'A')));
+  adjusted_val = (val & (1L << ('D' - 'A'))) ? adjusted_val : (val & ~(1L << ('Q' - 'A')));
```
scottj97: The key thing here is a dependency between F/D and Q/D. Can we find a good way to express this succinctly? For example the code could read something like:
```
adjusted_val = dependency(adjusted_val, 'D', 'F');
adjusted_val = dependency(adjusted_val, 'Q', 'D');
```
And if we could make it even more functional by removing the mutable variable that would be even better.

Then, make this change first as a refactor of the existing F/D bits, in its own commit, before adding the new functionality in later commits.

riscv/isa_parser.cc
```diff
@@ -81,6 +81,7 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
                 extension_table[EXT_ZPN] = true;
                 extension_table[EXT_ZPSFOPERAND] = true;
                 extension_table[EXT_ZMMUL] = true; break;
+      case 'v': // even rv32iv implies double float
```
scottj97, line 84: This should be its own commit

scottj97 (approved):
I think the commit messages are a bit misleading. The priv spec does not require any particular bits to be writable in `misa`. That's up to the implementation (or platform spec). 

> The Extensions field is a WARL field that can contain writable bits where the implementation allows the supported ISA to be modified.

What is changing here is to make certain bits writable.

What are these "Change-Id:" in your commit messages?

scottj97 (changes requested):
Mistakenly did "Approve" in my previous review submit.

marcfedorow:
@scottj97 
Here is what I think should be functionally done.
I can create some macro to make lines clearer (where should I put it?)
but I need some help to figure out how to write it reasonably on immutable variables
since one bit can depend on multiple bits (e.g. V depends on F, D)
and one bit can affect multiple bits (e.g. F affects D, V)
so variable names (even if generated by macro) could be complex or unclear
with new dependencies such as P depends on M.

I can calculate masks for each dependency first (as it is originally done)
and then apply all of them at once, but it is probably less efficient:
e.g. creation of misa.V mask requires to check both F and D
(or it is necessary to create V(F) and V(D) masks).

riscv/isa_parser.cc
marcfedorow: Done

scottj97:
> @scottj97 Here is what I think should be functionally done. I can create some macro to make lines clearer (where should I put it?) but I need some help to figure out how to write it reasonably on immutable variables since one bit can depend on multiple bits (e.g. V depends on F, D) and one bit can affect multiple bits (e.g. F affects D, V) so variable names (even if generated by macro) could be complex or unclear with new dependencies such as P depends on M.
> 
> I can calculate masks for each dependency first (as it is originally done) and then apply all of them at once, but it is probably less efficient: e.g. creation of misa.V mask requires to check both F and D (or it is necessary to create V(F) and V(D) masks).

Please no macros. C++ has much better ways of doing things.

The code here (as of 937baa91f0d47ef25ce37b4b29c5907d1d15ea5b) is:
```
  adjusted_val = (adjusted_val & (1L << ('F' - 'A'))) ? adjusted_val : (adjusted_val & ~(1L << ('D' - 'A')));
  adjusted_val = (adjusted_val & (1L << ('D' - 'A'))) ? adjusted_val : (adjusted_val & ~(1L << ('Q' - 'A')));
  adjusted_val = (adjusted_val & (1L << ('D' - 'A'))) ? adjusted_val : (adjusted_val & ~(1L << ('V' - 'A')));
```

I don't see any difficulty using a new method:
```
reg_t misa_csr_t::dependency(const reg_t val, const char feature, const char depends_on);
```
Which would enable the code I suggested yesterday.

Making this more functional (instead of repeatedly modifying `adjusted_val`) would be preferable but I don't see an obvious way to do that cleanly.

riscv/csrs.cc
```diff
@@ -488,13 +488,15 @@ bool sstatus_csr_t::enabled(const reg_t which) {
...
-  write_mask(max_isa & (0  // allow MAFDCH bits in MISA to be modified
+  write_mask(max_isa & (0  // allow MAFDQCHV bits in MISA to be modified
                         | (1L << ('M' - 'A'))
                         | (1L << ('A' - 'A'))
                         | (1L << ('F' - 'A'))
                         | (1L << ('D' - 'A'))
+                        | (1L << ('Q' - 'A'))
                         | (1L << ('C' - 'A'))
                         | (1L << ('H' - 'A'))
+                        | (1L << ('V' - 'A'))
```
scottj97, line 499: This is going to break [this code](https://github.com/riscv-software-src/riscv-isa-sim/blob/e6a2245bbd99482f32bdc736267ab97bbff9713f/riscv/csrs.cc#L323-L324) which assumes a constant `misa.V`. That code will need to be refactored first before we can enable writable `misa.V`.

scottj97 (changes requested):

riscv/csrs.cc
scottj97: It appears that code is only called during initialization, when `misa` will have the maximum features enabled, so simply removing the `_const` from the function call should suffice.

riscv/csrs.cc
```diff
@@ -320,8 +320,8 @@ reg_t base_status_csr_t::compute_sstatus_write_mask() const noexcept {
   // If a configuration has FS bits, they will always be accessible no
   // matter the state of misa.
   const bool has_fs = proc->extension_enabled('S') || proc->extension_enabled('F')
-              || proc->extension_enabled_const('V');
-  const bool has_vs = proc->extension_enabled_const('V');
+              || proc->extension_enabled('V');
+  const bool has_vs = proc->extension_enabled('V');
```
scottj97, line 324: Can you please split this change into its own commit? The commit message should explain that ever since ca08503 this code only runs at reset, so no longer depends on `misa.V` being constant.

scottj97 (approved):

riscv/csrs.cc
scottj97: At least since ca0850358e08397dcf712a42bc6631d283646e2b this has not depended on `misa.V` being constant.

scottj97 (approved):
Excellent, thank you for your patience.

aswaterman:
LGTM, too.

PR #964, opened 2022-04-06
make device_list.tick()  run when tohost is zero

liweiwei90:
My colleague stuck at step 15 when running xvisor on spike following the [riscv64-spike](https://github.com/xvisor/xvisor/blob/master/docs/riscv/riscv64-spike.txt) document. I help him debug the reason. Since it can run successfully on my local spike(equal to recently commit https://github.com/riscv-software-src/riscv-isa-sim/commit/38f085d2fc96fb7fb4c4a60834617446e424f23d), I tracked the commit history, and found this is introduced by commit https://github.com/riscv-software-src/riscv-isa-sim/commit/1fea2afbf46d2641d77f2db3d6108e0897431a84.
This fix is to restore the only changes  I found that may be caused by this commit to original logic . And it can make xvisor running successfully. However I don't quite understand why.

### DIFF
--- a/fesvr/htif.cc
+++ b/fesvr/htif.cc
@@ -245,7 +245,6 @@ int htif_t::run()
       try {
         command_t cmd(mem, tohost, fromhost_callback);
         device_list.handle_command(cmd);
-        device_list.tick();
       } catch (mem_trap_t& t) {
         std::stringstream tohost_hex;
         tohost_hex << std::hex << tohost;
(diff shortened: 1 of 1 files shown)


aswaterman:
Agreed, but we don't want to move `tick` outside of the `try` block.  Please verify that my alternate fix (1767a27ad4cf8f9eb1a30d9d9a2495477071339f) suffices for your purposes.

liweiwei90:
> Agreed, but we don't want to move `tick` outside of the `try` block. Please verify that my alternate fix ([1767a27](https://github.com/riscv-software-src/riscv-isa-sim/commit/1767a27ad4cf8f9eb1a30d9d9a2495477071339f)) suffices for your purposes.

Yeah. It's OK.

PR #968, opened 2022-04-09
[NFC] Clean up the redundant code by defining macros

4vtomat:
This commit reduce the code size(almost 10%) by defining new macros. By doing so, the code would be "more concise" and "more readable".

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -896,113 +896,34 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
   }
 
 // comparision result to masking register
+#define VI_LOOP_CMP_BODY(PARAMS, BODY) \
+  VI_LOOP_CMP_BASE \
+  INSNS_BASE(PARAMS, BODY) \
+  VI_LOOP_CMP_END
+
 #define VI_VV_LOOP_CMP(BODY) \
   VI_CHECK_MSS(true); \
-  VI_LOOP_CMP_BASE \
-  if (sew == e8){ \
-    VV_CMP_PARAMS(e8); \
-    BODY; \
-  }else if(sew == e16){ \
-    VV_CMP_PARAMS(e16); \
-    BODY; \
-  }else if(sew == e32){ \
-    VV_CMP_PARAMS(e32); \
-    BODY; \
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
I don't know what [NFC] means.

This PR is a single commit that is far too large to be reviewed.  Each individual refactoring should be separated into its own commit.  You have also made unrelated changes (deleting whitespace) that should be separated out into other commits.

How have you tested these changes, and why are you confident they are correct?

4vtomat:
> I don't know what [NFC] means.

This means "No Functional Change" lol~

> This PR is a single commit that is far too large to be reviewed. Each individual refactoring should be separated into its own commit. You have also made unrelated changes (deleting whitespace) that should be separated out into other commits.

Yes, sorry about that, I would divide the change into multiple much readable commits to be able to reason about.
 
> How have you tested these changes, and why are you confident they are correct?

Actually after I split up the change into multiple commits, I would use `clang++ -E` and check whether after macro expanding, the code remains exactly the same as the code before the change.

aswaterman:
> This means "No Functional Change" lol~

Ah, good acronym :)

> Yes, sorry about that, I would divide the change into multiple much readable commits to be able to reason about.

Cool, I'll review afterwards.

> check whether after macro expanding, the code remains exactly the same

Excellent strategy!

4vtomat:
I use the code below and the command: `clang++ -E <file>` to make sure after macro expansion, the code remains the same.
```
#include "riscv/decode.h"

int main() {
#include "riscv/insns/vmseq_vi.h"
#include "riscv/insns/vmseq_vv.h"
#include "riscv/insns/vmseq_vx.h"
#include "riscv/insns/vmsleu_vv.h"
#include "riscv/insns/vmsltu_vv.h"
#include "riscv/insns/vmsgtu_vx.h"
#include "riscv/insns/vmsleu_vx.h"
#include "riscv/insns/vmsltu_vx.h"
#include "riscv/insns/vmsgtu_vi.h"
#include "riscv/insns/vmsleu_vi.h"
}
```

aswaterman (approved):

PR #969, opened 2022-04-10
Adjust the access index of vs2 to zero in vmv_x_s.h

4vtomat:
According to [scalar move](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#161-integer-scalar-move-instructions), I guess the index of vs2 should always be 0 no matter what value is in rs1?

### DIFF
--- a/riscv/insns/vmv_x_s.h
+++ b/riscv/insns/vmv_x_s.h
@@ -1,4 +1,4 @@
-// vmv_x_s: rd = vs2[rs1]
+// vmv_x_s: rd = vs2[0]
 require_vector(true);
 require(insn.v_vm() == 1);
 uint64_t xmask = UINT64_MAX >> (64 - P.get_isa().get_max_xlen());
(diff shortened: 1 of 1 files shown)


aswaterman:
Fortunately, there's no bug here, because the illegal-instruction detection logic prevents rs1 from being nonzero.  Still, we should fix the code.

This patch appears to be incomplete, though.  Why is there any dependence on the value of rs1? Shouldn't we just delete the following? https://github.com/riscv-software-src/riscv-isa-sim/blob/b468631969d0bd3be85037021d7d7dba52e353fc/riscv/insns/vmv_x_s.h#L9-L11

4vtomat:
> Fortunately, there's no bug here, because the illegal-instruction detection logic prevents rs1 from being nonzero. Still, we should fix the code.
> 
> This patch appears to be incomplete, though. Why is there any dependence on the value of rs1? Shouldn't we just delete the following?
> 
> https://github.com/riscv-software-src/riscv-isa-sim/blob/b468631969d0bd3be85037021d7d7dba52e353fc/riscv/insns/vmv_x_s.h#L9-L11

Oh, I see. I'm not familiar with the hardware behavior lol, my information is all come from from the rvv spec.
Thanks for telling me, these lines should be deleted.

aswaterman:
Thanks for the quick fix!

PR #972, opened 2022-04-11
Fix hgatp CSR write

scottj97:
@avpatel's fix of #928, which broke the hgatp PPN field.

I'm embarrassed that I didn't catch this before approving #928.


avpatel:
@scottj97 Thanks for sending this as separate PR

aswaterman:
Thanks for accelerating this.

PR #973, opened 2022-04-11
Change processor_t to hold a pointer to an isa_parser_t

rswarbrick:
*This is split out of the long series of commits at #938*

Before, it had another copy, which is a little unnecessary.


aswaterman (approved):
OK by me.

scottj97 (approved):

PR #974, opened 2022-04-12
Teach cfg_t to "freeze" and add isa parsing to it

rswarbrick:
*Split out from #938*

The idea here is that we want to compute some "derived" configuration
in `cfg_t`, which can depend on more than one of the config arguments.
Specifically, we've got an `isa_parser_t` that we want to construct
based on the "isa" and "priv" strings.

One way to do that would be to parse the strings on the fly in some
sort of getter function. A problem there is that we'd end up parsing
them multiple times (on each call to the getter).

To avoid that re-computation, we could add a local cache. But that's a
bit awkward too: we'd like to pass a "`const cfg_t *`" around, but then
we'd have to mess around with `const_cast`s to allow us to mutate the
cache. What's more, it would probably be better to do the parsing
at a known time so that we can generate any error messages at a time
we expect.

The "proper RAII way" to do this is probably to define some expanded
version of `cfg_t`, whose constructor does all the parsing. Then we'd
pass that expanded configuration around afterwards. Here, we do
something slightly more lightweight and just expect users not to alter
any of the (public) configuration arguments once they've called
`freeze()`.

### DIFF
--- a/riscv/cfg.h
+++ b/riscv/cfg.h
@@ -5,6 +5,7 @@
 #include "decode.h"
 #include "mmu.h"
 #include <cassert>
+#include <memory>
 
 template <typename T>
 class cfg_arg_t {
@@ -13,6 +14,7 @@ public:
     : value(default_val), was_set(false) {}
 
   bool overridden() const { return was_set; }
+  void set_overridden() { was_set = true; }
 
   T operator()() const { return value; }
 
@@ -26,6 +28,8 @@ private:
   T value;
   bool was_set;
 };
+
+class isa_parser_t;
 
 // Configuration that describes a memory region
 class mem_cfg_t
@@ -61,8 +65,11 @@ public:
       nprocs(default_nprocs),
       isa(default_isa),
       priv(default_priv),
-      mem_layout(default_mem_layout)
+      mem_layout(default_mem_layout),
+      frozen(false)
   {}
+
+  ~cfg_t();
 
   cfg_arg_t<std::pair<reg_t, reg_t>> initrd_bounds;
   cfg_arg_t<const char *>            bootargs;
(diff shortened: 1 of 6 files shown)


riscv/cfg.h
```diff
@@ -61,15 +65,34 @@ class cfg_t
...
   cfg_arg_t<size_t>                  nprocs;
   cfg_arg_t<const char *>            isa;
   cfg_arg_t<const char *>            priv;
   cfg_arg_t<std::vector<mem_cfg_t>>  mem_layout;
+
+  // A function to fill out the private members that are derived from the
+  // cfg_arg_t<...> members above.
+  //
+  // Call this once after filling those out and then don't touch the
+  // configuration again.
```
aswaterman, line 85: "and then don't touch the configuration again" -- this seems ripe for misuse.  Would it not be better to have `freeze` return a copy, so that there's no possibility the frozen thing could change underfoot?

I'm not worried about the extra cost of the copy in this context.

rswarbrick: Hmm, maybe. I'd convinced myself that, since there was exactly one place that uses this, we could be reasonably sure things would work. But there's always scope for messing up! :-)

Alternatively, we could do what I alluded to in the commit message and make an "expanded config object" (so there's no frozen/unfrozen state to track). The awkward bit is that you either have to copy across all the fields (which adds a certain amount of boilerplate to the constructor) or you end up with things like `cfg.base_cfg.nprocs()` all over the place, which is horrible.

I'm slightly coming down on the idea of an expanded config object class (probably called `cfg_t`, and we'll call the thing we had `pre_cfg_t` or something). That can have const fields, meaning it's impossible to mess stuff up. And I think we should copy over the fields and just take the boilerplate hit :-) Does that sound reasonable to you? (I'd like to check before spending half an hour doing boring edits!)

scottj97 (changes requested):
I like the idea of freezing the config, but I don't like a `bool frozen` or another wrapper class.

Wouldn't it be sufficient to simply pass a `const cfg_t* const cfg` to `sim_t`?

For caching, the `mutable` keyword is available.

rswarbrick:
@scottj97: Passing the const object is essentially what we're doing already. The `bool frozen` field is just checking that we didn't mess up in `spike.cc` and forget to call the freeze method.

You're right that we could use `mutable` here, but I'm still not a big fan: the `isa_parser_t` constructor can fail (e.g. if the ISA string is bogus) and it feels cleaner to me to know when exactly we're calling it.

I'm happy to remove the frozen status and just leave it up to the callers. Or I'm happy to do the "frozen_cfg_t" class approach. But it looks like you and @aswaterman slightly disagree, which makes it difficult to know what's best to do :-D

Any thoughts?

aswaterman:
@scottj97's suggestion is also fine with me: it renders my suggestion moot, so you can ignore mine.

scottj97:
I think what I was picturing before all your recent work is a `cfg_t` class that could merge two instances of itself, creating a new merged `cfg_t` object. We would construct one from the cmdline, one from the device tree, then merge them.

Or perhaps there's a third one built first with all the default values? Then merge in the cmdline, then merge in the device tree. The tricky part will be dealing with conflicts, but I suspect this will simplify things versus the current direction.

Ideally, each object would be immutable after construction. Certainly, the merged object would be immutable after construction. This way you could do everything at a well-defined time in the constructor.

PR #975, opened 2022-04-13
add macro support for instructions with overlapping encodings and fix partial style problems

liweiwei90:
The first commit fix some style problems in decode.h and processor.cc, such as space before '{' and space before/after operator.
The second commit  tries to add support for overlap instructions(such as partial zce instructions overlap with c.fld{sp}/c.fsd{sp})
    * add DECLARE_OVERLAP_INSN to bind instructions with extension
    * add overlap_list.h to contain the declare of all overlapping instructions(only declare c.fld{sp}/c.fsd{sp} without their overlapping zce instructions currently)
    * make func function for overlapping instruction return NULL when the coresponding extension(s) is not supported.


scottj97:
Can you explain a little more about the motivation here? Is this a prerequisite to adding support for new Zce instructions?

riscv/processor.cc
```diff
@@ -942,16 +942,23 @@ void processor_t::register_extension(extension_t* x)
...
 {
   #define DECLARE_INSN(name, match, mask) \
-    insn_bits_t name##_match = (match), name##_mask = (mask);
+    insn_bits_t name##_match = (match), name##_mask = (mask); \
+    bool name##_supported = true;
+
   #include "encoding.h"
   #undef DECLARE_INSN
 
+  #define DECLARE_OVERLAP_INSN(name, ext) { name##_supported &= extension_enabled_const(ext); }
```
scottj97: This breaks everything, because this queries `state.misa` to check the writability of the extension bit, and `state.misa` is not initialized until `reset()`, which is called after `register_base_instructions()`.
```
Segmentation fault (core dumped)
```

Instead you should pass in and query `isa` (type `isa_parser_t`) which has been initialized at this point.

scottj97 (changes requested):

liweiwei90:
> Can you explain a little more about the motivation here? Is this a prerequisite to adding support for new Zce instructions?

Yeah. Zcmb, Zcmp(e) and Zcmt instructions overlap with c.fld{sp}/c.fsd{sp} instructions. We need a mechanism to support these overlapping instructions.

riscv/processor.cc
liweiwei90: Sorry. I didn't realize this. I'll fix it later.

riscv/decode.h
```diff
@@ -395,20 +395,20 @@ inline long double to_f(float128_t f){long double r; memcpy(&r, &f, sizeof(r));
...
+  printf("vfp(%lu) vd=%f vs1=%f vs2=%f \n", i, to_f(vd), to_f(vs1), to_f(vs2));
 #define DEBUG_RVV_FP_VF \
-  printf("vfp(%lu) vd=%f vs1=%f vs2=%f\n", i, to_f(vd), to_f(rs1), to_f(vs2));
+  printf("vfp(%lu) vd=%f vs1=%f vs2=%f \n", i, to_f(vd), to_f(rs1), to_f(vs2));
 #define DEBUG_RVV_FMA_VV \
-  printf("vfma(%lu) vd=%f vs1=%f vs2=%f vd_old=%f\n", i, to_f(vd), to_f(vs1), to_f(vs2), to_f(vd_old));
+  printf("vfma(%lu) vd=%f vs1=%f vs2=%f vd_old=%f \n", i, to_f(vd), to_f(vs1), to_f(vs2), to_f(vd_old));
 #define DEBUG_RVV_FMA_VF \
-  printf("vfma(%lu) vd=%f vs1=%f vs2=%f vd_old=%f\n", i, to_f(vd), to_f(rs1), to_f(vs2), to_f(vd_old));
+  printf("vfma(%lu) vd=%f vs1=%f vs2=%f vd_old=%f \n", i, to_f(vd), to_f(rs1), to_f(vs2), to_f(vd_old));
```
aswaterman: These aren't style changes - they're changing what gets printed!

aswaterman:
I would like to see an example of how this is used, given that the existing overlapping instructions (like `c.flw` vs. `c.ld`) already work fine without this PR.

riscv/decode.h
liweiwei90: Yeah, they were replaced when I tried to add space before '\'. I'll restore them later.

liweiwei90:
> I would like to see an example of how this is used, given that the existing overlapping instructions (like `c.flw` vs. `c.ld`) already work fine without this PR.

c.flw and c.ld have the same MATCH and MASK and can be distinguished by xlen. However the MASK and MATCH for Zt extensions is different from c.fld{sp}/c.fsd{sp}, for example:

```
#define MATCH_C_FSDSP 0xa002
#define MASK_C_FSDSP  0xe003

#define MATCH_CM_PUSH 0xb802
#define MASK_CM_PUSH  0xff03
#define MATCH_CM_POP 0xba02
#define MASK_CM_POP  0xff03
#define MATCH_CM_POPRET 0xbe02
#define MASK_CM_POPRET  0xff03
#define MATCH_CM_POPRETZ 0xbc02
#define MASK_CM_POPRETZ  0xff03
```
If we take the same way as c.flw and c.fld, we need parse the encoding in c_fsdsp.h. And we cannot distinguish them by xlen, so we also should distinguish them by supported extension.

aswaterman:
Makes sense.

scottj97 (approved):
Looks good now

aswaterman:
Agreed

PR #976, opened 2022-04-13
Fix for #873 AMO exception priorities

rbuchner-aril:
These changes are designed to fix the AMO fault priority bug described in #873. 

The solution implemented is to run through the store first before the AMO load-store sequence, but without storing any data. By doing this, we will check all the store attributes first (and throw the proper fault for them as well). This is a valid solution since the priority of the store fault will be greater than or equal to the load fault priority (since write permission can only be granted if read permissions are also granted). 

The MMIO store issues discussed in the original issue thread are not an issue because if the store permission check reaches the MMIO store, we can skip the MMIO store because we've already passed any possible store exceptions that could possibly be higher priority than any load exceptions.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -161,7 +161,7 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
   }
 }
 
-void mmu_t::store_slow_path(reg_t addr, reg_t len, const uint8_t* bytes, uint32_t xlate_flags)
+void mmu_t::store_slow_path(reg_t addr, reg_t len, const uint8_t* bytes, uint32_t xlate_flags, bool actually_store)
 {
   reg_t paddr = translate(addr, len, STORE, xlate_flags);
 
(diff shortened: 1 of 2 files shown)


scottj97 (approved):
I worked with @rbuchner-aril internally and have already reviewed this. I confirm that it fixes the misbehavior I reported in #873, and cannot see any other effects from it.

This could change when a dirty bit is written (if `RISCV_ENABLE_DIRTY`) if there's a page with inaccessible PMP. Previously the AMO would detect the PMP violation while doing its load and take the exception before setting the dirty bit. Now it will first set the dirty bit and then detect the PMP violation. I think this is an improvement.

aswaterman:
Thanks, @rbuchner-aril and @scottj97.

As an aside, this dirty-bit behavior could conceivably differ even within an implementation.  (Consider an implementation that caches the PMPs within the TLB.  If you get a TLB hit, you'll see the address doesn't have write perms due to PMP, so you could just take the access exception.  But if you get a TLB miss, you'll walk the page table, setting the D bit on the way, before you've checked the PMPs.)

PR #977, opened 2022-04-17
Adding more eviction policies to caches

denishoornaert:
Hi all!

Playing with Spike few weeks ago, I realized it was possible to get statistics about the caches. However, "only" the pseudo-random (i.e., lfsr) policy is available. Thus, I proceeded to extend the cache simulation part to include (1) more eviction policies, (2) a generic interface to enable the testing of new policies in the future, and (3) the performance counters (or cache statistics) in a dedicated class to allow for a cleaner integration of more complex statistics later.

Here is an overview of the changes:

Inspired by the `lfsr_t` class provided in `cachesim.h`, I created a virtual class called `eviction_policy_t` located in the `eviction_policies.h/cc` files. This virtual class provides three methods used by the cache:
 - `next()`: which returns the index of the cache-line to evict within a set
 - `update()`: is in charge of indicating to the eviction policy that a cache-line has been touched (useful for policies such as LRU)
 - `insert()`: is in charge of the insertion of the new cache-line and the updating of the metadata associated (useful for the implementation of Adaptive Insertion Policies such as LIP or BIP)

For now, five policies are available `lfsr` (as provided on the original `master` branch), `lru`, `fifo`, `lip`, and `bip`. The last two policies are issued from [1] and are often used in papers for comparison purposes. They can be removed if you do not consider adding them interesting :slightly_smiling_face: 

The selection of the eviction policy is possible at runtime thanks to the extension of the programs parameters. The setting of a the amount of sets, ways, and block size is now followed by the desired eviction policy. It is possible to set different policies for every caches (i.e., `--dc`, `--ic`, and `--l2`). For instance, `--dc=32:4:64:lru` sets the cache to use LRU. The usage of this parameters is in the helper and reads as follows:
```
    ...
    --ic=<S>:<W>:<B>:<P>   Instantiate a cache model with S sets,
    --dc=<S>:<W>:<B>:<P>  W ways, B-byte blocks (with S and B both powers of 2),
    --l2=<S>:<W>:<B>:<P>  and eviction policy P (e.g., lru, lfsr, fifo, lip, bip).
    ...
```

The cache statistics collected during the execution (i.e., performance counters) ave been moved from the `cache_sim_t` class to a dedicated call called `cache_sim_perf_counter_t`. The latter being aggregated into the former. While not strictly necessary right now, encapsulating the statistics in a class will enable the integration of more "complex" statistics such as *zero-reuse cache-lines* without polluting the `cache_sim_t` class.

Finally, because of the way I change the `cache_sim_t` class, I actually removed the specialized `fa_cahe_sim_t` class.
From what I understand, the latter was introduced 9 years ago with the specific purpose of optimizing execution time when simulating a fully associative cache. I do not know if this is still a concern nowadays but I am ready to re-integrate the class.

**Remark:** In addition, in an attempt to prevent the introduction of stupid bugs, I have relied on the C++ standard library for methods like `min_element` and `find`. This decision has led to the replacement of the dynamically allocated arrays such as `tags` by C++ standard library containers like `vector`.

[1] Moinuddin K. Qureshi, Aamer Jaleel, Yale N. Patt, Simon C. Steely, and Joel Emer. 2007. Adaptive insertion policies for high performance caching. In Proceedings of the 34th annual international symposium on Computer architecture (ISCA '07). Association for Computing Machinery, New York, NY, USA, 381–391. DOI:https://doi.org/10.1145/1250662.1250709


denishoornaert:
I left the "request" as a draft because I first want to understand if you are interested and if the code style fits your expectations.
I would also like to proceed to an in-depth proof-testing. Is there any set of benchmarks used to test modifications on spike?

Thanks for letting me know!

aswaterman:
Truth be told, I've never been too excited about pushing Spike further in the direction of simulating microarchitectural components.  It's already a fairly complicated piece of software in service of its main purpose as an ISA interpreter.

With that said, I don't view extending the cåche models to support other replacement policies as much of a stretch.  So I'm OK with this in principle (but I haven't yet reviewed the code).  I'd like to hear what others think.

jerryz123:
Agreed on not letting microarchitectural details leak into `riscv/` too much. However, since `cachesim` is pretty independent from the rest of the simulator implementation, this seems fine.

Perhaps `cachesim` can be moved into its own build directory (out of `riscv/`)?

scottj97:
I don't know much about performance modeling. Can an arch sim like Spike really provide useful information about caching, which is inherently microarchitectural? Spike cannot possibly perfectly match any real design's cache replacement policy.

As for the PR, thank you for splitting it into many commits. A good commit message should start with a verb, which most of these do, but "addition of X" would be better expressed as "add X". Also a PR should not have embedded merges such as [d346c92](https://github.com/riscv-software-src/riscv-isa-sim/pull/977/commits/d346c921344c33b81e163977342b05154293f5a7).

I tend to agree with @jerryz123 that the cache sim is well separated from the rest of the logic, so this probably adds little maintenance burden to anybody that doesn't care about cache modeling. (I haven't reviewed the code though.)

riscv/eviction_policies.cc
```diff
@@ -0,0 +1,90 @@
...
+{
+    size_t argmin = 0;
+    uint32_t min_value = 0xffffffff; // Abitrarily big
+    for (size_t i = 0; i < ways; i++) {
+      if (min_value > this->metadata[idx*ways+i]) {
+        min_value = this->metadata[idx*ways+i];
+        argmin = i;
+      }
+    }
+    return argmin;
```
aswaterman: I suggest using `min_element` to compute argmin.  It returns an iterator positioned at the minimum element, so then you can use iterator arithmetic to retrieve the index.

    auto begin = this->metadata + idx * ways;
    return std::min_element(begin, begin + ways) - begin;

riscv/eviction_policies.cc
```diff
@@ -0,0 +1,90 @@
...
+}
+
+void lru_t::promote(size_t idx, size_t way, uint32_t recency)
+{
+  this->metadata[idx*ways+way] = ways;
```
aswaterman: Some form of `this->metadata[idx*ways+ways]` shows up so often that you might want to define `get_metadata(idx, way)` and `set_metadata(idx, way, value)` protected members.

riscv/eviction_policies.cc
```diff
@@ -0,0 +1,90 @@
...
+void lip_t::insert(size_t idx, size_t way)
+{
+  // Inserted at LRU position
+  this->metadata[idx*ways+way] = 0;
+  // Other recencies remain unchanged
+}
+
+void bip_t::insert(size_t idx, size_t way)
+{
+  if (rand()%8 == 0) {
```
aswaterman: Does it make sense to make the LRU probability a constructor argument instead of hardcoding it?

spike_main/spike.cc
```diff
@@ -39,9 +39,9 @@ static void help(int exit_code = 1)
   fprintf(stderr, "  --varch=<name>        RISC-V Vector uArch string [default %s]\n", DEFAULT_VARCH);
   fprintf(stderr, "  --pc=<address>        Override ELF entry point\n");
   fprintf(stderr, "  --hartids=<a,b,...>   Explicitly specify hartids, default is 0,1,...\n");
-  fprintf(stderr, "  --ic=<S>:<W>:<B>      Instantiate a cache model with S sets,\n");
-  fprintf(stderr, "  --dc=<S>:<W>:<B>        W ways, and B-byte blocks (with S and\n");
-  fprintf(stderr, "  --l2=<S>:<W>:<B>        B both powers of 2).\n");
+  fprintf(stderr, "  --ic=<S>:<W>:<B>:<P>   Instantiate a cache model with S sets,\n");
+  fprintf(stderr, "  --dc=<S>:<W>:<B>:<P>  W ways, B-byte blocks (with S andB both powers of 2),\n");
+  fprintf(stderr, "  --l2=<S>:<W>:<B>:<P>  and an eviction policy P.\n");
```
aswaterman: Should we also print out the names of the available eviction policies below this line?

denishoornaert:
@aswaterman Looking at all your valid comments, I understand that you are currently looking at my oldest commits. Most of these remarks have been addressed in more recent commits.
I am used to commit at each step of the development, but I realize this might be problematic for the reviewing process...

Do you want me to squash all my commits?

scottj97:
Please do not squash them all into one. But if later commits improve/correct earlier ones, those should be squashed back so the bug/ugliness is completely removed from the history.

Think of the history as a [recipe, not a work log](https://www.bitsnbites.eu/git-history-work-log-vs-recipe/).

denishoornaert:
@scottj97 Thank you for the link! In its current state, I do not think the history matches your expectations. To ease the review process, I will rework the commits to create an adequate "recipe".

I will temporarily close the PR, work on the history, and come back to you once ready.
My apologies for the inconvenience.

riscv/cachesim.h
```diff
@@ -60,18 +60,6 @@ class cache_sim_t
...
- public:
-  fa_cache_sim_t(size_t ways, size_t linesz, const char* name);
-  fa_cache_sim_t(size_t ways, size_t linesz, const char* name, const std::string eviction_policy);
-  uint64_t* check_tag(uint64_t addr);
-  uint64_t victimize(uint64_t addr);
- private:
-  static bool cmp(uint64_t a, uint64_t b);
-  std::map<uint64_t, uint64_t> tags;
-};
-
```
aswaterman, line 78: If there isn't a compelling reason to make this change, I suggest backing it out.  My measurements suggest that for cache configuration `--ic=1:1024:32 --dc=1:1024:32`, with workload 403.gcc, there's a 2x perf hit if you use the naive linear-time algorithm.  (And the perf hit rises to 3x if I change `map` to `unordered_map`, which I guess I didn't know about at the time I wrote this.)

As an alternative, you could consider a more elaborate data structure that works well for both low-associativity and high-associativity caches, but that might be a bigger project.

aswaterman:
OK, I'll wait to review further until after you resubmit.  But I'm going out of town, so it might be a couple weeks.  Maybe someone else can step in.

th0ma5b:
> 
> I will temporarily close the PR, work on the history, and come back to you once ready. My apologies for the inconvenience.

Hi @denishoornaert,
Very useful feature. Can your PR be re-open so it eventually gets merged into master?

PR #978, opened 2022-04-21
Bug Fix for bug introduced in PR #976

rbuchner-aril:
In PR #976 we introduced a bug that would cause a store to occur on a misaligned AMO. To refresh, the strategy for the fix to bug #873 was to add a "Fake" store that would Table-walk but not actually store before the initial load in an AMO sequence. However, the case of a misaligned AMO was ignored, with the result being that the "fake" store during a misaligned AMO would actually modify the memory. This patch fixes that bug by adding a tag `require_alignment` to store_func similar to load_func. This will be set to true of the fake store, and will lead to a misaligned-fault/access-fault being thrown if the address is misaligned. This will take priority over any page-faults, the same behavior Spike exhibited prior to these changes.

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -147,7 +147,7 @@ public:
 
   // template for functions that store an aligned value to memory
   #define store_func(type, prefix, xlate_flags) \
-    void prefix##_##type(reg_t addr, type##_t val, bool actually_store=true) {   \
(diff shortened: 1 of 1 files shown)


scottj97 (approved):

aswaterman:
Cool, thanks. @scottj97 I’m OOO, so please merge if you approve.

PR #981, opened 2022-04-26
Add smepmp support to Spike

soberl:
Add smepmp support to Spike

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -256,12 +256,19 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
         if (proc->lg_pmp_granularity != PMP_SHIFT && (cfg & PMP_A) == PMP_NA4)
           cfg |= PMP_NAPOT; 
         /*
-        Adding a rule with executable privileges that either is M-mode-only or a locked Shared-Region is not possible 
-        and such pmpcfg writes are ignored, leaving pmpcfg unchanged. 
-        This restriction can be temporarily lifted e.g. during the boot process, by setting mseccfg.RLB.
-        */ 
-        if (rlb || !(mml && ((cfg & PMP_L) && ((cfg & PMP_X) || ((cfg & PMP_W) && !(cfg & PMP_R)))))) 
+         * Adding a rule with executable privileges that either is M-mode-only or a locked Shared-Region
+         * is not possible and such pmpcfg writes are ignored, leaving pmpcfg unchanged.
+         * This restriction can be temporarily lifted e.g. during the boot process, by setting mseccfg.RLB.
+         */
+        const bool cfgx = cfg & PMP_X;
+        const bool cfgw = cfg & PMP_W;
+        const bool cfgr = cfg & PMP_R;
+        if (rlb || !(mml && ((cfg & PMP_L)      // M-mode-only or a locked Shared-Region
+                && !(cfgx && cfgw && cfgr)      // RWX = 111 is allowed
+                && (cfgx || (cfgw && !cfgr))    // X=1 or RW=01 is not allowed
+        ))) {
           state->pmpaddr[i]->cfg = cfg;
+        }
       }
 
       write_success = true;
@@ -297,13 +304,8 @@ bool mseccfg_csr_t::unlogged_write(const reg_t val) noexcept {
   if (proc->n_pmp == 0)
     return false;
 
-  bool pmplock_recorded = false;
-  for (size_t i = 0; i < proc->n_pmp; i++) {
-    if (state->pmpaddr[i]->is_locked()) {
-      pmplock_recorded = true;
-      break;
-    }
-  }
+  bool pmplock_recorded = std::any_of(state->pmpaddr, state->pmpaddr + proc->n_pmp,
+          [](pmpaddr_csr_t_p & c) { return c->is_locked(); } );
 
   //When mseccfg.RLB is 0 and pmpcfg.L is 1 in any rule or entry (including disabled entries)
   if (!(pmplock_recorded && (mseccfg_val & MSECCFG_RLB)==0)) { 
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -117,8 +117,8 @@ bool pmpaddr_csr_t::unlogged_write(const reg_t val) noexcept {
   if (proc->n_pmp == 0)
     return false;
 
-  bool lock_bypass = state->mseccfg->get_rlb();  
-  bool locked = (cfg & PMP_L) && !lock_bypass;
+  const bool lock_bypass = state->mseccfg->get_rlb();
+  const bool locked = !lock_bypass && (cfg & PMP_L);
 
   if (pmpidx < proc->n_pmp && !locked && !next_locked_and_tor()) {
     this->val = val & ((reg_t(1) << (MAX_PADDR_BITS - PMP_SHIFT)) - 1);
@@ -131,9 +131,9 @@ bool pmpaddr_csr_t::unlogged_write(const reg_t val) noexcept {
 
 bool pmpaddr_csr_t::next_locked_and_tor() const noexcept {
   if (pmpidx+1 >= state->max_pmp) return false;  // this is the last entry
-  bool lock_bypass = state->mseccfg->get_rlb();
-  bool next_locked = (state->pmpaddr[pmpidx+1]->cfg & PMP_L) && !lock_bypass;
-  bool next_tor = (state->pmpaddr[pmpidx+1]->cfg & PMP_A) == PMP_TOR;
+  const bool lock_bypass = state->mseccfg->get_rlb();
+  const bool next_locked = !lock_bypass && (state->pmpaddr[pmpidx+1]->cfg & PMP_L);
+  const bool next_tor = (state->pmpaddr[pmpidx+1]->cfg & PMP_A) == PMP_TOR;
   return next_locked && next_tor;
 }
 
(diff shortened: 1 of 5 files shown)


riscv/csrs.cc
```diff
@@ -211,22 +240,85 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
...
+  bool rlb = state->mseccfg->get_rlb();
+  bool mml = state->mseccfg->get_mml();
   for (size_t i0 = (address - CSR_PMPCFG0) * 4, i = i0; i < i0 + proc->get_xlen() / 8; i++) {
     if (i < proc->n_pmp) {
-      if (!(state->pmpaddr[i]->cfg & PMP_L)) {
+      bool locked = (state->pmpaddr[i]->cfg & PMP_L);
+      bool next_locked = (i+1 < proc->n_pmp) && (state->pmpaddr[i+1]->cfg & PMP_L);
+      bool next_tor = (i+1 < proc->n_pmp) && (state->pmpaddr[i+1]->cfg & PMP_A) == PMP_TOR;
+      
+      if (rlb || (!locked && !(next_locked && next_tor))) {
```
scottj97: `pmpaddr_csr_t` has a `next_locked_and_tor()` method that should be used here, instead of duplicating that logic.

riscv/csrs.h
```diff
@@ -122,6 +122,23 @@ class pmpcfg_csr_t: public csr_t {
...
+    mseccfg_csr_t(processor_t* const proc, const reg_t addr);
+    virtual reg_t read() const noexcept override;
+    bool get_mml() const noexcept;
+    bool get_mmwp() const noexcept;
+    bool get_rlb() const noexcept; 
+  protected:
+    virtual bool unlogged_write(const reg_t val) noexcept override;
+  private:
+    reg_t mseccfg_val;
+    reg_t pmplock_recorded;
```
scottj97: This seems to be a bit field indexed by entry number? Since it's not modeling any hardware register, it's a bit fishy to implement it like one. It would be better to declare it as an array of bools.

But I'm not comfortable having this variable at all, since it's a duplicate of the Lock bits in all the `pmpcfg` CSRs. We should store only one copy of those bits so that there's no chance they ever get out of sync.

riscv/csrs.h
```diff
@@ -122,6 +122,23 @@ class pmpcfg_csr_t: public csr_t {
   virtual bool unlogged_write(const reg_t val) noexcept override;
 };
 
+class mseccfg_csr_t: public csr_t {
```
scottj97, line 129: I suspect it would be better to derive from `basic_csr_t` like many of the other CSRs here.

riscv/csrs.h
```diff
@@ -122,6 +122,23 @@ class pmpcfg_csr_t: public csr_t {
...
+    bool get_mml() const noexcept;
+    bool get_mmwp() const noexcept;
+    bool get_rlb() const noexcept; 
+  protected:
+    virtual bool unlogged_write(const reg_t val) noexcept override;
+  private:
+    reg_t mseccfg_val;
+    reg_t pmplock_recorded;
+    friend class pmpcfg_csr_t; //pmpcfg needs to access pmplock_recorded
+    //friend class pmpaddr_csr_t;
```
scottj97: Delete commented-out code

riscv/csrs.h
```diff
@@ -122,6 +122,23 @@ class pmpcfg_csr_t: public csr_t {
...
+    virtual reg_t read() const noexcept override;
+    bool get_mml() const noexcept;
+    bool get_mmwp() const noexcept;
+    bool get_rlb() const noexcept; 
+  protected:
+    virtual bool unlogged_write(const reg_t val) noexcept override;
+  private:
+    reg_t mseccfg_val;
+    reg_t pmplock_recorded;
+    friend class pmpcfg_csr_t; //pmpcfg needs to access pmplock_recorded
```
scottj97: Instead of `friend` how about a simple method: `record_lock(size_t idx, bool locked)`? But this is moot since I don't think the lock bits should be recorded in this class in the first place. (See comment on `pmplock_recorded`)

riscv/csrs.cc
```diff
@@ -211,22 +240,85 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
...
+        if (rlb || !(mml && ((cfg & PMP_L) && ((cfg & PMP_X) || ((cfg & PMP_W) && !(cfg & PMP_R)))))) 
+          state->pmpaddr[i]->cfg = cfg;
       }
+
       write_success = true;
+
+      state->mseccfg->pmplock_recorded &= ~(1ULL << i);
+      if (state->pmpaddr[i]->cfg & PMP_L) {
+        state->mseccfg->pmplock_recorded |= (1ULL << i);
+      }
```
scottj97: Elsewhere in the comments here I proposed a `record_lock()` making this moot, but for future reference: `set_field()` is a macro that can do all this in one line.

riscv/csrs.cc
```diff
@@ -186,11 +189,37 @@ bool pmpaddr_csr_t::subset_match(reg_t addr, reg_t len) const noexcept {
 
 
 bool pmpaddr_csr_t::access_ok(access_type type, reg_t mode) const noexcept {
-  return
-    (mode == PRV_M && !(cfg & PMP_L)) ||
-    (type == LOAD && (cfg & PMP_R)) ||
-    (type == STORE && (cfg & PMP_W)) ||
-    (type == FETCH && (cfg & PMP_X));
+  bool cfgx = cfg & PMP_X;
```
scottj97: These should all be `const` where possible

scottj97 (changes requested):
This is awfully large for a single commit. It should be broken down into multiple commits implementing this feature in steps. For example:

1. Add `mseccfg` CSR that does nothing useful
2. Revamp `access_ok` in prep for Smepmp (all the `cfgx` etc variables)
3. Add Smepmp functionality

And possibly even more than 3 commits, if you can find a good way to break it down.

riscv/csrs.cc
soberl: Agree, will do that.

riscv/csrs.h
soberl: From comments of basic_csr_t, it's used for "XLEN bits fully readable and writable". 
The mseccfg is not in the case. HW may implement only specified bits.
BTW, the existed pmpcfg CSRs are XLEN bits, while csr_t is selected currently.

riscv/csrs.h
soberl: Yes, maybe I can create a method at pmpcfg side to show the locked status, to use query instead of sync + friend.

riscv/csrs.h
soberl: Done. See 2nd commit.

riscv/csrs.h
scottj97: That's why you wouldn't use `basic_csr_t` directly; but a derived class can add any masking behavior needed. See the other classes that derive from `basic_csr_t`. There is also `masked_csr_t` which might be even more appropriate.

riscv/csrs.cc
```diff
@@ -211,22 +240,84 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
...
+  if (proc->n_pmp == 0)
+    return false;
+
+  bool pmplock_recorded = false;
+  for (size_t i = 0; i < proc->n_pmp; i++) {
+    if (state->pmpaddr[i]->is_locked()) {
+      pmplock_recorded = true;
+      break;
+    }
+  }
```
scottj97: Look into `std::any_of` for a much simpler way to do this.

scottj97 (changes requested):
Your second commit does three things; this should be three commits. (Which should then eventually be squashed back into a cleaner patch series.)

Can you tell us what you've done to test the new functionality? And make sure you haven't broken anything in the existing logic?

soberl:
> Your second commit does three things; this should be three commits. (Which should then eventually be squashed back into a cleaner patch series.)
> 
> Can you tell us what you've done to test the new functionality? And make sure you haven't broken anything in the existing logic?
Yes, we have created a lot of test cases for the new functionality. See https://github.com/joxie/riscv-isa-sim/tree/master/tests/mseccfg/gengen_src/outputs.  We are still considering the way to contribute this suite. While I am not sure for the test suite on existing logic.

riscv/csrs.cc
soberl: how about following? I would like to keep the is_locked() function since I don't hope to use PMP_L bit directly here.
  bool pmplock_recorded = std::any_of(state->pmpaddr, state->pmpaddr + proc->n_pmp,
          [](pmpaddr_csr_t_p & c) { return c->is_locked(); } );

riscv/csrs.h
soberl: Per my understanding, the key benefits from masked_csr_t is the mask function on unlogged_write(). While for mseccfg, the sticky bits can hardly reuse that. Anyway, I am still fine to use masked_csr_t as the class family to indicate the bits are partly/masked for a csr reg.

riscv/csrs.cc
scottj97: That looks good except the closure parameter can be const ref.

scottj97:
Again with [262419a](https://github.com/riscv-software-src/riscv-isa-sim/pull/981/commits/262419a78b8ed28fc6facadef1763398d3babb4d) you are mixing multiple things into a single commit. Please read [Writing Reviewable Code](https://secure.phabricator.com/book/phabflavor/article/writing_reviewable_code/) and keep each commit to a single thing, as small as possible.

This PR is definitely improving -- the final code is perhaps all good -- but the history needs a lot of work, as I [requested earlier](https://github.com/riscv-software-src/riscv-isa-sim/pull/981#pullrequestreview-953630049).

soberl:
> Again with [262419a](https://github.com/riscv-software-src/riscv-isa-sim/pull/981/commits/262419a78b8ed28fc6facadef1763398d3babb4d) you are mixing multiple things into a single commit. Please read [Writing Reviewable Code](https://secure.phabricator.com/book/phabflavor/article/writing_reviewable_code/) and keep each commit to a single thing, as small as possible.
> 
> This PR is definitely improving -- the final code is perhaps all good -- but the history needs a lot of work, as I [requested earlier](https://github.com/riscv-software-src/riscv-isa-sim/pull/981#pullrequestreview-953630049).

Yes, we see a lot of work to divide the PR into separated commits. Since it will be rework indeed, I just refine the coding style according the review comments firstly.
Later we will create another branch in my fork repo. from clean upstream, and will create the separated commits there. I will discuss with Joe Xie regarding whether to implement CSR access part firstly, or the pmp_ok() modification.
Thanks a lot for the review comments in such details and guide.

soberl:
https://github.com/riscv-software-src/riscv-isa-sim/pull/983 has been created. Close this one.

PR #985, opened 2022-04-30
Implement mcontrol.hit bit

timsifive:
Also clean up a few minor issues in the trigger code.

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -1,3 +1,4 @@
+#include "debug_defines.h"
 #include "processor.h"
 #include "triggers.h"
 
@@ -16,6 +17,7 @@ reg_t mcontrol_t::tdata1_read(const processor_t * const proc) const noexcept {
   v = set_field(v, MCONTROL_TYPE(xlen), 2);
   v = set_field(v, MCONTROL_DMODE(xlen), dmode);
   v = set_field(v, MCONTROL_MASKMAX(xlen), 0);
+  v = set_field(v, CSR_MCONTROL_HIT, hit);
   v = set_field(v, MCONTROL_SELECT, select);
   v = set_field(v, MCONTROL_TIMING, timing);
   v = set_field(v, MCONTROL_ACTION, action);
@@ -36,6 +38,7 @@ bool mcontrol_t::tdata1_write(processor_t * const proc, const reg_t val) noexcep
   }
   auto xlen = proc->get_xlen();
   dmode = get_field(val, MCONTROL_DMODE(xlen));
+  hit = get_field(val, CSR_MCONTROL_HIT);
   select = get_field(val, MCONTROL_SELECT);
   timing = get_field(val, MCONTROL_TIMING);
   action = (triggers::action_t) get_field(val, MCONTROL_ACTION);
@@ -131,6 +134,9 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
   }
 
   if (simple_match(xlen, value)) {
+    /* This is OK because this function is only called if the trigger was not
+     * inhibited by the previous trigger in the chain. */
+    hit = true;
     if (timing)
       return MATCH_FIRE_AFTER;
     else
@@ -165,6 +171,12 @@ match_result_t module_t::memory_access_match(action_t * const action, operation_
       continue;
     }
 
+    /* Note: We call memory_access_match for each trigger in a chain as long as
+     * the triggers are matching. This results in "temperature coding" so that
+     * `hit` is set on each of the consecutive triggers that matched, even if the
+     * entire chain did not match. This is allowed by the spec, because the final
+     * trigger in the chain will never get `hit` set unless the entire chain
+     * matches. */
     match_result_t result = triggers[i]->memory_access_match(proc, operation, address, data);
     if (result != MATCH_NONE && !triggers[i]->chain()) {
       *action = triggers[i]->action;
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -5,7 +5,7 @@ namespace triggers {
 
 mcontrol_t::mcontrol_t() :
   select(false), timing(false), chain_bit(false),
-  match(MATCH_EQUAL), m(false), h(false), s(false), u(false),
+  match(MATCH_EQUAL), m(false), s(false), u(false),
   execute_bit(false), store_bit(false), load_bit(false)
 {
 }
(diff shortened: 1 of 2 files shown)


riscv/triggers.cc
```diff
@@ -13,7 +13,7 @@ mcontrol_t::mcontrol_t() :
 reg_t mcontrol_t::tdata1_read(const processor_t * const proc) const noexcept {
   reg_t v = 0;
   auto xlen = proc->get_xlen();
-  v = set_field(v, MCONTROL_TYPE(xlen), type);
+  v = set_field(v, MCONTROL_TYPE(xlen), 2);
```
aswaterman: `MCONTROL_TYPE_MATCH` instead of `2`?

timsifive: Done. Thanks.

riscv/triggers.cc
```diff
@@ -22,7 +22,6 @@ reg_t mcontrol_t::tdata1_read(const processor_t * const proc) const noexcept {
   v = set_field(v, MCONTROL_CHAIN, chain_bit);
   v = set_field(v, MCONTROL_MATCH, match);
   v = set_field(v, MCONTROL_M, m);
-  v = set_field(v, MCONTROL_H, h);
```
scottj97, line 25: Shouldn't the `MCONTROL_H` define be removed too?

riscv/triggers.cc
```diff
@@ -37,6 +38,7 @@ bool mcontrol_t::tdata1_write(processor_t * const proc, const reg_t val) noexcep
   }
   auto xlen = proc->get_xlen();
   dmode = get_field(val, MCONTROL_DMODE(xlen));
+  hit = get_field(val, CSR_MCONTROL_HIT);
```
scottj97, line 41: Why is this `#define` of a different form than all the other fields?

riscv/triggers.cc
```diff
@@ -133,6 +134,9 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
   }
 
   if (simple_match(xlen, value)) {
+    /* This is OK because this function is only called if the trigger was not
+     * inhibited by the previous trigger in the chain. */
+    hit = true;
```
scottj97, line 139: IIUC this is implementing the "temperature coding" discussed [here](https://github.com/riscv/riscv-debug-spec/issues/550) and [here](https://github.com/riscv/riscv-debug-spec/pull/593), correct? It would be good to explicitly mention that choice in the commit log or comments.

scottj97 (changes requested):
Sadly, while I have many tests that cover chains and hit bits, they all rely on `tcontrol` which is not currently implemented here. So I haven't tested this thoroughly, but it looks correct to me.

riscv/triggers.cc
timsifive: In encoding.h? encoding.h is out of date wrt the debug spec in numerous ways, and I wasn't planning on fixing that.

I'm planning to move this code to use debug_defines.h, which is auto-generated from the same source that creates the debug spec, so is always correct.

riscv/triggers.cc
timsifive: This one comes from debug_defines.h, which is more correct and up-to-date than encoding.h. I'm planning to move all macros to the debug_defines.h ones, but before that I want https://github.com/riscv/riscv-debug-spec/pull/724 to merge over there.

riscv/triggers.cc
timsifive: That's right. I've added a comment.

scottj97 (approved):
Have you thought about how this affects CSR tracing? It would be good to emit a trace record whenever a Hit bit is set, which it seems it is not doing currently.

By "trace record" I mean something that shows up when running with `--log-commits`.

timsifive:
> Have you thought about how this affects CSR tracing? It would be good to emit a trace record whenever a Hit bit is set, which it seems it is not doing currently.

I didn't realize that even existed. Is there some example CSR implementation that you can point me to that shows me what function I should be calling?

On slightly more thought, this might not be an event that should be logged. Or at least, it's only a CSR write if the current value of tselect is the index of the trigger that was hit. Should the tdata1 write be logged if tselect indicates the trigger is currently selected, and not otherwise? Or is it better to never log this at all? I'm not sure what people use the log for.

scottj97:
> Is there some example CSR implementation that you can point me to that shows me what function I should be calling?

All of the classes in `csr.h` have logging via their `write()` method, so for most CSRs the logging happens automatically. The trigger CSRs are special because the CSR's value is stored externally in class `trigger_t` and family, and can be modified without calling the `csr_t::write()` method, which leads to CSRs changing without any logging event, which I consider a bug.

The logging should show any and all state changes. I consider the current tdata1/tdata2 logging to be broken. Indexed registers like these two should be reported as an array of states e.g. `tdata1[0]`. Instead they are reported as scalars and their value changes mysteriously when `tselect` changes.

If these registers were logged as array values I think that would answer all your questions, but since it's already broken I don't think we should try to tackle this as part of this PR.

[I use the log to compare state changes vs RTL hardware implementations, and there are several other teams that use Spike in the same way.]

timsifive:
It really sounds like what we want is a logging of the machine state. CSRs are a convenient short-hand for that, but for triggers we really want a trigger logging mechanisms, that tracks the state of the triggers. Is that right?

scottj97:
> It really sounds like what we want is a logging of the machine state. CSRs are a convenient short-hand for that, but for triggers we really want a trigger logging mechanisms, that tracks the state of the triggers. Is that right?

Yes we want machine state logging, but in this case I don't understand the distinction you're making. The trigger state consists entirely of the set of tdata1/tdata2 CSRs, does it not?

timsifive:
> Yes we want machine state logging, but in this case I don't understand the distinction you're making. The trigger state consists entirely of the set of tdata1/tdata2 CSRs, does it not?

I think of tdata1/tdata2 as windows into the actual trigger configuration (which is also how the recent refactor is structured).

There are multiple triggers, but only one tdata1/tdata2 CSR. This is handled by the tselect CSR. To access trigger N, write N to tselect, then access tdata1/tdata2.

So when hit is set in the trigger configuration, that doesn't modify a CSR, although it might have the effect that reading tdata1 now returns a different value than before (assuming tselect pointed to the trigger which hit).

If CSRs can contain an array of values (as the code was written before the refactor) then I can see how all the trigger state fits into CSRs, but that's not how I think of CSRs. Perhaps internal to spike they work that way.

scottj97:
Yes I see your point. You're right, we want to trace state, which I think of as `tdata1[0..N]` and `tdata2[0..N]`. A `csrw tdata1` will modify one of those. A Hit bit getting set will modify one of those. The trace in each case should look the same: a write to e.g. `tdata1[1]`.

But, I think all this is for a follow-on PR.

aswaterman:
> But, I think all this is for a follow-on PR.

Agreed.

aswaterman (approved):
LGTM, too.

PR #987, opened 2022-05-04
Fix the padding of register names in the log

fshaked:
Before this fix, register names with a single digit are printed with a space between the prefix and the digit (e.g. x5 is printed "x 5"). This fix moves the space after the digit (so x5 is now printed "x5 ").


scottj97 (approved):
Thank you for the contribution!

scottj97 (changes requested):
Can you please remove the "Change-Id:" line from the commit message? That doesn't really have any meaning in this repo.

aswaterman (approved):

aswaterman:
Merging, since Scott's concern was addressed.

PR #991, opened 2022-05-08
Disallow write to xs and add xs field checking for rocc

Phantom1003:
In the specification, both of the xs fields should be read-only, I summarized the permissions of xs in spike:
|  | spike | rocket |
|:-:|:-:|:-:|
| [m/s]status | RW | R |
| vsstatus | RW | R |

The following test case shows that both of them are writable.
```
core   0: 0x0000000080000188 (0x30001073) csrw    mstatus, zero
core   0: 0x000000008000018c (0x10001073) csrw    sstatus, zero
core   0: 0x0000000080000190 (0x20001073) csrw    vsstatus, zero
core   0: 0x0000000080000194 (0x300023f3) csrr    t2, mstatus
core   0: 0x0000000080000198 (0x10002e73) csrr    t3, sstatus
core   0: 0x000000008000019c (0x20002ef3) csrr    t4, vsstatus
: reg 0
t2: 0x0000000a00000000 t3: 0x0000000200000000  t4: 0x0000000200000000

core   0: 0x00000000800001a0 (0x000182b7) lui     t0, 0x18
core   0: 0x00000000800001a4 (0x30029073) csrw    mstatus, t0
core   0: 0x00000000800001a8 (0x10029073) csrw    sstatus, t0
core   0: 0x00000000800001ac (0x20029073) csrw    vsstatus, t0
core   0: 0x00000000800001b0 (0x300023f3) csrr    t2, mstatus
core   0: 0x00000000800001b4 (0x10002e73) csrr    t3, sstatus
core   0: 0x00000000800001b8 (0x20002ef3) csrr    t4, vsstatus
: reg 0
t2: 0x8000000a00018000 t3: 0x8000000200018000  t4: 0x8000000200018000
```
[xs-writable.tar.gz](https://github.com/chipsalliance/rocket-chip/files/8747904/xs-writable.tar.gz)

### DIFF
--- a/customext/cflush.cc
+++ b/customext/cflush.cc
@@ -12,7 +12,7 @@ static reg_t custom_cflush(processor_t* p, insn_t insn, reg_t pc)
 {
   require_privilege(PRV_M);
 
-  return pc + 4; \
+  return pc + 4;
 }
 
 class cflush_t : public extension_t
@@ -20,7 +20,7 @@ class cflush_t : public extension_t
  public:
   const char* name() { return "cflush"; }
 
-  cflush_t() {}
+  cflush_t() : extension_t(true) {}
 
   std::vector<insn_desc_t> get_instructions() {
     std::vector<insn_desc_t> insns;
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -13,6 +13,8 @@
 #include "trap.h"
 // For require():
 #include "insn_macros.h"
+// For xs_gatherer_t
+#include "extension.h"
 
 // STATE macro used by require_privilege() macro:
 #undef STATE
@@ -405,9 +407,15 @@ base_status_csr_t::base_status_csr_t(processor_t* const proc, const reg_t addr):
   has_page(proc->extension_enabled_const('S') && proc->supports_impl(IMPL_MMU)),
   sstatus_write_mask(compute_sstatus_write_mask()),
   sstatus_read_mask(sstatus_write_mask | SSTATUS_UBE | SSTATUS_UXL
+                    | (proc->extension_enabled('X') ? SSTATUS_XS : 0)
                     | (proc->get_const_xlen() == 32 ? SSTATUS32_SD : SSTATUS64_SD)) {
 }
 
+bool base_status_csr_t::field_exists(const reg_t which) {
+  return ((sstatus_write_mask 
+            | (proc->extension_enabled('X') ? SSTATUS_XS : 0))
+          & which) != 0;
+}
 
 reg_t base_status_csr_t::compute_sstatus_write_mask() const noexcept {
   // If a configuration has FS bits, they will always be accessible no
@@ -419,7 +427,6 @@ reg_t base_status_csr_t::compute_sstatus_write_mask() const noexcept {
     | (proc->extension_enabled('S') ? (SSTATUS_SIE | SSTATUS_SPIE | SSTATUS_SPP) : 0)
     | (has_page ? (SSTATUS_SUM | SSTATUS_MXR) : 0)
     | (has_fs ? SSTATUS_FS : 0)
-    | (proc->extension_enabled('X') ? SSTATUS_XS : 0)
     | (has_vs ? SSTATUS_VS : 0)
     ;
 }
@@ -467,6 +474,11 @@ vsstatus_csr_t::vsstatus_csr_t(processor_t* const proc, const reg_t addr):
   val(proc->get_state()->mstatus->read() & sstatus_read_mask) {
 }
 
+reg_t vsstatus_csr_t::read() const noexcept {
+  return val
+         | adjust_sd(set_field((reg_t)0, SSTATUS_XS, proc->get_xs_gatherer()->get_xs()));
+}
+
 bool vsstatus_csr_t::unlogged_write(const reg_t val) noexcept {
   const reg_t newval = (this->val & ~sstatus_write_mask) | (val & sstatus_write_mask);
   if (state->v) maybe_flush_tlb(newval);
@@ -503,6 +515,11 @@ mstatus_csr_t::mstatus_csr_t(processor_t* const proc, const reg_t addr):
   ) {
 }
 
+reg_t mstatus_csr_t::read() const noexcept {
+  return val 
+         | adjust_sd(set_field((reg_t)0, SSTATUS_XS, proc->get_xs_gatherer()->get_xs()));
+  
+}
 
 bool mstatus_csr_t::unlogged_write(const reg_t val) noexcept {
   const bool has_mpv = proc->extension_enabled('S') && proc->extension_enabled('H');
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -419,7 +419,7 @@ reg_t base_status_csr_t::compute_sstatus_write_mask() const noexcept {
     | (proc->extension_enabled('S') ? (SSTATUS_SIE | SSTATUS_SPIE | SSTATUS_SPP) : 0)
     | (has_page ? (SSTATUS_SUM | SSTATUS_MXR) : 0)
     | (has_fs ? SSTATUS_FS : 0)
-    | (proc->any_custom_extensions() ? SSTATUS_XS : 0)
+    | (proc->extension_enabled('X') ? SSTATUS_XS : 0)
     | (has_vs ? SSTATUS_VS : 0)
     ;
 }
@@ -689,7 +689,7 @@ mie_csr_t::mie_csr_t(processor_t* const proc, const reg_t addr):
 reg_t mie_csr_t::write_mask() const noexcept {
   const reg_t supervisor_ints = proc->extension_enabled('S') ? MIP_SSIP | MIP_STIP | MIP_SEIP : 0;
   const reg_t hypervisor_ints = proc->extension_enabled('H') ? MIP_HS_MASK : 0;
-  const reg_t coprocessor_ints = (reg_t)proc->any_custom_extensions() << IRQ_COP;
+  const reg_t coprocessor_ints = (reg_t)proc->extension_enabled('X') << IRQ_COP;
   const reg_t delegable_ints = supervisor_ints | coprocessor_ints;
   const reg_t all_ints = delegable_ints | hypervisor_ints | MIP_MSIP | MIP_MTIP | MIP_MEIP;
   return all_ints;
@@ -789,7 +789,7 @@ void mideleg_csr_t::verify_permissions(insn_t insn, bool write) const {
 
 bool mideleg_csr_t::unlogged_write(const reg_t val) noexcept {
   const reg_t supervisor_ints = proc->extension_enabled('S') ? MIP_SSIP | MIP_STIP | MIP_SEIP : 0;
-  const reg_t coprocessor_ints = (reg_t)proc->any_custom_extensions() << IRQ_COP;
+  const reg_t coprocessor_ints = (reg_t)proc->extension_enabled('X') << IRQ_COP;
   const reg_t delegable_ints = supervisor_ints | coprocessor_ints;
 
   return basic_csr_t::unlogged_write(val & delegable_ints);
--- a/riscv/extension.cc
+++ b/riscv/extension.cc
@@ -21,3 +21,25 @@ void extension_t::raise_interrupt()
 void extension_t::clear_interrupt()
 {
 }
+
+void xs_gatherer_t::reset() {
+  for (auto e : custom_extensions) {
+    e.second->reset();
+  }
+}
+
+void xs_gatherer_t::register_extension(extension_t* x)
+{
+  for (auto insn : x->get_instructions())
+    p->register_insn(insn);
+  p->build_opcode_map();
+
+  for (auto disasm_insn : x->get_disasms())
+    p->disassembler->add_insn(disasm_insn);
+
+  if (!custom_extensions.insert(std::make_pair(x->name(), x)).second) {
+    fprintf(stderr, "extensions must have unique names (got two named \"%s\"!)\n", x->name());
+    abort();
+  }
+  x->set_processor(p);
+}
--- a/riscv/extension.cc
+++ b/riscv/extension.cc
@@ -43,3 +43,23 @@ void xs_gatherer_t::register_extension(extension_t* x)
   }
   x->set_processor(p);
 }
+
+extension_t* xs_gatherer_t::get_extension()
+{
+  switch (custom_extensions.size()) {
+    case 0: return NULL;
+    case 1: return custom_extensions.begin()->second;
+    default:
+      fprintf(stderr, "xs_gatherer_t::get_extension() is ambiguous when multiple extensions\n");
+      fprintf(stderr, "are present!\n");
+      abort();
+  }
+}
+
+extension_t* xs_gatherer_t::get_extension(const char* name)
+{
+  auto it = custom_extensions.find(name);
+  if (it == custom_extensions.end())
+    abort();
+  return it->second;
+}
--- a/riscv/extension.cc
+++ b/riscv/extension.cc
@@ -28,6 +28,25 @@ void xs_gatherer_t::reset() {
   }
 }
 
+extension_state_t xs_gatherer_t::get_xs()
+{
+  bool clean = false, init = false;
+  for (auto e : custom_extensions) {
+      extension_state_t state = e.second->get_state();
+      if (state == EXT_STATE_DIRTY)
+        return EXT_STATE_DIRTY;
+      clean = (state == EXT_STATE_CLEAN) || clean;
+      init  = (state == EXT_STATE_INIT)  || init;
+  }
+
+  if (clean)
+    return EXT_STATE_CLEAN;
+  else if (init)
+    return EXT_STATE_INIT;
+  else
+    return EXT_STATE_OFF;
+}
+
 void xs_gatherer_t::register_extension(extension_t* x)
 {
   for (auto insn : x->get_instructions())
--- a/riscv/extension.h
+++ b/riscv/extension.h
@@ -61,6 +61,9 @@ public:
   void register_extension(extension_t*);
   extension_t* get_extension();
   extension_t* get_extension(const char* name);
+  bool any_custom_extensions() const {
+    return !custom_extensions.empty();
+  }
 protected:
   processor_t* p;
   std::unordered_map<std::string, extension_t*> custom_extensions;
--- a/riscv/extension.h
+++ b/riscv/extension.h
@@ -8,23 +8,39 @@
 #include <vector>
 #include <functional>
 
+typedef enum {
+  EXT_STATE_OFF,
+  EXT_STATE_INIT,
+  EXT_STATE_CLEAN,
+  EXT_STATE_DIRTY,
+
+  MAX_EXT_STATE
+} extension_state_t;
+
 class extension_t
 {
  public:
+  extension_t(bool always_dirty=false) : always_dirty(always_dirty) {}
   virtual std::vector<insn_desc_t> get_instructions() = 0;
   virtual std::vector<disasm_insn_t*> get_disasms() = 0;
   virtual const char* name() = 0;
-  virtual void reset() {};
+  virtual void reset() { state = always_dirty ? EXT_STATE_DIRTY : EXT_STATE_OFF; }
   virtual void set_debug(bool value) {};
   virtual ~extension_t();
 
   void set_processor(processor_t* _p) { p = _p; }
+  extension_state_t get_state() const { return state; }
+  void set_state(extension_state_t s) { state = s; }
  protected:
   processor_t* p;
 
   void illegal_instruction();
   void raise_interrupt();
   void clear_interrupt();
+
+ private:
+  const bool always_dirty;
+  extension_state_t state;
 };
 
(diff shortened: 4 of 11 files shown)


customext/dummy_rocc.cc
```diff
@@ -44,4 +44,4 @@ class dummy_rocc_t : public rocc_t
   reg_t acc[num_acc];
 };
 
-REGISTER_EXTENSION(dummy_rocc, []() { return new dummy_rocc_t; })
+REGISTER_EXTENSION(rocc, []() { return new dummy_rocc_t; })
```
aswaterman: This should not be part of the same commit, nor even the same PR.

riscv/rocc.cc
```diff
@@ -2,11 +2,13 @@
 
 #include "rocc.h"
 #include "trap.h"
+#include "insn_macros.h"
 #include <cstdlib>
 
 #define customX(n) \
   static reg_t c##n(processor_t* p, insn_t insn, reg_t pc) \
   { \
+    require_accelerator; \
```
aswaterman, line 10: This isn’t architecturally consistent. XS is not supposed to be a writable field: it’s a readable thing driven by the accelerator.

Phantom1003: I summarized the permissions of xs in different CSR:
|  | spike | rocket |
|:-:|:-:|:-:|
| [m/s]status | RW | R |
| vsstatus | RW | RW |

Can I consider all the RW in the table as inconsistent with the specification and report them as bugs?

Now the ROCC behavior of spike and rocket is not consistent, I want to submit a series of patches to make their behavior consistent for rocket verification, do you think it's worth it or is it intentional to make them inconsistent.

customext/dummy_rocc.cc
Phantom1003: Got it, I will remove this.

riscv/rocc.cc
Phantom1003: One more thing I would like to confirm with you is how we decide whether a Z* extension is on or off.
First, they cannot be represented in the misa.
And if XS is read-only, that means that for any Z* extension, it should have at least a 1-bit status register to indicate whether the extension is on or off, in addition to implementing its own logical functions. 
And even if the extension is purely combinatorial, such as Zk extension, it should have a status register to indicate its state. It looks like the majority of extensions are currently focused on the functional specification and have not yet declared the address of their status registers.

riscv/processor.cc
```diff
@@ -62,6 +59,10 @@ processor_t::processor_t(const isa_parser_t *isa, const char* varch,
   set_impl(IMPL_MMU_ASID, true);
   set_impl(IMPL_MMU_VMID, true);
 
+  xs_gatherer = new xs_gatherer_t(isa->get_max_xlen() == 32);
```
scottj97: This appears broken. The constructor parameter is `xlen`, but you are passing in a boolean?

Also, this code assumes that `xlen` will not change dynamically. For that, you should use `get_const_xlen()` to document that assumption.

customext/dummy_rocc.cc
```diff
@@ -44,4 +44,4 @@ class dummy_rocc_t : public rocc_t
   reg_t acc[num_acc];
 };
 
-REGISTER_EXTENSION(rocc, []() { return new dummy_rocc_t; })
+REGISTER_EXTENSION(dummy_rocc, []() { return new dummy_rocc_t; })
```
scottj97: There should not be reverts like a0c738a0a72e134e2bf6ce1c7c50b118327ea9c9; rebase and patch if you've changed your mind about the variable rename.

riscv/extension.h
```diff
@@ -5,23 +5,59 @@
...
 
   void set_processor(processor_t* _p) { p = _p; }
+  extension_state_t get_state() { return state; }
+  void set_state(extension_state_t s) { state = p->get_xs_gatherer()->update_state(state, s); }
+
  protected:
   processor_t* p;
 
+  bool always_dirty;
+  extension_state_t state;
```
scottj97: Seems like this should be private, so subclasses are required to use `set_state()`?

riscv/extension.cc
```diff
@@ -21,3 +21,29 @@ void extension_t::raise_interrupt()
 void extension_t::clear_interrupt()
 {
 }
+
+void xs_gatherer_t::reset() {
```
scottj97, line 25: Why isn't this in the constructor? Is there ever a need to call this more than once?

riscv/extension.h
```diff
@@ -5,23 +5,59 @@
...
+typedef enum {
+  EXT_STATE_OFF,
+  EXT_STATE_INIT,
+  EXT_STATE_CLEAN,
+  EXT_STATE_DIRTY,
+
+  MAX_EXT_STATE
+} extension_state_t;
+
+class xs_gatherer_t
```
scottj97: This class seems brittle. It's keeping track, independently, of the state of all extensions. It's very easy for an extension to screw up and have a subtle bug, such as neglecting to call `check_in()`, or changing their state without calling `update_state()`.

I'd rather see a design where `get_xs()` queries each extension for its state and combines them. That way, there's no possibility of this class's records getting out of date with the extensions.

riscv/csrs.cc
```diff
@@ -13,6 +13,8 @@
 #include "trap.h"
 // For require():
 #include "insn_macros.h"
+// For xs_gatherer_t
+#include "extension.h"
```
scottj97, line 17: Commit e4be43ca2ccba2188181d9f6079bd7d617f04573 does far more than what its commit message says. This is misleading. That commit should be broken up into many smaller pieces.

riscv/extension.cc
```diff
@@ -39,7 +39,7 @@ extension_state_t xs_gatherer_t::update_state(extension_state_t prev, extension_
 reg_t xs_gatherer_t::get_xs()
 {
   if (cnt[EXT_STATE_DIRTY] > 0)
-    return set_field(0, SSTATUS_XS, EXT_STATE_DIRTY);
+    return set_field(xlen == 32 ? SSTATUS32_SD : SSTATUS64_SD, SSTATUS_XS, EXT_STATE_DIRTY);
```
scottj97: The `SD` bit is not part of the `XS` field so it's misleading to set it here in a method called `get_xs()`. The `SD` bit is better set by the CSR-specific code. Same for shifting `XS` into the right bit position (the `set_field()` part).

riscv/extension.h
```diff
@@ -71,9 +72,4 @@ void register_extension(const char* name, std::function<extension_t*()> f);
     public: register_##name() { register_extension(#name, constructor); } \
   }; static register_##name dummy_##name;
 
-
-
```
scottj97: Squash this blank-line-removal back so there's not an addition of blank lines followed by a removal of blank lines in the commit history.

riscv/csrs.cc
```diff
@@ -407,7 +407,7 @@ base_status_csr_t::base_status_csr_t(processor_t* const proc, const reg_t addr):
   has_page(proc->extension_enabled_const('S') && proc->supports_impl(IMPL_MMU)),
   sstatus_write_mask(compute_sstatus_write_mask()),
   sstatus_read_mask(sstatus_write_mask | SSTATUS_UBE | SSTATUS_UXL
-                    | (proc->any_custom_extensions() ? SSTATUS_XS : 0)
+                    | (proc->get_xs_gatherer()->any_custom_extensions() ? SSTATUS_XS : 0)
```
scottj97: Commit 32c9728903f7640386766d1d236187a6e19bc0ad does far more than its commit message says, and needs to be broken up into many smaller commits to make it reviewable.

riscv/csrs.cc
```diff
@@ -427,6 +427,12 @@ reg_t base_status_csr_t::compute_sstatus_write_mask() const noexcept {
 }
 
 
+bool base_status_csr_t::field_exists(const reg_t which) {
```
scottj97: Commit 306671bb477d7adff9e3a07e71d1867dcb8107b6 does two things: Moves the implementation from `csrs.h` to `csrs.cc`, then changes that implementation. That should be two commits so that the change is apparent in the git diff.

In this case, however, shouldn't this change be part of an earlier commit? It looks like this is fixing a bug?

riscv/extension.h
```diff
@@ -5,23 +5,42 @@
...
 class extension_t
 {
  public:
+  extension_t(bool always_dirty=false)
+    : always_dirty(always_dirty), state(EXT_STATE_OFF) {}
   virtual std::vector<insn_desc_t> get_instructions() = 0;
   virtual std::vector<disasm_insn_t*> get_disasms() = 0;
   virtual const char* name() = 0;
-  virtual void reset() {};
+  virtual void reset() { state = always_dirty ? EXT_STATE_DIRTY : EXT_STATE_OFF; };
```
scottj97: Why isn't this part of the constructor? Is there ever a need to call this more than once?

riscv/extension.h
```diff
@@ -5,23 +5,42 @@
...
   virtual ~extension_t();
 
   void set_processor(processor_t* _p) { p = _p; }
+  extension_state_t get_state() { return state; }
+  void set_state(extension_state_t s);
+
  protected:
   processor_t* p;
 
+  bool always_dirty;
```
scottj97: This should be `const`

riscv/extension.h
```diff
@@ -5,23 +5,42 @@
...
   virtual std::vector<insn_desc_t> get_instructions() = 0;
   virtual std::vector<disasm_insn_t*> get_disasms() = 0;
   virtual const char* name() = 0;
-  virtual void reset() {};
+  virtual void reset() { state = always_dirty ? EXT_STATE_DIRTY : EXT_STATE_OFF; };
   virtual void set_debug(bool value) {};
   virtual ~extension_t();
 
   void set_processor(processor_t* _p) { p = _p; }
+  extension_state_t get_state() { return state; }
```
scottj97: This should be `const`

customext/cflush.cc
```diff
@@ -10,17 +10,17 @@ struct : public arg_t {
...
+  return pc + 4;
 }
 
 class cflush_t : public extension_t
 {
  public:
   const char* name() { return "cflush"; }
 
-  cflush_t() {}
+  cflush_t() { always_dirty = true; }
```
scottj97: This should be passed in as a constructor param, instead of allowing the parent class to set `always_dirty` incorrectly and then changing it here. Same for class `dummy_rocc_t`.

riscv/csrs.cc
```diff
@@ -467,6 +475,10 @@ vsstatus_csr_t::vsstatus_csr_t(processor_t* const proc, const reg_t addr):
   val(proc->get_state()->mstatus->read() & sstatus_read_mask) {
 }
 
+reg_t vsstatus_csr_t::read() const noexcept {
+  return val | proc->get_xs_gatherer()->get_xs();
```
scottj97: This precludes the possibility of an extension tracking virtualized state, like the VS and FS fields provide. (I.e. `sstatus.FS` and `vsstatus.FS` are separate states.) Is that okay?

scottj97 (changes requested):

riscv/processor.cc
Phantom1003: Yes, this is a bug, and it should be fixed at `32c972`. In the following commit, I move all extensions related to the xs_gatherer, it can dynamically get the xlen information.

customext/dummy_rocc.cc
Phantom1003: Thanks, I will rebase it.

riscv/extension.cc
Phantom1003: This commit is a simple hook just place the codes, I want to move all extensions into this class, so I need to reset after registering, just like the original code does.

riscv/extension.h
Phantom1003: The check_in is also a hook, after moving all extensions in the xs_gather, it has been removed.

I want to minimize the performance loss, if every time to read mstatus need to poll may affect the performance, so I designed this constant query time implementation.
At runtime, xs doesn't update for unrelated instruction and csr access, and if we can require users to update state only from specific api, I believe we can remove the performance loss caused by polling.

riscv/extension.h
Phantom1003: Thanks, I agree with you. (all the `const `)

riscv/csrs.cc
Phantom1003: Here I am following the rocket implementation, rocket really does not consider the virtualization of rocc

riscv/extension.cc
Phantom1003: Yes, I agree with you.

riscv/csrs.cc
Phantom1003: This place becomes a bug after disallowing write to xs, since it no longer appears in the write mask.

riscv/csrs.cc
```diff
@@ -419,7 +419,7 @@ reg_t base_status_csr_t::compute_sstatus_write_mask() const noexcept {
     | (proc->extension_enabled('S') ? (SSTATUS_SIE | SSTATUS_SPIE | SSTATUS_SPP) : 0)
     | (has_page ? (SSTATUS_SUM | SSTATUS_MXR) : 0)
     | (has_fs ? SSTATUS_FS : 0)
-    | (proc->any_custom_extensions() ? SSTATUS_XS : 0)
+    | (proc->extension_enabled('X') ? SSTATUS_XS : 0)
```
scottj97: The commit message for b26a2063384786752f7643a403369cbaf01539d4 should explain _why_ you are making this change, and what effect it will have.

riscv/extension.h
```diff
@@ -35,4 +51,27 @@ void register_extension(const char* name, std::function<extension_t*()> f);
     public: register_##name() { register_extension(#name, constructor); } \
   }; static register_##name dummy_##name;
 
+class xs_gatherer_t
```
scottj97, line 56: Commit d5a80cf4c26fde5d01233156c5643b286c680043 is far too big and needs to be broken up into smaller steps. Such as:

1. Introduce and instantiate new (empty) class `xs_gatherer_t`
2. Register all extensions in class `xs_gatherer_t` (only adding it to the local `custom_extensions` variable; leave old `register_extension()` in place until obsoleted)
3. Move `get_extension()` to class `xs_gatherer_t`
4. Move `any_custom_extensions()` to class `xs_gatherer_t`
5. Add `set_debug()` to class `xs_gatherer_t`
6. Move remaining body of old `register_extension()` to new `xs_gatherer_t::register_extension()`
7. Delete old `register_extension()` and `processor_t::custom_extensions` now that they're no longer needed
8. Add extension state tracking to `xs_gatherer_t`

riscv/csrs.cc
```diff
@@ -405,9 +407,15 @@ base_status_csr_t::base_status_csr_t(processor_t* const proc, const reg_t addr):
   has_page(proc->extension_enabled_const('S') && proc->supports_impl(IMPL_MMU)),
   sstatus_write_mask(compute_sstatus_write_mask()),
   sstatus_read_mask(sstatus_write_mask | SSTATUS_UBE | SSTATUS_UXL
+                    | (proc->extension_enabled('X') ? SSTATUS_XS : 0)
```
scottj97, line 410: This code is executed only at reset and therefore assumes that `proc->extension_enabled('X')` cannot change during runtime. Therefore this should use `proc->extension_enabled_const('X')` to document and verify that assumption.

scottj97: That is, assuming that dynamically setting `misa.X=0` should result in XS always reading 0? I'm not sure if that is allowed/required by the privileged spec or not.

If XS is allowed to be nonzero even after setting `misa.X=0` then this code is correct.

riscv/csrs.cc
```diff
@@ -503,6 +515,11 @@ mstatus_csr_t::mstatus_csr_t(processor_t* const proc, const reg_t addr):
   ) {
 }
 
+reg_t mstatus_csr_t::read() const noexcept {
+  return val 
+         | adjust_sd(set_field((reg_t)0, SSTATUS_XS, proc->get_xs_gatherer()->get_xs()));
```
scottj97, line 520: You may be right that this causes a significant performance loss. If so, optimization should come as a future commit or PR.

riscv/csrs.cc
```diff
@@ -13,6 +13,8 @@
 #include "trap.h"
 // For require():
 #include "insn_macros.h"
+// For xs_gatherer_t
```
scottj97, line 16: The commit message in ec708aa736c136603af019d65583e06d412f3dfd is misleading. It's not disallowing write to *status -- only the XS field. How about: "Compute *status.XS from extensions instead of being writable"

customext/cflush.cc
```diff
@@ -10,17 +10,17 @@ struct : public arg_t {
 
 static reg_t custom_cflush(processor_t* p, insn_t insn, reg_t pc)
 {
```
scottj97, line 12: Commit message for 289e47fdb51ae6d97b202fdf66cceb9fbe305ee1 is misleading because it updates more than just the rocc extension. Also: update for what? How about: "Update existing extensions for new API"

riscv/rocc.h
```diff
@@ -24,6 +24,7 @@ union rocc_insn_union_t
 class rocc_t : public extension_t
 {
  public:
+  rocc_t(bool always_dirty) : extension_t(always_dirty) {}
```
scottj97: Since `rocc_t` never calls `set_state()` it seems that the code assumes `always_dirty` and hence this should not be a constructor parameter. Instead:
```
rocc_t(): extension_t(true) {}
```

riscv/rocc.cc
```diff
@@ -7,6 +7,7 @@
...
```
scottj97, line 10: The ROCC extension can be in `EXT_STATE_OFF` even if other extensions are enabled (and hence XS != `EXT_STATE_OFF`). Shouldn't this only be checking for the state of ROCC and not all the other extensions too?

(I realize ROCC is always dirty but that might not always be the case for other extensions.)

scottj97 (changes requested):
Please cleanup whitespace errors. There are at least 4 lines with trailing whitespace and one file missing a final newline.

riscv/csrs.cc
Phantom1003: replace any_custom_extensions() to extension_enabled('X'), so that we can use --isa="xdummy" to enable related function, instead of loading a real extension.

riscv/csrs.cc
Phantom1003: This modification is the same as the original design, which uses any_cunstom_extension() in write_mask.
And in the misa_mask the X bit is not allowed to modify during runtime.

riscv/rocc.cc
Phantom1003: Yes, you are right, such situations do happen.
But I still insist on the check here, because rocc_t is a parent class, here it is a global check for performance optimization, (if xs==0, there is no need to do any following operations), and each extension's own state check should be placed inside its own implementation, e.g. dummy_rocc_t.

riscv/csrs.cc
Phantom1003: Thanks, I will optimize this in another PR after this one is merged.

riscv/rocc.cc
scottj97: I think it would be really misleading to someone who's using this code as a reference for adding a new extension that's not always dirty. If they include this line, their extension will be subtly broken.

What if you change the macro to be something like `require_accelerator(rocc)` which then tests `rocc->get_state() != OFF`? That would also be higher performance than calculating XS via the current system.

riscv/csrs.cc
scottj97: I don't understand what that means, but if you think that's a sufficient explanation, please update the commit message for b26a2063384786752f7643a403369cbaf01539d4 with that explanation.

riscv/csrs.cc
scottj97: The misa `write_mask` can change, and the purpose of `extension_enabled_const()` is to make sure such changes do not subtly break assumptions made elsewhere in the code.

The old `any_custom_extensions()` (which seems is dead now, and can be removed?) did not consider `misa.X` so didn't need to think about this.

The only line I can find in the spec about this is:

> In systems without additional user extensions requiring new state, the XS field is read-only zero.

And regarding bits in `misa` generally:

> When a standard extension is disabled by clearing its bit in misa, the instructions and CSRs defined or modified by the extension revert to their defined or reserved behaviors as if the extension is not implemented.

Which suggests to me that if `misa.X` is dynamically updated to 0, the XS field should also read as 0.

There's several options for achieving that, if we ever wanted to make `misa.X` writable, but at the very least we need to document the assumption here that `misa.X` is not writable, by using `extension_enabled_const()`.

scottj97:
As of 4e405e28a11cc99c8706d38c287f9305a2c87f97 this code is calling `p->register_insn()` twice and `p->disassembler->add_insn()` twice for each extension. That's not good. The calls to these methods should be moved to the new location in an atomic commit, not left in both places during the refactoring.

Also as of 4e405e28a11cc99c8706d38c287f9305a2c87f97 the other two places where `->register_extension()` are called (in `spike_log_parser.cc` and `spike.cc`) are broken since they are missing the call to the new `xs_gatherer_t::register_extension()`. These need to be updated here also to call both the old and the new during the transition.

scottj97:
4a3bd5422090831963cf7ebe9972039ea772869b still has three instances of trailing whitespace.

PR #992, opened 2022-05-09
Fix for issue #990, implements the PBMTE bit in the henvcfg /menvcfg reisters

### LINKED ISSUE #990: PBMTE bit not implemented in menvcfg and henvcfg
`PBMTE` needs to be added to both registers, `menvcfg` and `henvcfg`, and checked during TableWalks to determine if the `PBMT` bits can be set for leaf PTEs (see c4fdc8fe95e497ab6e107117a7de47ec40e7de7d). Currently Spike only checks to see if the extension is enabled.

In addition, the spec notes that `PBMT=3` is a reserved value, but doesn't specify behavior for that case. Currently Spike does nothing for this case (proceeds as if `PBMT=0`). Should a page fault be thrown in this case, as with other reserved values?

rbuchner-aril:
- Implements the `PBMTE` bit in the `henvcfg` /`menvcfg` registers and checks them during TableWalks
- Makes the value `PBMT==3`, which is reserved in the spec, cause a fault during a TableWalk
- Makes the reset value of `*envcfg.PBMTE = 1` for backward compatibility

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -302,6 +302,8 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
         break;
       } else if (!pbmte && (pte & PTE_PBMT)) {
         break;
+      } else if ((pte & PTE_PBMT) == PTE_PBMT) {
+        break;
       } else if (PTE_TABLE(pte)) { // next level of page table
         if (pte & (PTE_D | PTE_A | PTE_U | PTE_N | PTE_PBMT))
           break;
@@ -392,6 +394,8 @@ reg_t mmu_t::walk(reg_t addr, access_type type, reg_t mode, bool virt, bool hlvx
     } else if (!proc->extension_enabled(EXT_SVNAPOT) && (pte & PTE_N)) {
       break;
     } else if (!pbmte && (pte & PTE_PBMT)) {
+      break;
+    } else if ((pte & PTE_PBMT) == PTE_PBMT) {
       break;
     } else if (PTE_TABLE(pte)) { // next level of page table
       if (pte & (PTE_D | PTE_A | PTE_U | PTE_N | PTE_PBMT))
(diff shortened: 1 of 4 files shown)


riscv/mmu.cc
```diff
@@ -299,7 +299,7 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
         break;
       } else if (!proc->extension_enabled(EXT_SVNAPOT) && (pte & PTE_N)) {
         break;
-      } else if (!proc->extension_enabled(EXT_SVPBMT) && (pte & PTE_PBMT)) {
+      } else if (!(proc->get_state()->csrmap[CSR_HENVCFG]->read() & HENVCFG_PBMTE) && (pte & PTE_PBMT)) {
```
scottj97: Prefer accessing the CSR object directly, i.e. `proc->get_state()->henvcfg->read()`. The `csrmap` is there for dynamic access such as `csrw` instruction execution where the specific CSR is unknown at Spike compile time.

scottj97 (changes requested):

scottj97:
@daniellustig would you agree that PTE.PBMT=3, being a reserved value, should cause a page fault?

daniellustig:
> @daniellustig would you agree that PTE.PBMT=3, being a reserved value, should cause a page fault?

Yes, that matches the spec wording in the virtual address translation process: "...if any bits or encodings that are reserved for
future standard use are set within $pte$, stop and raise a page-fault exception corresponding to the original access type."

scottj97 (approved):

scottj97 (changes requested):
Internal discussions revealed some architectural problems with this implementation. @rbuchner-aril is working on it.

scottj97 (approved):
As of ccfeaa99732479b029ce95fc9b09968c7f3b1f64 this looks in good shape to me.

aswaterman (approved):
LGTM.

PR #994, opened 2022-05-11
rvv: fix the checking eew and elen for index load

chihminchao:
eew of index register can't be larger than elen

ex:
   elen = 32,  vloxei64.v is illegal


aswaterman (approved):
True.

PR #995, opened 2022-05-12
update disasm priority for custom extensions loaded from extlib

liweiwei90:
The disasms for custom extensions loaded from extlib are registered as the lowest priority currently which means they will be recognized as the unspecified instructions if they reuse the encode. 
For example, if they reuse the encode for custom* instructions(which is designed for custom instructions),  they will be recgonized as "custom*" instructions.
This patch update the priority for custom extensions to the highest priority to avoid this problem which is also in line with the comment: "highest priority: instructions explicitly enabled".

### DIFF
--- a/disasm/disasm.cc
+++ b/disasm/disasm.cc
@@ -2098,9 +2098,12 @@ disassembler_t::disassembler_t(const isa_parser_t *isa)
 {
   // highest priority: instructions explicitly enabled
   add_instructions(isa);
+}
 
+void disassembler_t::add_other_insns(unsigned xlen)
+{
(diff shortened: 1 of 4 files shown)


aswaterman (changes requested):
This makes the API weird: if you simply construct a `disassembler_t` but you don't call `add_other_insns`, then the disassembler doesn't work.  This doesn't follow the principle of least surprise.  I'll see if I can think of a cleaner solution.

aswaterman:
I've fixed this in an API-preserving way in #999.

aswaterman:
Superseded by #999.  Thanks for raising the issue!

PR #996, opened 2022-05-12
fix loop condition in decode_insn

liweiwei90:


scottj97:
This needs more context..."fix" what? What was wrong with the previous code?

scottj97:
It appears this might have been broken by 5b7cdbe1cf75112bd2a472b7490b15fa7078d798 but why haven't we seen any problems until now?

aswaterman:
I did break this in 5b7cdbe1cf75112bd2a472b7490b15fa7078d798.

I can see how this bug could manifest if a DECLARE_OVERLAP_INSN that isn't currently `supported` makes its way into the opcode_hash.  Not exactly sure how to force it to happen, though.

But... isn't the simpler and cleaner solution just to not register instructions that aren't supported, rather than carrying around the `supported` flag?  This allows us to get rid of all `nullptr`s from this code, eliminating this check altogether.  I've prototyped this idea in #997.

liweiwei90:
Sorry. This happens when a instruction loaded from extlib resue the encode from instructions in (unratified) standard extension which is not enabled by --isa.
By the way, the null check for desc.func in the loop seems not right or unnecessary.

aswaterman:
Superseded by #997.  Thanks for raising the issue!

PR #997, opened 2022-05-12
Simplify decode_insn and insn_desc_t

aswaterman:
... by removing the possibility of unsupported instructions in the instruction list, and hence the possibility of `nullptr`s within insn_desc_t.

Supersedes #996


liweiwei90:
Yeah, It's OK.

scottj97 (approved):

PR #998, opened 2022-05-12
Call throw_access_exception instead of duplicating

kwalker27:
Mark 'throw_access_exception' as NORETURN and call it from the end of
mmu_t::walk() instead of duplicating the contents of that function.


aswaterman (changes requested):
This isn't actually code duplication.  `throw_access_exception` throws access exceptions; the deleted code throws page-fault exceptions.

kwalker27:
Oops, missed that.

PR #999, opened 2022-05-12
Fix disassembly of custom instructions that overlap standard ones

aswaterman:
Iterate over the instruction chains in reverse order, prioritizing the last call to `disassembler_t::add_insn`.

To preserve behavior for the standard instructions, reverse the order in which we add instructions in the `disassembler_t` constructor.

Supersedes #995.


liweiwei90:
OK.

scottj97 (approved):

PR #1000, opened 2022-05-12
Disassemble Zicbop/Zihintpause HINT instructions

aswaterman:
We do not condition them on Zicbop/Zihintpause because, definitionally, all implementations provide them.


aswaterman:
Wow, PR number 1000!

scottj97 (approved):

PR #1006, opened 2022-05-18
Move ebreak* logic from take_trap into instructions.

### LINKED ISSUE #725: dcsr.ebreak* are not handled properly
The debug spec says that dcsr.ebreakm causes ebreak instructions in M mode to enter debug mode.  It should not affect other breakpoints.

execute.cc contains:
```
      switch (state.mcontrol[t.index].action) {
        case ACTION_DEBUG_MODE:
          enter_debug_mode(DCSR_CAUSE_HWBP);
          break;
        case ACTION_DEBUG_EXCEPTION: {
          insn_trap_t trap(CAUSE_BREAKPOINT, t.address);
          take_trap(trap, pc);
          break;
        }
```
So mcontrol will `trap(CAUSE_BREAKPOINT)` which is correct since mcause=3, etc.  But this is not an ebreak instruction, it's a PC match or whatever.

processor.cc has this in `processor_t::take_trap()`:
```
  if (t.cause() == CAUSE_BREAKPOINT && (
              (state.prv == PRV_M && state.dcsr.ebreakm) ||
              (state.prv == PRV_S && state.dcsr.ebreaks) ||
              (state.prv == PRV_U && state.dcsr.ebreaku))) {
    enter_debug_mode(DCSR_CAUSE_SWBP);
    return;
  }
```
So that means that dcsr.ebreakm applies to any CAUSE_BREAKPOINT, including PC breakpoints.

I don't currently have a test that does both self-hosted PC breakpoints (which should trap and not enter debug mode) and external debug (which is the only way to set those dcsr bits).  I found this by inspection.

I didn't trace this back but I assume that the ebreak instruction decode needs to have a switch statement similar to the execute.cc code I quoted above rather than calling trap() and letting that decide on debug mode or not.

timsifive:
Now that logic only affects ebreak instructions, and does not affect triggers that also cause a trap to be taken.

Fixes #725. Although like Paul, I don't have a test for this case.

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -337,6 +337,10 @@ void processor_t::step(size_t n)
           abort();
       }
     }
+    catch(trap_debug_mode&)
+    {
+      enter_debug_mode(DCSR_CAUSE_SWBP);
+    }
     catch (wait_for_interrupt_t &t)
     {
       // Return to the outer simulation loop, which gives other devices/harts a
--- a/riscv/insns/c_ebreak.h
+++ b/riscv/insns/c_ebreak.h
@@ -1,2 +1,9 @@
 require_extension('C');
-throw trap_breakpoint(STATE.v, pc);
+if (!STATE.debug_mode &&
+    ((STATE.prv == PRV_M && STATE.dcsr->ebreakm) ||
+     (STATE.prv == PRV_S && STATE.dcsr->ebreaks) ||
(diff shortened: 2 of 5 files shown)


scottj97 (approved):
Thank you for splitting this up into multiple commits; however in this case the first two commits are only dead code until the third one. So I think it's better to just squash this all into one.

aswaterman (approved):
This is cleaner than the old approach.

PR #1009, opened 2022-05-22
Reorder 'H'-extension

a4lg:
Based on current consensus made in:
<https://github.com/riscv/riscv-isa-manual/issues/837>, this commit changes the canonical order of 'H'.

Of course, making `--isa` string more permissive can be a solution to possible incompatibility caused by this.

But it'll be a TODO.


aswaterman:
Even though this PR is right, it will break some people’s workflows, so my preference is to instead relax the ordering. I guess doing so would supersede this PR. Do you want to handle that task, or should I take a stab?

PR #1013, opened 2022-05-25
Check triggers before checking alignment

### LINKED ISSUE #971: [Bug Report] Wrong exception priority during access memory
Let's take the load instruction as an example:

https://github.com/riscv-software-src/riscv-isa-sim/blob/0f15aa09005a12521c27005515e088cbd1b81629/riscv/mmu.h#L99-L125

At line 101, load will first check if it is aligned, then at line 122 it will try to access the address in the load_slow_path function.

https://github.com/riscv-software-src/riscv-isa-sim/blob/0f15aa09005a12521c27005515e088cbd1b81629/riscv/mmu.cc#L142-L162

In load_slow_path, it will first check if it is legal address at line 153, and the watch point will be checked at the end of the function.

Briefly, the order of priority is as follows: trap_load_address_misaligned > trap_load_access_fault > trap_breakpoint

However, in the specification, trap_breakpoint has a higher priority than the others:
image

We also co-simulate with rocket to check this point, rocket threw a breakpoint exception, while spike threw an error misaligned exception.
The test point is at 0x800001c0 where loading a misaligned illegal address 0x100004001:
```
3 0x00800001bc (0x7a261073)
core   0: 0x00000000800001bc (0x7a261073) csrw    tdata2, a2
3 0x0080000004 (0x34302f73) x30 0x0000000100004001
core   0: 0x00000000800001c0 (0x00062603) lw      a2, 0(a2)
core   0: exception trap_load_address_misaligned, epc 0x00000000800001c0
core   0:           tval 0x0000000100004001
core   0: 0x0000000080000004 (0x34302f73) csrr    t5, mtval
3 0x0080000008 (0x34202f73) x30 0x0000000000000003
core   0: 0x0000000080000008 (0x34202f73) csrr    t5, mcause
[error] WDATA SIM 0000000000000004, DUT 0000000000000003
[error] check board clear 30 error
[CJ] integer register Judge Failed
```
[spike-0.zip](https://github.com/riscv-software-src/riscv-isa-sim/files/8464609/spike-0.zip)

timsifive:
Fixes #971.

Running the towers benchmark now takes 11% longer.

Tested this by running `make run` in riscv-tests/isa. I'm not sure this code is correct on both endiannesses.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -143,11 +143,8 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
 {
   reg_t paddr = translate(addr, len, LOAD, xlate_flags);
 
-  if (!matched_trigger) {
-    matched_trigger = check_trigger_address_before(triggers::OPERATION_LOAD, addr);
-    if (matched_trigger)
-      throw *matched_trigger;
-  }
+  if (!matched_trigger)
+    check_trigger_address_before(triggers::OPERATION_LOAD, addr);
 
   if (unlikely(addr & (len-1))) {
     if (require_alignment) {
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -139,7 +139,7 @@ bool mmu_t::mmio_store(reg_t addr, size_t len, const uint8_t* bytes)
   return sim->mmio_store(addr, len, bytes);
 }
 
-void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate_flags)
+void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate_flags, bool require_alignment)
 {
   reg_t paddr = translate(addr, len, LOAD, xlate_flags);
 
@@ -149,8 +149,26 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
       throw *matched_trigger;
   }
 
+  if (unlikely(addr & (len-1))) {
+    if (require_alignment) {
+      load_reserved_address_misaligned(addr);
+    } else {
+      reg_t value = misaligned_load(addr, len, xlate_flags);
+      for (size_t i = 0; i < len; i++) {
+        if (target_big_endian) {
+          bytes[len - i - 1] = value & 0xff;
+        } else {
+          bytes[i] = value & 0xff;
+        }
+        value >>= 8;
+      }
+      return;
+    }
+  }
+
   if (auto host_addr = sim->addr_to_mem(paddr)) {
     memcpy(bytes, host_addr, len);
+    if (proc) READ_MEM(addr, len);
     if (tracer.interested_in_range(paddr, paddr + PGSIZE, LOAD))
       tracer.trace(paddr, len, LOAD);
     else if (xlate_flags == 0)
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -185,7 +185,8 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
   }
 }
 
-void mmu_t::store_slow_path(reg_t addr, reg_t len, const uint8_t* bytes, uint32_t xlate_flags, bool actually_store)
+void mmu_t::store_slow_path(reg_t addr, reg_t len, const uint8_t* bytes, uint32_t xlate_flags, bool actually_store,
+    bool require_alignment)
 {
   reg_t paddr = translate(addr, len, STORE, xlate_flags);
 
@@ -196,9 +197,21 @@ void mmu_t::store_slow_path(reg_t addr, reg_t len, const uint8_t* bytes, uint32_
       throw *matched_trigger;
   }
 
+  if (unlikely(addr & (len-1))) {
+    if (require_alignment) {
+      store_conditional_address_misaligned(addr);
+    } else if (actually_store) {
+      reg_t val = reg_from_bytes(len, bytes);
+      misaligned_store(addr, val, len, xlate_flags);
+    }
+    return;
+  }
+
   if (actually_store) {
     if (auto host_addr = sim->addr_to_mem(paddr)) {
       memcpy(host_addr, bytes, len);
+      if (proc)
+        WRITE_MEM(addr, reg_from_bytes(len, bytes), len);
       if (tracer.interested_in_range(paddr, paddr + PGSIZE, STORE))
         tracer.trace(paddr, len, STORE);
       else if (xlate_flags == 0)
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -143,6 +143,12 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
 {
   reg_t paddr = translate(addr, len, LOAD, xlate_flags);
 
+  if (!matched_trigger) {
+    matched_trigger = check_trigger_address_before(triggers::OPERATION_LOAD, addr);
+    if (matched_trigger)
+      throw *matched_trigger;
+  }
+
   if (auto host_addr = sim->addr_to_mem(paddr)) {
     memcpy(bytes, host_addr, len);
     if (tracer.interested_in_range(paddr, paddr + PGSIZE, LOAD))
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -150,16 +150,7 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
     if (require_alignment) {
       load_reserved_address_misaligned(addr);
     } else {
-      reg_t value = misaligned_load(addr, len, xlate_flags);
-      for (size_t i = 0; i < len; i++) {
-        if (target_big_endian) {
-          bytes[len - i - 1] = value & 0xff;
-        } else {
-          bytes[i] = value & 0xff;
-        }
-        value >>= 8;
-      }
-      return;
+      return misaligned_load(addr, len, bytes, xlate_flags);
     }
   }
 
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -160,38 +160,48 @@ public:
   proc->state.log_mem_write.push_back(std::make_tuple(addr, val, size));
 #endif
 
+  template<class T, unsigned xlate_flags>
+  inline void store_fast(reg_t addr, T val, bool actually_store=true, bool require_alignment=false) {
+    const size_t size = sizeof(T);
+    if (unlikely(addr & (size-1))) {
+      if (require_alignment)
+        store_conditional_address_misaligned(addr);
+      else
+        return misaligned_store(addr, val, size, xlate_flags, actually_store);
+    }
+    reg_t vpn = addr >> PGSHIFT;
+    if (xlate_flags == 0 && likely(tlb_store_tag[vpn % TLB_ENTRIES] == vpn)) {
+      if (actually_store) {
+        if (proc)
+          WRITE_MEM(addr, val, size);
+        *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr) = to_target(val);
+      }
+    }
+    else if (xlate_flags == 0 && unlikely(tlb_store_tag[vpn % TLB_ENTRIES] == (vpn | TLB_CHECK_TRIGGERS))) {
+      if (actually_store) {
+        if (!matched_trigger) {
+          matched_trigger = trigger_exception(triggers::OPERATION_STORE, addr, val);
+          if (matched_trigger)
+            throw *matched_trigger;
+        }
+        if (proc)
+          WRITE_MEM(addr, val, size);
+        *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr) = to_target(val);
+      }
+    }
+    else {
+      target_endian<T> target_val = to_target(val);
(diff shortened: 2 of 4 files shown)


aswaterman:
The template version is certainly an improvement over the macro version :)

I feel like we can get rid of a lot of complexity here by moving misaligned accesses into the slow path.  This is justifiable on the basis that, on average, they account for a significant minority of accesses even on systems that support them with reasonable performance.

I've made a branch off of this `trigger_priority` branch where I've mocked it up (and I've checked that the perf is roughly back to where it used to be for both gcc/x86 and clang/arm).  https://github.com/riscv-software-src/riscv-isa-sim/tree/simplify-misaligned

scottj97 (changes requested):
Commit 1890710e93cad46491782db258921928159af6cf does not compile:
```
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/triggers.cc: In member function ‘triggers::match_result_t triggers::module_t::memory_access_match(triggers::action_t*, triggers::operation_t, reg_t, reg_t)’:                                                                                                      /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/triggers.cc:218:29: error: ‘trigger’ was not declared in this scope
     match_result_t result = trigger->memory_access_match(proc, operation, address, data);
                             ^~~~~~~
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/triggers.cc:218:29: note: suggested alternative: ‘triggers’
     match_result_t result = trigger->memory_access_match(proc, operation, address, data);
                             ^~~~~~~
                             triggers
```

Also the commit message is incorrect (wrong method name)

riscv/mmu.h
```diff
@@ -477,6 +477,22 @@ class mmu_t
...
+      return NULL;
+    }
+    triggers::action_t action;
+    auto match = proc->TM.address_match(&action, operation, address);
+    if (match == triggers::MATCH_NONE)
+      return NULL;
+    if (match == triggers::MATCH_FIRE_BEFORE) {
+      throw triggers::matched_t(operation, address, 0, action);
+    }
+    return new triggers::matched_t(operation, address, 0, action);
```
scottj97: Does this ever get `delete`d?

riscv/mmu.cc
```diff
@@ -149,6 +149,23 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
...
+    } else {
+      reg_t value = misaligned_load(addr, len, xlate_flags);
+      for (size_t i = 0; i < len; i++) {
+        if (target_big_endian) {
+          bytes[len - i - 1] = value & 0xff;
+        } else {
+          bytes[i] = value & 0xff;
+        }
+        value >>= 8;
+      }
```
scottj97: Can't this just be `memcpy(bytes, &value, len)`?

riscv/mmu.cc
```diff
@@ -149,6 +149,23 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
       throw *matched_trigger;
   }
 
+  if (unlikely(addr & (len-1))) {
```
scottj97, line 149: It seems this will never match as of 894ab32c66c4046d7ac405944e3a7b1ad4651fcc? That's not good because then `git bisect` wouldn't be able to identify this new code if it had a bug.

riscv/mmu.h
```diff
@@ -94,15 +94,19 @@ class mmu_t
...
+        else return misaligned_load(addr, size, xlate_flags);
+      }
       if (proc) READ_MEM(addr, size);
       return from_target(*(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr));
     }
     if ((xlate_flags) == 0 && unlikely(tlb_load_tag[vpn % TLB_ENTRIES] == (vpn | TLB_CHECK_TRIGGERS))) {
+      if (unlikely(addr & (size-1))) {
+        if (require_alignment) load_reserved_address_misaligned(addr);
+        else return misaligned_load(addr, size, xlate_flags);
+      }
```
scottj97: This duplicated paragraph is really unfortunate, and likely to lead to subtle bugs where turning on a trigger in a page changes the behavior of misaligneds. (Because someone in the future updated one copy of this code and neglected the other.)

riscv/mmu.h
```diff
@@ -87,11 +87,18 @@ class mmu_t
   #define load_func(type, prefix, xlate_flags) \
     inline type##_t prefix##_##type(reg_t addr, bool require_alignment = false) { \
       const size_t size = sizeof(type##_t); \
+      const reg_t vpn = addr >> PGSHIFT; \
+      if ((xlate_flags) == 0 && unlikely(tlb_load_tag[vpn % TLB_ENTRIES] == (vpn | TLB_CHECK_TRIGGERS))) { \
+        if (!matched_trigger) { \
+          matched_trigger = check_trigger_address_before(triggers::OPERATION_LOAD, addr); \
+          if (matched_trigger) \
+            throw *matched_trigger; \
+        } \
```
scottj97: Both times this new method is called, the return value is immediately thrown. Why not `throw` from within the `check_trigger_address_before()` function?

scottj97 (changes requested):

timsifive:
> I feel like we can get rid of a lot of complexity here by moving misaligned accesses into the slow path. This is justifiable on the basis that, on average, they account for a significant minority of accesses even on systems that support them with reasonable performance.

Eyeballing that branch, it looks like a nice improvement.

timsifive:
> Commit [1890710](https://github.com/riscv-software-src/riscv-isa-sim/commit/1890710e93cad46491782db258921928159af6cf) does not compile:

Fixed.

riscv/mmu.cc
timsifive: I think that would do the wrong thing if the host endianness differs from the endianness of the simulated target.

riscv/mmu.h
timsifive: I wouldn't swear that this catches all of them, but https://github.com/riscv-software-src/riscv-isa-sim/blob/trigger_priority/riscv/execute.cc#L324 does delete the ones that are caught there.

riscv/mmu.h
aswaterman: The strategy I suggested on my `simplify-misaligned` branch eliminates this redundancy.

riscv/mmu.cc
aswaterman: It seems to me the problem is that this approach handles endianness 3 times (within `misaligned_load`, right here, and upon return from this function).  While correct, it's quite confusing.  We can get rid of two of them by replacing lines 156 - 164 with

    misaligned_load(addr, len, xlate_flags); // possibly take alignment trap; discard result

    for (size_t i = 0; i < len; i++)
      bytes[i] = load_uint8(addr + i);

and then the endian swap will occur on return from this function.

riscv/mmu.cc
timsifive: I'm still trying to learn exactly what you're looking for in a set of changes. IIRC it was that each stage compiles, runs, and also is tiny. Can you point me at a list of all the rules somewhere?

I've squashed this change into another one.

riscv/mmu.h
timsifive: Agreed, but I was trying to keep the code fast and readable, although I didn't benchmark.

In any case, Andrew's changes solve this nicely and I think we should make that change after this merges.

riscv/mmu.h
timsifive: Good point. Done.

riscv/mmu.cc
timsifive: Hmm... but then you won't get the exception if misaligned handling is disabled. You'd end up copying misaligned_load() into load_slow_path(), and then you have the same problem that @scottj97 pointed out about duplicating code which leads to problems down the road when one copy is changed and not another.

I agree this is a mess. I had started to change load_slow_path() to be a template and return a type, like load_fast() does. Then only load_fast() needs to worry about endianness, and the code gets a bit more consistent as well. But then I discarded it as too big a change that is unrelated.

aswaterman: You might’ve missed that I edited my post. I added back the call to misaligned_load, which we evaluate only for the side effects. It’s weird, but it’s way less of a mess.

timsifive: Ah. That works. Implemented.

riscv/mmu.cc
scottj97: We ought to have a Contributing Guide here, but we don't. :(

I think ["On Sausage Making"](https://sethrobertson.github.io/GitBestPractices/#sausage) explains it best:

> Another good reason is to ensure each commit compiles and/or passes regression tests, and represents a different easily understood concepts. The former allows git-bisect to choose any commit and have a good chance of that commit doing something useful, and the latter allows for easy change/commit/code review, understanding, archeology, and cherry-picking. When reviewing commits, for example the reviewer might see something suspicious in a commit and then have to spend time tracking down their suspicions and write them up, only to discover five commits later that the original developer subsequently found and fixed the problem, wasting the reviewer’s time (reviewing the entire patch series as a diff fixes this problem but greatly adds complexity as multiple concepts get muddled). By cleaning up patches into single, logical changes that build on one another, and which don’t individually regress (i.e., they are always moving towards some desirable common endpoint), the author is writing a chronological story not of what happened, but what should happen, with the intent that the audience (i.e., reviewers) are convinced that the change is the right thing to do. Proponents claim it is all about leaving a history others can later use to understand why the code became the way it is now, to make it less likely for others to break it.

scottj97:
Commit fb8eb2ac9d5348a707e11864db5357565e2d4347 has broken a few things.

For unaligned loads, an extra memory access is now shown in the commit log:
```
 core   0: 0x00000000800025ec (0x00913183) ld      gp, 9(sp)
-core   0: 1 0x00000000800025ec (0x00913183) x3  0xffffffffffffffff mem 0x0000000080003299 mem 0x000000008000329a mem 0x000000008000329b mem 0x000000008000329c mem 0x000000008000329d mem 0x000000008000329e mem 0x000000008000329f mem 0x00000000800032a0
+core   0: 1 0x00000000800025ec (0x00913183) x3  0xffffffffffffffff mem 0x0000000080003299 mem 0x000000008000329a mem 0x000000008000329b mem 0x000000008000329c mem 0x000000008000329d mem 0x000000008000329e mem 0x000000008000329f mem 0x00000000800032a0 mem 0x0000000080003299
```

The difference is the extra report of the first byte at `0x0000000080003299`.

The second issue is an incorrect `tval` when there's a page crossing where the second page has a problem:
```
 core   0: 0x0000000080004098 (0x0005b203) ld      tp, 0(a1)
 core   0: exception trap_load_access_fault, epc 0x0000000080004098
-core   0:           tval 0x0000000080005000
+core   0:           tval 0x0000000080004ffd
```

The correct `tval` here is the second page at 0x80005000 (which has a PMP issue) but after this commit it's incorrectly reporting it on the first page at 0x80004ffd.

I apologize again for the proprietary tests that I am unable to share.

scottj97:
Commit 6043dab46b4f1c4f7fd2d9fe4c03604e6b414066 causes more problems with the commit log; unaligned accesses now have each byte reported twice.

riscv/mmu.h
scottj97: Now that you've made me aware of [this rather crazy system](https://github.com/riscv-software-src/riscv-isa-sim/blob/6043dab46b4f1c4f7fd2d9fe4c03604e6b414066/riscv/execute.cc#L314-L326), I'm no longer sure this is a good idea. Don't we need to set `mmu->matched_trigger` sometimes before throwing?

timsifive: Good point. I missed this because it's only relevant when timing=after, which isn't covered by the spike tests and OpenOCD doesn't use it either.

On my todo list is to make both timing=before and timing=after work correctly in spike. When I get around to that I'll also revisit this logic, which seems suspect but currently harmless.

timsifive:
> For unaligned loads, an extra memory access is now shown in the commit log:

I can't easily explain that. It looks like there aren't *any* tests for unaligned accesses in riscv-tests. Is that right? I guess I get to write some...

timsifive:
> For unaligned loads, an extra memory access is now shown in the commit log:

I cannot reproduce that. I get:
```
core   0: 0x000000008000057c (0x0010b703) ld      a4, 1(ra)
core   0: 0 0x000000008000057c (0x0010b703) x14 0x0000ff00ff00ff00 mem 0x0000000080002001 mem 0x0000000080002002 mem 0x0000000080002003 mem 0x0000000080002004 mem 0x0000000080002005 mem 0x0000000080002006 mem 0x0000000080002007 mem 0x0000000080002008
```

Can you help me reproduce this problem?

timsifive:
> I cannot reproduce that. I get:

I can reproduce it. It only happens when the access goes through the slow path.

timsifive:
@scottj97 I think I have this all worked out again. Can you run it through your own tests?

scottj97 (changes requested):
bb65ba4add396fcb8981e08d676dc0c715c08632 breaks some memory tracing -- the `mem` part of the commit log disappears. (Seems to occur only on the first load of each sim?) Also this breaks `tval` like reported earlier, when there's a PMP access fault on the second half of a page crossing.

60b0caa722782dd8f0edd983096b581fb031e109 breaks some store tracing -- again the `mem` part of the commit log disappears -- perhaps only for the first store of the sim? Only for certain tests, and I can't see any pattern.

4ea4ec0eac6b3df76a0502b339c05a44bfaed02d causes bogus load access faults to occur on previously-successful loads on page crossings. It also changes `tval` on some page-crossing accesses, though I can't immediately tell if it's fixing the earlier problem, or making things worse.

aswaterman:
Can we close this since https://github.com/riscv-software-src/riscv-isa-sim/pull/1105 was merged, or does it address an additional bug?

timsifive:
Addressed by #1105.

PR #1014, opened 2022-05-26
Fix RV32 hgatp write mask computation

aswaterman:
clang with `-Wall` happened to catch this.


scottj97 (approved):
Obviously hypervisor on RV32 is not well-tested.

aswaterman:
Indeed.

PR #1018, opened 2022-06-01
Remove the now-unused PC_SERIALIZE_WFI

kwalker27:
When WFI was changed to throw a C++ exception, the special-npc
signaling became obsolete.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -315,8 +315,7 @@ private:
 class wait_for_interrupt_t {};
 
 #define wfi() \
-  do { set_pc_and_serialize(npc); \
-       npc = PC_SERIALIZE_WFI; \
+  do { STATE.pc = npc; \
        throw wait_for_interrupt_t(); \
      } while (0)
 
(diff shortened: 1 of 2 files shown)


kwalker27:
Not sure if there's a reason this code is still there, but as far as I can tell it can safely go away. I didn't see a reason to keep `set_pc_and_serialize` in favor of a simple STATE.pc update.

aswaterman:
@kwalker27 Right, there's no need for `set_pc_and_serialize` because (a) `npc` is dead past the `throw` and (b) enforcing alignment isn't necessary, since the instruction after WFI is guaranteed to be aligned if the WFI is aligned.

aswaterman:
However, it seems we only do the direct `STATE.pc` assignment in the one place, so the code is cleaner if we keep the `set_pc_and_serialize` invocation.  I'll manually merge the PR with that modification.  Thank you!

aswaterman:
Manually merged via 7e9da9966856743fb17c42a0a2c2c915e3a3e2b9

Issue #1022, opened 2022-06-03
48bit illegal instruction behavior


neelgala:
I have a test which checks for illegal instruction behavior which is running on rv32imc config of spike (`spike --log ref.dump --log-commits --priv=mu --isa=rv32imc ref.elf`). The following is the test-case of interest:

```
core   0: 0x80000048 (0x430000003f) unknown
core   0: exception trap_illegal_instruction, epc 0x80000048
core   0:           tval 0x00000000
```

I wanted to understand why tval is `0x00000000` instead of `0x0000003f`?

My understanding so far is that the instruction size and csr sizes are both uint64_t throughout. And for an rv32-config system it should be reading the lower 32-bit of the csr . What am I missing ?

neelgala:
going back to spec I see the following which indicates the mtval should be `0x000003f`

> If mtval is written with a nonzero value when an illegal-instruction exception occurs, then mtval will contain the shortest of:
> • the actual faulting instruction
> • the first ILEN bits of the faulting instruction
> • the first MXLEN bits of the faulting instruction

But then there is an immediate commentary as below which could indicate setting the value of mtval to 0: 

> A value of zero in mtval signifies either that the feature is not supported, or an illegal zero instruction was fetched.

Which of the above scenarios should be applied when ??

aswaterman:
I think it should be legal to write 0 some of the time, but it still looks like a Spike bug.

aswaterman:
@scottj97 I need some C++ help.  I tracked the problem down and observe the following makes the bug disappear.

To me, this reeks of undefined behavior somewhere.  What is it about explicitly creating the object then throwing it, vs. constructing in the throw statement, that makes a difference?  https://github.com/riscv-software-src/riscv-isa-sim/commit/32e925d6e12fbdd2d101c002c993131c23575338

aswaterman:
Perhaps the issue is that the object’s lifetime only lasts til the end of the statement? If so, I’ll need to audit all throw statements. I guess I really don’t know C++.

neelgala:
> A value of zero in mtval signifies either that the feature is not supported, or an illegal zero instruction was fetched.

Sorry for treading on this a little more - but are I find the two statements `the first ILEN bits of the faulting instruction` and `A value of zero in mtval signifies either that the feature is not supported` conflicting. Are both behaviors considered legal ? Is this an implementation-defined choice ?

[...]

PR #1023, opened 2022-06-03
Remove nonstandard length encoding

aswaterman:
This was an artifact of an old P-extension draft that erroneously allocated a reserved major opcode.  The newer draft uses a different opcode in the 32-bit encoding space, so this hack is no longer needed.


scottj97:
Does this have anything to do with #1022?

aswaterman:
No, I just happened to spot it while looking into #1022.

scottj97 (approved):

PR #1024, opened 2022-06-04
fix cmo support

liweiwei90:
This patchset will [use different cache block size for zicbom and zicboz operations and update the diasam for cbo.* instructions.

### DIFF
--- a/disasm/disasm.cc
+++ b/disasm/disasm.cc
@@ -2037,13 +2037,13 @@ void disassembler_t::add_instructions(const isa_parser_t* isa)
   }
 
   if (isa->extension_enabled(EXT_ZICBOM)) {
-    DISASM_INSN("cbo.clean", cbo_clean, 0, {&xrs1});
-    DISASM_INSN("cbo.flush", cbo_flush, 0, {&xrs1});
-    DISASM_INSN("cbo.inval", cbo_inval, 0, {&xrs1});
+    DISASM_INSN("cbo.clean", cbo_clean, 0, {&base_only_address});
+    DISASM_INSN("cbo.flush", cbo_flush, 0, {&base_only_address});
+    DISASM_INSN("cbo.inval", cbo_inval, 0, {&base_only_address});
   }
 
   if (isa->extension_enabled(EXT_ZICBOZ)) {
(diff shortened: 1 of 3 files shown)


aswaterman (changes requested):
These should be separate PRs.

The disassembler change looks fine.

I don't support the second change.  The RISC-V profiles are going to use a consistent cache-block size for the two extensions, so I doubt anyone will leverage the theoretical possibility that they might differ.  Just because someone can do something in RISC-V doesn't mean we need to model it in Spike.

liweiwei90:
> 



> I don't support the second change. The RISC-V profiles are going to use a consistent cache-block size for the two extensions, so I doubt anyone will leverage the theoretical possibility that they might differ. Just because someone can do something in RISC-V doesn't mean we need to model it in Spike.

Sorry. I don't know how the RISC-V profiles are going to address this problem (@dkruckemyer-ventana). This is suggested in https://github.com/riscv-software-src/riscv-config/pull/79 by @pdonahue-ventana . And I remember this was also suggested in the QEMU PR for cmo support several month ago.

aswaterman:
Paul’s not wrong, but it’s only a theoretical concern AFAICS. We can address it if it ever becomes relevant in practice.

Can you split the disassembler changes into a separate PR?

pdonahue-ventana:
> Paul’s not wrong, but it’s only a theoretical concern AFAICS. We can address it if it ever becomes relevant in practice.

That's fine with me.  There are probably lots of other things that are theoretical concerns that we haven't even identified.  I just wanted to point this one out since I did happen to notice it.

liweiwei90:
OK. I have sent a new PR(https://github.com/riscv-software-src/riscv-isa-sim/pull/1026) for it.

> Paul’s not wrong, but it’s only a theoretical concern AFAICS. We can address it if it ever becomes relevant in practice.
> 
> Can you split the disassembler changes into a separate PR?

aswaterman:
Thanks.  I've merged #1026.

I'm going to close this for now.  In the unlikely event that realistic systems with divergent block sizes for Zicbom and Zicboz emerge, we can revisit the rest.

PR #1027, opened 2022-06-07
Fix #1022 by removing need to mask insn_t bits

### LINKED ISSUE #1022: 48bit illegal instruction behavior

aswaterman:
#1022 is ultimately caused by undefined behavior in left-shifting by >= the datatype's width.  This only manifests for 8-byte instructions, which is why it wasn't noticed until now.

Directly fixing the bug would be easy, but it seemed better to me to attack the need to shift & mask, by arranging for the MSBs of `insn_t::b` to be zero.

As a side effect, this also gets rid of a few dependences on the right-shifting of negative values.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -79,7 +79,7 @@ class insn_t
 public:
   insn_t() = default;
   insn_t(insn_bits_t bits) : b(bits) {}
-  insn_bits_t bits() { return b & ~((UINT64_MAX) << (length() * 8)); }
+  insn_bits_t bits() { return b; }
   int length() { return insn_length(b); }
   int64_t i_imm() { return xs(20, 12); }
   int64_t shamt() { return x(20, 6); }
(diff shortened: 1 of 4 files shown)


scottj97 (approved):

aswaterman:
@scottj97 Sorry if this presents any new bugs, but I'll try to be prompt about fixing them if it does.

PR #1035, opened 2022-07-01
Add support for smstateen 1.0.0

liweiwei90:
Add support for smstateen 1.0.0:
- remove multi blink lines in riscv/
- add support for smstateen related CSRs
- add smstateen check for fcsr, s/henvcfg


scottj97:
`encoding.h` is generated by [riscv-opcodes](https://github.com/riscv/riscv-opcodes) package; it should not be hand-edited.

If this was indeed generated from `riscv-opcodes` please record which commit hash in the commit message here.

riscv/csrs.cc
```diff
@@ -1240,3 +1240,110 @@ bool vxsat_csr_t::unlogged_write(const reg_t val) noexcept {
...
+  mask(mask) {
+}
+
+reg_t stateenh_csr_t::read() const noexcept {
+  return (stateen->read() >> 32) & mask;
+}
+
+bool stateenh_csr_t::unlogged_write(const reg_t val) noexcept {
+  return stateen->unlogged_write((stateen->written_value() & ~(mask << 32)) | ((val & mask) << 32));
+}
```
scottj97: This is unfortunate that each of the upper 32-bit *h regs have to have all this boilerplate. Can you find a way to modify & rename `mstatush_csr_t` so that you can use that class for each of these *h regs?

riscv/csrs.cc
```diff
@@ -1240,3 +1240,110 @@ bool vxsat_csr_t::unlogged_write(const reg_t val) noexcept {
...
+reg_t hstateen_csr_t::read() const noexcept {
+  // For every bit in an mstateen CSR that is zero (whether read-only zero or set to zero),
+  // the same bit appears as read-only zero in the matching hstateen and sstateen CSRs
+  return stateen_csr_t::read() & state->mstateen[index]->read();
+}
+
+bool hstateen_csr_t::unlogged_write(const reg_t val) noexcept {
+  // For every bit in an mstateen CSR that is zero (whether read-only zero or set to zero),
+  // the same bit appears as read-only zero in the matching hstateen and sstateen CSRs
+  return stateen_csr_t:: unlogged_write(val & state->mstateen[index]->read());
```
scottj97: This will actually write to those "read-only zero" bits and set them to 0. For example,
```
  li x1,-1
  csrw mstateen0,x1  // write all 1s
  csrw hstateen0,x1  // write all 1s
  csrw mstateen0,x0  // write all 0s
  csrr x2,hstateen0  // sees bits 0
  csrw hstateen,x1  // try to write all 1s
  csrr x3,hstateen0  // sees bits 0 still
  csrw mstateen0,x1  // re-enable
  csrr x4,hstateen0  // sees 0 bits
```

Should x4 see the 1 bits previously written into `hstateen0`? It will if you remove the "try to write all 1s" line.

riscv/csrs.cc
```diff
@@ -1240,3 +1240,110 @@ bool vxsat_csr_t::unlogged_write(const reg_t val) noexcept {
...
+// implement class hstateenh_csr_t
+hstateenh_csr_t::hstateenh_csr_t(processor_t* const proc, const reg_t addr, const reg_t mask,
+                                 stateen_csr_t_p stateen, uint8_t index):
+  stateenh_csr_t(proc, addr, mask, stateen),
+  index(index) {
+}
+
+void hstateenh_csr_t::verify_permissions(insn_t insn, bool write) const {
+  if ((state->prv < PRV_M) && !(state->mstateen[index]->read() & MSTATEEN_HSTATEEN))
+    throw trap_illegal_instruction(insn.bits());
```
scottj97, line 1271: Won't this get checked when `stateenh_csr_t::verify_permissions()` calls `stateen->verify_permissions(insn, write);` which will call `hstateen_csr_t::verify_permissions()`?

In other words, it seems this entire method is unnecessary.

riscv/csrs.cc
```diff
@@ -1240,3 +1240,110 @@ bool vxsat_csr_t::unlogged_write(const reg_t val) noexcept {
...
+    return hstateen_csr_t::read();
+}
+
+bool sstateen_csr_t::unlogged_write(const reg_t val) noexcept {
+  // For every bit in an mstateen CSR that is zero (whether read-only zero or set to zero),
+  // the same bit appears as read-only zero in the matching hstateen and sstateen CSRs
+  // For every bit in an hstateen CSR that is zero (whether read-only zero or set to zero),
+  // the same bit appears as read-only zero in sstateen when accessed in VS-mode
+  if (state->v && state->prv == PRV_S)
+    return hstateen_csr_t:: unlogged_write(val & state->hstateen[index]->read());
```
scottj97: This could write bits to 0 in `sstateen` that will affect HS-mode too, which seems wrong.

riscv/csrs.cc
```diff
@@ -1240,3 +1240,110 @@ bool vxsat_csr_t::unlogged_write(const reg_t val) noexcept {
...
+  else
+    return hstateen_csr_t:: unlogged_write(val);
+}
+
+void sstateen_csr_t::verify_permissions(insn_t insn, bool write) const {
+  if (state->v && !(state->hstateen[index]->read() & MSTATEEN_HSTATEEN))
```
scottj97: Should this be **H**STATEEN_**S**STATEEN?

riscv/processor.cc
```diff
@@ -393,6 +393,44 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   const reg_t henvcfg_init = (proc->extension_enabled(EXT_SVPBMT) ? HENVCFG_PBMTE : 0);
   csrmap[CSR_HENVCFG] = henvcfg = std::make_shared<henvcfg_csr_t>(proc, CSR_HENVCFG, henvcfg_mask, henvcfg_init, menvcfg);
 
+  const reg_t mstateen0_mask = (proc->extension_enabled(EXT_ZFINX) ? MSTATEEN0_FCSR : 0) |
+                               MSTATEEN0_CS | MSTATEEN0_HENVCFG | MSTATEEN_HSTATEEN;
```
scottj97: All these masks should be calculated in the class constructors, such as `mstatush_csr_t::mstatush_csr_t()` does.

riscv/processor.cc
```diff
@@ -393,6 +393,44 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
+    const reg_t mstateen0h_mask = MSTATEEN0H_HENVCFG | MSTATEENH_HSTATEEN;
+    csrmap[CSR_MSTATEEN0H] = std::make_shared<stateenh_csr_t>(proc, CSR_MSTATEEN0H, mstateen0h_mask, mstateen[0]);
+  }
+
+  const reg_t hstateen0_mask = (proc->extension_enabled(EXT_ZFINX) ? HSTATEEN0_FCSR : 0) |
+                               HSTATEEN0_CS | HSTATEEN0_SENVCFG | HSTATEEN_SSTATEEN;
+  csrmap[CSR_HSTATEEN0] = hstateen[0] = std::make_shared<hstateen_csr_t>(proc, CSR_MSTATEEN0, hstateen0_mask, 0, 0);
+  if (xlen == 32) {
+    const reg_t hstateen0h_mask = HSTATEEN0H_SENVCFG | HSTATEENH_SSTATEEN;
+    csrmap[CSR_MSTATEEN0H] = std::make_shared<hstateenh_csr_t>(proc, CSR_HSTATEEN0H, hstateen0h_mask, hstateen[0], 0);
```
scottj97: Shouldn't this be csrmap[CSR_**H**STATEEN0H]?

riscv/processor.cc
```diff
@@ -393,6 +393,44 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
+  }
+
+  const reg_t sstateen0_mask = (proc->extension_enabled(EXT_ZFINX) ? SSTATEEN0_FCSR : 0) |
+                               SSTATEEN0_CS;
+  csrmap[CSR_SSTATEEN0] = sstateen[0] = std::make_shared<sstateen_csr_t>(proc, CSR_SSTATEEN0, sstateen0_mask, 0, 0);
+
+  for (int i = 1; i < 4; i++) {
```
scottj97: Once the masks are moved to the constructors, and calculated using the index, it seems this loop could cover 0-3 instead of just 1-3?

riscv/csrs.cc
```diff
@@ -1347,3 +1347,54 @@ void sstateen_csr_t::verify_permissions(insn_t insn, bool write) const {
...
   hstateen_csr_t::verify_permissions(insn, write);
 }
+
+// implement class fcsr_csr_t
+fcsr_csr_t::fcsr_csr_t(processor_t* const proc, const reg_t addr, csr_t_p upper_csr, csr_t_p lower_csr, const unsigned upper_lsb):
+  composite_csr_t(proc, addr, upper_csr, lower_csr, upper_lsb) {
+}
+
+void fcsr_csr_t::verify_permissions(insn_t insn, bool write) const {
+  if (proc->extension_enabled(EXT_SMSTATEEN) && proc->extension_enabled(EXT_ZFINX)) {
```
scottj97, line 1317: Commit 402fccb0747334fabac6ad152192765543e27f7b needs to be broken up because it's not clear where the behavioral changes are. First, implement `fcsr_csr_t`, `senvcfg_t`, and `henvcfg_t` as blank placeholder classes with no behavior changes. Then add the Smstateen changes as a separate commit so it's clear what's new.

riscv/processor.cc
```diff
@@ -393,6 +393,44 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
+    }
+
+    const reg_t hstateen_mask = HSTATEEN_SSTATEEN;
+    csrmap[CSR_HSTATEEN0 + i] = hstateen[i] = std::make_shared<hstateen_csr_t>(proc, CSR_HSTATEEN0 + i, hstateen_mask, 0, i);
+    if (xlen == 32) {
+      const reg_t hstateenh_mask =  HSTATEENH_SSTATEEN;
+      csrmap[CSR_HSTATEEN0H + i] = std::make_shared<hstateenh_csr_t>(proc, CSR_HSTATEEN0H + i, hstateenh_mask, hstateen[i], i);
+    }
+
+    csrmap[CSR_SSTATEEN0 + i] = sstateen[i] = std::make_shared<sstateen_csr_t>(proc, CSR_HSTATEEN0 + i, hstateen_mask, 0, i);
```
scottj97: Remove trailing whitespace

scottj97 (changes requested):

liweiwei90:
> `encoding.h` is generated by [riscv-opcodes](https://github.com/riscv/riscv-opcodes) package; it should not be hand-edited.
> 
> If this was indeed generated from `riscv-opcodes` please record which commit hash in the commit message here.

OK, I have sent a PR(https://github.com/riscv/riscv-opcodes/pull/134) to riscv-opcodes to support this.

riscv/csrs.cc
liweiwei90: OK.

riscv/csrs.cc
liweiwei90: I have a similar question about this, and sent a issue(https://github.com/riscv/riscv-state-enable/issues/9). I think this will also be unspecified.

riscv/csrs.cc
liweiwei90: Yeah. I'll delete this.

riscv/csrs.cc
liweiwei90: As comments, these bits in sstateen will be "read-only zero", Similar to above question, this values for 'read-only zero' will also be unspecified.

riscv/csrs.cc
liweiwei90: Yeah. I'll fix this.

riscv/processor.cc
liweiwei90: Why? I want to use a general class for all mstateen csrs. If we caculated the masks in the class constructor, we should use different class for different mstateen since they have different mask.

riscv/processor.cc
liweiwei90: Yeah, I'll fix this.

riscv/processor.cc
liweiwei90: stateen 1~3 have different mask from 0.

riscv/csrs.cc
liweiwei90: OK.

riscv/processor.cc
liweiwei90: OK. I'll fix this.

scottj97:
In the first commit, s/blink/blank/

riscv/csrs.cc
```diff
@@ -1303,3 +1244,127 @@ bool vxsat_csr_t::unlogged_write(const reg_t val) noexcept {
...
+reg_t hstateen_csr_t::read() const noexcept {
+  // For every bit in an mstateen CSR that is zero (whether read-only zero or set to zero),
+  // the same bit appears as read-only zero in the matching hstateen and sstateen CSRs
+  return stateen_csr_t::read() & state->mstateen[index]->read();
+}
+
+bool hstateen_csr_t::unlogged_write(const reg_t val) noexcept {
+  // For every bit in an mstateen CSR that is zero (whether read-only zero or set to zero),
+  // the same bit appears as read-only zero in the matching hstateen and sstateen CSRs
+  return stateen_csr_t:: unlogged_write(val & state->mstateen[index]->read());
```
scottj97: Remove space after :: (and 5 other places in this file)

riscv/processor.cc
scottj97: Because you have 99% of the CSR details in the class, where they belong, and then 1% of the details over here a mile away. Those details are best kept in the class.

And because you have this big paragraph of code duplicated here, once for index 0 and another copy for 1-3. That's not [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

You'd have to use a new class because one integer changes by index? :raised_eyebrow:
```
const reg_t mask = (index == 0) ? mask_for_mstatus0 : mask_for_mstatus1_2_3;
```

riscv/processor.cc
scottj97: One integer difference is a ridiculously bad reason for copy/pasting 15 lines of code.

riscv/csrs.cc
```diff
@@ -1317,3 +1317,16 @@ void sstateen_csr_t::verify_permissions(insn_t insn, bool write) const {
 
   hstateen_csr_t::verify_permissions(insn, write);
 }
+
+// implement class fcsr_csr_t
+fcsr_csr_t::fcsr_csr_t(processor_t* const proc, const reg_t addr, csr_t_p upper_csr, csr_t_p lower_csr, const unsigned upper_lsb):
+  composite_csr_t(proc, addr, upper_csr, lower_csr, upper_lsb) {
+}
+
+
```
scottj97, line 1332: An earlier commit removed multiple blank lines, and this adds some back!

scottj97 (changes requested):
It would be best to get that `riscv-opcodes` change merged first, just in case it needs rebasing and the commit hash changes (in which case the hash here in encoding.h would be stale).

aswaterman:
I've merged the corresponding riscv-opcodes PR.

riscv/processor.cc
liweiwei90: Yeah, It's OK to calculate reg mask this way to reuse the code. My question is whether is necessary to calculate mask in the constructor just like mstatush_csr_t::mstatush_csr_t() does.

riscv/csrs.cc
liweiwei90: OK

riscv/csrs.cc
liweiwei90: Sorry for this mistake. I'll fix it.

liweiwei90:
> I've merged the corresponding riscv-opcodes PR.

Ok. Thanks a lot, I'll update the encoding.h.

riscv/csrs.cc
```diff
@@ -1323,8 +1323,48 @@ fcsr_csr_t::fcsr_csr_t(processor_t* const proc, const reg_t addr, csr_t_p upper_
...
 
+void fcsr_csr_t::verify_permissions(insn_t insn, bool write) const {
+  if (proc->extension_enabled(EXT_SMSTATEEN) && proc->extension_enabled(EXT_ZFINX)) {
+    if ((state->prv < PRV_M) && !(state->mstateen[0]->read() & MSTATEEN0_FCSR))
+      throw trap_illegal_instruction(insn.bits());
+
+    if (state->v && !(state->hstateen[0]->read() & HSTATEEN0_FCSR))
+      throw trap_virtual_instruction(insn.bits());
+
+    if ((proc->extension_enabled('S') && state->prv < PRV_S) && !(state->sstateen[0]->read() & HSTATEEN0_FCSR)) {
```
scottj97: Should this be `SSTATEEN0_FCSR`? (I suppose it doesn't matter since its value is the same as `HSTATEEN0_FCSR` but it looks suspicious like this.)

riscv/processor.cc
```diff
@@ -397,6 +397,33 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
+      const reg_t mstateenh_mask = i == 0 ? mstateen0h_mask : MSTATEENH_HSTATEEN;
+      csrmap[CSR_MSTATEEN0H + i] = std::make_shared<rv32_high_csr_t>(proc, CSR_MSTATEEN0H + i, mstateenh_mask, mstateen[i]);
+    }
+
+    const reg_t hstateen_mask = i == 0 ? hstateen0_mask : HSTATEEN_SSTATEEN;
+    csrmap[CSR_HSTATEEN0 + i] = hstateen[i] = std::make_shared<hstateen_csr_t>(proc, CSR_HSTATEEN0 + i, hstateen_mask, 0, i);
+    if (xlen == 32) {
+      const reg_t hstateenh_mask =  i == 0 ? hstateen0h_mask : HSTATEENH_SSTATEEN;
+      csrmap[CSR_HSTATEEN0H + i] = std::make_shared<rv32_high_csr_t>(proc, CSR_HSTATEEN0H + i, hstateenh_mask, hstateen[i]);
+    }
```
scottj97: If hypervisor extension is not implemented, these CSRs should not exist.

This also means that `hstateen_csr_t::verify_permissions()` needs to dynamically check for the H extension, since `misa.H` is writable.

Edit: it's sufficient to do this check only inside `hstateen_csr_t::verify_permissions()`; no need to skip their creation here.

riscv/processor.cc
```diff
@@ -397,6 +397,33 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   const reg_t henvcfg_init = (proc->extension_enabled(EXT_SVPBMT) ? HENVCFG_PBMTE : 0);
   csrmap[CSR_HENVCFG] = henvcfg = std::make_shared<henvcfg_csr_t>(proc, CSR_HENVCFG, henvcfg_mask, henvcfg_init, menvcfg);
 
+  const reg_t mstateen0_mask = (proc->extension_enabled(EXT_ZFINX) ? MSTATEEN0_FCSR : 0) |
+                               MSTATEEN0_CS | MSTATEEN0_HENVCFG | MSTATEEN_HSTATEEN;
+  const reg_t mstateen0h_mask = MSTATEEN0H_HENVCFG | MSTATEENH_HSTATEEN;
```
scottj97: Since the high32 CSRs are reading & writing through the full 64-bit CSR, do they really need their own mask? Can you make the mask -1 for these? (`mstatush` might be a special case because the fields are slightly different than RV64.)

riscv/csrs.cc
scottj97: I believe Smstateen requires S-mode, yes? If so the `extension_enabled()` check here is unnecessary.

riscv/processor.cc
```diff
@@ -397,6 +397,33 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
+  const reg_t hstateen0_mask = (proc->extension_enabled(EXT_ZFINX) ? HSTATEEN0_FCSR : 0) |
+                               HSTATEEN0_CS | HSTATEEN0_SENVCFG | HSTATEEN_SSTATEEN;
+  const reg_t hstateen0h_mask = HSTATEEN0H_SENVCFG | HSTATEENH_SSTATEEN;
+  const reg_t sstateen0_mask = (proc->extension_enabled(EXT_ZFINX) ? SSTATEEN0_FCSR : 0) |
+                               SSTATEEN0_CS;
+  for (int i = 0; i < 4; i++) {
```
scottj97: If Smstateen is not configured, none of these CSRs should exist.

riscv/processor.cc
scottj97: It's not strictly "necessary" but if it's not a big burden then it's better to put all the details of the CSR into one place (its class) instead of spreading them out. But see my other comment about calculating masks; it might make more sense to leave this here.

riscv/processor.cc
```diff
@@ -397,6 +397,33 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
   csrmap[CSR_HENVCFG] = henvcfg = std::make_shared<henvcfg_csr_t>(proc, CSR_HENVCFG, henvcfg_mask, henvcfg_init, menvcfg);
 
+  const reg_t mstateen0_mask = (proc->extension_enabled(EXT_ZFINX) ? MSTATEEN0_FCSR : 0) |
+                               MSTATEEN0_CS | MSTATEEN0_HENVCFG | MSTATEEN_HSTATEEN;
+  const reg_t mstateen0h_mask = MSTATEEN0H_HENVCFG | MSTATEENH_HSTATEEN;
+  const reg_t hstateen0_mask = (proc->extension_enabled(EXT_ZFINX) ? HSTATEEN0_FCSR : 0) |
+                               HSTATEEN0_CS | HSTATEEN0_SENVCFG | HSTATEEN_SSTATEEN;
+  const reg_t hstateen0h_mask = HSTATEEN0H_SENVCFG | HSTATEENH_SSTATEEN;
+  const reg_t sstateen0_mask = (proc->extension_enabled(EXT_ZFINX) ? SSTATEEN0_FCSR : 0) |
+                               SSTATEEN0_CS;
```
scottj97: Would it make sense to build the masks upon each other? For example:
```
const reg_t sstateen0_mask = (proc->extension_enabled(EXT_ZFINX) ? SSTATEEN0_FCSR : 0) |
                                 SSTATEEN0_CS;
const reg_t hstateen0_mask = sstateen0_mask | HSTATEEN0_SENVCFG | HSTATEEN_SSTATEEN;
const reg_t mstateen0_mask = hstateen0_mask;
```

scottj97 (changes requested):
I have tested this only so far as to see that it doesn't break any other features. I have not tested the Smstateen functionality.

@liweiwei90 can you tell us how you have tested this code?

riscv/csrs.cc
liweiwei90: Yeah. I'll fix this.

riscv/processor.cc
liweiwei90: > This also means that `hstateen_csr_t::verify_permissions()` needs to dynamically check for the H extension, since `misa.H` is writable.

OK. I'll add this.



> Edit: it's sufficient to do this check only inside `hstateen_csr_t::verify_permissions()`; no need to skip their creation here.

It's OK. However, I found the check for xlen is added here for other similar csrs.

riscv/csrs.cc
liweiwei90: Sorry. I didn't find this in the spec. I'll check this later.

riscv/processor.cc
liweiwei90: I think so.

riscv/processor.cc
liweiwei90: Yeah. It's OK currently. However I don't know whether the fields in different csrs can be different, such as SSTATEEN0_FCSR,  HSTATEEN0_FCSR, MSTATEEN0_FCSR bits.

liweiwei90:
> I have tested this only so far as to see that it doesn't break any other features. I have not tested the Smstateen functionality.
> 
> @liweiwei90 can you tell us how you have tested this code?

Sorry. I didn't test the code systematically.  Partial of the code was tested in an ugly way, such as for the check for fcsr, I set the bit in mstateen/sstateen to different values in reset to test the access from U mode. All of the functions are privilege-related. I didn't know a good way to test them. Any suggestions?

riscv/processor.cc
liweiwei90: Yeah. It's OK currently.

riscv/csrs.cc
liweiwei90: I think it doesn't require S-mode, since in chapter 1 of the spec, it have a description for sstateen csrs: "If supervisor mode is implemented, another four CSRs are defined at supervisor level"

riscv/processor.cc
liweiwei90: > If hypervisor extension is not implemented, these CSRs should not exist.
> 
> This also means that `hstateen_csr_t::verify_permissions()` needs to dynamically check for the H extension, since `misa.H` is writable.

This have been checked in the base class csr_t, so I think it's unnecessary here

```
void csr_t::verify_permissions(insn_t insn, bool write) const {
  // Check permissions. Raise virtual-instruction exception if V=1,
  // privileges are insufficient, and the CSR belongs to supervisor or
  // hypervisor. Raise illegal-instruction exception otherwise.
  unsigned priv = state->prv == PRV_S && !state->v ? PRV_HS : state->prv;

  if ((csr_priv == PRV_S && !proc->extension_enabled('S')) ||
      (csr_priv == PRV_HS && !proc->extension_enabled('H')))
```

scottj97: Yes, you're right, since the CSR's address indicates privilege, this is already handled.

riscv/processor.cc
scottj97: My concern is that the `mstaten0_mask` and `mstateen0h_mask` could easily get out of sync as new bits are added to the spec. As I understand it, RV32's `mstateen0h` will always be identical to the upper 32 bits of RV64's `mstateen`, so by using -1 as the mask, it will avoid any possibility of the two masks getting unsynchronized in the future.

riscv/processor.cc
scottj97: I see there is a check in `stateen_csr_t::verify_permissions()` that will trap in this case. But that dynamic check isn't necessary since Smstateen can't be dynamically disabled. Instead the check could be here in `processor_t::reset()` and skip the creation of the CSRs in the first place.

riscv/processor.cc
scottj97: If I understand the spec correctly, the bit positions are deliberately aligned in the three CSRs, and will always be so.

scottj97:
> Sorry. I didn't test the code systematically. Partial of the code was tested in an ugly way, such as for the check for fcsr, I set the bit in mstateen/sstateen to different values in reset to test the access from U mode. All of the functions are privilege-related. I didn't know a good way to test them. Any suggestions?

All of these features need architectural tests, which are out of scope for Spike contributors, but such tests are sorely lacking in the open-source world. :(

riscv/csrs.cc
```diff
@@ -975,6 +975,8 @@ bool counter_proxy_csr_t::myenable(csr_t_p counteren) const noexcept {
 }
 
 void counter_proxy_csr_t::verify_permissions(insn_t insn, bool write) const {
+  proxy_csr_t::verify_permissions(insn, write);
```
scottj97: What is the effect of this? Was there a bug here?

If this is not related to Smstateen, it does not belong in this PR.

riscv/processor.cc
liweiwei90: OK. I'll change the check to reset later.

riscv/csrs.cc
liweiwei90: I think it's a bug here. I'll send another PR later.

scottj97 (approved):
All good now. Thank you for the contribution and for your patience with the review process.

aswaterman (approved):
🎉

PR #1036, opened 2022-07-04
add support for sscofpmf extension v0.5.2

liweiwei90:
This patchset add support for sscofpmf extension v0.5.2.  since spike doesn't truly support counting of hardware performance events, only csr related read/write functions is supported currently.


liweiwei90:
reuse part of commits from https://github.com/riscv-software-src/riscv-isa-sim/pull/1035

YenHaoChen:
I am interested in the sscofpmf extension. How is this PR going? Will this PR be merged recently?

liweiwei90:
NO, I think. The review has not started currently. Any comments are welcome.

> I am interested in the sscofpmf extension. How is this PR going? Will this PR be merged recently?

liweiwei90:
@scottj97 @aswaterman Hi, are there any comments for this PR? Or we should I do next?

riscv/csrs.cc
```diff
@@ -1356,3 +1359,34 @@ void henvcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
...
+  /* when mcounteren bit X and hcounteren bit X are both set, and otherwise reads as zero. */
+  val &= state->mcounteren->read();
+  if (state->v)
+    val &= state->hcounteren->read();
+  return val;
+}
+
+bool scountovf_csr_t::unlogged_write(const reg_t val) noexcept {
+  /* this function is unused */
+  return true;
```
scottj97: Even though this is never called, it would be better to `return false;` to document that no write has occurred.

riscv/processor.cc
```diff
@@ -237,15 +240,23 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
         auto counterh = std::make_shared<counter_proxy_csr_t>(proc, which_counterh, mcounterh);
         csrmap[which_counterh] = counterh;
       }
+      if (proc->extension_enabled_const(EXT_SSCOFPMF)) {
+        const reg_t meventh_mask = MHPMEVENT_VUINH | MHPMEVENT_VSINH | MHPMEVENTH_UINH |
+                                   MHPMEVENT_UINH | MHPMEVENT_MINH | MHPMEVENT_OF;
```
scottj97: This looks like exactly the same value as `mevent_mask`...is that correct? If so, why not use that value instead of repeating it here?

riscv/processor.cc
```diff
@@ -217,13 +217,16 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
     const reg_t which_mcounterh = CSR_MHPMCOUNTER3H + i - 3;
     const reg_t which_counter = CSR_HPMCOUNTER3 + i - 3;
     const reg_t which_counterh = CSR_HPMCOUNTER3H + i - 3;
-    auto mevent = std::make_shared<const_csr_t>(proc, which_mevent, 0);
+    const reg_t mevent_mask = proc->extension_enabled_const(EXT_SSCOFPMF) ? MHPMEVENT_VUINH | MHPMEVENT_VSINH | MHPMEVENTH_UINH |
+                                                                            MHPMEVENT_UINH | MHPMEVENT_MINH | MHPMEVENT_OF : 0;
+    mevent[i - 3] = std::make_shared<masked_csr_t>(proc, which_mevent, mevent_mask, 0);
     auto mcounter = std::make_shared<const_csr_t>(proc, which_mcounter, 0);
-    csrmap[which_mevent] = mevent;
+    csrmap[which_mevent] = mevent[i - 3];
```
scottj97: Is there any reason we might need `rv32_low_csr_t` for the csrmap here?

scottj97 (changes requested):
I don't know anything about this extension, so I'm only reviewing the code and not its RISCV functionality.

Can you tell us how this has been tested?

riscv/csrs.cc
liweiwei90: OK. I'll update this.

riscv/processor.cc
liweiwei90: Sorry. a mistake here. They should be shift right 32 bits here. Meventh_mask has the high 32 bits of  mevet_mask. Maybe I can use -1 directly here.

riscv/processor.cc
liweiwei90: Do you mean why I didn't use rv32_low_csr_t here? I think I should use rv32_low_csr_t here since high bits is added in sscofpmf extension and they are writable. I'll fix it later.

liweiwei90:
> I don't know anything about this extension, so I'm only reviewing the code and not its RISCV functionality.
> 
> Can you tell us how this has been tested?

The spec can be found in  https://github.com/riscv/riscv-count-overflow/releases/tag/v0.5.2
Currently. only a few feather of sscofpmf is added:
- new csr and csr fields for mevent: this is tested in a dirty way as Smstateen
- new irq type is added, it will not be triggered since pmu counter doesn't work currently

riscv/processor.cc
```diff
@@ -228,35 +228,46 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
+      csrmap[which_mevent] = std::make_shared<rv32_low_csr_t>(proc, which_mevent, mevent[i - 3]);;
       auto mcounterh = std::make_shared<const_csr_t>(proc, which_mcounterh, 0);
       csrmap[which_mcounterh] = mcounterh;
       if (proc->extension_enabled_const(EXT_ZICNTR) && proc->extension_enabled_const(EXT_ZIHPM)) {
         auto counterh = std::make_shared<counter_proxy_csr_t>(proc, which_counterh, mcounterh);
         csrmap[which_counterh] = counterh;
       }
+      if (proc->extension_enabled_const(EXT_SSCOFPMF)) {
+        auto eventh = std::make_shared<rv32_high_csr_t>(proc, which_meventh, mevent[i - 3]);
+        csrmap[which_meventh] = eventh;
```
scottj97: Can we please name this variable `meventh`?

Because there's some user-level CSRs with similar names (like `mcounterh` vs `counterh`) and even though `meventh` doesn't have a user-level `eventh` counterpart, this could be confused for that without the `m` in front.

scottj97 (approved):

riscv/processor.cc
liweiwei90: OK. I'll update it.

YenHaoChen:
Hello, 

I am looking forward to the Sscofpmf support in Spike.
Any further update on this PR?

Best regards,
YenHaoChen

scottj97 (approved):

YenHaoChen:
@aswaterman @scottj97 @liweiwei90 
This PR has been approved for over a month but is still not merged. Is there any other consideration?

scottj97:
It's all good from my perspective, but I assumed @aswaterman would want to take a look.

aswaterman (approved):
Sorry, it slipped through the cracks.  Looks good to me, too.

PR #983, opened 2022-04-30
Epmp updates 2

soberl:
Add smepmp support to Spike.
This is a rebased version. Existed review comments can be found at https://github.com/riscv-software-src/riscv-isa-sim/pull/981.

PR #1039, opened 2022-07-11
Allow writes to pmp(i-1)cfg on locked pmp(i)cfg

YenHaoChen:
The privileged spec allows writes to pmp(i-1)cfg with locked pmp(i)cfg
(According to a recent discussion: https://github.com/riscv/riscv-isa-manual/issues/866)

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -248,7 +248,7 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
   for (size_t i0 = (address - CSR_PMPCFG0) * 4, i = i0; i < i0 + proc->get_xlen() / 8; i++) {
     if (i < proc->n_pmp) {
       const bool locked = (state->pmpaddr[i]->cfg & PMP_L);
-      if (rlb || (!locked && !state->pmpaddr[i]->next_locked_and_tor())) {
+      if (rlb || !locked) {
         uint8_t cfg = (val >> (8 * (i - i0))) & (PMP_R | PMP_W | PMP_X | PMP_A | PMP_L);
         // Drop R=0 W=1 when MML = 0
         // Remove the restriction when MML = 1


aswaterman:
This change was added by @soberl as part of ePMP.

You're obviously right that this change is incorrect when ePMP is not implemented.  One of the ePMP folks will need to verify that it is correct for ePMP.  Then, we will need to conditionalize it on whether ePMP is implemented.

cc @gfavor, since he responded to the ISA spec thread.

aswaterman (changes requested):
Don't merge until the ePMP question is answered, and until this check is conditionalized by ePMP being implemented if appropriate.

YenHaoChen:
@aswaterman @soberl @gfavor
Thank you for your time and kindness. Please let me know if I can help with the PR. To my understanding, the locking behavior of this PR also conforms to the ePMP spec, Section 2.2 (https://github.com/riscv/riscv-tee/blob/main/Smepmp/Smepmp.pdf).

soberl:
@YenHaoChen Thanks for the point. I think you are correct. I just checked the ePMP manual and run ePMP testsuite with you change. I am fine to accept your commits. Looks like we don't need to conditionalize it.
@aswaterman Also thanks for the help here.

scottj97:
I have (proprietary) test cases that should cover this, but for some reason I do not see any change in their commit logs when I include this PR. I'm investigating.

scottj97 (approved):
I was using a seed for my (randomized) test case that did not exercise this logic. I've fixed that and I can now see the problem, which was originally caused by [this commit](https://github.com/riscv-software-src/riscv-isa-sim/commit/af500657c3ecde57fb88ed027dbdaa5da72a1c4b) (part of [PR 983](https://github.com/riscv-software-src/riscv-isa-sim/pull/983)).

I confirm this PR changes the PMP behavior back to the previous behavior, before af500657c3ecde57fb88ed027dbdaa5da72a1c4b, as expected.

I don't know anything about ePMP.

aswaterman (approved):
Cool, thanks for the quick resolution.

PR #1040, opened 2022-07-12
Update for counter related CSR 

liweiwei90:
This PR do some update for counter related csrs:

-  modify minstreth/mcycleh to reuse rv32_high_csr_t
-  add support for time/timeh/htimedelta/htimedeltah csrs
-  add base verify_permission in counter_proxy_csr_t::verify_permissions

### DIFF
--- a/riscv/clint.cc
+++ b/riscv/clint.cc
@@ -82,6 +82,7 @@ void clint_t::increment(reg_t inc)
     mtime += inc;
   }
   for (size_t i = 0; i < procs.size(); i++) {
+    procs[i]->state.time->sync(mtime);
     procs[i]->state.mip->backdoor_write_with_mask(MIP_MTIP, 0);
     if (mtime >= mtimecmp[i])
       procs[i]->state.mip->backdoor_write_with_mask(MIP_MTIP, MIP_MTIP);
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -515,6 +515,10 @@ bool rv32_low_csr_t::unlogged_write(const reg_t val) noexcept {
   return orig->unlogged_write((orig->written_value() >> 32 << 32) | (val & 0xffffffffU));
 }
 
+reg_t rv32_low_csr_t::written_value() const noexcept {
+  return orig->written_value() & 0xffffffffU;
+}
+
 // implement class rv32_high_csr_t
 rv32_high_csr_t::rv32_high_csr_t(processor_t* const proc, const reg_t addr, csr_t_p orig):
   csr_t(proc, addr),
@@ -531,6 +535,10 @@ void rv32_high_csr_t::verify_permissions(insn_t insn, bool write) const {
 
 bool rv32_high_csr_t::unlogged_write(const reg_t val) noexcept {
   return orig->unlogged_write((orig->written_value() << 32 >> 32) | ((val & 0xffffffffU) << 32));
+}
+
+reg_t rv32_high_csr_t::written_value() const noexcept {
+  return (orig->written_value() >> 32) & 0xffffffffU;
 }
 
 // implement class sstatus_csr_t
@@ -924,10 +932,7 @@ void wide_counter_csr_t::bump(const reg_t howmuch) noexcept {
 }
 
 bool wide_counter_csr_t::unlogged_write(const reg_t val) noexcept {
-  if (proc->get_xlen() == 32)
-    this->val = (this->val >> 32 << 32) | (val & 0xffffffffU);
-  else
-    this->val = val;
+  this->val = val;
   // The ISA mandates that if an instruction writes instret, the write
   // takes precedence over the increment to instret.  However, Spike
   // unconditionally increments instret after executing an instruction.
@@ -939,27 +944,6 @@ bool wide_counter_csr_t::unlogged_write(const reg_t val) noexcept {
 reg_t wide_counter_csr_t::written_value() const noexcept {
   // Re-adjust for upcoming bump()
   return this->val + 1;
-}
-
-void wide_counter_csr_t::write_upper_half(const reg_t val) noexcept {
-  this->val = (val << 32) | (this->val << 32 >> 32);
-  this->val--; // See comment above.
-  // Log upper half only.
-  log_special_write(address + (CSR_MINSTRETH - CSR_MINSTRET), written_value() >> 32);
-}
-
-counter_top_csr_t::counter_top_csr_t(processor_t* const proc, const reg_t addr, wide_counter_csr_t_p parent):
-  csr_t(proc, addr),
-  parent(parent) {
-}
-
-reg_t counter_top_csr_t::read() const noexcept {
-  return parent->read() >> 32;
-}
-
-bool counter_top_csr_t::unlogged_write(const reg_t val) noexcept {
-  parent->write_upper_half(val);
-  return true;
 }
 
 proxy_csr_t::proxy_csr_t(processor_t* const proc, const reg_t addr, csr_t_p delegate):
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -946,6 +946,25 @@ reg_t wide_counter_csr_t::written_value() const noexcept {
   return this->val + 1;
 }
 
+// implement class time_counter_csr_t
+time_counter_csr_t::time_counter_csr_t(processor_t* const proc, const reg_t addr):
+  csr_t(proc, addr),
+  shadow_val(0) {
+}
+
+reg_t time_counter_csr_t::read() const noexcept {
+  // reading the time CSR in VS or VU mode returns the sum of the contents of
+  // htimedelta and the actual value of time.
+  if (state->v)
+    return shadow_val + state->htimedelta->read();
+  else
+    return shadow_val;
+}
+
+void time_counter_csr_t::sync(const reg_t val) noexcept {
+  shadow_val = val;
+}
+
 proxy_csr_t::proxy_csr_t(processor_t* const proc, const reg_t addr, csr_t_p delegate):
   csr_t(proc, addr),
   delegate(delegate) {
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1001,11 +1001,13 @@ bool counter_proxy_csr_t::myenable(csr_t_p counteren) const noexcept {
 }
 
 void counter_proxy_csr_t::verify_permissions(insn_t insn, bool write) const {
+  proxy_csr_t::verify_permissions(insn, write);
+
   const bool mctr_ok = (state->prv < PRV_M) ? myenable(state->mcounteren) : true;
   const bool hctr_ok = state->v ? myenable(state->hcounteren) : true;
   const bool sctr_ok = (proc->extension_enabled('S') && state->prv < PRV_S) ? myenable(state->scounteren) : true;
 
-  if (write || !mctr_ok)
+  if (!mctr_ok)
     throw trap_illegal_instruction(insn.bits());
   if (!hctr_ok)
       throw trap_virtual_instruction(insn.bits());
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -204,9 +204,11 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   csrmap[CSR_MCAUSE] = mcause = std::make_shared<cause_csr_t>(proc, CSR_MCAUSE);
   minstret = std::make_shared<wide_counter_csr_t>(proc, CSR_MINSTRET);
   mcycle = std::make_shared<wide_counter_csr_t>(proc, CSR_MCYCLE);
+  time = std::make_shared<time_counter_csr_t>(proc, CSR_TIME);
   if (proc->extension_enabled_const(EXT_ZICNTR)) {
     csrmap[CSR_INSTRET] = std::make_shared<counter_proxy_csr_t>(proc, CSR_INSTRET, minstret);
     csrmap[CSR_CYCLE] = std::make_shared<counter_proxy_csr_t>(proc, CSR_CYCLE, mcycle);
+    csrmap[CSR_TIME] = std::make_shared<counter_proxy_csr_t>(proc, CSR_TIME, time);
   }
   if (xlen == 32) {
     csr_t_p minstreth, mcycleh;
@@ -215,8 +217,10 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
     csrmap[CSR_MCYCLE] = std::make_shared<rv32_low_csr_t>(proc, CSR_MCYCLE, mcycle);
     csrmap[CSR_MCYCLEH] = mcycleh = std::make_shared<rv32_high_csr_t>(proc, CSR_MCYCLEH, mcycle);
     if (proc->extension_enabled_const(EXT_ZICNTR)) {
+      auto timeh = std::make_shared<rv32_high_csr_t>(proc, CSR_TIMEH, time);
       csrmap[CSR_INSTRETH] = std::make_shared<counter_proxy_csr_t>(proc, CSR_INSTRETH, minstreth);
       csrmap[CSR_CYCLEH] = std::make_shared<counter_proxy_csr_t>(proc, CSR_CYCLEH, mcycleh);
+      csrmap[CSR_TIMEH] = std::make_shared<counter_proxy_csr_t>(proc, CSR_TIMEH, timeh);
     }
   } else {
     csrmap[CSR_MINSTRET] = minstret;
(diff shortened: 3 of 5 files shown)


aswaterman:
Are these changes necessarily coupled?  Can they not be split into several independent PRs?

liweiwei90:
> Are these changes necessarily coupled? Can they not be split into several independent PRs?

It's OK to split into independent PRs. I'll do this later.

riscv/csrs.cc
```diff
@@ -973,6 +973,8 @@ bool counter_proxy_csr_t::myenable(csr_t_p counteren) const noexcept {
 }
 
 void counter_proxy_csr_t::verify_permissions(insn_t insn, bool write) const {
+  proxy_csr_t::verify_permissions(insn, write);
```
scottj97, line 1004: Can you clarify what problem/bug this solves? It seems there was a bug here with CSR access permissions but we need to explain how to trigger the bug.

liweiwei90:
I have split it into three PRs. This PR is for counter csr related update. https://github.com/riscv-software-src/riscv-isa-sim/pull/1041 is for new added csrs. https://github.com/riscv-software-src/riscv-isa-sim/pull/1042 is for checking H extension implicitly requires S mode.

riscv/processor.cc
```diff
@@ -202,17 +202,22 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
   if (proc->extension_enabled_const(EXT_ZICNTR)) {
     csrmap[CSR_INSTRET] = std::make_shared<counter_proxy_csr_t>(proc, CSR_INSTRET, minstret);
     csrmap[CSR_CYCLE] = std::make_shared<counter_proxy_csr_t>(proc, CSR_CYCLE, mcycle);
+    csrmap[CSR_TIME] = std::make_shared<counter_proxy_csr_t>(proc, CSR_TIME, time);
   }
   if (xlen == 32) {
     csr_t_p minstreth, mcycleh;
     csrmap[CSR_MINSTRETH] = minstreth = std::make_shared<rv32_high_csr_t>(proc, CSR_MINSTRETH, -1, minstret);
     csrmap[CSR_MCYCLEH] = mcycleh = std::make_shared<rv32_high_csr_t>(proc, CSR_MCYCLEH, -1, mcycle);
+    auto timeh = std::make_shared<rv32_high_csr_t>(proc, CSR_CYCLEH, -1, time);
```
scottj97: Should be CSR_TIMEH?

riscv/processor.cc
```diff
@@ -202,17 +202,22 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
     if (proc->extension_enabled_const(EXT_ZICNTR)) {
       csrmap[CSR_INSTRETH] = std::make_shared<counter_proxy_csr_t>(proc, CSR_INSTRETH, minstreth);
       csrmap[CSR_CYCLEH] = std::make_shared<counter_proxy_csr_t>(proc, CSR_CYCLEH, mcycleh);
+      csrmap[CSR_TIMEH] = std::make_shared<counter_proxy_csr_t>(proc, CSR_CYCLEH, timeh);
```
scottj97: Should be CSR_TIMEH?

riscv/processor.cc
```diff
@@ -342,6 +347,8 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
     (1 << CAUSE_STORE_PAGE_FAULT);
   csrmap[CSR_HEDELEG] = hedeleg = std::make_shared<masked_csr_t>(proc, CSR_HEDELEG, hedeleg_mask, 0);
   csrmap[CSR_HCOUNTEREN] = hcounteren = std::make_shared<masked_csr_t>(proc, CSR_HCOUNTEREN, counteren_mask, 0);
+  csrmap[CSR_HTIMEDELTA] = htimedelta = std::make_shared<basic_csr_t>(proc, CSR_HTIMEDELTA, 0);
+  if (xlen == 32) csrmap[CSR_HTIMEDELTAH] = std::make_shared<rv32_high_csr_t>(proc, CSR_HTIMEDELTA, 0, htimedelta);
```
scottj97: Should be CSR_HTIMEDELTAH?

riscv/clint.cc
```diff
@@ -83,6 +83,7 @@ void clint_t::increment(reg_t inc)
   }
   for (size_t i = 0; i < procs.size(); i++) {
     procs[i]->state.mip->backdoor_write_with_mask(MIP_MTIP, 0);
+    procs[i]->state.time->sync(mtime);
```
scottj97, line 85: This smells funny to me...what about initializing `time_counter_csr_t` with a `mtime_t*` that points to the clint's `mtime` variable? No sync needed then.

In any case, this call to `sync()` shouldn't be right in the middle of the two writes to MTIP; put it before or after.

scottj97 (changes requested):

riscv/processor.cc
```diff
@@ -207,9 +207,9 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
     csrmap[CSR_CYCLE] = std::make_shared<counter_proxy_csr_t>(proc, CSR_CYCLE, mcycle);
   }
   if (xlen == 32) {
-    counter_top_csr_t_p minstreth, mcycleh;
-    csrmap[CSR_MINSTRETH] = minstreth = std::make_shared<counter_top_csr_t>(proc, CSR_MINSTRETH, minstret);
-    csrmap[CSR_MCYCLEH] = mcycleh = std::make_shared<counter_top_csr_t>(proc, CSR_MCYCLEH, mcycle);
+    csr_t_p minstreth, mcycleh;
+    csrmap[CSR_MINSTRETH] = minstreth = std::make_shared<rv32_high_csr_t>(proc, CSR_MINSTRETH, -1, minstret);
+    csrmap[CSR_MCYCLEH] = mcycleh = std::make_shared<rv32_high_csr_t>(proc, CSR_MCYCLEH, -1, mcycle);
```
scottj97: Commit 136121d2fe8d1699527789405b5807ba5097e2ef breaks `mcycleh` and `minstreth`; they both read & write only 0 now. Test `rv32mi-p-csr` from `riscv-tests` shows this misbehavior.

I can't see what's wrong with the code, though.

riscv/csrs.cc
liweiwei90: Yeah, the base verify_permissions will check some basic properties of csrs, such as whether the csr is read-only.  "counter_proxy_csr_t" is used for read-only csrs (INSTRET/CYCLE/TIME). Without the base check, write access will be allowed in current verify_permissions.

riscv/processor.cc
liweiwei90: Sorry for this and following typos. I'll fix them later

riscv/clint.cc
liweiwei90: It seems more difficult and will take more code modification to pass mtime_t* from clint to the initializing time_counter_csr_t. Initializing time_counter_csr_t is done when construction of processor. Clint is initialized later in current implementation and may not exist. So I prefer current way to sync mtime.

riscv/processor.cc
liweiwei90: OK. I'll test this later.

riscv/clint.cc
scottj97: OK, that's reasonable. Please move the call to `sync()` either above or below the two updates to MTIP (which it seems could be easily improved to be a single update, but that's a different problem).

riscv/csrs.cc
scottj97: Line 982 `if (write || !mctr_ok)` handles this case.

I have tested this thoroughly, and I cannot see any change in behavior from 2bc41bb1814a70d39ee93aac53e90c28d0102676.

I think it is an improvement, but the commit message should clarify that this is not expected to cause any functional change.

riscv/processor.cc
liweiwei90: Sorry. When I running rv32mi-p-csr with following command:
`spike --isa=rv32im /workspace/riscv/riscv-tools/riscv-tests/isa/rv32mi-p-csr`
It seems pass(no fail related message). And I don't find cycleh in the code of isa/rv64si/csr.S which is reused by rv32mi/csr.S in the latest version of riscv-tests(commit 56c2471)

riscv/csrs.cc
liweiwei90: Yeah. The other checks in csr_t::verify_permissions will not work on INSTRET/CYCLE/TIME{H}, since they are user-level csrs.

riscv/processor.cc
scottj97: My mistake; I have local modifications to the test to cover this. I will open a PR on riscv-tests with these changes.

scottj97: Using [this branch](https://github.com/scottj97/riscv-tests/tree/mcycleh) ([this commit](https://github.com/scottj97/riscv-tests/commit/1d669b9b42e82987101db8692ac1583324625423)) I ran the following (after building riscv-tests):

```
./configure --enable-commitlog --enable-misaligned
make
./spike -l --log-commits --isa=rv32imafdc riscv-tests/isa/rv32mi-p-csr > rv32mi-p-csr.out 2>&1
```

The additions I made to the test are not self-checking. The test "passes" both before and after 136121d2fe8d1699527789405b5807ba5097e2ef, but you can see the difference in the log before and after:
```
--- a/rv32mi-p-csr.out
+++ b/rv32mi-p-csr.out
@@ -161,17 +161,17 @@ core   0: 3 0x80000194 (0xb0202273) x4  0x000000b0
 core   0: 0x80000198 (0x00100193) li      gp, 1
 core   0: 3 0x80000198 (0x00100193) x3  0x00000001
 core   0: 0x8000019c (0xb8019073) csrw    mcycleh, gp
-core   0: 3 0x8000019c (0xb8019073) c2944_mcycleh 0x00000001
+core   0: 3 0x8000019c (0xb8019073) c2944_mcycleh 0x00000000
 core   0: 0x800001a0 (0xb8002273) csrr    tp, mcycleh
-core   0: 3 0x800001a0 (0xb8002273) x4  0x00000001
+core   0: 3 0x800001a0 (0xb8002273) x4  0x00000000
 core   0: 0x800001a4 (0x00118193) addi    gp, gp, 1
```

liweiwei90: OK. Thanks. I'll test it later.

liweiwei90: In the wide_counter_csr_t::unlogged_write, the upper 32 bits for written val will be ignored. So the mcycleh and minstreth which reuse the high 32 bits of mcycle and minstret will be read & write only 0.

scottj97: Your change in 58706130eca53189d0f073cf13a09572a9bedd49 is not correct; privileged spec says of RV32 counters (emphasis added):

> When MXLEN=32, reads of the mcycle, minstret, and mhpmcountern CSRs return bits 31–0 of the corresponding counter, and **_writes change only bits 31–0_**; reads of the mcycleh, minstreth, and mhpmcounternh CSRs return bits 63–32 of the corresponding counter, and writes change only bits 63–32.

(It's possible RV32 `mstatus` is broken when hypervisor's implemented...it looks like `csrw mstatus` might write the upper 32 bits (aka `mstatush`) to 0?)

liweiwei90: Yeah, that is a question. Maybe we should make rv32_high_csr_t to store the high part additionally. or modify the unlogged_write of original csr to write only the low part for RV32.

scottj97: If `unlogged_write()` only writes the low part for RV32 (and I think it must) then the `rv32_high_csr_t` can't use `orig->unlogged_write()` to update the upper 32 bits.

I've confirmed that RV32 `mstatus` is broken, and always has been, ever since `mstatush` first had writable bits back when #812 was fixed. Doing `csrw mstatus` will erroneously clear the upper 32 bits. (Note that `mstatush` only has writable bits if Hypervisor is enabled.)

liweiwei90: Yes, so I add two virtual function to csr_t:

- write_upper_half : update the high half of csr
- write_full: update the whole 64 bits of csr

riscv/csrs.cc
```diff
@@ -798,8 +817,19 @@ masked_csr_t::masked_csr_t(processor_t* const proc, const reg_t addr, const reg_
   mask(mask) {
 }
 
+void masked_csr_t::write_full(const reg_t val) noexcept {
+  bool success = basic_csr_t::unlogged_write((read() & ~mask) | (val & mask));
+  if (success)
+    log_write();
+}
```
scottj97: This is a bit messy, having to deal with logging specially here. This also breaks logging of `csrw mstatush` which now logs `mstatus` (the low 32 bits) instead of `mstatush`.

See #1045 for my alternative proposal. (Your review appreciated!)

liweiwei90: I have updated based on the new rv32_low/high_csr_t way.

scottj97 (changes requested):
Commit c6d107ffa6eb0c94d908e0a87762f159bd3caa8f breaks the logged value of `minstret` on a CSRW. It's logging (val-1) instead of the written value.

```
 core   0: 0x800001c8 (0x06410113) addi    sp, sp, 100
 core   0: 3 0x800001c8 (0x06410113) x2  0x000000bf
 core   0: 0x800001cc (0xb0211073) csrw    minstret, sp
-core   0: 3 0x800001cc (0xb0211073) c2818_minstret 0x000000bf
+core   0: 3 0x800001cc (0xb0211073) c2818_minstret 0x000000be
 core   0: 0x800001d0 (0xb0002273) csrr    tp, mcycle
 core   0: 3 0x800001d0 (0xb0002273) x4  0x0000005e
```

riscv/processor.cc
scottj97: mstatush was fixed by #1045 and this was rebased on top. Resolving this conversation.

riscv/processor.cc
```diff
@@ -202,20 +202,29 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
   if (xlen == 32) {
-    counter_top_csr_t_p minstreth, mcycleh;
-    csrmap[CSR_MINSTRETH] = minstreth = std::make_shared<counter_top_csr_t>(proc, CSR_MINSTRETH, minstret);
-    csrmap[CSR_MCYCLEH] = mcycleh = std::make_shared<counter_top_csr_t>(proc, CSR_MCYCLEH, mcycle);
+    csr_t_p minstreth, mcycleh;
+    csrmap[CSR_MINSTRET] = std::make_shared<rv32_low_csr_t>(proc, CSR_MINSTRET, minstret);
+    csrmap[CSR_MINSTRETH] = minstreth = std::make_shared<rv32_high_csr_t>(proc, CSR_MINSTRETH, minstret);
+    csrmap[CSR_MCYCLE] = std::make_shared<rv32_low_csr_t>(proc, CSR_MCYCLE, mcycle);
+    csrmap[CSR_MCYCLEH] = mcycleh = std::make_shared<rv32_high_csr_t>(proc, CSR_MCYCLEH, mcycle);
+    auto timeh = std::make_shared<rv32_high_csr_t>(proc, CSR_TIMEH, time);
```
scottj97: Can the `timeh` construction be moved inside the next line's conditional? It doesn't look like it's used otherwise.

liweiwei90: OK. I have updated it.

liweiwei90:
> Commit [c6d107f](https://github.com/riscv-software-src/riscv-isa-sim/commit/c6d107ffa6eb0c94d908e0a87762f159bd3caa8f) breaks the logged value of `minstret` on a CSRW. It's logging (val-1) instead of the written value.
> 
> ```
>  core   0: 0x800001c8 (0x06410113) addi    sp, sp, 100
>  core   0: 3 0x800001c8 (0x06410113) x2  0x000000bf
>  core   0: 0x800001cc (0xb0211073) csrw    minstret, sp
> -core   0: 3 0x800001cc (0xb0211073) c2818_minstret 0x000000bf
> +core   0: 3 0x800001cc (0xb0211073) c2818_minstret 0x000000be
>  core   0: 0x800001d0 (0xb0002273) csrr    tp, mcycle
>  core   0: 3 0x800001d0 (0xb0002273) x4  0x0000005e
> ```
The bug is that rv32_low_csr_t doesn't have the written_value function to re-adjust the instret value just as wide_counter_csr_t. So I think we can add written_value function in rv32_low_csr_t to get the written value from orig->written_value. I have fixed it in this way currently.

scottj97:
> The bug is that rv32_low_csr_t doesn't have the written_value function to re-adjust the instret value just as wide_counter_csr_t. So I think we can add written_value function in rv32_low_csr_t to get the written value from orig->written_value. I have fixed it in this way currently.

Seems like `rv32_high_csr_t` will have the same problem, yes? If the low half happens to be 0 when we write the upper half, then the [decrement](https://github.com/riscv-software-src/riscv-isa-sim/blob/b0d9782e13156abd5884fa73017a0286441202d1/riscv/csrs.cc#L945) will modify the upper half, and the commit log will show the wrong value.

liweiwei90:
No, rv32_high_csr_t doesn't have this problem. 

> > The bug is that rv32_low_csr_t doesn't have the written_value function to re-adjust the instret value just as wide_counter_csr_t. So I think we can add written_value function in rv32_low_csr_t to get the written value from orig->written_value. I have fixed it in this way currently.
> 
> Seems like `rv32_high_csr_t` will have the same problem, yes? If the low half happens to be 0 when we write the upper half, then the [decrement](https://github.com/riscv-software-src/riscv-isa-sim/blob/b0d9782e13156abd5884fa73017a0286441202d1/riscv/csrs.cc#L945) will modify the upper half, and the commit log will show the wrong value.

Yeah. I didn't realize this special case. I've add it to rv32_high_csr_t.

riscv/processor.cc
```diff
@@ -202,20 +202,29 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
   csrmap[CSR_MCAUSE] = mcause = std::make_shared<cause_csr_t>(proc, CSR_MCAUSE);
-  csrmap[CSR_MINSTRET] = minstret = std::make_shared<wide_counter_csr_t>(proc, CSR_MINSTRET);
-  csrmap[CSR_MCYCLE] = mcycle = std::make_shared<wide_counter_csr_t>(proc, CSR_MCYCLE);
+  minstret = std::make_shared<wide_counter_csr_t>(proc, CSR_MINSTRET);
+  mcycle = std::make_shared<wide_counter_csr_t>(proc, CSR_MCYCLE);
+  time = std::make_shared<time_counter_csr_t>(proc, CSR_TIME);
   if (proc->extension_enabled_const(EXT_ZICNTR)) {
     csrmap[CSR_INSTRET] = std::make_shared<counter_proxy_csr_t>(proc, CSR_INSTRET, minstret);
     csrmap[CSR_CYCLE] = std::make_shared<counter_proxy_csr_t>(proc, CSR_CYCLE, mcycle);
+    csrmap[CSR_TIME] = std::make_shared<counter_proxy_csr_t>(proc, CSR_TIME, time);
```
scottj97, line 211: How come these three don't need to use `rv32_low_csr_t` for RV32?

scottj97 (changes requested):

riscv/processor.cc
liweiwei90: All of them are read-only. So they seem unnecessary to use rv32_low_csr_t.

scottj97 (approved):

aswaterman (approved):

PR #1041, opened 2022-07-12
add support for m/henvcfgh and mconfigptr CSRs 

liweiwei90:


riscv/csrs.cc
```diff
@@ -467,7 +467,10 @@ mstatus_csr_t::mstatus_csr_t(processor_t* const proc, const reg_t addr):
       | (proc->extension_enabled_const('S') && (proc->get_const_xlen() != 32) ? set_field((reg_t)0, MSTATUS_SXL, xlen_to_uxl(proc->get_const_xlen())) : 0)
 
 #ifdef RISCV_ENABLE_DUAL_ENDIAN
-      | (proc->get_mmu()->is_target_big_endian() ? MSTATUS_UBE | MSTATUS_SBE | MSTATUS_MBE : 0)
+      | (proc->get_mmu()->is_target_big_endian() ? (proc->extension_enabled_const('U') ? MSTATUS_UBE : 0)
```
scottj97: b570b4493768ca6f563b08e7a0c2f7c90fcb6739 looks like 4 separate commits to me

riscv/csrs.cc
```diff
@@ -858,7 +862,7 @@ satp_csr_t::satp_csr_t(processor_t* const proc, const reg_t addr):
 void satp_csr_t::verify_permissions(insn_t insn, bool write) const {
   base_atp_csr_t::verify_permissions(insn, write);
   if (get_field(state->mstatus->read(), MSTATUS_TVM))
-    require(state->prv >= PRV_M);
+    require(state->prv == PRV_M);
```
scottj97, line 888: This should be its own commit with a message such as: "Clarify permission check since prv can never be larger than PRV_M" (if I'm understanding this correctly?)

riscv/processor.cc
```diff
@@ -388,8 +388,11 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
                              (proc->extension_enabled(EXT_ZICBOZ) ? MENVCFG_CBZE : 0) |
                              (proc->extension_enabled(EXT_SVPBMT) ? MENVCFG_PBMTE : 0);
   const reg_t menvcfg_init = (proc->extension_enabled(EXT_SVPBMT) ? MENVCFG_PBMTE : 0);
-  csrmap[CSR_MENVCFG] = menvcfg = std::make_shared<masked_csr_t>(proc, CSR_MENVCFG, menvcfg_mask, menvcfg_init);
-  if (xlen == 32) csrmap[CSR_MENVCFGH] = std::make_shared<rv32_high_csr_t>(proc, CSR_MENVCFGH, -1, menvcfg);
+  menvcfg = std::make_shared<masked_csr_t>(proc, CSR_MENVCFG, menvcfg_mask, menvcfg_init);
```
scottj97: Would it be okay to never initialize `menvcfg` if there's no U-mode? I can't see that any code would access it in that case.

scottj97 (changes requested):

riscv/csrs.cc
liweiwei90: Yeah. It contains four point of u/s related check.

riscv/csrs.cc
liweiwei90: OK. I'll do this later.

riscv/processor.cc
liweiwei90: Leaving menvcfg here is for compatibility. menvcfg will be used to check whether PBMT is supported (in walk and s2xlate) , such as in s2xlate:
`      bool pbmte = proc->get_state()->menvcfg->read() & MENVCFG_PBMTE;`
If we never initialize menvcfg if there's no U-mode, this function of PBMT will be affected.

scottj97: If there's no U-mode then there's no S-mode either (right?) so paging will never be enabled. The code in `s2xlate()` and friends should never be executed.

liweiwei90: Yeah. I agree.  I'll update this.

riscv/csrs.cc
```diff
@@ -862,7 +862,7 @@ satp_csr_t::satp_csr_t(processor_t* const proc, const reg_t addr):
...
```
scottj97, line 888: The commit message of "fix permission check" is misleading, even alarming, because it makes it sound like there was a bug with this CSR's permission checks. But this commit doesn't actually change any behavior. It clarifies a confusing line of code, that's all.

riscv/csrs.cc
```diff
@@ -467,7 +467,10 @@ mstatus_csr_t::mstatus_csr_t(processor_t* const proc, const reg_t addr):
       | (proc->extension_enabled_const('S') && (proc->get_const_xlen() != 32) ? set_field((reg_t)0, MSTATUS_SXL, xlen_to_uxl(proc->get_const_xlen())) : 0)
 
 #ifdef RISCV_ENABLE_DUAL_ENDIAN
-      | (proc->get_mmu()->is_target_big_endian() ? MSTATUS_UBE | MSTATUS_SBE | MSTATUS_MBE : 0)
+      | (proc->get_mmu()->is_target_big_endian() ? (proc->extension_enabled_const('U') ? MSTATUS_UBE : 0)
+                                                 | (proc->extension_enabled_const('S') ? MSTATUS_SBE : 0)
+                                                 | MSTATUS_MBE
+                                                 : 0)
```
scottj97: The syntax is awkward here because of the ternary operator. What if you made a separate const `big_endian_bits` first, and then use that to make this ternary into a single line again?

riscv/csrs.cc
liweiwei90: OK. I'll update this.

riscv/csrs.cc
liweiwei90: OK.

riscv/csrs.cc
```diff
@@ -465,24 +465,28 @@ bool sstatus_proxy_csr_t::unlogged_write(const reg_t val) noexcept {
...
+    const reg_t big_endian_bits = (proc->extension_enabled_const('U') ? MSTATUS_UBE : 0)
+                                | (proc->extension_enabled_const('S') ? MSTATUS_SBE : 0)
+                                | MSTATUS_MBE;
+#endif
+    val = 0
+          | (proc->extension_enabled_const('U') && (proc->get_const_xlen() != 32) ? set_field((reg_t)0, MSTATUS_UXL, xlen_to_uxl(proc->get_const_xlen())) : 0)
+          | (proc->extension_enabled_const('S') && (proc->get_const_xlen() != 32) ? set_field((reg_t)0, MSTATUS_SXL, xlen_to_uxl(proc->get_const_xlen())) : 0)
 #ifdef RISCV_ENABLE_DUAL_ENDIAN
-      | (proc->get_mmu()->is_target_big_endian() ? MSTATUS_UBE | MSTATUS_SBE | MSTATUS_MBE : 0)
+          | (proc->get_mmu()->is_target_big_endian() ? big_endian_bits : 0)
```
scottj97: Sorry, I did not realize when I requested this change that it was part of an initializer list. `val` really ought to be initialized in the initializer list, not in the constructor body.

You could use a [delegating constructor](https://stackoverflow.com/a/17375163/202201) though perhaps that's too much. Maybe there's a way to fit this all in one expression that would be easier to read than the original?

liweiwei90: Can we move all this into a separate function, and initialize the val with this function in the initializer list? just like the way for sstatus_write_mask of base_status_csr_t:
`sstatus_write_mask(compute_sstatus_write_mask())`

scottj97: Yes, good idea!

riscv/csrs.cc
```diff
@@ -497,6 +490,20 @@ bool mstatus_csr_t::unlogged_write(const reg_t val) noexcept {
   return true;
 }
 
+reg_t mstatus_csr_t::compute_mstatus_initial_value() const noexcept {
+#ifdef RISCV_ENABLE_DUAL_ENDIAN
```
scottj97: I don't the either of these `#ifdef`s are needed here. `is_target_big_endian()` will always be false if this macro is undefined, and it doesn't hurt to always calculate these values.

scottj97 (changes requested):
Commit 022e5ac2afffce2a3bf2eb7e288a000c9aaf343d combines a refactoring (extract new function) with functional changes. Please split into two so the functional changes are more apparent.

riscv/csrs.cc
liweiwei90: Yeah, I think so and remove the similar  #ifdef RISCV_ENABLE_DUAL_ENDIAN in other functions.

fesvr/htif.h
```diff
@@ -31,26 +31,18 @@ class htif_t : public chunked_memif_t
 
   template<typename T> inline T from_target(target_endian<T> n) const
   {
-#ifdef RISCV_ENABLE_DUAL_ENDIAN
```
scottj97, line 34: This change I'm not sure about. It is functionally correct, but might cause a performance degradation. `from_target()` and `to_target()` are called frequently.

This is probably why there was a compile-time define in the first place, versus only the dynamic runtime endianness choice.

scottj97 (changes requested):

fesvr/htif.h
liweiwei90: It seems only be used by to_host/from_host related read/write.

scottj97: ??

`from_target()` is used in the `load_func()` macro which means every load calls it. `to_target()` is used in the `store_func()` macro which means every store calls it.

liweiwei90: I think load/store_func use the  from/to_target in mmu.h itself not this one.

scottj97: Yes, my mistake.

scottj97 (approved):

aswaterman:
Thanks @liweiwei90 and @scottj97!

PR #1042, opened 2022-07-12
add check for H extension implicitly requires S mode

liweiwei90:
Even though the spec doesn't explicitly say this,It seems that Hypervisor extension cannot work without S-mode.


aswaterman (approved):
I suggest deleting the word "implicitly", but otherwise I agree.

liweiwei90:
> I suggest deleting the word "implicitly", but otherwise I agree.

OK. Done.

PR #1043, opened 2022-07-13
Conditionalize Smepmp extension (ePMP) support

YenHaoChen:
I want to conditionalize Smepmp extension (ePMP) support, which provides a more flexible ISA configuration. This PR includes two modifications. First, it includes Smepmp to the ISA string parser. Second, it adds verifiy_permissions() for mseccfg_csr_t, which checks the permission for the mseccfg.


YenHaoChen:
@soberl How do you feel about this?

soberl:
@YenHaoChen  Looks good for me. Just one question from my side, what's the order for enum value for EXT_SMEPMP? Is there any rules for that?

YenHaoChen:
@soberl I don't know whether there are rules for the order for the enum of extensions. Since the EXT_SMEPMP starts with 'EXT_S,' I put it with other extensions that begin with 'EXT_S.'

soberl:
@YenHaoChen  Thanks. Maybe there is no rules needed since it's just an internal status. For users, the config interface is the ISA string. I am file with that.

scottj97 (approved):
I don't know anything about ePMP but I assume if `mseccfg==0` then the behavior matches the old PMP?

YenHaoChen:
@scottj97 Yes, if mseccfg==0, the behavior matches the old PMP. That's why the Smepmp spec suggests implementing the reset value of mseccfg to zero for backward compatibility.
@soberl Would you help confirm this?
image

soberl:
> @scottj97 Yes, if mseccfg==0, the behavior matches the old PMP. That's why the Smepmp spec suggests implementing the reset value of mseccfg to zero for backward compatibility. @soberl Would you help confirm this? image

Yes, the behavior should keep same when mseccfg==0. New features are based on RLB, MMWP and MML bits of mseccfg.

aswaterman (approved):
LGTM.

PR #1045, opened 2022-07-13
Fix #1044

### LINKED ISSUE #1044: RV32 csrw to mstatus erroneously clears mstatush
On RV32, a `csrw mstatus` should not touch the bits in `mstatush`, but ever since we first added writable bits in `mstatush` (in dac0d3ce7161d480c06b8a19675f66ae636bc20d for #812) this has been broken. A `csrw mstatus` will erroneously clear the writable bits in `mstatush`.

Note that `mstatush` only has writable bits if hypervisor is enabled.

scottj97:
This fixes #1044 by creating a proxy CSR object for `csrr/csrw` instructions to use when accessing the low 32 bits of `mstatus` et al.

Other indirect users ([like this](https://github.com/riscv-software-src/riscv-isa-sim/blob/0ad0d9733b41e666172ff55ee946253c6dbdd080/riscv/mmu.cc#L62)) of the `mstatus` object will still see a 64-bit value, even in RV32.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -493,6 +493,24 @@ bool mstatus_csr_t::unlogged_write(const reg_t val) noexcept {
   return true;
 }
 
+// implement class rv32_low_csr_t
+rv32_low_csr_t::rv32_low_csr_t(processor_t* const proc, const reg_t addr, csr_t_p orig):
+  csr_t(proc, addr),
+  orig(orig) {
+}
+
+reg_t rv32_low_csr_t::read() const noexcept {
+  return orig->read() & 0xffffffffU;
+}
+
+void rv32_low_csr_t::verify_permissions(insn_t insn, bool write) const {
+  orig->verify_permissions(insn, write);
+}
+
+bool rv32_low_csr_t::unlogged_write(const reg_t val) noexcept {
+  return orig->unlogged_write((orig->written_value() >> 32 << 32) | (val & 0xffffffffU));
+}
+
 // implement class rv32_high_csr_t
 rv32_high_csr_t::rv32_high_csr_t(processor_t* const proc, const reg_t addr, csr_t_p orig):
   csr_t(proc, addr),
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -500,7 +500,7 @@ rv32_high_csr_t::rv32_high_csr_t(processor_t* const proc, const reg_t addr, csr_
 }
 
 reg_t rv32_high_csr_t::read() const noexcept {
-  return (orig->read() >> 32) & mask;
+  return (orig->read() >> 32) & 0xffffffffU;
 }
 
 void rv32_high_csr_t::verify_permissions(insn_t insn, bool write) const {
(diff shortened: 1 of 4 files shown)


riscv/csrs.cc
```diff
@@ -455,6 +455,10 @@ sstatus_proxy_csr_t::sstatus_proxy_csr_t(processor_t* const proc, const reg_t ad
 bool sstatus_proxy_csr_t::unlogged_write(const reg_t val) noexcept {
   const reg_t new_mstatus = (mstatus->read() & ~sstatus_write_mask) | (val & sstatus_write_mask);
 
+  // On RV32 this will only log the low 32 bits, so make sure we're
+  // not modifying anything in the upper 32 bits.
+  assert((sstatus_write_mask & 0xffffffffU) == sstatus_write_mask);
+
   mstatus->write(new_mstatus);
```
liweiwei90, line 462: Is this necessary? sstatus_write_mask seems to be a nearly const value calculated by compute_sstatus_write_mask:
```
  return 0
    | (has_fs ? SSTATUS_FS : 0)
    | (proc->any_custom_extensions() ? SSTATUS_XS : 0)
    | (has_vs ? SSTATUS_VS : 0)
    ;
```

riscv/processor.cc
```diff
@@ -189,11 +189,13 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
   csrmap[CSR_MISA] = misa = std::make_shared<misa_csr_t>(proc, CSR_MISA, max_isa);
-  csrmap[CSR_MSTATUS] = mstatus = std::make_shared<mstatus_csr_t>(proc, CSR_MSTATUS);
+  mstatus = std::make_shared<mstatus_csr_t>(proc, CSR_MSTATUS);
 
   if (xlen == 32) {
-    const reg_t mstatush_mask = MSTATUSH_MPV | MSTATUSH_GVA | MSTATUSH_SBE | MSTATUSH_MBE;
-    csrmap[CSR_MSTATUSH] = std::make_shared<rv32_high_csr_t>(proc, CSR_MSTATUSH, mstatush_mask, mstatus);
+    csrmap[CSR_MSTATUS] = std::make_shared<rv32_low_csr_t>(proc, CSR_MSTATUS, mstatus);
+    csrmap[CSR_MSTATUSH] = mstatush = std::make_shared<rv32_high_csr_t>(proc, CSR_MSTATUSH, mstatus);
+  } else {
```
liweiwei90, line 197: The mask for mstatush for RV32 is a little different from the high half of mstatus in RV64. Maybe we need a mask here to mask the unused fields in mstatush before writing to the whole mstatus.

liweiwei90 (changes requested):

riscv/csrs.cc
scottj97: The `sstatus` write mask is computed a mile away and can change in the future. The assumption in this code (about none of the upper bits changing) is not at all obvious, so I wanted to document and enforce it with the assertion.

One alternative is to move the assertion into `compute_sstatus_write_mask()` so it's only executed once, and update the comment to explain specifically what code is reliant on this assumption.

What do you think?

liweiwei90: OK. I see. Just leave it here.

riscv/processor.cc
scottj97: AFAICT the only differences are:
* SD, which is never writable
* SXL and UXL, which are not writable today in Spike, and even if Spike was enhanced to support that, those fields would/could/should be masked to 0 when MXLEN=32.

I have tested writing 0xffffffff to `mstatush` and I can see that it writes only the two hypervisor bits (MPV and GVA), as expected.

So I don't think any mask is needed.

liweiwei90: Yeah. A mistake myself. I didn't realize SXL and UXL are not writable currently.

liweiwei90 (approved):

aswaterman (approved):
LGTM.

PR #1047, opened 2022-07-16
Fix misaligned HLV/HLVX/HSV

scottj97:
If the address was misaligned, these three instructions would access memory using the current privilege, instead of the guest privilege as intended.

This fix is a little messy but will probably be much simpler once we solve #872.

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -71,8 +71,11 @@ public:
   inline void misaligned_store(reg_t addr, reg_t data, size_t size, uint32_t xlate_flags, bool actually_store=true)
   {
 #ifdef RISCV_ENABLE_MISALIGNED
-    for (size_t i = 0; i < size; i++)
-      store_uint8(addr + (target_big_endian? size-1-i : i), data >> (i * 8), actually_store);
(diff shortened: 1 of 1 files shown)


riscv/mmu.h
```diff
@@ -56,8 +56,11 @@ class mmu_t
   {
 #ifdef RISCV_ENABLE_MISALIGNED
     reg_t res = 0;
-    for (size_t i = 0; i < size; i++)
-      res += (reg_t)load_uint8(addr + (target_big_endian? size-1-i : i)) << (i * 8);
+    for (size_t i = 0; i < size; i++) {
+      const reg_t byteaddr = addr + (target_big_endian? size-1-i : i);
+      const reg_t bytedata = (reg_t)load_uint8(byteaddr);
```
scottj97: Maybe the `(reg_t)` typecast is no longer necessary? Since the shifting is done in a later step now.

aswaterman: Yeah, I'm OK with this sort of cast being implicit in the assignment.

aswaterman (approved):
LGTM.

PR #1049, opened 2022-07-22
Pay attention to dmcs2.grouptype.

timsifive:


aswaterman:
Looks fine, but can you mention how this was tested?

timsifive:
The history behind this change is that a bug crept into OpenOCD which inverted dmcs2.grouptype. Because spike ignored this bit, the bug didn't show up in testing against spike, but it did show up when running against hardware.

Now that spike pays attention to the bit, I've rerun it with OpenOCD and confirmed that OpenOCD correctly configures halt groups by observing the value OpenOCD gets after writing to dmcs2.

aswaterman (approved):
Cool.

Issue #668, opened 2021-03-05
Allowed exception delegations


scottj97:
The values allowed for `medeleg` in Spike are a bit mysterious to me.

[Table 5.2](https://github.com/riscv/riscv-isa-manual/blob/0453d462a180927169656e6e3f7faf3042b23e5b/src/hypervisor.tex#L386-L409) sets strict requirements for `hedeleg` but `medeleg` leaves many details up to the implementation.

Assuming hypervisor support (misa.H=1), we have:

| Bit | Cause | Allowed in `medeleg`? ([source](https://github.com/riscv/riscv-isa-sim/blob/0cd5a8bf3bdc13979a355e3d87e43a7697edb153/riscv/processor.cc#L996-L1010)) | Allowed in `hedeleg`? ([source](https://github.com/riscv/riscv-isa-sim/blob/853320fc115da85da34c24ae08be583b51a877b0/riscv/processor.cc#L1121-L1133)) |
| --- | ----- | ------------------------------ | ----------------------------- |
| 0   | Instruction address misaligned   | Yes | Yes |
| 1   | Instruction access fault         | No (?) | Yes |
| 2   | Illegal instruction              | No (?) | Yes |
| 3   | Breakpoint                       | Yes | Yes |
| 4   | Load address misaligned          | No (?) | Yes |
| 5   | Load access fault                | No (?) | Yes |
| 6   | Store/AMO address misaligned     | No (?) | Yes |
| 7   | Store/AMO access fault           | No (?)| Yes |
| 8   | Environment call from U-mode     | Yes | Yes |
| 9   | Environment call from S-mode     | Yes | No |
| 10  | Environment call from VS-mode    | Yes | No ~~(though allowed by spec)~~ |
| 11  | Environment call from M-mode     | No (as recommended by spec) | No |
| 12  | Instruction page fault           | Yes | Yes |
| 13  | Load page fault                  | Yes | Yes |
| 15  | Store/AMO page fault             | Yes | Yes |
| 20 | Instruction guest page fault | Yes | No |
| 21 | Load guest page fault | Yes | No |
| 22 | Virtual instruction fault | Yes | No |
| 23 | Store/AMO guest page fault | Yes | No |

[...]

Issue #194, opened 2018-04-20
`medeleg` doesn't support most exceptions


msuozzo:
Currently, QEMU supports a large number of delegable exceptions [1] ([src](https://github.com/riscv/riscv-qemu/blob/riscv-all/target/riscv/csr.c#L218-L233)) while spike only supports a handful [2] ([src](https://github.com/riscv/riscv-isa-sim/blob/master/riscv/processor.cc#L393-L399)). One notable exclusion is `ILLEGAL_INST`. Any reason for this disparity?

[1]: From riscv-qemu/targets/riscv/csr.c
```
static const target_ulong delegable_excps =
    (1ULL << (RISCV_EXCP_INST_ADDR_MIS)) |
    (1ULL << (RISCV_EXCP_INST_ACCESS_FAULT)) |
    (1ULL << (RISCV_EXCP_ILLEGAL_INST)) |
    (1ULL << (RISCV_EXCP_BREAKPOINT)) |
    (1ULL << (RISCV_EXCP_LOAD_ADDR_MIS)) |
    (1ULL << (RISCV_EXCP_LOAD_ACCESS_FAULT)) |
    (1ULL << (RISCV_EXCP_STORE_AMO_ADDR_MIS)) |
    (1ULL << (RISCV_EXCP_STORE_AMO_ACCESS_FAULT)) |
    (1ULL << (RISCV_EXCP_U_ECALL)) |
    (1ULL << (RISCV_EXCP_S_ECALL)) |
    (1ULL << (RISCV_EXCP_H_ECALL)) |
    (1ULL << (RISCV_EXCP_M_ECALL)) |
    (1ULL << (RISCV_EXCP_INST_PAGE_FAULT)) |
    (1ULL << (RISCV_EXCP_LOAD_PAGE_FAULT)) |
    (1ULL << (RISCV_EXCP_STORE_PAGE_FAULT));
```

[2]: From riscv-isa-sim
```
reg_t mask =
    (1 << CAUSE_MISALIGNED_FETCH) |
    (1 << CAUSE_BREAKPOINT) |
    (1 << CAUSE_USER_ECALL) |
    (1 << CAUSE_FETCH_PAGE_FAULT) |
    (1 << CAUSE_LOAD_PAGE_FAULT) |
    (1 << CAUSE_STORE_PAGE_FAULT);
```

aswaterman:
[...]

PR #1050, opened 2022-07-26
Add additional bits to medeleg

mehnadnerd:
Pursuant to https://github.com/riscv-software-src/riscv-isa-sim/issues/668 and https://github.com/riscv-software-src/riscv-isa-sim/issues/194, allowing for additional exceptions to be delegated from M-mode.
It is implementation-defined whether these bits are defined or are read-only-zero.
QEMU implements the added bits (Fetch/Load/StoreAMO access, Load/StoreAMO misalignment, and illegal instruction). (https://github.com/qemu/qemu/blob/f6cce6bcb2ef959cdd4da0e368f7c72045f21d6d/target/riscv/csr.c#L813)
ECALL_FROM_M is not implemented here because it would have no effect, although QEMU does implement it.
This allows Spike to emulate QEMU and other systems which allow for the delegation of such exceptions.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -822,7 +822,13 @@ void medeleg_csr_t::verify_permissions(insn_t insn, bool write) const {
 bool medeleg_csr_t::unlogged_write(const reg_t val) noexcept {
   const reg_t mask = 0
     | (1 << CAUSE_MISALIGNED_FETCH)
+    | (1 << CAUSE_FETCH_ACCESS)
+    | (1 << CAUSE_ILLEGAL_INSTRUCTION)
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
No objections here (even though I continue to expect many implementations won't be quite so flexible).

OK with you, too, @scottj97?

scottj97:
No objections, but in #194 you didn't like allowing illegal instruction to be delegated.

aswaterman:
Yeah, I think what I wrote in #194 is still true, but I'm OK with leaving this optionality up to M-mode software, rather than enforcing policy through Spike.

PR #1054, opened 2022-07-30
Move debug_rom fence inside entry_loop.

timsifive:
This greatly reduces the amount of spinning in the debug loop spike does when DEBUG_ROM_FLAGS are written by the debug module, because now it's not getting a cached version over and over again.

Test using riscv-tests/debug. Debugging still works.

Also two changes to make this code build again.


aswaterman (approved):
If this is empirically better, I’m fine with it. But Spike does no data caching and FENCE is a no-op on Spike (since it doesn’t model a relaxed memory consistency model), so I don’t understand why it’s better.

timsifive:
I swear this did make a difference, but now it doesn't anymore. Clearly I don't understand the problem I'm tracking down yet.

PR #1057, opened 2022-08-01
Add Sstc support.

i2h2:
This PR implements Sstc. Tested with OpenSBI and Linux kernel changes from Atish.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -679,7 +679,8 @@ void mip_csr_t::backdoor_write_with_mask(const reg_t mask, const reg_t val) noex
 }
 
 reg_t mip_csr_t::write_mask() const noexcept {
-  const reg_t supervisor_ints = proc->extension_enabled('S') ? MIP_SSIP | MIP_STIP | MIP_SEIP : 0;
+  // MIP_STIP is writable unless SSTC exists and STCE is set in MENVCFG
+  const reg_t supervisor_ints = proc->extension_enabled('S') ? MIP_SSIP | ((state->menvcfg->read() &  MENVCFG_STCE) ? 0 : MIP_STIP) | MIP_SEIP : 0;
   const reg_t vssip_int = proc->extension_enabled('H') ? MIP_VSSIP : 0;
   const reg_t hypervisor_ints = proc->extension_enabled('H') ? MIP_HS_MASK : 0;
   // We must mask off sgeip, vstip, and vseip. All three of these
@@ -979,6 +980,11 @@ reg_t time_counter_csr_t::read() const noexcept {
 
 void time_counter_csr_t::sync(const reg_t val) noexcept {
   shadow_val = val;
+  if (proc->extension_enabled(EXT_SSTC)) {
+    const reg_t mip_val = (shadow_val >= state->stimecmp->read() ? MIP_STIP : 0) |
+      (shadow_val + state->htimedelta->read() >= state->vstimecmp->read() ? MIP_VSTIP : 0);
+    state->mip->backdoor_write_with_mask(MIP_STIP | MIP_VSTIP, mip_val);
+  }
 }
 
 proxy_csr_t::proxy_csr_t(processor_t* const proc, const reg_t addr, csr_t_p delegate):
@@ -1398,5 +1404,34 @@ void henvcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
   if (proc->extension_enabled(EXT_SMSTATEEN)) {
     if ((state->prv < PRV_M) && !(state->mstateen[0]->read() & MSTATEEN0_HENVCFG))
       throw trap_illegal_instruction(insn.bits());
+  }
+}
+
+stimecmp_csr_t::stimecmp_csr_t(processor_t* const proc, const reg_t addr, const reg_t imask):
+  basic_csr_t(proc, addr, 0), intr_mask(imask) {
+}
+
+bool stimecmp_csr_t::unlogged_write(const reg_t val) noexcept {
+  state->mip->backdoor_write_with_mask(intr_mask, state->time->read() >= val ? intr_mask : 0);
+  return basic_csr_t::unlogged_write(val);
+}
+
+virtualized_stimecmp_csr_t::virtualized_stimecmp_csr_t(processor_t* const proc, csr_t_p orig, csr_t_p virt):
+  virtualized_csr_t(proc, orig, virt) {
+}
+
+void virtualized_stimecmp_csr_t::verify_permissions(insn_t insn, bool write) const {
+  virtualized_csr_t::verify_permissions(insn, write);
+
+  // check for read permission to time as enabled by xcounteren
+  state->time_proxy->verify_permissions(insn, false);
+
+  if (!(state->menvcfg->read() & MENVCFG_STCE)) {
+    // access to (v)stimecmp with MENVCFG.STCE = 0
+    if (state->prv < PRV_M)
+      throw trap_illegal_instruction(insn.bits());
+  } else if (state->v && !(state->henvcfg->read() & HENVCFG_STCE)) {
+    // access to vstimecmp with MENVCFG.STCE = 1 and HENVCFG.STCE = 0 when V = 1
+    throw trap_virtual_instruction(insn.bits());
   }
 }
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -208,7 +208,7 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   if (proc->extension_enabled_const(EXT_ZICNTR)) {
     csrmap[CSR_INSTRET] = std::make_shared<counter_proxy_csr_t>(proc, CSR_INSTRET, minstret);
     csrmap[CSR_CYCLE] = std::make_shared<counter_proxy_csr_t>(proc, CSR_CYCLE, mcycle);
-    csrmap[CSR_TIME] = std::make_shared<counter_proxy_csr_t>(proc, CSR_TIME, time);
+    csrmap[CSR_TIME] = time_proxy = std::make_shared<counter_proxy_csr_t>(proc, CSR_TIME, time);
   }
   if (xlen == 32) {
     csr_t_p minstreth, mcycleh;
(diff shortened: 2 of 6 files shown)


riscv/csrs.cc
```diff
@@ -1400,3 +1412,39 @@ void henvcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
       throw trap_illegal_instruction(insn.bits());
   }
 }
+
+bool henvcfg_csr_t::unlogged_write(const reg_t val) noexcept {
+  // mask out read-only 0 bits if menvcfg has them turned off
+  return masked_csr_t::unlogged_write((menvcfg->read() | ~(MENVCFG_PBMTE | MENVCFG_STCE)) & val);
+}
```
scottj97: This does not seem necessary when you are also masking on reads.

riscv/csrs.cc
```diff
@@ -1400,3 +1412,39 @@ void henvcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
...
+stimecmp_csr_t::stimecmp_csr_t(processor_t* const proc, const reg_t addr, const reg_t imask):
+  basic_csr_t(proc, addr, 0), intr_mask(imask) {
+}
+
+bool stimecmp_csr_t::unlogged_write(const reg_t val) noexcept {
+  state->mip->backdoor_write_with_mask(intr_mask, 0);
+  // time will return time + htimedelta if v = true
+  if (state->time->read() >= val) {
+    state->mip->backdoor_write_with_mask(intr_mask, intr_mask);
+  }
```
scottj97: Instead of sometimes doing two writes to `mip`, compute one value (either `intr_mask` or 0) and do one write every time.
```
bool pending = state->time->read() >= val;
state->mip->backdoor_write_with_mask(intr_mask, pending ? intr_mask : 0);
```

riscv/processor.cc
```diff
@@ -457,6 +459,12 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
     }
   }
 
+  if (proc->extension_enabled_const(EXT_SSTC)) {
+    stimecmp = std::make_shared<stimecmp_csr_t>(proc, CSR_STIMECMP, MIP_STIP);
+    csrmap[CSR_VSTIMECMP] = vstimecmp = std::make_shared<stimecmp_csr_t>(proc, CSR_VSTIMECMP, MIP_VSTIP);
+    csrmap[CSR_STIMECMP] = std::make_shared<virtualized_stimecmp_csr_t>(proc, stimecmp, vstimecmp);
```
scottj97: What about `stimecmph` and `vstimecmph` for RV32?

scottj97 (changes requested):
Can you point us to the Sstc spec that this is based on? I can't find an up-to-date Sstc spec that explains the `menvcfg.STCE` bit.

gfavor:
The public review spec is at https://drive.google.com/file/d/1m84Re2yK8m_vbW7TspvevCDR82MOBaSX/view.  It says the following regarding the *envcfg.STCE bits:

**Environment Config (menvcfg/henvcfg) Support**
_Enable/disable bits for this extension are provided in the new menvcfg / henvcfg CSRs._

_Bit 63 of menvcfg (or bit 31 of menvcfgh) - named STCE (STimecmp Enable) - enables stimecmp
for S-mode when set to one, and the same bit of henvcfg enables vstimecmp for VS-mode.
These STCE bits are WARL and are hard-wired to 0 when this extension is not implemented._

_When STCE in menvcfg is zero, an attempt to access stimecmp or vstimecmp in a mode other
than M-mode raises an illegal instruction exception, STCE in henvcfg is read-only zero, and STIP
in mip and sip reverts to its defined behavior as if this extension is not implemented._

_When STCE in menvcfg is one but STCE in henvcfg is zero, an attempt to access stimecmp
(really vstimecmp) when V = 1 raises a virtual instruction exception, and VSTIP in hip reverts to
its defined behavior as if this extension is not implemented._


**For stimecmph it says:**

_The stimecmp CSR is a 64-bit register and has 64-bit precision on all RV32 and RV64 systems.
In RV32 only, accesses to the stimecmp CSR access the low 32 bits, while accesses to the
stimecmph CSR access the high 32 bits of stimecmp._

_The CSR numbers for stimecmp / stimecmph are 0x14D / 0x15D (within the Supervisor Trap
Setup block of CSRs)._


**For vstimecmph it says:**

_The vstimecmp CSR is a 64-bit register and has 64-bit precision on all RV32 and RV64 systems.
In RV32 only, accesses to the vstimecmp CSR access the low 32 bits, while accesses to the
vstimecmph CSR access the high 32 bits of vstimecmp._

_The proposed CSR numbers for vstimecmp / vstimecmph are 0x24D / 0x25D (within the Virtual
Supervisor Registers block of CSRs, and mirroring the CSR numbers for stimecmp/stimecmph)._

riscv/csrs.cc
```diff
@@ -679,7 +679,8 @@ void mip_csr_t::backdoor_write_with_mask(const reg_t mask, const reg_t val) noex
 }
 
 reg_t mip_csr_t::write_mask() const noexcept {
-  const reg_t supervisor_ints = proc->extension_enabled('S') ? MIP_SSIP | MIP_STIP | MIP_SEIP : 0;
+  // MIP_STIP is writable unless SSTC exists and STCE is set in MENVCFG
+  const reg_t supervisor_ints = proc->extension_enabled('S') ? MIP_SSIP | (proc->extension_enabled(EXT_SSTC) && (state->menvcfg->read() &  MENVCFG_STCE) ? 0 : MIP_STIP) | MIP_SEIP : 0;
```
scottj97: Since `menvcfg.STCE` is masked to 0 when Sstc is not enabled, there's no need to condition the STIP bit with `proc->extension_enabled(EXT_SSTC)`.

riscv/csrs.cc
```diff
@@ -1400,3 +1412,39 @@ void henvcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
...
+}
+
+stimecmp_csr_t::stimecmp_csr_t(processor_t* const proc, const reg_t addr, const reg_t imask):
+  basic_csr_t(proc, addr, 0), intr_mask(imask) {
+}
+
+bool stimecmp_csr_t::unlogged_write(const reg_t val) noexcept {
+  state->mip->backdoor_write_with_mask(intr_mask, 0);
+  // time will return time + htimedelta if v = true
+  if (state->time->read() >= val) {
```
scottj97: What if `val==0`? This is special-cased in `time_counter_csr_t::sync()` so why not here?

riscv/csrs.cc
```diff
@@ -979,6 +980,17 @@ reg_t time_counter_csr_t::read() const noexcept {
 
 void time_counter_csr_t::sync(const reg_t val) noexcept {
   shadow_val = val;
+  if (proc->extension_enabled(EXT_SSTC)) {
+    auto timer_val = state->stimecmp->read();
+    // do nothing if timer is 0
+    if (timer_val && shadow_val >= timer_val) {
```
scottj97: Is there a case where `shadow_val` has wrapped around to 0, so we should turn off STIP?

riscv/csrs.cc
i2h2: I'm being pedantic here because writes to henvcfg would retain values to those bits when the corresponding bits are off in menvcfg and later turned on (not technically read-only).

riscv/csrs.cc
i2h2: 0 is technically not a special value and there's no architectural reset value for xtimecmp - I just didn't want the interrupts to fire spuriously at boot time. I can make the reset value -1 or have the interrupts assert the way CLINT is implemented.

scottj97: In `time_counter_csr_t::sync()` you will never turn on STIP if `stimecmp==0`. But here, you will. Why is that behavior different?

It's okay to have interrupts firing at boot time, and MTIP already does (iirc), because `mstatus.MIE=0` will prevent any trap from being taken.

riscv/csrs.cc
scottj97: If no CSRW is performed while `menvcfg` is masking these bits, the old values will be revealed when `menvcfg` re-enables these bits. If any CSRW happens, this code will 0 those two bits. That is pretty bizarre behavior.

I'm not sure what the spec requires in this case, but I'm pretty sure it doesn't require this behavior. If it's like other bits that are read-only 0 and later become writable, the bits become undefined at the time they become writable. (For example, hypervisor-only state [such as `mstatus.GVA`] when `misa.H` goes from 0 to 1.)

riscv/processor.cc
```diff
@@ -462,7 +462,12 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   if (proc->extension_enabled_const(EXT_SSTC)) {
     stimecmp = std::make_shared<stimecmp_csr_t>(proc, CSR_STIMECMP, MIP_STIP);
     csrmap[CSR_VSTIMECMP] = vstimecmp = std::make_shared<stimecmp_csr_t>(proc, CSR_VSTIMECMP, MIP_VSTIP);
-    csrmap[CSR_STIMECMP] = std::make_shared<virtualized_stimecmp_csr_t>(proc, stimecmp, vstimecmp);
+    auto virtualized_stimecmp = std::make_shared<virtualized_stimecmp_csr_t>(proc, stimecmp, vstimecmp);
+    csrmap[CSR_STIMECMP] = virtualized_stimecmp;
```
scottj97: You need to use `rv32_low_csr_t` for the lower half; otherwise CSRW will erroneously zero the upper 32 bits. See above for `CSR_HENVCFG` for an example.

scottj97:
Excellent; this is much improved. Can you please squash the commits together into one, so we don't have the bugs+corrections in the history anymore.

riscv/csrs.cc
```diff
@@ -1400,3 +1406,30 @@ void henvcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
...
+}
+
+virtualized_stimecmp_csr_t::virtualized_stimecmp_csr_t(processor_t* const proc, csr_t_p orig, csr_t_p virt):
+  virtualized_csr_t(proc, orig, virt) {
+}
+
+void virtualized_stimecmp_csr_t::verify_permissions(insn_t insn, bool write) const {
+  // check for read permission to time as enabled by xcounteren
+  state->time_proxy->verify_permissions(insn, false);
+
```
liweiwei90, line 1428: This seems not right. CSR_TIME is a user mode CSR. If only  verify_permissions for CSR_TIME is reused here, access from U/VU mode for stimecmp/vstimecmp may be allowed.

liweiwei90 (changes requested):

riscv/csrs.cc
i2h2: stimecmp is in S-mode R/W space and vstimecmp is in HS-mode R/W space. All the verify_permission virtual methods (including counter_proxy_csr_t's) should call the base class' csr_t::verify_permission that checks csr_priv. Please correct me if I'm wrong.

scottj97: But this method does not call its own base class's `verify_permission()`, only the `verify_permission()` of `time_proxy`, which will check the U-mode permissions appropriate to `time`. (Right?)

i2h2: `timer_proxy->verify_permission` -> `counter_proxy_csr_t::verify_permissions` -> `proxy_csr_t::verify_permissions` -> `csr_t::verify_permissions`, which should throw an exception if U mode is accessing S/HS mode CSRs.

`counter_proxy_csr_t::verify_permissions` then does extra checking of xcounteren and throws the appropriate exception.

scottj97: The `time_proxy` object has a CSR address of `CSR_TIME` and therefore a `csr_priv` of `PRV_U` so its parent classes' methods will never check the access permissions of `CSR_STIMECMP` or `CSR_VSTIMECMP`.

i2h2: You're right. I need to call `virtualized_csr_t::verify_permissions` explicitly.

scottj97 (approved):

aswaterman (approved):

PR #1059, opened 2022-08-03
add stateen related check to frm/fflags

liweiwei90:
Add stateen related check to frm/fflags and then apply to fcsr implicitly
Fix exception type for accessing senvcfg/henvcfg/hstateen: Illegal instruciton trap should be raised when accessing senvcfg/henvcfg/hstateen if related bit of mstateen is zero in VU mode

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1324,10 +1324,9 @@ bool hstateen_csr_t::unlogged_write(const reg_t val) noexcept {
 }
 
 void hstateen_csr_t::verify_permissions(insn_t insn, bool write) const {
-  masked_csr_t::verify_permissions(insn, write);
-
   if ((state->prv < PRV_M) && !(state->mstateen[index]->read() & MSTATEEN_HSTATEEN))
     throw trap_illegal_instruction(insn.bits());
+  masked_csr_t::verify_permissions(insn, write);
 }
 
 // implement class sstateen_csr_t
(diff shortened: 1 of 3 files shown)


scottj97:
[a7de776](https://github.com/riscv-software-src/riscv-isa-sim/pull/1059/commits/a7de776de66a1c1caea8d896e6ff51503b0a46bf) is kind of a big deal...are you sure about this exception cause change?

liweiwei90:
> [a7de776](https://github.com/riscv-software-src/riscv-isa-sim/pull/1059/commits/a7de776de66a1c1caea8d896e6ff51503b0a46bf) is kind of a big deal...are you sure about this exception cause change?

According the section 8.6 of riscv-privileged spec , virtual instruction exception is raised for following cases:

- in VS-mode or VU-mode, attempts to access an implemented non-high-half hypervisor CSR or VS CSR when the same access (read/write) would be allowed in HS-mode, assuming mstatus.TVM=0;
- in VS-mode or VU-mode, if XLEN=32, attempts to access an implemented high-half hypervisor CSR or high-half VS CSR when the same access (read/write) to the CSR’s low-half- partner would be allowed in HS-mode, assuming mstatus.TVM=0;

- in VU-mode, attempts to access an implemented non-high-half supervisor CSR when the same access (read/write) would be allowed in HS-mode, assuming mstatus.TVM=0;
- in VU-mode, if XLEN=32, attempts to access an implemented high-half supervisor CSR when the same access to the CSR’s low-half partner would be allowed in HS-mode, assuming mstatus.TVM=0;

All of them have the same condition the access should be allowed in HS-mode.
If we do the base verify_permissions first, if bit in mstateen0 zero, which means HS-mode access is disallowed, virtual instruction exception  will still be raised if access is from VU mode.

scottj97 (approved):

aswaterman (approved):

PR #1063, opened 2022-08-07
Revert "remove unnecessary ifdef for RISCV_ENABLE_DUAL_ENDIAN"

geekLucian:
This reverts commit https://github.com/riscv-software-src/riscv-isa-sim/commit/f82e54124345f348abaa80ec82d67528a9a8f774. 

[playground](https://github.com/chipsalliance/playground)'s CI [failed](https://github.com/chipsalliance/playground/runs/7710418003?check_suite_focus=true) due to https://github.com/riscv-software-src/riscv-isa-sim/commit/f82e54124345f348abaa80ec82d67528a9a8f774. Reverting this commit solves the problem.

### DIFF
--- a/fesvr/htif.h
+++ b/fesvr/htif.h
@@ -31,18 +31,26 @@ class htif_t : public chunked_memif_t
 
   template<typename T> inline T from_target(target_endian<T> n) const
   {
+#ifdef RISCV_ENABLE_DUAL_ENDIAN
     memif_endianness_t endianness = get_target_endianness();
     assert(endianness == memif_endianness_little || endianness == memif_endianness_big);
 
     return endianness == memif_endianness_big? n.from_be() : n.from_le();
+#else
+    return n.from_le();
+#endif
   }
 
   template<typename T> inline target_endian<T> to_target(T n) const
   {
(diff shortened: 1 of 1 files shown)


sequencer:
This does break the playground regression CI. Glad you have figured the commit.
However I'm not so sure about reverting is the correct way or not.
@liweiwei90 remind me to find an another way by setting `set_target_endianness` in the DTM codebase, this might be a better solution than reverting this commit.

aswaterman:
I will wait for y’all to resolve that question before proceeding.

liweiwei90:
> @liweiwei90 remind me to find an another way by setting `set_target_endianness` in the DTM codebase, this might be a better solution than reverting this commit.

Yeah. Or we can revert partial of the commit. I think only the modification in htif.h really matters.

geekLucian:
> > @liweiwei90 remind me to find an another way by setting `set_target_endianness` in the DTM codebase, this might be a better solution than reverting this commit.
> 
> Yeah. Or we can revert partial of the commit. I think only the modification in htif.h really matters.

I think changing DTM codebase prevails, downstream projects shouldn't interfere with upstream's incompatible changes.

PR #1068, opened 2022-08-11
Unify PMPCFGx behaviour with PMPADDRx where PMP is disabled

GregAC:
Previously any access to the PMPADDRx CSRs when no PMP regions were
configured would result in an illegal instruction trap, whilst
PMPCFGx registers would act as WARL, ignoring writes and reading as 0.

This unifies the behaviour so both PMPADDRx and PMPCFGx CSRs produce an
illegal instruction trap when accessed when no PMP regions are
configured.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -218,6 +218,16 @@ pmpcfg_csr_t::pmpcfg_csr_t(processor_t* const proc, const reg_t addr):
   csr_t(proc, addr) {
 }
 
+void pmpcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
+  csr_t::verify_permissions(insn, write);
+  // If n_pmp is zero, that means pmp is not implemented hence raise
+  // trap if it tries to access the csr. I would prefer to implement
+  // this by not instantiating any pmpcfg_csr_t for these regs, but
+  // n_pmp can change after reset() is run.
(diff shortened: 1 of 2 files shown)


riscv/csrs.cc
```diff
@@ -218,6 +218,16 @@ pmpcfg_csr_t::pmpcfg_csr_t(processor_t* const proc, const reg_t addr):
   csr_t(proc, addr) {
 }
 
+void pmpcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
+  csr_t::verify_permissions(insn, write);
+  // If n_pmp is zero, that means pmp is not implemented hence raise
+  // trap if it tries to access the csr. I would prefer to implement
+  // this by not instantiating any pmpaddr_csr_t for these regs, but
```
scottj97: Should refer to pmpcfg_csr_t here in the comment

scottj97 (approved):
I'm not excited about the copy/paste from `pmpaddr_csr_t` but in this case (1) it's simple enough, (2) the alternative is more obscure, and (3) this is code I plan to eliminate once we improve the configuration system (as the comment explains).

riscv/csrs.cc
GregAC: Ah yes, well spotted

aswaterman (approved):
Agreed.

PR #1072, opened 2022-08-22
Fix redundant loops when calculating vrgather.vi.

yorange1:
There is a redundant for loop in the vrgather_vi calculation insns file, which may lead to many repeated calculations. So I removed the redundant loop and made it the same as vrgather_vx.


aswaterman:
Agreed the `vrgather.vi` implementation should look just like the `vrgather.vx` implementation, modulo the rs1 operand.

aswaterman (approved):

PR #1074, opened 2022-08-26
Fix tval on illegal instruction faults with long illegal instruction

YenHaoChen:
The current spike implementation does not include the ILEN (maximum
instruction length supported by the implementation), which is required
to constrain the value of tval on an illegal instruction exception.

Consider an implementation supporting only an RV64I base instruction
set. The ILEN is 32 bits (spec sec. 1.5), and the MXLEN is 64 bits (spec
sec. 5.1). Under an illegal instruction exception with the instruction
longer than the ILEN, the mtval should contain the first ILEN (32 bits)
of the faulting instruction. However, the current spike implementation
lets the mtval be the instruction's first MXLEN (64 bits).

To fix this bug, this PR adds a new variable ilen into the class
processor_t. The processor_t::reset() obtains the ilen from the class
isa_parser_t (along with the xlen) because the ilen depends on the
supported extensions. The class isa_parser_t decides the max_ilen during
the construction (along with the max_xlen).

To my knowledge, all extensions have at most 32-bit instructions. Thus,
the max_xlen is always 32 bits.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -949,7 +949,10 @@ reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek)
 
 reg_t illegal_instruction(processor_t* p, insn_t insn, reg_t pc)
 {
-  throw trap_illegal_instruction(insn.bits());
+  // The illegal instruction can be longer than ILEN bits, where the tval will
+  // contain the first ILEN bits of the faulting instruction. We hard-code the
+  // ILEN to 32 bits since all official instructions have at most 32 bits.
+  throw trap_illegal_instruction(insn.bits() & 0xffffffffULL);
 }
 
 insn_func_t processor_t::decode_insn(insn_t insn)


aswaterman:
Thanks for making this PR.

This is only one possible fix. Another is to say that Spike is always an ILEN=64 machine, regardless of which extensions are enabled.

Yet another fix is to remove all extensions with instructions longer than 32 bits, since no standard ones exist, anyway. (The length encoding isn’t even a ratified standard.)

I’ll let others weigh in, but my preference is to avoid adding extra complexity for a case that isn’t standard.

scottj97:
> To my knowledge, all extensions have at most 32-bit instructions. Thus, the max_xlen is always 32 bits.

I assume this was a typo, and that's supposed to say max_ilen?

riscv/processor.cc
```diff
@@ -949,7 +950,8 @@ reg_t processor_t::get_csr(int which, insn_t insn, bool write, bool peek)
 
 reg_t illegal_instruction(processor_t* p, insn_t insn, reg_t pc)
 {
-  throw trap_illegal_instruction(insn.bits());
+  // the illegal instruction may longer than ILEN bits
+  throw trap_illegal_instruction(insn.bits() & ((reg_t(1)<<(p->get_ilen()))-1));
```
scottj97: This line is the only functional change here, yes? The rest looks like a lot of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) only to ultimately hard-code an ilen of 32.

Why not just hard-code the 32-bit ilen here, and leave a comment to that effect.

scottj97 (changes requested):

riscv/processor.cc
aswaterman: Unrelated comment: this will overflow the shift amount for ILEN=64. Can be avoided by writing (2<<(x-1))-1 instead of (1<<x)-1. But probably irrelevant for the reason Scott mentioned.

YenHaoChen:
> > To my knowledge, all extensions have at most 32-bit instructions. Thus, the max_xlen is always 32 bits.
> 
> I assume this was a typo, and that's supposed to say max_ilen?

Yes, that's a typo and supposed to be max_ilen.

riscv/processor.cc
YenHaoChen: @aswaterman @scottj97 
Thanks for your comments. Since there is no instruction longer than 32 bits, I think simplicity assuming 32-bit ILEN makes sense for most implementations.
I updated the patch and commit log with your comments, which avoids complex modification by hard-coding the ILEN.

aswaterman (approved):
SGTM.  @scottj97 can you merge once you approve?

scottj97 (approved):

PR #1079, opened 2022-09-06
vmv<nf>r.v depends on vtype, and therefore should check vill

### LINKED ISSUE #1078: Is 'require_vector_novtype'  correct for  vmvnfr.v instruction now ?
I have seen  that vmvnfr_v.h  was fixed to take vsew into account,  but it still use 'require_vector_novtype' function , which won't raise a trap if vill is set.  As spec said, "If the vill bit is set, then any attempt to execute a vector instruction that depends upon vtype will raise an illegal-instruction exception". 
Also, "These instructions are intended to aid compilers to shuffle vector registers without needing to know or change vl or vtype." seems contradictory now.

aswaterman:
Resolves #1078

### DIFF
--- a/riscv/insns/vmvnfr_v.h
+++ b/riscv/insns/vmvnfr_v.h
@@ -1,4 +1,4 @@
-// vmv1r.v vd, vs2
+// vmv<nf>r.v vd, vs2
 require_vector(true);
 const reg_t baseAddr = RS1;
 const reg_t vd = insn.rd();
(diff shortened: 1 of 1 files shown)


aswaterman:
cc @chihminchao

scottj97 (approved):
I haven't bothered to understand this change, but I confirm it doesn't seem to break anything.

aswaterman:
WFM.

PR #1081, opened 2022-09-09
Non-functional changes for RVV

liweiwei90:
Remove redundant require_vm/require_vector in current RVV macros. and remove unnecessary argument alu from require_vector_notype.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -306,12 +306,10 @@ do { \
     WRITE_VSTATUS; \
     dirty_vs_state; \
   } while (0);
-#define require_vector_novtype(is_log, alu) \
+#define require_vector_novtype(is_log) \
   do { \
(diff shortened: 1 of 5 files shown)


aswaterman (changes requested):
I would support merging the first three commits, but I don't think the last three commits are an improvement.

liweiwei90:
The last three are to remove the duplicate require_vector.  However, it needs introduce require_vector to some macros explicitly instead of reusing it in previous macro. I think it's a trade-off.

aswaterman:
Yeah, I understand the motivation.  I think static code duplication is worse than the redundant check (which compilers can optimize).

liweiwei90:
OK. I'll drop them.

aswaterman (approved):

Issue #627, opened 2021-01-13
Breakpoint chain not supported in Spike


TracyCai-16:
In type 2 trigger (address match control), when set one breakpoint with tselect=0 and chain=1, another with tselect=1 and chain=0, the trigger should only fire satisfying both breakpoint settings, according to [debug spec](https://github.com/riscv/riscv-debug-spec/blob/release/riscv-debug-release.pdf) (5.2.9 Match Control - Field: chain).
However, running spike model, it seems spike ignores the first breakpoint (with tselect=0 and chain=1), only takes the second one (with tselect=1 and chain=0) into acount.

Issue #599, opened 2020-11-18
trigger mcontrol chain match issue


KingFrige:
hello, we test spike trigger mcontrol.chain function, set trigger[0-2]'s mcontrol.chain =1  trigger[3]'s mcontrol.chain = 0, match **the value equals tdata2**, addressees(tdata2 of trigger[0-3]) are not equals, according to riscv-debug-manual, **it can't match**, but trigger3 is matched and generate exception.

 Has the feature of trigger mcontrol chain  been implemented?

[trigger_match](https://github.com/riscv/riscv-isa-sim/blob/a4994cba0d26915f46292a9ac7c4ce80094c4755/riscv/processor.h#L341)

timsifive:
In theory I implemented support for mcontrol.chain, but since gdb can't set data+value hardware breakpoints, OpenOCD doesn't use mcontrol.chain so it has not been tested. OpenOCD does have a `wp` command that would let you set such a trigger, but RISC-V OpenOCD hasn't implemented this.

KingFrige:
Thanks

PR #1083, opened 2022-09-13
Fix trigger mcontrol.chain match issue #599 #627

YenHaoChen:
A false condition of variable chain_ok is missing. The false condition should be mcontrol.chain=1 and not matching; otherwise, the chain_ok=true (including initialization).

Related issues:
https://github.com/riscv-software-src/riscv-isa-sim/issues/627
https://github.com/riscv-software-src/riscv-isa-sim/issues/599

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -183,7 +183,7 @@ match_result_t module_t::memory_access_match(action_t * const action, operation_
       return result;
     }
 
-    chain_ok = true;
+    chain_ok = (result!=MATCH_NONE || !triggers[i]->chain());
   }
   return MATCH_NONE;
 }
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -183,7 +183,7 @@ match_result_t module_t::memory_access_match(action_t * const action, operation_
       return result;
     }
 
-    chain_ok = (result!=MATCH_NONE || !triggers[i]->chain());
+    chain_ok = result != MATCH_NONE || !triggers[i]->chain();
   }
   return MATCH_NONE;
 }


scottj97:
I don't really understand the logic here but I do observe a useful change in triggers.

I have a chained breakpoint set up like so:
tdata1[0]=0x200000000000090b # chain=1 match=2(ge) u=1 store=1 load=1
tdata2[0]=0x000000008000300a
tdata1[1]=0x200000000000018b # match=3(lt) u=1 store=1 load=1
tdata2[1]=0x0000000080003013

This represents a breakpoint on 0x8000300a <= addr < 0x80003013.

Now execute a `ld` to 0x80003002, which loads bytes up through 0x80003009.

Before this PR, this `ld` would trigger a breakpoint trap. Why? It wasn't touching any of the bytes in the range.

After this PR, no breakpoint trap is taken on this instruction.

Note that for aligned accesses, Spike isn't checking every byte in the range -- only the first. For misaligned, it checks every byte. For example, `ld` to 0x80003008 does not trigger the above breakpoint following this PR. (The debug spec under `mcontrol.select` recommends but does not require checking every byte.)

YenHaoChen:
*Update commit message.

```
The variable chain_ok is used to indicate if the current trigger
is suppressed by the trigger chain. A true value means the trigger
is either un-chained or matches all previous triggers in the chain,
and a false value means the trigger is chained and mismatches
previous triggers.

A false condition of variable chain_ok is missing. The false
condition should be mcontrol.chain=1 and not matching; otherwise,
the chain_ok=true (including initialization). The bug results in
issues https://github.com/riscv-software-src/riscv-isa-sim/pull/559 and https://github.com/riscv-software-src/riscv-isa-sim/issues/627.

Related issues:
- https://github.com/riscv-software-src/riscv-isa-sim/issues/599
- https://github.com/riscv-software-src/riscv-isa-sim/issues/627

This PR fixes the issues https://github.com/riscv-software-src/riscv-isa-sim/pull/559 and https://github.com/riscv-software-src/riscv-isa-sim/issues/627.
```

YenHaoChen:
@timsifive @scottj97 
Is there anything missing from my side?

riscv/triggers.cc
```diff
@@ -183,7 +183,7 @@ match_result_t module_t::memory_access_match(action_t * const action, operation_
       return result;
     }
 
-    chain_ok = true;
+    chain_ok = (result!=MATCH_NONE || !triggers[i]->chain());
```
timsifive:
```suggestion
    chain_ok = result != MATCH_NONE || !triggers[i]->chain();
```

timsifive (changes requested):
The logic looks right. Did you check that https://github.com/riscv-software-src/riscv-tests/tree/master/debug tests still pass?

YenHaoChen:
> The logic looks right. Did you check that https://github.com/riscv-software-src/riscv-tests/tree/master/debug tests still pass?

I have checked that riscv-tests still passes with ```make all```.

timsifive (approved):

PR #1084, opened 2022-09-13
Fix trigger never fire on executing an instruction on devices

YenHaoChen:
The trigger matching logic searches TLB for instruction data. However, the TLB will not be filled with instructions on devices; thus, triggers cannot fire when executing instructions on devices.

The PR moves the trigger matching logic to another (upper) function, i.e., from translate_insn_addr() to refill_icache(). The movement makes the instruction data naturally available without searching the TLB and, thus, resolves the issue.

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -477,13 +477,11 @@ private:
     } else {
       result = tlb_data[vpn % TLB_ENTRIES];
     }
-    if (unlikely(tlb_insn_tag[vpn % TLB_ENTRIES] == (vpn | TLB_CHECK_TRIGGERS))) {
-      target_endian<uint16_t>* ptr = (target_endian<uint16_t>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr);
-      triggers::action_t action;
-      auto match = proc->TM.memory_access_match(&action, triggers::OPERATION_EXECUTE, addr, from_target(*ptr));
-      if (match != triggers::MATCH_NONE) {
-        throw triggers::matched_t(triggers::OPERATION_EXECUTE, addr, from_target(*ptr), action);
-      }
+    target_endian<uint16_t>* ptr = (target_endian<uint16_t>*)(result.host_offset + addr);
+    triggers::action_t action;
+    auto match = proc->TM.memory_access_match(&action, triggers::OPERATION_EXECUTE, addr, from_target(*ptr));
+    if (match != triggers::MATCH_NONE) {
+      throw triggers::matched_t(triggers::OPERATION_EXECUTE, addr, from_target(*ptr), action);
     }
     return result;
   }


scottj97:
This seems like it could be a significant performance degradation, because every single fetch will have to be checked for triggers now. (Versus before, when only fetches from a page with active triggers would check.)

scottj97 (changes requested):
This change causes some breakpoint traps to change.

My test is proprietary so I cannot share it (sorry) but the setup is:

tdata1[0]=0x2000000000000814
tdata2[0]=0x000000008000281e
tdata1[1]=0x2000000000000000
tdata2[1]=0x000000008000329c
tdata1[2]=0x200000000000010c
tdata2[2]=0x000000008000275d
tdata1[3]=0x2000000000000000
tdata2[3]=0x0000000080003299

Breakpoint 2 is the relevant one here.

pc=8000275a commits with no trap. (compressed instruction)
pc=8000275c takes a breakpoint trap before this change, but no longer does after this change.
pc=80002760 takes a breakpoint trap both before and after this change.

I think it's debatable whether this change is an improvement or not, but at least it ought to be discussed and explained in the PR or commit messages.

aswaterman:
I agree it isn't acceptable for perf reasons to check the triggers on every instruction fetch.

It should be possible to mitigate the performance issue by only checking the triggers on TLB misses.

YenHaoChen:
> This change causes some breakpoint traps to change.
> 
> My test is proprietary so I cannot share it (sorry) but the setup is:
> 
> tdata1[0]=0x2000000000000814 tdata2[0]=0x000000008000281e tdata1[1]=0x2000000000000000 tdata2[1]=0x000000008000329c tdata1[2]=0x200000000000010c tdata2[2]=0x000000008000275d tdata1[3]=0x2000000000000000 tdata2[3]=0x0000000080003299
> 
> Breakpoint 2 is the relevant one here.
> 
> pc=8000275a commits with no trap. (compressed instruction) pc=8000275c takes a breakpoint trap before this change, but no longer does after this change. pc=80002760 takes a breakpoint trap both before and after this change.
> 
> I think it's debatable whether this change is an improvement or not, but at least it ought to be discussed and explained in the PR or commit messages.

The triggers indexed 0, 1, and 3 take no effect. The trigger indexed 2 matches PC greater or equal to 0x8000275d. The PC 0x8000275c is less than 0x8000275d and, thus, should not match regardless of this PR.
``` 
tdata1[0]=0x2000000000000814 # chain, s, execute
tdata2[0]=0x000000008000281e
tdata1[1]=0x2000000000000000 # (never match)
tdata2[1]=0x000000008000329c
tdata1[2]=0x200000000000010c # match=2 (ge), u, executeg
tdata2[2]=0x000000008000275d
tdata1[3]=0x2000000000000000 # (never match)
tdata2[3]=0x0000000080003299
```

Update 2022.09.16: The PC 0x8000275c matches the trigger indexed 2. (https://github.com/riscv-software-src/riscv-isa-sim/pull/1084#issuecomment-1248799393)

scottj97:
Any behavior change in Spike needs to be deliberate, understood, and explained. The commit message here makes it sound like the only change is that breakpoints in devices will now work. But as I've shown, there are more changes. These changes are not explained and therefore this PR is unacceptable in its current form.

YenHaoChen:
I agree PR should clarify all the changes being made. The case seems to demonstrate another bug, which was not expected by this PR and is challenging to analyze with only the above information. Can you share some more information for analysis?

scottj97:
> I agree PR should clarify all the changes being made. The case seems to demonstrate another bug, which was not expected by this PR and is challenging to analyze with only the above information. Can you share some more information for analysis?

Is the information above not sufficient to reproduce the issue? What more information can I provide?

YenHaoChen:
The debug spec recommended comparing multiple accessed addresses. Thus, Accessing pc=0x8000275c will be recommended to compare 0x8000275d as well.

image

The PR is revised to check triggers on TLB misses (and hits) in the same function translate_insn_addr(), which mitigates the performance impact and keeps checking the recommended addresses.

scottj97 (approved):
It looks like this will only check triggers on a TLB miss, because of the `return` on line 473 above.

I confirm this undoes the earlier behavior change that I noticed. I haven't tested the performance impact.

aswaterman (approved):
No discernible perf impact.

aswaterman:
Thanks!

PR #1085, opened 2022-09-15
Detect loading isa-incompatible code

ImanHosseini:
This would detect incompatibility between loaded code and the specified '--isa' option. (only checks xlen)
Issues stemming from unsupported extension are easy to detect from error messages, but xlen is not. See: https://github.com/riscv-software-src/riscv-pk/issues/283 (it really doesn't make sense to have spike load incompatible code, you'd expect spike to just throw an error in this case)

### DIFF
--- a/fesvr/elfloader.cc
+++ b/fesvr/elfloader.cc
@@ -16,7 +16,7 @@
 #include <vector>
 #include <map>
 
-std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry)
+std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry, unsigned required_xlen = 0)
 {
   int fd = open(fn, O_RDONLY);
   struct stat s;
@@ -32,6 +32,10 @@ std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* 
   assert(size >= sizeof(Elf64_Ehdr));
   const Elf64_Ehdr* eh64 = (const Elf64_Ehdr*)buf;
   assert(IS_ELF32(*eh64) || IS_ELF64(*eh64));
+  unsigned xlen = IS_ELF32(*eh64) ? 32 : 64;
+  if (required_xlen != 0 && required_xlen != xlen) {
+    throw incompat_xlen(required_xlen, xlen);
+  }
   assert(IS_ELFLE(*eh64) || IS_ELFBE(*eh64));
   assert(IS_ELF_EXEC(*eh64));
   assert(IS_ELF_RISCV(*eh64) || IS_ELF_EM_NONE(*eh64));
--- a/fesvr/htif.cc
+++ b/fesvr/htif.cc
@@ -82,8 +82,14 @@ htif_t::~htif_t()
 
 void htif_t::start()
 {
-  if (!targs.empty() && targs[0] != "none")
+  if (!targs.empty() && targs[0] != "none") {
+    try {
       load_program();
+    } catch (const incompat_xlen & err) {
+      fprintf(stderr, "Error: cannot execute %d-bit program on RV%d hart\n", err.actual_xlen, err.expected_xlen);
+      exit(1);
+    }
+  }
 
   reset();
 }
@@ -129,7 +135,7 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
   } preload_aware_memif(this);
 
   try {
-    return load_elf(path.c_str(), &preload_aware_memif, entry);
+    return load_elf(path.c_str(), &preload_aware_memif, entry, expected_xlen);
   } catch (mem_trap_t& t) {
     bad_address("loading payload " + payload, t.get_tval());
     abort();
--- a/fesvr/htif.h
+++ b/fesvr/htif.h
@@ -26,7 +26,7 @@ class htif_t : public chunked_memif_t
   int run();
   bool done();
   int exit_code();
-
+  void set_expected_xlen(unsigned int m) { expected_xlen = m; }
   virtual memif_t& memif() { return mem; }
 
   template<typename T> inline T from_target(target_endian<T> n) const
@@ -74,7 +74,7 @@ class htif_t : public chunked_memif_t
   void parse_arguments(int argc, char ** argv);
   void register_devices();
   void usage(const char * program_name);
-
+  unsigned int expected_xlen = 0;
   memif_t mem;
   reg_t entry;
   bool writezeros;
--- a/fesvr/memif.h
+++ b/fesvr/memif.h
@@ -5,6 +5,7 @@
 
 #include <stdint.h>
 #include <stddef.h>
+#include <stdexcept>
 #include "byteorder.h"
 
 typedef uint64_t reg_t;
@@ -77,6 +78,13 @@ public:
 
 protected:
   chunked_memif_t* cmemif;
+};
+
+class incompat_xlen : public std::exception {
+public:
+  const unsigned expected_xlen;
+  const unsigned actual_xlen;
+  incompat_xlen(unsigned _expected_xlen, unsigned _actual_xlen) : expected_xlen(_expected_xlen), actual_xlen(_actual_xlen) {}
 };
 
 #endif // __MEMIF_H
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -202,6 +202,7 @@ int sim_t::run()
 {
   host = context_t::current();
   target.init(sim_thread_main, this);
+  htif_t::set_expected_xlen(isa.get_max_xlen());
   return htif_t::run();
 }
 
--- a/fesvr/elfloader.h
+++ b/fesvr/elfloader.h
@@ -8,6 +8,6 @@
 #include <string>
 
 class memif_t;
-std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry);
+std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry, unsigned required_xlen = 0);
 
 #endif


fesvr/htif.h
```diff
@@ -26,7 +26,8 @@ class htif_t : public chunked_memif_t
   int run();
   bool done();
   int exit_code();
-
+  virtual unsigned int get_max_xlen() {return max_xlen;};
```
aswaterman: Is this actually used?

aswaterman:
This will presumably break non-Spike uses of HTIF, since they don’t know to call set_max_xlen. Perhaps you need to default-initialize max_XLEN to 0, and skip these new checks if it is 0.

I also suggest just using “xlen” instead of “max_xlen” since they are equivalent in this case (the program begins executing under the max XLEN).

fesvr/elfloader.cc
```diff
@@ -32,6 +32,11 @@ std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t*
   assert(size >= sizeof(Elf64_Ehdr));
   const Elf64_Ehdr* eh64 = (const Elf64_Ehdr*)buf;
   assert(IS_ELF32(*eh64) || IS_ELF64(*eh64));
+  if(IS_ELF32(*eh64)){
+    is32bit = true;
+  }else{
+    is32bit = false;
+  }
```
aswaterman, line 38: This can just be a single statement, is32bit = IS_ELF32(*eh64);

fesvr/htif.cc
```diff
@@ -138,7 +138,12 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
 
 void htif_t::load_program()
 {
-  std::map<std::string, uint64_t> symbols = load_payload(targs[0], &entry);
+  bool is32bit = false;
+  std::map<std::string, uint64_t> symbols = load_payload(targs[0], &entry, is32bit);
+  if(is32bit != (32 == max_xlen)){
+    printf("mixing 32bit with 64bit payload!\n");
```
aswaterman: This is an unclear error message; exactly what is being mixed?

Can’t we just delete this code altogether and only have the one error message below?

fesvr/htif.cc
```diff
@@ -157,7 +162,11 @@ void htif_t::load_program()
   for (auto payload : payloads)
   {
     reg_t dummy_entry;
-    load_payload(payload, &dummy_entry);
+    load_payload(payload, &dummy_entry, is32bit);
+    if(is32bit != (32 == max_xlen)){
+      printf("loading %dbit payload on %dbit core!\n", is32bit?32:64, max_xlen);
```
aswaterman: “Error: cannot execute %d-bit program on RV%d hart”

fesvr/htif.cc
```diff
@@ -129,7 +129,7 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
   } preload_aware_memif(this);
 
   try {
-    return load_elf(path.c_str(), &preload_aware_memif, entry);
+    return load_elf(path.c_str(), &preload_aware_memif, entry, is32bit);
```
scottj97: I don't like passing variables by non-const ref because it's not clear at the call site here that `is32bit` will be modified by the function.

Why don't you pass an `unsigned required_xlen` and do the checking inside `load_elf()`?

Maybe inside `load_elf()` isn't the right place to print error messages, so throw a new exception and catch it where appropriate.

scottj97 (changes requested):

fesvr/htif.cc
aswaterman: Agreed

aswaterman: (But my previous comment about non-Spike use cases still stands: after taking Scott’s suggestion into account, only perform the check / throw the exception if xlen has actually been set)

fesvr/htif.h
ImanHosseini: no, removed.

ImanHosseini:
Thanks for taking the time to review. I tried to incorporate all the comments. (responded inline to each one)

fesvr/htif.cc
ImanHosseini: Done: so I only check if xlen was actually set, and no more passing refs: I only pass required_xlen to load_elf and that is also an optional arg so if you don't pass it it wouldn't check (e.g. for elf2hex), and throw an exception if there is an issue.

fesvr/htif.cc
ImanHosseini: Changed to using this format.

fesvr/elf2hex.cc
```diff
@@ -40,6 +40,7 @@ int main(int argc, char** argv)
   htif_hexwriter_t htif(base, width, depth);
   memif_t memif(&htif);
   reg_t entry;
+  bool is32bit;
```
scottj97: No longer needed?

fesvr/elfloader.cc
```diff
@@ -32,6 +34,12 @@ std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t*
   assert(size >= sizeof(Elf64_Ehdr));
   const Elf64_Ehdr* eh64 = (const Elf64_Ehdr*)buf;
   assert(IS_ELF32(*eh64) || IS_ELF64(*eh64));
+  unsigned xlen = IS_ELF32(*eh64)?32:64;
+  if(required_xlen != 0){
+    if(required_xlen != xlen){
```
scottj97: Please format code consistent with the rest of Spike: spaces around binary or ternary operators, space after `if` and before open bracket.

No need for nested `if` here, this is simply `&&`.

fesvr/htif.cc
```diff
@@ -82,8 +82,14 @@ htif_t::~htif_t()
 
 void htif_t::start()
 {
-  if (!targs.empty() && targs[0] != "none")
-      load_program();
+  if (!targs.empty() && targs[0] != "none"){
+      try{
+        load_program();
+      } catch( const std::runtime_error& err){
```
scottj97: It's not a good idea to use `std::runtime_error` and assume it could only come from the one new `throw` that you added. This could inadvertently catch unrelated things. The [C++ Core Guidelines](http://isocpp.github.io//CppCoreGuidelines/CppCoreGuidelines#Re-exception-types) best practice is to declare a new custom exception class.

It's a few extra lines of code but it will make this much safer.

scottj97: Use of spaces is erratic and inconsistent

fesvr/htif.h
```diff
@@ -26,7 +26,7 @@ class htif_t : public chunked_memif_t
   int run();
   bool done();
   int exit_code();
-
+  virtual void set_max_xlen(unsigned int m) {max_xlen = m;};
```
scottj97: No need for `virtual`, or the last semicolon

riscv/sim.cc
```diff
@@ -202,6 +202,8 @@ int sim_t::run()
 {
   host = context_t::current();
   target.init(sim_thread_main, this);
+  unsigned int max_xlen = isa.get_max_xlen();
```
scottj97: No need for variable `max_xlen`

scottj97 (changes requested):
Please squash fixes together; we don't need the first implementation followed by the second in the commit history for eternity.

fesvr/elf2hex.cc
ImanHosseini: removed.

fesvr/htif.h
ImanHosseini: Oops! Fixed.

fesvr/htif.cc
ImanHosseini: I've always had trouble with this. I ran this through https://codebeautify.org/cpp-formatter-beautifier for the next commit, if it still looks erratic & there is a specific formatting tool you use, let me know.

ImanHosseini (commented):
Fixed issues. Regarding squashing commits, I thought this is something that maintainer does when merging? The dropdown they use to 'merge' has a 'sqaush and merge' option:
https://cloudfour.com/wp-content/uploads/2019/10/github-merge-options.png
https://cloudfour.com/thinks/squashing-your-pull-requests/

scottj97:
> Fixed issues. Regarding squashing commits, I thought this is something that maintainer does when merging? The dropdown they use to 'merge' has a 'sqaush and merge' option:

We can do it when merging but a PR is not usually best as a single giant commit. In this case, it is reasonably small enough, but generally in Git -- especially for public review like this -- you want a series of simple commits. See [Writing Reviewable Code](https://secure.phabricator.com/book/phabflavor/article/writing_reviewable_code/).

fesvr/htif.cc
```diff
@@ -138,6 +144,7 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
 
 void htif_t::load_program()
 {
+  unsigned required_xlen = max_xlen;
```
scottj97: Delete unused variable

fesvr/htif.h
```diff
@@ -26,7 +27,7 @@ class htif_t : public chunked_memif_t
   int run();
   bool done();
   int exit_code();
-
+  void set_max_xlen(unsigned int m) {max_xlen = m;};
```
scottj97: Add spaces inside `{}`; delete extra semicolon

fesvr/elfloader.cc
```diff
@@ -32,6 +32,10 @@ std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t*
   assert(size >= sizeof(Elf64_Ehdr));
   const Elf64_Ehdr* eh64 = (const Elf64_Ehdr*)buf;
   assert(IS_ELF32(*eh64) || IS_ELF64(*eh64));
+  unsigned xlen = IS_ELF32(*eh64) ? 32 : 64;
+  if(required_xlen != 0 && required_xlen != xlen){
+      throw incompat_xlen(std::to_string(xlen));
+  }
```
scottj97: Indent 2 spaces like the surrounding code; add space after `if`; add space before opening `{`

fesvr/htif.cc
```diff
@@ -82,8 +82,14 @@ htif_t::~htif_t()
 
 void htif_t::start()
 {
-  if (!targs.empty() && targs[0] != "none")
-      load_program();
+  if (!targs.empty() && targs[0] != "none"){
```
scottj97: Add space before opening `{`

fesvr/htif.cc
```diff
@@ -82,8 +82,14 @@ htif_t::~htif_t()
...
+      try {
```
scottj97: Indent 2 spaces like (most of) surrounding code

fesvr/htif.cc
```diff
@@ -82,8 +82,14 @@ htif_t::~htif_t()
...
+        load_program();
+      } catch (incompat_xlen & err) {
```
scottj97: Catch by `const` ref

fesvr/memif.h
```diff
@@ -79,4 +80,15 @@ class memif_t
   chunked_memif_t* cmemif;
 };
 
+class incompat_xlen : public std::exception {
+  private:
+  std::string msg;
+
```
scottj97: Delete trailing whitespace

fesvr/memif.h
```diff
@@ -79,4 +80,15 @@ class memif_t
   chunked_memif_t* cmemif;
 };
 
+class incompat_xlen : public std::exception {
+  private:
```
scottj97: See neighboring classes for the way public/private should be indented. (Other code in Spike is inconsistent here, but let's not add even more diverse indenting styles.)

fesvr/memif.h
```diff
@@ -79,4 +80,15 @@ class memif_t
...
+  std::string msg;
+  
+  public:
+  incompat_xlen(const std::string& _msg) : msg(_msg) {};
```
scottj97: Delete extra semicolon

fesvr/memif.h
```diff
@@ -79,4 +80,15 @@ class memif_t
...
+  const char* what() {
```
scottj97: Having `what()` return a string that's actually a number obscures what the reported problem is. Why not have two public fields in the exception class:
```
  const unsigned expected_xlen;
  const unsigned actual_xlen;
```
Then the contents of the exception object become self-explanatory. No need for `what()` at all.

fesvr/htif.h
```diff
@@ -74,7 +75,7 @@ class htif_t : public chunked_memif_t
   void parse_arguments(int argc, char ** argv);
   void register_devices();
   void usage(const char * program_name);
-
+  unsigned int max_xlen = 0;
```
scottj97: I wonder if HTIF should really describe this as the "maximum" xlen? Here in HTIF, it defines the _only_ xlen allowed in the loaded ELF. Should we call this `expected_xlen` or `enforce_xlen` or perhaps simply `xlen`?

(Also, I considered asking for `std::optional<unsigned>` here, instead of treating 0 as a special value, but in this case it's obvious that 0 is not a valid xlen, so I think the current system is acceptable.)

scottj97 (changes requested):

scottj97:
> https://cloudfour.com/thinks/squashing-your-pull-requests/

Interesting article, but I disagree with their premise:
> As a general rule, when merging a pull request from a feature branch **with a messy commit history**, you should squash your commits. There are exceptions, but in most cases, squashing results in a cleaner Git history that’s easier for the team to read.

The problem is the messy commit history! Fix that, and there's no need to squash. See:
* [The Case for Git Rebase](http://www.darwinweb.net/articles/the-case-for-git-rebase)
* [Understanding the Git Workflow](https://sandofsky.com/workflow/git-workflow/)

fesvr/htif.h
ImanHosseini: Fair point. I also think expected_xlen is more fitting.

aswaterman:
Can you fix the title of the commit?  I'm otherwise satisfied with the contents now.

scottj97 (approved):
Yes, this is perfect now except for the commit message.

Thank you for your patience and persistence with the review process.

ImanHosseini:
I hope I did the squashing right :) (did this: https://stackoverflow.com/a/5201642)
Thank you guys for your patience - I learned so much from you from just 1 PR! - & sorry for the numerous mistakes, will get better next time!

aswaterman (approved):
Yeah, looks good.  Echoing what @scottj97 said, thanks for your patience in dealing with the review process!

PR #1086, opened 2022-09-15
Initialize triggers using default member initializers

scottj97:
Instead of constructor member initializer lists, which require repeating the list of members, and cause ugly merge conflicts when the list of members changes.


aswaterman (approved):
I was unaware of this C++ feature until just now. Agreed with your remarks.

timsifive (approved):

PR #1088, opened 2022-09-22
Run CI with -Werror

aswaterman:
I don't want to enable `-Werror` by default, because we don't know what compilers our users are using and we don't want to make their builds fail for no good reason.  But I still think we should run CI with `-Werror` to avoid new warnings creeping in.

Over time, we can add extra flags to be more pedantic, but this will require experimentation and code improvements.


aswaterman:
Unilaterally merging this, since it's just a CI change and it avoids a merge conflict I'm about to otherwise create.

PR #1089, opened 2022-09-23
Fix or work around ignored-qualifiers, unused-function, unused-parameter, and unused-variable warnings

aswaterman:
Also, add these to the `-Werror` list so CI tests so we won't regress.


scottj97 (changes requested):
I don't have time today for a thorough review but I ran some quick tests.

13dac73a4300b499428dba26568b714b543891e2 breaks everything; segfault on every spike run.

aswaterman:
@scottj97 Sorry for the bother, but can you dig into that further?  I'm not getting any segfaults on either ARM+clang or x86+GCC.  It's possible you've exposed another incremental build bug and this will disappear with a clean build.  If not, can you gdb it and figure out where I screwed up?

aswaterman:
@jerinjoy I reverted the commit that adds the warning to the CI script, since we're using GCC in CI and GCC doesn't understand that flag.

scottj97:
> @scottj97 Sorry for the bother, but can you dig into that further? I'm not getting any segfaults on either ARM+clang or x86+GCC. It's possible you've exposed another incremental build bug and this will disappear with a clean build. If not, can you gdb it and figure out where I screwed up?

Clean build still fails. Running `spike` with no options is successful, but `spike -h` core dumps. My system doesn't have full debug-ability (I'll work on that) so this is all I got from gdb:
```
0x00007ffff6d8b4a7 in ____strtoull_l_internal () from /lib64/libc.so.6
Missing separate debuginfos, use: debuginfo-install glibc-2.17-325.el7_9.x86_64
```

aswaterman:
Oh, sorry for failing to try the obvious.  I'll make a separate PR to add `spike -h` to the CI script.

aswaterman:
@scottj97 this is ready for review whenever you have time.  I know it's a wide-reaching and large PR, so I can split it up if you prefer, but if so, you should recommend the partitioning strategy.  (My own preference is to keep it intact, just for my own productivity, and since I intend no functional changes.)

scottj97:
Core dumps are fixed, but I won't have time to look at the source changes in detail until end of this week at the earliest. Feel free to merge if you need these changes in sooner.

aswaterman:
I'm fine with waiting a week.  Since I made plenty of judgment calls about whether to suppress, fix, or disable the warnings, it would be good to get a second opinion on my heuristics.

aswaterman:
@scottj97 merge conflicts have been fixed

fesvr/device.cc
```diff
@@ -35,6 +35,7 @@ void device_t::handle_command(command_t cmd)
 
 void device_t::handle_null_command(command_t cmd)
 {
+  (void) cmd;
```
scottj97: Would it solve the problem to instead leave the parameters unnamed?
```
void device_t::handle_null_command(command_t)
```

Or maybe (for better documentation):
```
void device_t::handle_null_command(command_t /*cmd*/)
```

riscv/v_ext_macros.h
```diff
@@ -208,9 +208,9 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
   require(P.VU.vsew >= e8 && P.VU.vsew <= e64); \
   require_vector(true); \
   reg_t vl = P.VU.vl->read(); \
-  reg_t sew = P.VU.vsew; \
+  reg_t UNUSED sew = P.VU.vsew; \
```
scottj97, line 211: I assume these are used by some instructions but not all. What happens if an UNUSED variable is actually used? Is it possible the compiler could emit bad code in that case?

aswaterman: Right, `__attribute__((unused))` means the variable _might_ be unused; it doesn't result in undefined behavior if the variable actually is used.

fesvr/device.cc
aswaterman: Agreed, I'll make that change.

softfloat/specialize.h
```diff
@@ -295,6 +298,9 @@ INLINE
 struct uint128 softfloat_commonNaNToF128UI( const struct commonNaN *aPtr )
 {
     struct uint128 uiZ;
+
+    (void) aPtr;
```
aswaterman, line 302: @scottj97 Note I didn't change this one, because unnamed parameters are a C++ feature.  (They are coming to C2x, but they aren't standard yet.)

ci-tests/test-spike
```diff
@@ -6,7 +6,7 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 mkdir build
 cd build
 mkdir install
-CFLAGS="-Werror" $DIR/../configure --prefix=`pwd`/install
+CFLAGS="-Werror -Wignored-qualifiers -Wunused-function -Wunused-parameter -Wunused-variable" $DIR/../configure --prefix=`pwd`/install
```
scottj97: Should we add `-Wnon-virtual-dtor` to this?

scottj97 (approved):

ci-tests/test-spike
aswaterman: Yep, I had excluded it because I thought our CI's compiler didn't support it, but that was my error.  I needed to add it to CXXFLAGS, not CFLAGS.

PR #1093, opened 2022-09-23
Build and install the lib and header dependencies required by Hammer

jerinjoy:
https://github.com/rivosinc/hammer

Signed-off-by: Jerin Joy <joy@rivosinc.com>

### DIFF
--- a/riscv/riscv.mk.in
+++ b/riscv/riscv.mk.in
@@ -3,7 +3,13 @@ get_opcode = $(shell grep ^DECLARE_INSN.*\\\<$(2)\\\> $(1) | sed 's/DECLARE_INSN
 
 riscv_subproject_deps = \
 	fdt \
+	disasm \
+	fesvr \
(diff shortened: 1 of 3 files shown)


aswaterman (changes requested):
It looks like we are currently using tabs, rather than spaces, in these Makefiles.  Maybe that was a questionable decision, but can you follow it for consistency?

Also, please add a trailing slash to the final items in these lists to reduce the diff when the lists are expanded.

aswaterman:
Pleasantly surprised to see that `-fPIC` does not cause a performance regression.

jerinjoy:
I've replaced the spaces with tabs and I've added the trailing slash.

jerinjoy:
I accidentally pushed an empty commit and the PR got closed. Reopened with the requested changes.

aswaterman (approved):
Thank you!

PR #1095, opened 2022-09-23
Fixed vill check in set_vl()

jerinjoy:
vill was being incorrectly set to true by this condition: 

```
vsew > std::min(vflmul, 1.0f) * ELEN
```

The LMUL tells us what fraction of the vector register will be used so VLEN (vector register size) should be used instead of ELEN (maximum element size).

Signed-off-by: Jerin Joy <joy@rivosinc.com>

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -531,7 +531,7 @@ reg_t processor_t::vectorUnit_t::set_vl(int rd, int rs1, reg_t reqVL, reg_t newT
     vma = extract64(newType, 7, 1);
 
     vill = !(vflmul >= 0.125 && vflmul <= 8)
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
This is actually correct as-is.  As the spec says, "The use of vtype encodings with LMUL < SEWMIN/ELEN is reserved."

aswaterman:
Note that means that it's not technically incorrect for an implementation to do what you are proposing--it's a valid implementation decision to define the reserved behavior in that way.  But Spike should be more conservative to catch software portability problems.

jerinjoy:
Thanks for the clarification. Closing.

PR #1099, opened 2022-09-26
support arrow/home/end keys in debug mode

XiaJin-RiVAI:
support the arrow/home/end keys to move the cursor and trace back the history commands in the debug mode

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -21,11 +21,39 @@
 #include <math.h>
 
 #define MAX_CMD_STR 40 // maximum possible size of a command line
+#define BITS_PER_CHAR 8
 
 #define STR_(X) #X      // these definitions allow to use a macro as a string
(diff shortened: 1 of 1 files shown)


scottj97 (approved):
This change is nicely isolated to one file, but it does seem awfully big for what it does.

The alternative we discussed, of integrating a third-party library, would probably be much fewer lines of code in Spike, but would introduce another dependency, which we'd probably want to make optional, which means we'd need to make it work with or without that library. Which can be a mess too.

So I'm okay with this as it is. Reviewed, but untested.

aswaterman (approved):

aswaterman (approved):
@scottj97 IMO, this is compact enough that adding another external dependence seems worse.

aswaterman:
This introduces some new warnings when compiling with Clang, but I'll fix them separately.

Issue #1102, opened 2022-09-28
Maby a bug about `bne` instruction?


lhtin:
Hi, when I run a user program like the one below. I got an unexpected behavior: the control flow goto PC `0x10102` instead of `0x100e8` when running the instruction in PC `0x100e4` where a4 and a3 are the same value `#afee619d`. You can download [the zip elf](https://github.com/riscv-software-src/riscv-isa-sim/files/9662521/testO3.zip) and reproduce use this command: `spike -d --isa=rv32gcv_zfh pk testO3.elf`. Thanks.

spike debug output:

```
: until pc 0 100e4
bbl loader
: reg 0
zero: 0x00000000  ra: 0x00010160  sp: 0x7ffffd80  gp: 0x00020c60
  tp: 0x00000000  t0: 0x00010140  t1: 0x0000000f  t2: 0x00000000
  s0: 0x00000000  s1: 0x00000000  a0: 0x00000001  a1: 0x7ffffda4
  a2: 0x00000000  a3: 0xafee619d  a4: 0xafee619d  a5: 0x00006b5a
  a6: 0x00000000  a7: 0x00000000  s2: 0x00000000  s3: 0x00000000
  s4: 0x00000000  s5: 0x00000000  s6: 0x00000000  s7: 0x00000000
  s8: 0x00000000  s9: 0x00000000 s10: 0x00000000 s11: 0x00000000
  t3: 0x00000000  t4: 0x00000000  t5: 0x00000000  t6: 0x00000000
:
core   0: 0x000100e4 (0x00d71f63) bne     a4, a3, pc + 30
:
core   0: 0x00010102 (0x1a41a683) lw      a3, 420(gp)
:
core   0: 0x00010106 (0x1a01a603) lw      a2, 416(gp)
:
```

objdump output:

```
000100b2 <main>:
   100b2:       1a01a703                lw      a4,416(gp) # 20e00 <__SDATA_BEGIN__>
   100b6:       1a41a783                lw      a5,420(gp) # 20e04 <__SDATA_BEGIN__+0x4>
   100ba:       1101                    addi    sp,sp,-32
   100bc:       c43a                    sw      a4,8(sp)
   100be:       c63e                    sw      a5,12(sp)
   100c0:       ce06                    sw      ra,28(sp)
   100c2:       003c                    addi    a5,sp,8
   100c4:       c590f057                vsetivli        zero,1,e64,m2,ta,mu
[...]

PR #1103, opened 2022-09-28
Fix vmv.x.s for RV32

aswaterman:
The Spike internals require that, when XLEN is narrower than reg_t, values be sign-extended to the width of reg_t.

### DIFF
--- a/riscv/insns/vmv_x_s.h
+++ b/riscv/insns/vmv_x_s.h
@@ -1,27 +1,28 @@
 // vmv_x_s: rd = vs2[0]
 require_vector(true);
 require(insn.v_vm() == 1);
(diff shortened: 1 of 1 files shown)


aswaterman:
Merging, since reporter of #1102 confirmed the fix, and it seems pretty cut and dry.

lhtin:
@aswaterman Hello, Can this bugfix patch apply to v1.1.0? Because I got another bug(I opened another [issue](https://github.com/riscv-software-src/riscv-isa-sim/issues/1106) on it) when updating to master but doesn't exist on v1.1.0.

aswaterman:
We don't have resources to do backports.

PR #1105, opened 2022-09-29
Fix trigger priority

YenHaoChen:
This PR corrects the mcontrol trigger priority with other traps, e.g., page fault and address misaligned. (Debug spec, Table 5.2)
image

The spec defines that the mcontrol execute address has a higher priority over the instruction page fault. Additionally, the mcontrol load address has a higher priority over the load page fault. Furthermore, the mcotnrol store address/data have higher priorities over the store page fault. The current implementation does not correctly satisfy the above priorities.

This PR fixes the priority issue with four commits. It adds address-only trigger checking functions first, then performs the address trigger checking with those functions for execution, load, and store in separate commits.

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -471,6 +471,11 @@ private:
     reg_t vpn = addr >> PGSHIFT;
     if (likely(tlb_insn_tag[vpn % TLB_ENTRIES] == vpn))
       return tlb_data[vpn % TLB_ENTRIES];
+    triggers::action_t action;
+    auto match = proc->TM.memory_access_match(&action, triggers::OPERATION_EXECUTE, addr, false);
+    if (match != triggers::MATCH_NONE) {
+      throw triggers::matched_t(triggers::OPERATION_EXECUTE, addr, 0, action);
+    }
     tlb_entry_t result;
     if (unlikely(tlb_insn_tag[vpn % TLB_ENTRIES] != (vpn | TLB_CHECK_TRIGGERS))) {
       result = fetch_slow_path(addr);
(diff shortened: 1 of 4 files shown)


riscv/mmu.h
```diff
@@ -100,6 +100,11 @@ class mmu_t
   // template for functions that load an aligned value from memory
   #define load_func(type, prefix, xlate_flags) \
     type##_t ALWAYS_INLINE prefix##_##type(reg_t addr, bool require_alignment = false) { \
+      if (!matched_trigger) { \
+        matched_trigger = trigger_exception(triggers::OPERATION_LOAD, addr); \
+        if (matched_trigger) \
+          throw *matched_trigger; \
+      } \
```
aswaterman: This needs to be done in such a way as to not slow down the common case.  For a test that is heavy in loads and stores, it results in a 200% increase in runtime, since it calls `trigger_exception` on every load and store!

We need to handle the common case (aligned + TLB hit) with no additional overhead.  It should even be possible to do without any additional `if` statements on the critical path by reordering the code.

aswaterman (changes requested):
In addition to the performance problem, there is a lot of code duplication.  The new functions are all basically duplicates of existing functions with only minor changes.  This needs to be refactored so that there is almost no new code.

YenHaoChen:
Thanks for the comments. The revised PR moves the trigger checking to non-critical paths to mitigate the performance impact. Additionally, it reused the code by adding a new argument instead of copying whole functions.

scottj97 (approved):
Code looks decently good. Can you tell us how you've tested this?

aswaterman (approved):
Thanks, this refactored version addresses my concerns.  I'll wait for @scottj97 to merge after you answer his question about testing.

aswaterman:
Merging now, but please follow up with an explanation of how this was tested.

YenHaoChen:
Sorry for the lateness. I am preparing a more solid test plan for the commits and will follow up.

YenHaoChen:
Hello, the code has been tested with the following cases.
1. An instruction with both fetch page fault and address trigger breakpoint
2. An instruction with both load page fault and address trigger breakpoint
3. An instruction with both store page fault and address trigger breakpoint

Before the patch, the above cases trap to page faults. After the patch, the above cases fires trigger breakpoint as required by the RISC-V spec.

The patch aims to fix the priority between the misaligned fault and the trigger address breakpoint, which is passed with normal accesses. Unfortunately, it fails to fix the priority between misaligned AMO and trigger breakpoints. I will have another PR for the issue.

Is the above description enough for you? @scottj97 @aswaterman

aswaterman:
Any chance you can contribute your tests to https://github.com/riscv-software-src/riscv-tests?

YenHaoChen:
Unfortunately, my tests are not compatible with the repository.

scottj97:
Thank you for the explanation.

PR #1107, opened 2022-10-06
Simplify handling of load/store/fetch slow-path cases; fix two minor trigger bugs

aswaterman:
The load/store code in particular has been becoming difficult to maintain because of the competing interests of keeping the common case fast and correctly handling the less-common cases (triggers and misaligned accesses).  Refactor so that only the fast path is in the harder-to-maintain inlined code.

As a side effect, misaligned accesses now behave the same as aligned accesses with respect to triggers: only the first byte is checked.

This refactoring suggested several opportunities for DRY in the handling of triggers.  Address that, too.

Finally, being forced to understand the code revealed two bugs:
- Fetch triggers on big-endian machines were using the wrong endianness (instruction fetch is always little-endian)
- Trigger-after was relying on memory accesses being idempotent, which is not entirely true: MMIO loads can have side effects, and re-executing them because of a data value trigger isn't correct.  Avoiding re-execution is also a bit cleaner.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -194,15 +194,13 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
   }
 }
 
-void mmu_t::store_slow_path(reg_t addr, reg_t len, const uint8_t* bytes, uint32_t xlate_flags, bool actually_store)
+void mmu_t::store_slow_path_intrapage(reg_t addr, reg_t len, const uint8_t* bytes, uint32_t xlate_flags, bool actually_store)
 {
-  if (actually_store) {
-    if (!matched_trigger) {
-      reg_t data = reg_from_bytes(len, bytes);
-      matched_trigger = trigger_exception(triggers::OPERATION_STORE, addr, true, data);
-      if (matched_trigger)
-        throw *matched_trigger;
-    }
+  reg_t vpn = addr >> PGSHIFT;
+  if (xlate_flags == 0 && vpn == (tlb_store_tag[vpn % TLB_ENTRIES] & ~TLB_CHECK_TRIGGERS)) {
+    auto host_addr = tlb_data[vpn % TLB_ENTRIES].host_offset + addr;
+    memcpy(host_addr, bytes, len);
+    return;
   }
 
   reg_t paddr = translate(addr, len, STORE, xlate_flags);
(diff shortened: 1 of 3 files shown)


scottj97 (commented):
0a57a7950d84e074de7bf1fd198162bc4c245b9b causes a change in breakpoint traps, though it is not a bug since the debug spec is flexible here.

I have a chained breakpoint set up like so:
tdata1[0]=0x200000000000090b # chain=1 match=2(ge) u=1 store=1 load=1
tdata2[0]=0x000000008000300a
tdata1[1]=0x200000000000018b # match=3(lt) u=1 store=1 load=1
tdata2[1]=0x0000000080003013

This represents a breakpoint on 0x8000300a <= addr < 0x80003013.

LD to 0x80003004, which loads 0x80003004-0x8000300b, overlaps this range. Before 0a57a7950d84e074de7bf1fd198162bc4c245b9b this was taking a breakpoint trap. After 0a57a7950d84e074de7bf1fd198162bc4c245b9b it no longer takes a breakpoint trap.

This has to do with whether we check all bytes of the access for breakpoints, versus just the first. The debug spec under mcontrol.select recommends but does not require checking every byte.

Spike used to check each byte only for unaligned accesses. For aligned accesses, it was checking only the first. It appears that after 0a57a7950d84e074de7bf1fd198162bc4c245b9b, it is checking only the first byte even for unaligned.

I suppose this is an improvement, because at least it's now consistent about only checking the first byte.

aswaterman:
@scottj97 I realized the same thing around the same time you were writing your comment.  I'll update the commit description to remark on the behavioral change.

I agree the new behavior is preferable to the old behavior for consistency.

Just now, I briefly considered changing the behavior to be consistent but strict (i.e. always check every byte).  But given this PR's behavior is cheaper to implement in HW, for misaligned accesses in particular, I suspect it will be the more popular choice, and therefore it's justifiable to stick with it in Spike.

scottj97:
> Just now, I briefly considered changing the behavior to be consistent but strict (i.e. always check every byte). But given this PR's behavior is cheaper to implement in HW, for misaligned accesses in particular, I suspect it will be the more popular choice, and therefore it's justifiable to stick with it in Spike.

Checking every byte is more desirable to the debugger, though. If it sets a breakpoint on a variable at 0x80004004, then a SD to 0x80004000 doesn't trap, the debugger will miss the change in that variable. I think the debug spec is flawed to allow this behavior.

But, that's a different PR.

I'll try to take a closer look at this change later today.

aswaterman:
Yeah, the alternative is obviously preferable for debugging.  Nevertheless, it's a judgment call whether to model the preferable behavior vs. the behavior more likely to be implemented.

aswaterman:
BTW, thanks for taking the time to analyze this PR in more detail... I'm definitely leaning on both your more extensive tests and your willingness to perform a careful review.

timsifive (approved):
This looks right to me, but most of this code I'm not very familiar with.

riscv/mmu.cc
```diff
@@ -139,12 +149,37 @@ bool mmu_t::mmio_store(reg_t addr, size_t len, const uint8_t* bytes)
   return sim->mmio_store(addr, len, bytes);
 }
 
-void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate_flags)
+void mmu_t::check_triggers(triggers::operation_t operation, reg_t address, bool has_data, reg_t data)
```
scottj97, line 152: Instead of `has_data/data` how about `std::optional<reg_t> data`? Then the code can never accidentally use `data` when it's meaningless.

scottj97 (approved):
This is great.

riscv/mmu.cc
aswaterman: Agreed… I would’ve written it that way in Scala.

aswaterman: It's a larger refactoring than I thought, so I'll make it a second PR.

scottj97:
For the record: 6311f7513aa150797f69ecac906978bb9e9fecbd introduces a bug with AMOs, which is fixed by 98552cdcd0b39ec492531c3ca871041a9df0fb9e (aka 2b7edec86) as part of #1123.

PR #1111, opened 2022-10-10
Report error if an unsupported memory configuration is detected

EccoTheDolphin:
This commit adds error message for the case when an incorrect memory region is
detected. The check for invariant was moved to a separate function to simplify
the code of mem_cfg_t users.

### DIFF
--- a/riscv/cfg.h
+++ b/riscv/cfg.h
@@ -32,18 +32,22 @@ private:
 class mem_cfg_t
 {
 public:
-  mem_cfg_t(reg_t base, reg_t size)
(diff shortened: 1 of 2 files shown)


EccoTheDolphin:
@scottj97 this PR is a result of our discussion [here](https://github.com/riscv-software-src/riscv-isa-sim/pull/1100). Would you kindly review this one?

scottj97 (approved):
Sorry to take so long. Looks good to me!

aswaterman (approved):

PR #1112, opened 2022-10-11
set tval on illegal instruction of aes64ks1i invalid opcode (rcon)

YenHaoChen:
Although the tval of illegal instruction can be either 0 or instruction bits, I think spike prefers the instruction bits more. This commit changes the tval of illegal instruction on an invalid opcode of aes64ks1i.


aswaterman (changes requested):
This should just be `require(enc_rcon <= 0xA);`, which will fill in the `tval` register correctly.

aswaterman:
Manually merged my proposed fix via 86d9fe49eda1fff863a43e682015216a25cc72f3

Thanks for finding this!

PR #1114, opened 2022-10-13
In triggers, use optional<data> instead of {has_data, data}

aswaterman:
No functional change intended.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -162,13 +162,13 @@ void mmu_t::check_triggers(triggers::operation_t operation, reg_t address, bool 
       return;
 
     case triggers::MATCH_FIRE_BEFORE:
-      throw triggers::matched_t(operation, address, data, action);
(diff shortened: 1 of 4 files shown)


timsifive (approved):

riscv/mmu.cc
```diff
@@ -76,7 +76,7 @@ reg_t mmu_t::translate(reg_t addr, reg_t len, access_type type, uint32_t xlate_f
 
 tlb_entry_t mmu_t::fetch_slow_path(reg_t vaddr)
 {
-  check_triggers(triggers::OPERATION_EXECUTE, vaddr, false);
+  check_triggers(triggers::OPERATION_EXECUTE, vaddr, std::nullopt);
```
scottj97: Want to make `std::nullopt` the default value for `data`, so we can just leave off the last parameter here?

aswaterman: sgtm

PR #1115, opened 2022-10-14
Support command-line configuration of number of pmpregions

jerryz123:


scottj97 (changes requested):
There needs to be some kind of sanity check on the requested number.

I know Spike supports 8 and 16. I'm not sure if it supports any other numbers.

spike_main/spike.cc
```diff
@@ -36,6 +36,7 @@ static void help(int exit_code = 1)
   fprintf(stderr, "  --log=<name>          File name for option -l\n");
   fprintf(stderr, "  --debug-cmd=<name>    Read commands from file (use with -d)\n");
   fprintf(stderr, "  --isa=<name>          RISC-V ISA string [default %s]\n", DEFAULT_ISA);
+  fprintf(stderr, "  --pmpregions-<n>      Number of PMP regions [default 16]\n");
```
aswaterman: there's a hyphen where there should be an equal sign

jerryz123:
> There needs to be some kind of sanity check on the requested number.
> 
> I know Spike supports 8 and 16. I'm not sure if it supports any other numbers.

There is an existing sanity check on the requested number where Spike parses the DTS. Since this PR sets pmpregions by changing the DTS, the existing check will still apply here.

https://github.com/riscv-software-src/riscv-isa-sim/blob/26c6795f2f72a369c5c192a5d53643b9558a2691/riscv/sim.cc#L114-L128

scottj97 (approved):

PR #1116, opened 2022-10-14
Add dump memory command to interactive mode

jerryz123:


aswaterman (approved):
I'm fine with this.  Out of curiosity, what's the use case?

Any thoughts, @scottj97?

riscv/interactive.cc
```diff
@@ -774,3 +776,16 @@ void sim_t::interactive_until(const std::string& cmd, const std::vector<std::str
...
+void sim_t::interactive_dumpmems(const std::string& cmd, const std::vector<std::string>& args)
+{
+  for (unsigned i = 0; i < mems.size(); i++) {
+    std::stringstream mem_fname;
+    mem_fname << "mem.0x" << std::hex << mems[i].first << ".bin";
+
+    std::ofstream mem_file;
+    mem_file.open(mem_fname.str());
+    mems[i].second->dump(mem_file);
+    mem_file.close();
```
scottj97: Pass in the filename to the constructor, which will open the file. Then let the destructor do the close.

scottj97 (dismissed):

scottj97:
> I'm fine with this. Out of curiosity, what's the use case?
> 
> Any thoughts, @scottj97?

🤷‍♂️

jerryz123:
> I'm fine with this. Out of curiosity, what's the use case?

This makes it easier to generate an architectural checkpoint from spike execution.

aswaterman:
Makes sense.

PR #1117, opened 2022-10-14
Add command to display PRV in interactive mode

jerryz123:


aswaterman (commented):
Makes sense, but I think we'd also want to print out the virtualization state.  That further suggests pretty-printing the combination of `prv` and `v` as one of M, S, U, VS, or VU.

aswaterman (commented):
IMO, it would be cleaner to separate generating the string from the interactive gunk (e.g. add a method `const char* processor_t::privilege_string()` method).  Also, I think it should `abort()` in the unknown cases, as that represents an internal error.

aswaterman (approved):

PR #1118, opened 2022-10-14
Add interactive mode commands to read mtime/mtimecmp

jerryz123:


aswaterman (changes requested):
`mtime` and `mtimecmp` are 64 bits irrespective of XLEN.  Otherwise, I'm fine with this.

aswaterman (commented):
The two calls to `zext` shouldn't be necessary anymore, either, right?

aswaterman (approved):

PR #1119, opened 2022-10-17
Do not allow memory regions larger than (1ull << MAX_PADDR_BITS)

aap-sc:
Such regions are inaccessible to the hardware being modeled.

### DIFF
--- a/spike_main/spike.cc
+++ b/spike_main/spike.cc
@@ -3,6 +3,7 @@
 #include "cfg.h"
 #include "sim.h"
 #include "mmu.h"
+#include "arith.h"
 #include "remote_bitbang.h"
 #include "cachesim.h"
 #include "extension.h"
@@ -14,6 +15,8 @@
 #include <string>
 #include <memory>
 #include <fstream>
+#include <limits>
+#include <cinttypes>
 #include "../VERSION"
 
 static void help(int exit_code = 1)
@@ -188,7 +191,24 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
       exit(EXIT_FAILURE);
     }
 
-    res.push_back(mem_cfg_t(reg_t(base), reg_t(size)));
+    const unsigned long long max_allowed_pa = (1ull << MAX_PADDR_BITS) - 1ull;
+    assert(max_allowed_pa <= std::numeric_limits<reg_t>::max());
+    mem_cfg_t mem_region(base, size);
+    auto last_pa_region = mem_region.base + mem_region.size - 1;
+    if (last_pa_region > max_allowed_pa) {
+      int bits_required = 64 - clz(last_pa_region);
+      fprintf(stderr, "Unsupported memory region "
+                      "{base = 0x%" PRIX64 ", size = 0x%" PRIX64 "} specified,"
+                      " which requires %d bits of physical address\n"
+                      "    The largest accessible physical address "
+                      "is 0x%llX (defined by MAX_PADDR_BITS constant, which is %d)\n",
+              mem_region.base, mem_region.size, bits_required,
+              max_allowed_pa, MAX_PADDR_BITS);
+      exit(EXIT_FAILURE);
+    }
+
+    res.push_back(mem_region);
+
     if (!*p)
       break;
     if (*p != ',')
--- a/spike_main/spike.cc
+++ b/spike_main/spike.cc
@@ -176,16 +176,15 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
       size += PGSIZE - size % PGSIZE;
 
     if (size != size0) {
-      fprintf(stderr, "Warning: the memory at  [0x%llX, 0x%llX] has been realigned\n"
+      fprintf(stderr, "Warning: the memory at [0x%llX, 0x%llX] has been realigned\n"
                       "to the %ld KiB page size: [0x%llX, 0x%llX]\n",
               base0, base0 + size0 - 1, long(PGSIZE / 1024), base, base + size - 1);
     }
 
     if (!mem_cfg_t::check_if_supported(base, size)) {
-      fprintf(stderr, "unsupported memory region "
+      fprintf(stderr, "Unsupported memory region "
                       "{base = 0x%llX, size = 0x%llX} specified\n",
-                      (unsigned long long)base,
-                      (unsigned long long)size);
+              base, size);
       exit(EXIT_FAILURE);
     }
 


aap-sc:
This commit addresses an issue  (discussed in #1100) when user can configure memory region to have addresses larger than (1 << MAX_PADDR_BITS), while the current spike implementation does not really allow for such accesses.

spike_main/spike.cc
```diff
@@ -189,7 +189,20 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
       exit(EXIT_FAILURE);
     }
 
-    res.push_back(mem_cfg_t(reg_t(base), reg_t(size)));
+    mem_cfg_t mem_region(base, size);
+    const uint64_t max_allowed_pa = (1ull << MAX_PADDR_BITS) - 1ull;
+    if ((mem_region.base + mem_region.size - 1) > max_allowed_pa) {
+      fprintf(stderr, "unsupported memory region "
+                      "{ base = 0x%llX, size = 0x%llX } specified\n"
```
scottj97: [Use `cinttypes`](https://stackoverflow.com/a/9225648/202201) for printing `uint64_t`: `PRIX64`

Or C++ streams to `cerr`.

spike_main/spike.cc
```diff
@@ -189,7 +189,20 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
...
+                      "currently, spike does not support physical adresses "
```
scottj97: addresses is misspelled

spike_main/spike.cc
```diff
@@ -189,7 +189,20 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
...
+                      "larger than 0x%llX)\n",
```
scottj97: closing parenthesis has no matching open parenthesis

spike_main/spike.cc
scottj97: How about we give the user a clue within the error message about why the physical address is limited?

"Based on MAX_PADDR_BITS, the largest accessible physical address is 0x%x"

scottj97 (changes requested):
I'd like the commit message to explain better why this change is good. It's not that "currently, spike may not support such memory regions properly" it's that "Such regions are inaccessible to the hardware being modeled."

liweiwei90:
I just have another question. Do we needs to take the other device memory map into consideration? For example,

- If we use the memory cfg: 0:0x100000000, if we try to access 0x02000004(in both the address space for memory and clint), it will access clint (upper_bound seems take the base address to compare, any misunderstanding?)

- And, if we try to access 0x02100000(out of the clint address space, but in memory address space), however, it seems also will access clint and fail.

spike_main/spike.cc
aap-sc: Thanks! Decided to use PRIX64 since we tend to use fprintf in other places.

spike_main/spike.cc
aap-sc: my bad, fixed.

spike_main/spike.cc
aap-sc: fixed

spike_main/spike.cc
aap-sc: Sure. Updated the error message.

aap-sc:
> I just have another question. Do we needs to take the other device memory map into consideration?

I guess that we should, though I don't think that the current spike interface allows that. To add a device to the bus we specify only the base address. The size of the underlying memory region is not specified.

> And, if we try to access 0x02100000(out of the clint address space, but in memory address space), however, it seems also will access clint and fail.

I guess this is what will happen. 

It is also may be the case that your question should be treated a different issue - that is it does not look like it is **directly** related to this PR.

spike_main/spike.cc
aap-sc: One thing to note - to avoid compiler warnings I had to update the surrounding code a little bit (auto -> explicit uint64_t and started to use PRIx64 in another error message). Hope this is not  a big deal.

liweiwei90:
> It is also may be the case that your question should be treated a different issue - that is it does not look like it is **directly** related to this PR.

Yeah. The current PR is LGTM.  It's just another question that may also be used to limit the memory region.

aap-sc:
@scottj97 , gentle ping. Could you please take a look once again?

spike_main/spike.cc
```diff
@@ -176,20 +177,32 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
       size += PGSIZE - size % PGSIZE;
 
     if (size != size0) {
-      fprintf(stderr, "Warning: the memory at  [0x%llX, 0x%llX] has been realigned\n"
-                      "to the %ld KiB page size: [0x%llX, 0x%llX]\n",
+      fprintf(stderr, "Warning: the memory at "
+                      "[0x%" PRIx64 ", 0x%" PRIx64 "] has been realigned\n"
+                      "to the %ld KiB page size: "
+                      "[0x%" PRIx64 ", 0x%" PRIx64 "]\n",
```
scottj97: These changes are beneficial but independent of the new functionality you've added, so should be in its own commit.

spike_main/spike.cc
```diff
@@ -176,20 +177,32 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
...
       exit(EXIT_FAILURE);
     }
 
-    res.push_back(mem_cfg_t(reg_t(base), reg_t(size)));
+    mem_cfg_t mem_region(base, size);
+    const uint64_t max_allowed_pa = (1ull << MAX_PADDR_BITS) - 1ull;
+    if ((mem_region.base + mem_region.size - 1) > max_allowed_pa) {
+      fprintf(stderr, "Unsupported memory region "
+                      "{base = 0x%" PRIx64 ", size = 0x%" PRIx64 "} specified\n"
+                      "    The largets accessible physical address "
```
scottj97: "Largest" is misspelled

scottj97 (changes requested):

spike_main/spike.cc
aap-sc: ok, will move this part to a separate PR then

scottj97: Not a separate PR. It's related to the other change, so it's reasonable to include here in this PR.

spike_main/spike.cc

spike_main/spike.cc
aap-sc: The patch was refactored (a little bit)

aap-sc:
To improve portability I've decided to stick with `unsigned long long` types and use PRIx64 only when it's required.

spike_main/spike.cc
```diff
@@ -188,7 +190,20 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
...
 
-    res.push_back(mem_cfg_t(reg_t(base), reg_t(size)));
+    const unsigned long long max_allowed_pa = (1ull << MAX_PADDR_BITS) - 1ull;
+    assert(max_allowed_pa <= std::numeric_limits<reg_t>::max());
+    mem_cfg_t mem_region(base, size);
+    if ((mem_region.base + mem_region.size - 1) > max_allowed_pa) {
+      fprintf(stderr, "Unsupported memory region "
+                      "{base = 0x%" PRIx64 ", size = 0x%" PRIx64 "} specified\n"
+                      "    The largest accessible physical address "
+                      "is 0x%llX (defined by MAX_PADDR_BITS constant)\n",
```
scottj97: Be consistent in usage of lower vs uppercase hex digits. This prints:
```
Unsupported memory region {base = 0xffffffffffff0000, size = 0x4000} specified
    The largest accessible physical address is 0xFFFFFFFFFFFFFF (defined by MAX_PADDR_BITS constant)
```

I also wonder if we could make the error message a little easier to read, instead of the user having to count all those digits? Such as:
```
Unsupported memory region {base = 0xffffffffffff0000, size = 0x4000} specified, which requires 64 paddr bits
    The largest accessible physical address is 0xFFFFFFFFFFFFFF (defined by MAX_PADDR_BITS constant, which is 56)
```

scottj97 (changes requested):
A few nits and I think this is good.

"Redundant" is misspelled in the commit message for d2e1173f7ce1773aca1d3d3bb32b7d07bbd1938b

spike_main/spike.cc
aap-sc: > "Redundant" is misspelled

My bad. fixed.

> Be consistent in usage of lower vs uppercase hex digits.

Done.

> I also wonder if we could make the error message a little easier to read

Sure, however I'd like to use `__builtin_clzll` to calculate the number of required physical bits. Do you think it's appropriate? We already use `__builtin_expect` in likely/unlikely macro, so probably this should be ok.. Still I want to double-check

scottj97: I'm okay with that, since it works in both GCC and clang. @aswaterman ?

aswaterman: I've been trying to avoid that kind of thing.  Note that we only use `__builtin_expect` if `__GNUC__` is defined.

Please use this one: https://github.com/riscv-software-src/riscv-isa-sim/blob/ab3225a3ff687fda8b4180f9e4e0949a400d1247/riscv/arith.h#L174

aap-sc: @aswaterman , 
> Please use this one:

Thanks!

> I also wonder if we could make the error message a little easier to read, instead of the user having to count all those digits? 

Addressed

spike_main/spike.cc
```diff
@@ -176,20 +179,36 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
...
-    res.push_back(mem_cfg_t(reg_t(base), reg_t(size)));
+    const unsigned long long max_allowed_pa = (1ull << MAX_PADDR_BITS) - 1ull;
+    assert(max_allowed_pa <= std::numeric_limits<reg_t>::max());
+    mem_cfg_t mem_region(base, size);
+    auto last_pa_region = mem_region.base + mem_region.size - 1;
+    if (last_pa_region > max_allowed_pa) {
+      int bits_required = 64 - clz(last_pa_region);
+      fprintf(stderr, "Unsupported memory region "
+                      "{base = 0x%" PRIX64 ", size = 0x%" PRIX64 "} specified,"
+                      " which requires %d paddr bits\n"
```
aswaterman: I suggest spelling out “physical address” on this line rather than abbreviating it “paddr”.

I’m happy with this PR otherwise. Thank you!

scottj97 (approved):
Excellent, thank you for your patience.

aswaterman (approved):

PR #1120, opened 2022-10-18
Slightly simplify LR/SC implementation

aswaterman:
The call to refill_tlb is a bit baroque and isn't necessary. (Also, the LR instruction will have populated the TLB anyway by invoking load_uintXX, so it doesn't even help performance.)


aswaterman:
In a moment, I'll supersede this with a PR that fixes a bug I just found by inspection.

PR #1122, opened 2022-10-20
Fix minor MMU bugs; clean up MMU some more

aswaterman:
- Load-reserved to MMIO would action side effects despite trapping, i.e. an uncommon precise-exception bug
- Undefined behavior on AMOs due to `memcpy(nullptr, ..., 0)`, though it's unlikely this would've manifested
- Further simplify macros and templates in `mmu_t`
- Template-ize some of the macros

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -199,7 +199,7 @@ void mmu_t::load_slow_path_intrapage(reg_t addr, reg_t len, uint8_t* bytes, uint
   }
 }
 
-void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate_flags, bool UNUSED require_alignment)
+void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate_flags)
 {
   check_triggers(triggers::OPERATION_LOAD, addr);
 
@@ -210,7 +210,7 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate
 #ifndef RISCV_ENABLE_MISALIGNED
     throw trap_load_address_misaligned(gva, addr, 0, 0);
 #else
-    if (require_alignment)
+    if (xlate_flags & RISCV_XLATE_LR)
       throw trap_load_access_fault(gva, addr, 0, 0);
 
     reg_t len_page0 = std::min(len, PGSIZE - addr % PGSIZE);
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -190,7 +190,11 @@ void mmu_t::load_slow_path_intrapage(reg_t addr, reg_t len, uint8_t* bytes, uint
       tracer.trace(paddr, len, LOAD);
     else if (xlate_flags == 0)
       refill_tlb(addr, paddr, host_addr, LOAD);
-  } else if (!mmio_load(paddr, len, bytes)) {
+
+    if (xlate_flags & RISCV_XLATE_LR) {
+      load_reservation_address = paddr;
+    }
+  } else if ((xlate_flags & RISCV_XLATE_LR) || !mmio_load(paddr, len, bytes)) {
     throw trap_load_access_fault((proc) ? proc->state.v : false, addr, 0, 0);
   }
 }
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -59,23 +59,28 @@ public:
   proc->state.log_mem_read.push_back(std::make_tuple(addr, 0, size));
 #endif
 
+  template<typename T>
+  T ALWAYS_INLINE load(reg_t addr, bool require_alignment = false, uint32_t xlate_flags = 0) {
+    target_endian<T> res;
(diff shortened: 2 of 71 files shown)


aswaterman:
Supersedes #1120

riscv/mmu.cc
```diff
@@ -224,7 +228,9 @@ void mmu_t::store_slow_path_intrapage(reg_t addr, reg_t len, const uint8_t* byte
   reg_t vpn = addr >> PGSHIFT;
   if (xlate_flags == 0 && vpn == (tlb_store_tag[vpn % TLB_ENTRIES] & ~TLB_CHECK_TRIGGERS)) {
     auto host_addr = tlb_data[vpn % TLB_ENTRIES].host_offset + addr;
-    memcpy(host_addr, bytes, len);
+    if (actually_store) {
+      memcpy(host_addr, bytes, len);
+    }
```
liweiwei90: Sorry. I didn't reallize this fix and send a similar fix in a new PR.

riscv/mmu.h
```diff
@@ -59,75 +60,88 @@ class mmu_t
...
-        return from_target(res); \
-      } \
-    }
+    type##_t ALWAYS_INLINE prefix##_##type(reg_t addr) { return load<type##_t>(addr, xlate_flags); }
 
   // load value from memory at aligned address; zero extend to register width
   load_func(uint8, load, 0)
   load_func(uint16, load, 0)
   load_func(uint32, load, 0)
   load_func(uint64, load, 0)
```
liweiwei90: I don't know why use two kinds of way here. We can use load\<uint8/16/32/64> directly just as guest_load<T>. Or let guest_load* remain the original way using load_func.

riscv/mmu.cc
aswaterman: No problem, I accepted your version.

riscv/mmu.h
aswaterman: I agree.  I was going to save that for a follow-up PR, but I guess I can handle it now.

aswaterman:
Thanks for the feedback, @liweiwei90!

riscv/mmu.cc
```diff
@@ -190,7 +190,11 @@ void mmu_t::load_slow_path_intrapage(reg_t addr, reg_t len, uint8_t* bytes, uint
       tracer.trace(paddr, len, LOAD);
     else if (xlate_flags == 0)
       refill_tlb(addr, paddr, host_addr, LOAD);
-  } else if (!mmio_load(paddr, len, bytes)) {
+
+    if (xlate_flags & RISCV_XLATE_LR) {
+      load_reservation_address = paddr;
+    }
+  } else if ((xlate_flags & RISCV_XLATE_LR) || !mmio_load(paddr, len, bytes)) {
     throw trap_load_access_fault((proc) ? proc->state.v : false, addr, 0, 0);
```
liweiwei90, line 198: While LR operation cannot access mmio, how about amo operation?
If it can, how to handle the cases when the mmio address is read-only or write-only?

aswaterman: Yeah, supporting AMO MMIOs is not uncommon.

AMOs will check for write permissions before performing the read.  Either the write-permission check or the read might trap, and that is OK.

aswaterman: (Note this PR makes no functional change with respect to AMOs on MMIO regions.)

aswaterman: Note this PR does not change any policy re: AMO/LR/SC to MMIO.  It only fixes the LR precise exception bug.  Otherwise, the exception behavior for all of these operations is unchanged.

liweiwei90: Before performing the read, write permissions haven't been checked for MMIOs currently. So the actual store may fail after the load success.

liweiwei90: > Note this PR does not change any policy re: AMO/LR/SC to MMIO. It only fixes the LR precise exception bug. Otherwise, the exception behavior for all of these operations is unchanged.

Yeah. I know. I just thought of this.

aswaterman: > Before performing the read, write permissions haven't been checked for MMIOs currently. So the actual store may fail after the load success.

This might be acceptable for MMIO, as often this kind of thing is dealt with using (imprecise) bus errors.  We should always be precise when possible, but some amount of imprecision is inevitable.

liweiwei90: OK. Thanks for your patient explanation.

scottj97:
The good news is: I observe no behavior change from any of these commits.

The bad news is: I won't have time to go through these commits in detail any time soon.

aswaterman:
@scottj97 what’s your recommendation as to how to proceed? I think leaving this for a week is fine, but it’ll be a merge-conflict tar pit before too long.

scottj97:
> @scottj97 what’s your recommendation as to how to proceed? I think leaving this for a week is fine, but it’ll be a merge-conflict tar pit before too long.

Go ahead and merge if you're ready

aswaterman:
@scottj97 Will do.  I'm happy to address any concerns you have after the fact.

PR #1126, opened 2022-10-21
fix value printing function: commit_log_print_value()

YenHaoChen:
This PR includes two commits to the memory value printing function, commit_log_print_value(). The first commit corrects the number of digits of width=8 accesses. The second commit enhances the capability of irregular accesses with width={24,40,48,56}, required by the previous commit (https://github.com/riscv-software-src/riscv-isa-sim/commit/e2e66015af4b5fce4bc958e70398f1fb7af7bcd9).

update 2022.10.24
Another three commits merges the switch cases. Additionally, they correct the abort condition on width={1,2,4}.

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -25,44 +25,26 @@ static void commit_log_print_value(FILE *log_file, int width, const void *data)
 {
   assert(log_file);
 
-  switch (width) {
-    case 8:
-      fprintf(log_file, "0x%02" PRIx8, *(const uint8_t *)data);
-      break;
-    case 16:
-      fprintf(log_file, "0x%04" PRIx16, *(const uint16_t *)data);
-      break;
-    case 24:
-      fprintf(log_file, "0x%06" PRIx32, *(const uint32_t *)data & ((uint32_t(1)<<width)-1));
-      break;
-    case 32:
-      fprintf(log_file, "0x%08" PRIx32, *(const uint32_t *)data);
-      break;
-    case 40:
-      fprintf(log_file, "0x%010" PRIx64, *(const uint64_t *)data & ((uint64_t(1)<<width)-1));
-      break;
-    case 48:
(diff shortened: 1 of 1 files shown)


riscv/execute.cc
```diff
@@ -27,14 +27,26 @@ static void commit_log_print_value(FILE *log_file, int width, const void *data)
...
       break;
+    case 40:
+      fprintf(log_file, "0x%010" PRIx64, *(const uint64_t *)data & ((uint64_t(1)<<width)-1));
+      break;
+    case 48:
+      fprintf(log_file, "0x%012" PRIx64, *(const uint64_t *)data & ((uint64_t(1)<<width)-1));
+      break;
+    case 56:
+      fprintf(log_file, "0x%014" PRIx64, *(const uint64_t *)data & ((uint64_t(1)<<width)-1));
+      break;
```
scottj97: Instead of adding so many more cases here, why not rewrite this as a loop?

I would do that as an additional commit on top of this one, in order to demonstrate that it's just a refactoring and not a functional change from the old system. (At least, not for the aligned sizes.)

YenHaoChen: New commits merges the switch cases. Additionally, it corrects the abort condition on width={1,2,4}.

scottj97 (changes requested):
I can't see that 37620e1049473402fea17b74007ddb4e1c346717 has any effect.

The call to WRITE_MEM() is outside of `store_slow_path()` where the store gets split into multiple accesses. So I'm not seeing any case where this `commit_log_print_value()` ever needs to print any of these other sizes.

What am I missing here?

YenHaoChen:
@scottj97 Yes, you are right. This PR does not affect this since the WRITE_MEM() is outside store_slow_path(). I want to log down the physical address of memory writes. Thus, I move the WRITE_MEM() into the store_slow_path(). Sorry for the misleading.

PR #1128, opened 2022-10-25
Implement itrigger and etrigger

YenHaoChen:
Previous implementation only has the mcontrol trigger described in debug specification, Section 5.5.11. This PR enhances the trigger support of the itrgger and etrigger described in debug specification, Section 5.5.14 and Section 5.5.15.

Previously, all triggers were mcontrol triggers. After this PR, the spike can configure each trigger to mcontrol, itrigger, etrigger, or disable dynamically. 

A significant difference is that writing 0 to (or resetting) trigger CSRs with multiple types results in disabled, i.e., type=15 instead of type=2. By the specification, the software should take responsibility for understanding the support of the trigger module. There are PRs to the riscv-test and riscv-openocd to fix the historical assumption issue.

The commits has passed the riscv-tests/debug/ (https://github.com/riscv-software-src/riscv-tests/tree/master/debug) with patches (https://github.com/riscv-software-src/riscv-tests/pull/425 and https://github.com/riscv/riscv-openocd/pull/753).


riscv/triggers.cc
```diff
@@ -141,9 +151,115 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
...
+  dmode = proc->get_state()->debug_mode ? get_field(val, CSR_ITRIGGER_DMODE(xlen)) : 0;
+  hit = get_field(val, CSR_ITRIGGER_HIT(xlen));
+  vs = proc->extension_enabled('H') ? get_field(val, CSR_ITRIGGER_VS) : 0;
+  vu = proc->extension_enabled('H') ? get_field(val, CSR_ITRIGGER_VU) : 0;
+  nmi = get_field(val, CSR_ITRIGGER_NMI);
+  m = get_field(val, CSR_ITRIGGER_M);
+  s = proc->extension_enabled('S') ? get_field(val, CSR_ITRIGGER_S) : 0;
+  u = proc->extension_enabled('U') ? get_field(val, CSR_ITRIGGER_U) : 0;
+  action = (action_t)get_field(val, CSR_ITRIGGER_ACTION);
+  if (action > 4 || (action==1 && dmode==0))
```
timsifive:
```suggestion
  if (action > 4 || (action==ACTION_DEBUG_MODE && dmode==0))
```

riscv/triggers.cc
```diff
@@ -141,9 +151,115 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
...
+  assert(get_field(val, CSR_ETRIGGER_TYPE(xlen)) == 5);
+  dmode = proc->get_state()->debug_mode ? get_field(val, CSR_ETRIGGER_DMODE(xlen)) : 0;
+  hit = get_field(val, CSR_ETRIGGER_HIT(xlen));
+  vs = proc->extension_enabled('H') ? get_field(val, CSR_ETRIGGER_VS) : 0;
+  vu = proc->extension_enabled('H') ? get_field(val, CSR_ETRIGGER_VU) : 0;
+  m = get_field(val, CSR_ETRIGGER_M);
+  s = proc->extension_enabled('S') ? get_field(val, CSR_ETRIGGER_S) : 0;
+  u = proc->extension_enabled('U') ? get_field(val, CSR_ETRIGGER_U) : 0;
+  action = (action_t)get_field(val, CSR_ETRIGGER_ACTION);
+  if (action > 4 || (action==1 && dmode==0))
```
timsifive: Same as above.

riscv/triggers.h
```diff
@@ -42,30 +42,90 @@ class matched_t
...
 
-  virtual bool chain() const { return false; }
-  virtual bool execute() const { return false; }
-  virtual bool store() const { return false; }
-  virtual bool load() const { return false; }
+  virtual bool get_dmode() const = 0;
+  virtual bool get_chain() const { return false; }
+  virtual bool get_execute() const { return false; }
+  virtual bool get_store() const { return false; }
+  virtual bool get_load() const { return false; }
```
timsifive: Why rename all this? Does this make things more consistent with other code?

timsifive (changes requested):
Looks good overall. I just have a few minor comments below.

riscv/triggers.h
YenHaoChen: Yes, the renaming is to make it more consistent with other codes. For instance, we have get_dts() in sim.h, get_xlen() in processor.h, get_tval() in trap.h, etc. Additionally, renaming the functions avoids conflict in naming the variables, chain, execute, store, and load. Otherwise, those variables must be named chain_bit, execute_bit, store_bit, and load_bit, which is inconsistent with other codes.

Does this make sense to you?

timsifive: That makes sense. Thanks for the explanation.

timsifive (approved):

YenHaoChen:
@aswaterman @scottj97 @timsifive 
Does this RP have any missing or concerns that I could help with?

aswaterman (approved):
I'm OK with this, but I only reviewed the changes to processor.cc and execute.cc.  I'll give @scottj97 a chance to chime in.

scottj97 (changes requested):
I haven't gotten very far, only looking at a4ac7a46ac313ffa46c636f26a1f0199df672f72, but it  is problematic.

1. It does too much, moving code within the file and simultaneously moving methods from one class to another. It makes it hard to tell from the diff what has changed in that code. (It appears nothing has changed.)
2. New class `tdata2_csr_t` uses `trigger_t` as a `public virtual` parent, instead of simply `public`, without explanation or justification that I can see.
3. It fails compilation, since `mcontrol_t` still declares methods `tdata2_read` and `tdata2_write` with no definition.

scottj97 (changes requested):
Commit 06d3ba53942b613da6e6a4b2331d1be11c3f3ffa does at least 11 different things in a single commit. That should be 11 individual commits.

Commit ec7132a3b0face3ecb33afbaaa585cba996ff003 adds only dead code, which means it cannot be tested and `git bisect` won't properly be able to identify the source of any regressions caused by that code.

Commit de313f276c098e9fef50c849a7149b88babdbb95 adds two new features at once. That should be two separate commits. It's also all dead code since nowhere are these new classes used that I can find. It also declares new method `module_t::trap_taking_match` that is not defined until the subsequent commit.

The commit message for 0c8e4ea77b93d66093cf161e632295a803f67440 doesn't tell me anything useful about what this commit does. It looks like it a major change, but it also reorganizes the code within the file, making the diff unreadable. (Example: `module_t::tdata1_read` seems to have no changes? But `module_t::tdata1_write` has major changes. Neither of these are easy to see in the diff.)

Commit 6ab12fe59b4916c03a5ce78acd336b4bda85f2e9 combines a refactoring (extract new method `take_trigger_action`) with additional debug logging. That's two separate changes.

Commit d1b4c57d59bbe4c9c1ee233ab5099e2a6450e242 finally uses some of the new code. This change should be way earlier in the series, before any new behavior is added.

Commit 076e62fbf9725591b54863b302997b4fdb577779 patches up a bug introduced by an earlier commit. This should be squashed so the bug does not appear at all anymore.

The commit message in d36a791dd4a17d9e72e23107d77d076e850a7e7b is useless. And it should be squashed.

Same for 9dcf92421bce84d9ddcf17b67589cbb9421f183a.

timsifive:
I've started cleaning up the history for this PR at https://github.com/riscv-software-src/riscv-isa-sim/tree/itrigger-etrigger-cleanup

YenHaoChen:
@timsifive sorry for the lateness. I am working on rebasing in my local. Do you have any requirement or consideration for the rebasing?

YenHaoChen:
@scottj97 @timsifive 
Sorry for the lateness. I was on a vacation trip.
Thanks for the comments and suggestions. I do the rebase based on (https://github.com/riscv-software-src/riscv-isa-sim/tree/itrigger-etrigger-cleanup).
This revised PR avoids dead code in all commits and separates modifications into different commits for better readability with git diff.
The virtual class  ```trigger_t``` depends on ```tdata1_csr_t```,  ```tdata2_csr_t```, and ```tdata3_csr_t```. I plan to use the virtual class to support ```tdata2_csr_t``` and ```tdata3_csr_t```. The current spike does not implement the ```tdata3_csr_t```. Thus, ```public virtual``` does not make any difference to simply ```public```. Nevertheless, the ```public virtual``` logically makes more sense according to the debug spec. I am preparing another PR that implements ```tdata3_csr_t```. Should we use ```public``` and change to ```public virtual``` when implementing ```tdata3_csr_t```?

timsifive:
> @timsifive Sorry for the lateness. I was on a vacation trip.

No worries. I'm glad you were able to use some of the refactoring I did.

riscv/triggers.cc
```diff
@@ -191,6 +188,9 @@ reg_t module_t::tdata1_read(const processor_t * const proc, unsigned index) cons
 
 bool module_t::tdata1_write(processor_t * const proc, unsigned index, const reg_t val) noexcept
 {
+  if (triggers[index]->get_dmode() && !proc->get_state()->debug_mode) {
+    return false;
+  }
```
scottj97: This is much more than just a rename of some fields/methods and should have its own commit with an explanation.

riscv/triggers.cc
```diff
@@ -17,6 +17,22 @@ bool tdata2_csr_t::tdata2_write(processor_t UNUSED * const proc, const reg_t UNU
   return true;
 }
 
+reg_t disabled_trigger_t::tdata1_read(const processor_t * const proc) const noexcept
+{
+  auto xlen = proc->get_xlen();
+  reg_t tdata1 = 0;
+  tdata1 = set_field(tdata1, CSR_TDATA1_TYPE(xlen), 15);
```
scottj97: Use `CSR_TDATA1_TYPE_DISABLED`

riscv/triggers.cc
```diff
@@ -17,6 +17,22 @@ bool tdata2_csr_t::tdata2_write(processor_t UNUSED * const proc, const reg_t UNU
...
+  tdata1 = set_field(tdata1, CSR_TDATA1_DMODE(xlen), dmode);
+  return tdata1;
+}
+
+bool disabled_trigger_t::tdata1_write(processor_t * const proc, const reg_t val) noexcept
+{
+  auto xlen = proc->get_xlen();
+  dmode = proc->get_state()->debug_mode ? get_field(val, CSR_TDATA1_DMODE(xlen)) : 0;
```
scottj97: The debug spec (20220912) says:

> 15 (disabled): This trigger is disabled. In this state, tdata2 and tdata3 can be written with any value that is supported for any of the types this trigger implements. **The remaining bits in this register are ignored.**

So why is there a `dmode` bit here?

riscv/triggers.cc
```diff
@@ -191,7 +207,19 @@ bool module_t::tdata1_write(processor_t * const proc, unsigned index, const reg_
...
-  bool result = triggers[index]->tdata1_write(proc, val);
+
+  auto xlen = proc->get_xlen();
+
+  unsigned type = get_field(val, CSR_TDATA1_TYPE(xlen));
+  reg_t tdata1 = val;
+
+  delete triggers[index];
+  switch (type) {
+    case 2: triggers[index] = new mcontrol_t(); break;
```
scottj97: Use `CSR_TDATA1_TYPE_MCONTROL`

riscv/triggers.cc
```diff
@@ -191,7 +207,19 @@ bool module_t::tdata1_write(processor_t * const proc, unsigned index, const reg_
...
     return false;
   }
-  bool result = triggers[index]->tdata1_write(proc, val);
+
+  auto xlen = proc->get_xlen();
+
+  unsigned type = get_field(val, CSR_TDATA1_TYPE(xlen));
+  reg_t tdata1 = val;
+
+  delete triggers[index];
```
scottj97, line 307: This doesn't seem right. Every time we write to any `tdata1`, it deletes and recreates the object? Doesn't that mean `tdata2`'s current value will also be reset?

The debug spec says:

> Writes to one tdata register must not modify the contents of other tdata registers, nor the configuration of any trigger besides the one that is currently selected.

scottj97 (changes requested):
I've reviewed through 7c80bb79f15adf1decd4e38b538853cabcafc91d but I have some work to do on my end before I can review any further.

I think it would be good to document the software-observable changes in the README as these occur. For this, I've created #1145. Once that is merged, please rebase and then keep the README updated as the trigger configuration changes through the commits. That will be a big help in explaining each change.

riscv/triggers.cc
timsifive: The spec was recently updated to clarify that dmode does exist for type 15: https://github.com/riscv/riscv-debug-spec/commit/68aa0dc808dc604b5429a5b21ab83327e2f9f439

scottj97:
> The virtual class `trigger_t` depends on `tdata1_csr_t`, `tdata2_csr_t`, and `tdata3_csr_t`. I plan to use the virtual class to support `tdata2_csr_t` and `tdata3_csr_t`. The current spike does not implement the `tdata3_csr_t`. Thus, `public virtual` does not make any difference to simply `public`. Nevertheless, the `public virtual` logically makes more sense according to the debug spec. I am preparing another PR that implements `tdata3_csr_t`. Should we use `public` and change to `public virtual` when implementing `tdata3_csr_t`?

(1) Yes, change it to `public virtual` only once you need that functionality. Though it would be good to make that change by itself in a commit right before it's needed, in case it exposes some issue elsewhere.

(2) The class named `triggers::tdata2_csr_t` is misnamed because that name suggests that it inherits from `csr_t` which it does not. It's actually referring to a trigger that has `tdata2` functionality. Call it `trigger_with_tdata2_t`? (Note there already is a `tdata2_csr_t` declared in `csrs.h` that is indeed a descendant of `csr_t`.)

(3) I don't understand why multiple inheritance would be useful for triggers with `tdata3`. Wouldn't `trigger_with_tdata3_t` simply inherit from `trigger_with_tdata2_t`? Is there ever a trigger that has `tdata3` but not `tdata2`?

YenHaoChen:
Update the PR.
- Use ```public``` instead of ```public virtual```
- Use ```trigger_with_tdata2_t``` instead of ```tdata2_csr_t```

> (3) I don't understand why multiple inheritance would be useful for triggers with `tdata3`. Wouldn't `trigger_with_tdata3_t` simply inherit from `trigger_with_tdata2_t`? Is there ever a trigger that has `tdata3` but not `tdata2`?

Agree.

riscv/triggers.cc
scottj97: Why was this marked as resolved? The code is still there.

scottj97:
Please rebase to the latest master and keep the README updated through the changes, [as discussed here](https://github.com/riscv-software-src/riscv-isa-sim/pull/1128#pullrequestreview-1183207568).

riscv/triggers.cc
YenHaoChen: >   reg_t tdata2 = triggers[index]->tdata2_read(proc);
> ...
>   delete triggers[index];
> ...
>   result &= triggers[index]->tdata2_write(proc, tdata2);

The revised commit merges with the one of writing tdata2. It stores the previous tdata2 value, deletes and recreates the object, and then writes the original value with legalization. Thus, it resolves the issue of the tdata2 reseting.

YenHaoChen:
Update README.md in commits "add disabled_trigger_t, itrigger_t, and etrigger_t".

scottj97 (changes requested):
Something's not working right in 416875c40c7989440343400c621440d0138d6bcf. I see this in my commit log:
```
core   0: 0x000000008000225a (0x0002b383) ld      t2, 0(t0)
core   0: 3 0x000000008000225a (0x0002b383) x7  0x2f978d8719999e3f mem 0x0000000080003128
core   0: 0x000000008000225e (0x000002a1) c.addi  t0, 8
core   0: 3 0x000000008000225e (0x02a1) x5  0x0000000080003130
core   0: 0x0000000080002260 (0x7a101073) csrw    tdata1, zero
core   0: 3 0x0000000080002260 (0x7a101073) c1953_tdata1 0xf000000000000000
core   0: 0x0000000080002264 (0x7a239073) csrw    tdata2, t2
core   0: 3 0x0000000080002264 (0x7a239073) c1954_tdata2 0x2f978d8719999e3f
core   0: 0x0000000080002268 (0x7a202473) csrr    s0, tdata2
core   0: 3 0x0000000080002268 (0x7a202473) x8  0x2f978d8719999e3f
core   0: 0x000000008000226c (0x7a139073) csrw    tdata1, t2
core   0: 3 0x000000008000226c (0x7a139073)
core   0: 0x0000000080002270 (0x7a1023f3) csrr    t2, tdata1
core   0: 3 0x0000000080002270 (0x7a1023f3) x7  0x280000000019981f
core   0: 0x0000000080002274 (0x7a2024f3) csrr    s1, tdata2
core   0: 3 0x0000000080002274 (0x7a2024f3) x9  0x0000000000000000
```

You can see that `tdata2` reads as 0x2f978d8719999e3f, but then after I write to `tdata1`, somehow `tdata2` has changed and now reads 0.

This behavior changes in 5b7318a3bd8c4e75e579b01480317fc40c286e1a -- which is supposed to be only a refactoring!

YenHaoChen:
> Something's not working right in [416875c](https://github.com/riscv-software-src/riscv-isa-sim/commit/416875c40c7989440343400c621440d0138d6bcf). I see this in my commit log:
> 
> ```
> core   0: 0x000000008000225a (0x0002b383) ld      t2, 0(t0)
> core   0: 3 0x000000008000225a (0x0002b383) x7  0x2f978d8719999e3f mem 0x0000000080003128
> core   0: 0x000000008000225e (0x000002a1) c.addi  t0, 8
> core   0: 3 0x000000008000225e (0x02a1) x5  0x0000000080003130
> core   0: 0x0000000080002260 (0x7a101073) csrw    tdata1, zero
> core   0: 3 0x0000000080002260 (0x7a101073) c1953_tdata1 0xf000000000000000
> core   0: 0x0000000080002264 (0x7a239073) csrw    tdata2, t2
> core   0: 3 0x0000000080002264 (0x7a239073) c1954_tdata2 0x2f978d8719999e3f
> core   0: 0x0000000080002268 (0x7a202473) csrr    s0, tdata2
> core   0: 3 0x0000000080002268 (0x7a202473) x8  0x2f978d8719999e3f
> core   0: 0x000000008000226c (0x7a139073) csrw    tdata1, t2
> core   0: 3 0x000000008000226c (0x7a139073)
> core   0: 0x0000000080002270 (0x7a1023f3) csrr    t2, tdata1
> core   0: 3 0x0000000080002270 (0x7a1023f3) x7  0x280000000019981f
> core   0: 0x0000000080002274 (0x7a2024f3) csrr    s1, tdata2
> core   0: 3 0x0000000080002274 (0x7a2024f3) x9  0x0000000000000000
> ```
> 
> You can see that `tdata2` reads as 0x2f978d8719999e3f, but then after I write to `tdata1`, somehow `tdata2` has changed and now reads 0.
> 
> This behavior changes in [5b7318a](https://github.com/riscv-software-src/riscv-isa-sim/commit/5b7318a3bd8c4e75e579b01480317fc40c286e1a) -- which is supposed to be only a refactoring!

That's interesting.

The reason is due to the incorrect dmode checking in trigger_with_tdata2_t::tdata2_write in https://github.com/riscv-software-src/riscv-isa-sim/commit/416875c40c7989440343400c621440d0138d6bcf. Updating the tdata1 and tdata2 should be logically simultaneous. However, after updating the tdata1, the dmode is set to 1, and M-mode cannot update the tdata2. In https://github.com/riscv-software-src/riscv-isa-sim/commit/5b7318a3bd8c4e75e579b01480317fc40c286e1a, the checking moves to module_t::tdata2_write, which avoids incorrect checking.

@scottj97 Can you confirm that the execution was in M-mode (instead of Debug mode)? If the execution is in M-mode, there are two solutions to this issue.
1. Move tdata2_write before tdata1_write in https://github.com/riscv-software-src/riscv-isa-sim/commit/416875c40c7989440343400c621440d0138d6bcf.
2. Move commit https://github.com/riscv-software-src/riscv-isa-sim/commit/5b7318a3bd8c4e75e579b01480317fc40c286e1a before https://github.com/riscv-software-src/riscv-isa-sim/commit/416875c40c7989440343400c621440d0138d6bcf.

I choose the second solution.

riscv/triggers.cc
```diff
@@ -153,6 +269,71 @@ module_t::~module_t() {
...
+  switch (type) {
+    case CSR_TDATA1_TYPE_MCONTROL: triggers[index] = new mcontrol_t(); break;
+    case CSR_TDATA1_TYPE_ITRIGGER: triggers[index] = new itrigger_t(); break;
+    case CSR_TDATA1_TYPE_ETRIGGER: triggers[index] = new etrigger_t(); break;
+    default: triggers[index] = new disabled_trigger_t(); break;
+  }
+
+  bool result = triggers[index]->tdata1_write(proc, tdata1);
+  result &= triggers[index]->tdata2_write(proc, tdata2);
```
scottj97: This looks suspicious, because `module_t::tdata1_write()` is supposed to return `true` iff `tdata1` changed, and as we discussed earlier, `tdata2` should never appear to be changed by that CSRW.

From what I can tell, every `trigger_t`'s implementation of both `tdata1_write()` and `tdata2_write()` always `return true`. If so, there's no longer any reason for them to return a `bool`.

It looks like you can remove those return values immediately after d0584f8aa62f7ca2361f016679ebc9d5a4b4a853. Then you wouldn't need this `bool result` and could always `return true` here.

scottj97 (changes requested):
I still have lots more to review on this, but I'll be mostly unavailable this week due to the Thanksgiving holiday.

YenHaoChen:
1. Update tinfo along the commits (https://github.com/riscv-software-src/riscv-isa-sim/pull/1153).
2. Add a refactor commit: use CSR_TDATA1_TYPE_MCONTROL instead of MCONTROL_TYPE_MATCH for consistency

riscv/triggers.cc
```diff
@@ -34,7 +34,7 @@ reg_t mcontrol_t::tdata1_read(const processor_t * const proc) const noexcept {
   v = set_field(v, MCONTROL_U, u);
   v = set_field(v, MCONTROL_EXECUTE, execute);
   v = set_field(v, MCONTROL_STORE, store_bit);
-  v = set_field(v, MCONTROL_LOAD, load_bit);
+  v = set_field(v, MCONTROL_LOAD, load);
```
scottj97, line 49: The commit message for 22f26e3b01097a72626d7d8c252631939a94ea3a has a typo:
```
load_bit -> bit
```

That should be:
```
load_bit -> load
```

riscv/triggers.cc
```diff
@@ -209,7 +238,7 @@ bool module_t::tdata2_write(processor_t * const proc, unsigned index, const reg_
 reg_t module_t::tinfo_read(UNUSED const processor_t * const proc, unsigned UNUSED index) const noexcept
 {
   /* In spike, every trigger supports the same types. */
-  return 1<<MCONTROL_TYPE_MATCH;
+  return (1<<MCONTROL_TYPE_MATCH) | (1<<CSR_TDATA1_TYPE_DISABLED);
```
scottj97: Put spaces around binary operators like `<<`

riscv/triggers.cc
```diff
@@ -177,6 +177,15 @@ bool module_t::tdata1_write(processor_t * const proc, unsigned index, const reg_
   reg_t tdata1 = val;
   reg_t tdata2 = triggers[index]->tdata2_read(proc);
 
+  // hardware must zero chain in writes that set dmode to 0 if the next trigger has dmode of 1
+  if (index+1 < triggers.size())
+    if (triggers[index+1]->get_dmode())
+      if (!get_field(val, CSR_TDATA1_DMODE(xlen)))
+        if (type == 2 || type == 6) {
```
scottj97: Don't use magic numbers.

This doesn't need to be 4 nested `if` statements, it's just a compound expression (`&&`).

I don't think this is the right place to do this. It should be down inside `mcontrol_t` where the `chain` bit is implemented -- but that doesn't have enough information. Add a `bool allow_chain` parameter to `trigger_t::tdata1_write()`?

riscv/triggers.cc
```diff
@@ -13,6 +13,21 @@ void trigger_with_tdata2_t::tdata2_write(processor_t UNUSED * const proc, const
...
+  reg_t tdata1 = 0;
+  tdata1 = set_field(tdata1, CSR_TDATA1_TYPE(xlen), CSR_TDATA1_TYPE_DISABLED);
+  tdata1 = set_field(tdata1, CSR_TDATA1_DMODE(xlen), dmode);
+  return tdata1;
+}
+
+void disabled_trigger_t::tdata1_write(processor_t * const proc, const reg_t val) noexcept
+{
+  auto xlen = proc->get_xlen();
+  dmode = proc->get_state()->debug_mode ? get_field(val, CSR_TDATA1_DMODE(xlen)) : 0;
```
scottj97: This checking of `debug_mode` becomes obsolete after ca58beb7fb2e737054d61f05937e9a05a1333882

riscv/triggers.cc
```diff
@@ -173,6 +173,12 @@ bool module_t::tdata1_write(processor_t * const proc, unsigned index, const reg_
 
   auto xlen = proc->get_xlen();
 
+  // hardware should ignore writes that set dmode to 1 if the previous trigger has both dmode of 0 and chain of 1
+  if (index > 0)
+    if (!triggers[index-1]->get_dmode() && triggers[index-1]->get_chain())
+      if (get_field(val, CSR_TDATA1_DMODE(xlen)))
+        return false;
```
scottj97: No need for nested `if`; use `&&`

riscv/triggers.cc
```diff
@@ -148,6 +148,59 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
...
+  nmi = get_field(val, CSR_ITRIGGER_NMI);
+  m = get_field(val, CSR_ITRIGGER_M);
+  s = proc->extension_enabled('S') ? get_field(val, CSR_ITRIGGER_S) : 0;
+  u = proc->extension_enabled('U') ? get_field(val, CSR_ITRIGGER_U) : 0;
+  action = (action_t)get_field(val, CSR_ITRIGGER_ACTION);
+  if (action > 4 || (action==ACTION_DEBUG_MODE && dmode==0))
+    action = ACTION_DEBUG_EXCEPTION;
+}
+
+match_result_t itrigger_t::trap_taking_match(processor_t * const proc, trap_t& t)
```
scottj97: Can this be `const trap_t& t`?

riscv/triggers.cc
```diff
@@ -254,10 +308,26 @@ match_result_t module_t::memory_access_match(action_t * const action, operation_
   return MATCH_NONE;
 }
 
+match_result_t module_t::trap_taking_match(action_t * const action, trap_t& t)
```
scottj97: Can this be `const trap_t& t`?

Rather than modifying `*action`, can `action` be part of the `match_result_t`?

riscv/triggers.cc
```diff
@@ -254,10 +308,26 @@ match_result_t module_t::memory_access_match(action_t * const action, operation_
...
+{
+  state_t * const state = proc->get_state();
+  if (state->debug_mode)
+    return MATCH_NONE;
+
+  for (unsigned int i = 0; i < triggers.size(); i++) {
```
scottj97: Use modern C++ loop: `for (auto trigger: triggers) {`

riscv/triggers.h
```diff
@@ -58,6 +58,7 @@ class trigger_t {
 
   virtual match_result_t memory_access_match(processor_t UNUSED * const proc,
       operation_t UNUSED operation, reg_t UNUSED address, std::optional<reg_t> UNUSED data) { return MATCH_NONE; }
+  virtual match_result_t trap_taking_match(processor_t UNUSED * const proc, trap_t UNUSED & t) { return MATCH_NONE; }
```
scottj97: I don't understand the name of this method. Method names should be verbs. What does `trap_taking_match` do? It sounds like it should throw a trap, but it doesn't.

riscv/triggers.cc
```diff
@@ -148,6 +148,59 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
...
+  tdata1 = set_field(tdata1, CSR_ITRIGGER_U, u);
+  tdata1 = set_field(tdata1, CSR_ITRIGGER_ACTION, action);
+  return tdata1;
+}
+
+void itrigger_t::tdata1_write(processor_t * const proc, const reg_t val) noexcept
+{
+  auto xlen = proc->get_xlen();
+  assert(get_field(val, CSR_ITRIGGER_TYPE(xlen)) == 4);
+  dmode = proc->get_state()->debug_mode ? get_field(val, CSR_ITRIGGER_DMODE(xlen)) : 0;
```
scottj97: The check of `debug_mode` here is already handled in `module_t::tdata1_write()`.

riscv/triggers.cc
```diff
@@ -148,6 +148,59 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
...
+  hit = get_field(val, CSR_ITRIGGER_HIT(xlen));
+  vs = proc->extension_enabled('H') ? get_field(val, CSR_ITRIGGER_VS) : 0;
+  vu = proc->extension_enabled('H') ? get_field(val, CSR_ITRIGGER_VU) : 0;
```
scottj97: This is a bit fishy...if `misa.H` is changed from 1 to 0, some previously-set triggers could still have `vs` and `vu` set, and that would be visible via `csrr tdata1`.

Would it be sufficient to do the masking only in `tdata1_read()` based on the current `misa.H`?

riscv/triggers.cc
```diff
@@ -148,6 +148,59 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
...
+  nmi = get_field(val, CSR_ITRIGGER_NMI);
+  m = get_field(val, CSR_ITRIGGER_M);
+  s = proc->extension_enabled('S') ? get_field(val, CSR_ITRIGGER_S) : 0;
+  u = proc->extension_enabled('U') ? get_field(val, CSR_ITRIGGER_U) : 0;
+  action = (action_t)get_field(val, CSR_ITRIGGER_ACTION);
+  if (action > 4 || (action==ACTION_DEBUG_MODE && dmode==0))
```
scottj97: Magic number 4 means what?

riscv/triggers.cc
```diff
@@ -177,6 +177,10 @@ bool module_t::tdata1_write(processor_t * const proc, unsigned index, const reg_
   reg_t tdata1 = val;
   reg_t tdata2 = triggers[index]->tdata2_read(proc);
 
+  // dmode only writable from debug mode
+  if (!proc->get_state()->debug_mode)
+    tdata1 = set_field(tdata1, CSR_TDATA1_DMODE(xlen), 0);
```
scottj97, line 304: There is an unwritten assumption here that `CSR_TDATA1_DMODE(xlen)` and `MCONTROL_DMODE(xlen)` and `CSR_ITRIGGER_DMODE(xlen)` and `CSR_ETRIGGER_DMODE(xlen)` are all equal. Should we have assertion(s) to that effect?

riscv/triggers.cc
```diff
@@ -148,6 +148,59 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
...
+  tdata1 = set_field(tdata1, CSR_ITRIGGER_S, s);
+  tdata1 = set_field(tdata1, CSR_ITRIGGER_U, u);
+  tdata1 = set_field(tdata1, CSR_ITRIGGER_ACTION, action);
+  return tdata1;
+}
+
+void itrigger_t::tdata1_write(processor_t * const proc, const reg_t val) noexcept
+{
+  auto xlen = proc->get_xlen();
+  assert(get_field(val, CSR_ITRIGGER_TYPE(xlen)) == 4);
```
scottj97: Magic number 4

riscv/triggers.cc
```diff
@@ -201,6 +201,57 @@ match_result_t itrigger_t::trap_taking_match(processor_t * const proc, trap_t& t
   return MATCH_NONE;
 }
 
+reg_t etrigger_t::tdata1_read(const processor_t * const proc) const noexcept
+{
+  auto xlen = proc->get_xlen();
+  reg_t tdata1 = 0;
+  tdata1 = set_field(tdata1, CSR_ETRIGGER_TYPE(xlen), 5);
```
scottj97: Magic number 5

riscv/triggers.cc
```diff
@@ -201,6 +201,57 @@ match_result_t itrigger_t::trap_taking_match(processor_t * const proc, trap_t& t
...
+}
+
+void etrigger_t::tdata1_write(processor_t * const proc, const reg_t val) noexcept
+{
+  auto xlen = proc->get_xlen();
+  assert(get_field(val, CSR_ETRIGGER_TYPE(xlen)) == CSR_TDATA1_TYPE_ETRIGGER);
+  dmode = proc->get_state()->debug_mode ? get_field(val, CSR_ETRIGGER_DMODE(xlen)) : 0;
+  hit = get_field(val, CSR_ETRIGGER_HIT(xlen));
+  vs = proc->extension_enabled('H') ? get_field(val, CSR_ETRIGGER_VS) : 0;
+  vu = proc->extension_enabled('H') ? get_field(val, CSR_ETRIGGER_VU) : 0;
```
scottj97: Same comment as `itrigger_t`

riscv/triggers.cc
```diff
@@ -201,6 +201,57 @@ match_result_t itrigger_t::trap_taking_match(processor_t * const proc, trap_t& t
...
+  m = get_field(val, CSR_ETRIGGER_M);
+  s = proc->extension_enabled('S') ? get_field(val, CSR_ETRIGGER_S) : 0;
+  u = proc->extension_enabled('U') ? get_field(val, CSR_ETRIGGER_U) : 0;
+  action = (action_t)get_field(val, CSR_ETRIGGER_ACTION);
+  if (action > 4 || (action==ACTION_DEBUG_MODE && dmode==0))
```
scottj97: Magic number 4 means what?

riscv/triggers.cc
```diff
@@ -201,6 +201,57 @@ match_result_t itrigger_t::trap_taking_match(processor_t * const proc, trap_t& t
...
+  state_t * const state = proc->get_state();
+  if ((state->prv == PRV_M && !m) ||
+      (state->prv == PRV_S && !s) ||
+      (state->prv == PRV_U && !u)) {
+    return MATCH_NONE;
+  }
+
+  reg_t bit = t.cause();
+  bool interrupt = (bit & ((reg_t)1 << (proc->get_const_xlen() - 1))) != 0;
+  assert((bit&0xff) < proc->get_const_xlen());
```
scottj97: Spaces around binary operators

scottj97 (changes requested):

riscv/triggers.cc
YenHaoChen: This requires trap_t::cause() to be a const function. I add a refactoring commit to do so.

riscv/triggers.h
YenHaoChen: How about trap_match?

riscv/triggers.cc
```diff
@@ -176,6 +176,11 @@ bool module_t::tdata1_write(processor_t * const proc, unsigned index, const reg_
   reg_t tdata1 = val;
   reg_t tdata2 = triggers[index]->tdata2_read(proc);
 
+  // hardware must zero chain in writes that set dmode to 0 if the next trigger has dmode of 1
+  bool allow_chain = true;
+  if (index+1 < triggers.size() && triggers[index+1]->get_dmode() && !get_field(val, CSR_TDATA1_DMODE(xlen)))
+    allow_chain = false;
```
scottj97: No need for a mutable variable here.
```
const bool allow_chain = <expression>;
```

riscv/triggers.cc
```diff
@@ -22,7 +22,7 @@ reg_t disabled_trigger_t::tdata1_read(const processor_t * const proc) const noex
   return tdata1;
 }
 
-void disabled_trigger_t::tdata1_write(processor_t * const proc, const reg_t val) noexcept
+void disabled_trigger_t::tdata1_write(processor_t * const proc, const reg_t val, bool UNUSED allow_chain) noexcept
```
scottj97: The new param should be `const` everywhere.

riscv/triggers.cc
```diff
@@ -100,8 +100,8 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
       (operation == triggers::OPERATION_STORE && !store) ||
       (operation == triggers::OPERATION_LOAD && !load) ||
       (state->prv == PRV_M && !m) ||
-      (state->prv == PRV_S && !s) ||
-      (state->prv == PRV_U && !u)) {
+      (!state->v && state->prv == PRV_S && !s) ||
+      (!state->v && state->prv == PRV_U && !u)) {
```
scottj97: This logic does not look correct. Each line here is another reason for not matching. When in VS-mode, none of these lines will match anymore, so the trigger will incorrectly match.

I think you just want a line that says `state->v ||`, i.e. when V=1 then no trigger should match.

riscv/triggers.h
```diff
@@ -24,10 +24,20 @@ typedef enum
...
-  MATCH_NONE,
-  MATCH_FIRE_BEFORE,
-  MATCH_FIRE_AFTER
-} match_result_t;
+  TIMING_BEFORE = 0,
+  TIMING_AFTER = 1
+} timing_t;
+
+struct match_result_t {
+  match_result_t(const bool &f, const timing_t &t=TIMING_BEFORE, const action_t &a=ACTION_DEBUG_EXCEPTION) {
```
scottj97: Why pass all these by const ref? They are all tiny in size so could be passed by value cheaply.

riscv/triggers.h
```diff
@@ -24,10 +24,20 @@ typedef enum
 } action_t;
 
 typedef enum {
-  MATCH_NONE,
-  MATCH_FIRE_BEFORE,
-  MATCH_FIRE_AFTER
-} match_result_t;
```
scottj97, line 30: I like the change in 073e7bc3bfd0b1f2eca2c42b5f55a35a86a44d2f but it is very large for one commit. I won't postpone this merge anymore by asking this to be broken up, but there's really about 5 different things changing at once here, which next time should be 5 separate commits.

riscv/triggers.h
scottj97: I guess the old `memory_access_match` is not the greatest name either, but at least "memory" can't be misinterpreted as a verb like "trap" can.

How about `find_trap_match`? `get_trap_match`? `detect_trap_match`?

riscv/triggers.cc
```diff
@@ -147,6 +147,61 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
...
+  if (action > ACTION_TRACE_EMIT || (action==ACTION_DEBUG_MODE && dmode==0))
+    action = ACTION_DEBUG_EXCEPTION;
+}
+
+match_result_t itrigger_t::trap_match(processor_t * const proc, const trap_t& t)
+{
+  state_t * const state = proc->get_state();
+  if ((state->prv == PRV_M && !m) ||
+      (!state->v && state->prv == PRV_S && !s) ||
+      (!state->v && state->prv == PRV_U && !u) ||
```
scottj97, line 192: Logic looks incorrect, like earlier comment on `mcontrol_t`

riscv/triggers.cc
```diff
@@ -147,6 +147,61 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
...
+  dmode = get_field(val, CSR_ITRIGGER_DMODE(xlen));
+  hit = get_field(val, CSR_ITRIGGER_HIT(xlen));
+  vs = get_field(val, CSR_ITRIGGER_VS);
+  vu = get_field(val, CSR_ITRIGGER_VU);
+  nmi = get_field(val, CSR_ITRIGGER_NMI);
+  m = get_field(val, CSR_ITRIGGER_M);
+  s = proc->extension_enabled_const('S') ? get_field(val, CSR_ITRIGGER_S) : 0;
+  u = proc->extension_enabled_const('U') ? get_field(val, CSR_ITRIGGER_U) : 0;
+  action = (action_t)get_field(val, CSR_ITRIGGER_ACTION);
+  if (action > ACTION_TRACE_EMIT || (action==ACTION_DEBUG_MODE && dmode==0))
```
scottj97: I understand `ACTION_TRACE_EMIT` is the largest allowed value for this field. I think it would be better to have an enum value defined for that purpose, i.e.:
```
typedef enum
{
  ACTION_DEBUG_EXCEPTION = MCONTROL_ACTION_DEBUG_EXCEPTION,
  ACTION_DEBUG_MODE = MCONTROL_ACTION_DEBUG_MODE,
  ACTION_TRACE_START = MCONTROL_ACTION_TRACE_START,
  ACTION_TRACE_STOP = MCONTROL_ACTION_TRACE_STOP,
  ACTION_TRACE_EMIT = MCONTROL_ACTION_TRACE_EMIT,
  ACTION_MAXVAL = MCONTROL_ACTION_TRACE_EMIT
} action_t;
```

That way when someone updates the enum to add a new action in the future they will update this WARL check automatically.

riscv/triggers.cc
```diff
@@ -147,6 +147,61 @@ match_result_t mcontrol_t::memory_access_match(processor_t * const proc, operati
...
+  if ((state->prv == PRV_M && !m) ||
+      (!state->v && state->prv == PRV_S && !s) ||
+      (!state->v && state->prv == PRV_U && !u) ||
+      (state->v && state->prv == PRV_S && !vs) ||
+      (state->v && state->prv == PRV_U && !vu)) {
+    return match_result_t(false);
+  }
+
+  bool interrupt = (t.cause() & ((reg_t)1 << (proc->get_const_xlen() - 1))) != 0;
+  reg_t bit = t.cause() & 0xff;
```
scottj97: Why 0xff? The privileged spec defines the exception code to be all bits below the interrupt bit.

riscv/triggers.cc
```diff
@@ -25,6 +25,7 @@ reg_t disabled_trigger_t::tdata1_read(const processor_t * const proc) const noex
 void disabled_trigger_t::tdata1_write(processor_t * const proc, const reg_t val) noexcept
 {
   auto xlen = proc->get_xlen();
+  assert(get_field(val, CSR_TDATA1_TYPE(xlen)) == CSR_TDATA1_TYPE_DISABLED);
```
scottj97: This assertion is firing whenever I write `tdata1` with an unsupported `type` field.

scottj97 (changes requested):
I'm sorry this PR is taking so many iterations, but it keeps changing! New assertions are causing new failures as of 52460e47241f04d31ace4b8422b6749dfc2ea3fd

riscv/triggers.cc
YenHaoChen: The itrigger and etrigger have separated m, s, u, vs, and vu bits. They indicate whether the trigger will match on different modes. Thus, for any of the following conditions, the trigger will not match:
1. In M-mode (state->prv==PRV_M), m is not set (!m).
2. In S/HS-mode (!state->v && state->prv==PRV_S), s is not set (!s).
3. In U-mode (!state->v && state->prv==PRV_U), u is not set (!u).
4. In VS-mode (state->v && state->prv==PRV_S), vs is not set (!vs).
5. In VU-mode (state->v && state->prv==PRV_U), vu is not set (!vu).

Do I understand this correctly?

scottj97: Yes, my mistake, this looks correct to me. (Though I think `mcontrol_t` is still problematic.)

riscv/triggers.h
YenHaoChen: Thank you for the comment. I split the commit into two (https://github.com/riscv-software-src/riscv-isa-sim/pull/1128/commits/b81652954ef91c6ab169efaae050825d159fdf05 and https://github.com/riscv-software-src/riscv-isa-sim/pull/1128/commits/a399d65d9fab4c38839a71dbd38d3c04c3494514), which seem to be minimal to me. Are they small enough for you?

riscv/triggers.h
YenHaoChen: I choose ```detect_trap_match``` and add a refactor commit for ```detect_memory_access_match```.
BTW, do you have any suggestions for the match function of icount? ```detect_icount_match```? ```detect_countdown_match```? (We can do this later, though.)

riscv/triggers.cc
```diff
@@ -153,65 +268,117 @@ module_t::~module_t() {
...
+  // hardware should ignore writes that set dmode to 1 if the previous trigger has both dmode of 0 and chain of 1
+  if (index > 0 && !triggers[index-1]->get_dmode() && triggers[index-1]->get_chain() && get_field(val, CSR_TDATA1_DMODE(xlen)))
+    return false;
+
+  unsigned type = get_field(val, CSR_TDATA1_TYPE(xlen));
+  reg_t tdata1 = val;
+  reg_t tdata2 = triggers[index]->tdata2_read(proc);
+
+  // hardware must zero chain in writes that set dmode to 0 if the next trigger has dmode of 1
+  const bool allow_chain = (index+1 < triggers.size() && triggers[index+1]->get_dmode() && !get_field(val, CSR_TDATA1_DMODE(xlen))) ? false : true;
```
scottj97: `x ? false : true` is the same as `!x`.

Untested:
```
const bool allow_chain = (index+1 >= triggers.size()) || !triggers[index+1]->get_dmode() || get_field(val, CSR_TDATA1_DMODE(xlen));
```

scottj97: What about setting chain=1 on the highest-index trigger? There is no next trigger to chain with. Should that index have chain forced to 0?

scottj97 (changes requested):
I think this is very close to perfect.

riscv/triggers.cc
timsifive: The debug spec is silent about this. Forcing it to 0 seems like reasonable hardware behavior, especially since it might save a bit that is otherwise useless.

YenHaoChen: Should we do this undocumented behavior?

timsifive: When the spec doesn't say, choose the simplest implementation. I wouldn't make a special case for this.

YenHaoChen:
I want to stabilize this PR.
This PR discussed the action legalization and S/U/VS/VU extension support. Thus, I add another commits to fix those issues in mcontrol trigger.

scottj97 (approved):

scottj97:
I have not tested the itrigger or etrigger functionality, except reads/writes to `tdata1` and `tdata2`, but I have tested that `mcontrol` still works as expected.

@YenHaoChen thank you for the contribution and for your patience with the review process.

timsifive:
I'm working on OpenOCD and riscv-tests/debug coverage of itrigger and etrigger. I have an itrigger test working well enough that I think spike is doing the right thing.

aswaterman:
Thanks again, @YenHaoChen (and @scottj97)!

PR #1130, opened 2022-10-25
Clean up dual-endian support

### LINKED ISSUE #1129: Assertion `endianness == memif_endianness_little || endianness == memif_endianness_big' failed.
Hello, I am getting the error described in the title when I am trying to run tests in rocket chip. 

In order to get past this error and continue my work I hardcoded `virtual memif_endianness_t get_target_endianness()` to return `memif_endianness_little` in memif.h.

In order to build riscv-isa-sim I followed the instructions in the README. Have i configured something wrong, or is this a bug?

Assertion error:

>  emulator-freechips.rocketchip.system-freechips.rocketchip.system.DefaultConfig: ../fesvr/htif.h:35: T htif_t::from_target(target_endian<T>) const [with T = long unsigned int]: Assertion `endianness == memi      f_endianness_little || endianness == memif_endianness_big' failed.

Thanks!

jerryz123:
This PR reduces the brittleness of the interfaces for dual-endian support.

- Remove `memif_endianness_undecided`. The default endianness is changed to little endian.
- Remove `set_target_endianness`. As this was only called at object initialization, this behavior can be moved into the constructors of relevant classes. Now, an additional flag `--big-endian` constructs the simulator with a big-endian memory interface, and is necessary for loading of big-endian ELFs.

Resolves #1129, https://github.com/chipsalliance/rocket-chip/issues/3159,

### DIFF
--- a/fesvr/elfloader.cc
+++ b/fesvr/elfloader.cc
@@ -98,7 +98,9 @@ std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* 
   } while (0)
 
   if (IS_ELFLE(*eh64)) {
-    memif->set_target_endianness(memif_endianness_little);
+    if (memif->get_target_endianness() != memif_endianness_little) {
+      throw std::invalid_argument("Specified ELF is little endian, but system uses a big-endian memory system. Rerun without --big-endian");
+    }
     if (IS_ELF32(*eh64))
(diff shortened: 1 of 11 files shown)


aswaterman (approved):
This is an improvement all around.  I agree the endianness should be a property of the target machine, and that should be sanity-checked against the loaded ELF, rather than vice-versa.

Defaulting to little-endian is appropriate, given the PrivArch spec says that harts reset into little-endian mode if both endiannesses are supported.

aswaterman:
I thought about suggesting the second commit be split up, but the two major changes really are coupled.

PR #1131, opened 2022-10-25
Add Spike support for Zvfh and Zvfhmin extensions

aswaterman:
99% of this support already existed (thanks to @chihminchao et al.).  But RVIA has recently standardized these as separate extensions, rather than being implied by the combination of V + Zfh.  Instead, the FP16 <-> FP32 vector conversion instructions are now part of Zvfhmin, and all of the other FP16 vector instructions are part of Zvfh.  Finally, Zvfh implies Zfhmin (and of course it also implies Zvfhmin).


liweiwei90:
LGTM. By the way, where can I find the spec for them? I didn't find it in riscv-isa-manual and riscv-v-spec. And Is there zfinx version for them? like Zvhinx{min}?

aswaterman:
https://github.com/riscv/riscv-v-spec/blob/d404111e51298b428f5bba6041d6288a6a209664/v-spec.adoc#zvfhmin-vector-extension-for-minimal-half-precision-floating-point-arithmetic

aswaterman:
No, there is no Zfinx version.  The V extension requires F and D.

liweiwei90:
OK. Thanks a lot.

aswaterman:
@scottj97 feel free to merge if you approve

scottj97 (approved):
I don't know anything about the extension changes, but this doesn't seem to break anything.

PR #1132, opened 2022-10-26
fix print message of width=8 accesses

YenHaoChen:
The number of digits indicates the width of accesses. An access with width=8 needs two digits instead of one.


liweiwei90:
LGTM.

PR #1133, opened 2022-10-26
fix mevent_mask for Sscofpmf extension

YenHaoChen:
The mevent_mask of Sscofpmf depends on the support of privilege modes. The UINH, SINH, VSINH, and VUINH are read-only 0 if unsupporting the corresponding modes.
(https://github.com/riscv/riscv-count-overflow/issues/3#issuecomment-1291247070)

This commit fixes the mevent_mask of Sscofpmf accordingly. The previous MHPMEVENTH_UINH is a typo, which should be MHPMEVENT_SINH instead.


liweiwei90:
LGTM.

aswaterman (approved):

PR #1136, opened 2022-10-27
Allow stdin to be a different device from stderr for debug mode

### LINKED ISSUE #1134: Unattended debug mode with piped stdin/stderr
Hello,

I’m making a toy RISC-V implementation as a learning project. To test it, I’d like to run my simulation in lock-step with Spike and inspect registers between each instruction. In an interactive terminal, I can run `spike -d` and use debug mode and manually do this kind of inspection. However I haven’t managed to automate it.

My test harness runs spike as a sub-process, and communicates with it through a pair of Unix pipes connected to its stdin and stderr. However instead of blocking while waiting for input, it looks like every [`read` call from Spike’s stdin](https://github.com/riscv-software-src/riscv-isa-sim/blob/5a50590f25a932cc1f25fe78b9912e9661d37d30/riscv/interactive.cc#L122) returns immediately with a zero-size read. Spike treats this the same as an empty line, which is the same as `run 1`. So the simulation runs quickly and ends (it’s a short program) before any command was read.

My test harness is in Rust, the subprocess creation looks like this:

```rust
std::process::Command::new("spike")
    // […]
    .arg("-d")
    .arg(elf_file_path)
    .stdin(std::process::Stdio::piped())
    .stderr(std::process::Stdio::piped())
    .stdout(std::process::Stdio::null())
    .spawn()
```

Is there some way to configure the stdin pipe so that reads are blocking? (If necessary instead of `Stdio::piped()` I can pass a file descriptor created with libc calls.) Or should I rather do lock-step simulation in an entirely different way? I found https://github.com/riscv-software-src/riscv-isa-sim/issues/911 but the approach there seems to involve hacking the build system and using Spike as a C++ library, I’d rather avoid that if possible.

SimonSapin:
Fixes https://github.com/riscv-software-src/riscv-isa-sim/issues/1134

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -98,8 +98,10 @@ static void send_key(bool noncanonical, int fd, keybuffer_t key_code, const int 
   }
 }
 
-static std::string readline(int fd)
+static std::string readline()
 {
+  int input_fd = STDIN_FILENO;
+  int fd = STDERR_FILENO;
   struct termios tios;
   // try to make sure the terminal is noncanonical and nonecho
   if (tcgetattr(fd, &tios) == 0)
@@ -119,7 +121,7 @@ static std::string readline(int fd)
   size_t cursor_pos = s.size();
   const size_t initial_s_len = cursor_pos;
   std::cerr << s << std::flush;
-  for (char ch; read(fd, &ch, 1) == 1; )
+  for (char ch; read(input_fd, &ch, 1) == 1; )
   {
     uint32_t keycode = key_buffer << BITS_PER_CHAR | ch;
     switch (keycode)
@@ -274,7 +276,7 @@ std::string sim_t::rin(boost::asio::streambuf *bout_ptr) {
     // output goes to socket
     sout_.rdbuf(bout_ptr);
   } else { // if we are not listening on a socket, get commands from terminal
-    s = readline(2); // 2 is stderr, but when doing reads it reverts to stdin
+    s = readline();
     // output goes to stderr
     sout_.rdbuf(std::cerr.rdbuf());
   }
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
This breaks a different use case we care about: using interactive mode to debug processes in the target machine that read from the host's stdin (as in, `cat input | spike -d program_that_consumes_input`).  Effectively, the target machine is granted stdin and stdout, and stderr is reserved for the host.  See https://github.com/riscv-software-src/riscv-isa-sim/blob/ab3225a3ff687fda8b4180f9e4e0949a400d1247/fesvr/syscall.cc#L172

Unfortunately, I think any solution to your problem will need to involve sticking to stderr.

SimonSapin:
Is there some way to make stderr of a child process be a read-write device (like tty in "normal" interactive mode) that communicates both ways with its parent process?

Alternatively, is there another way to have another program control a Spike simulation? Preferably without calling into `libriscv.so`’s C++ API

aswaterman:
Unfortunately, I'm not an enough of an expert on the TTY stuff to know.

At the moment, the C++ bindings _are_ the way of controlling Spike through an alternate interface.  If the goal is to avoid infecting your Rust project with C++, then conceivably you could build a different frontend around Spike that uses the C++ bindings but runs in a separate process and communicates over a pipe... which I guess is essentially doing what this PR does, just not upstream.

SimonSapin:
> At the moment, the C++ bindings _are_ the way of controlling Spike through an alternate interface.

That’s good to know, thanks

> to avoid infecting your Rust project with C++

"Infecting" is not the word I would use. Integrating multiple build systems and doing FFI can be complex and fragile, whereas `spike -d` already exits.

> some way to make stderr of a child process be a read-write device

I’ve managed to make this work by creating an anonymous Unix socket (which is bidirectional unlike pipes), turning one end into a file descriptor, and setting that as Spike’s stderr

> This breaks a different use case we care about

Given this I’ll close the PR. Thanks for the feedback!

aswaterman:
Cool, that sounds like an elegant solution to me.

PR #1137, opened 2022-10-29
Add hit-count by instruction name to HISTOGRAM

Incarnation-p-lee:
* Controlled by HISTOGRAM-related macro.
* Disabled by default, requires configuring option '--enable-histogram'.

Signed-off-by: Pan Li <pan2.li@intel.com>


aswaterman (changes requested):
I think it would be cleaner and faster to do this in a post-processing script instead of as part of Spike.  (Of course, that only works if you have a correspondence between PC and static instruction, which means you are running only a single program... but I think that is a common case.)

riscv/processor.cc
```diff
@@ -77,6 +77,13 @@ processor_t::~processor_t()
     fprintf(stderr, "PC Histogram size:%zu\n", pc_histogram.size());
     for (auto it : pc_histogram)
       fprintf(stderr, "%0" PRIx64 " %" PRIu64 "\n", it.first, it.second);
+
+    fprintf(stderr, "Instruction Name Histogram size:%zu\n", insn_name_histogram.size());
+    for (auto it : insn_name_histogram) {
+      auto insn_name = new char[it.first.size() + 1] { '\0' };
```
aswaterman: memory leak?

Incarnation-p-lee: I see, C++ requires explicit deconstruct. Thank you, let me update it.

Incarnation-p-lee:
> I think it would be cleaner and faster to do this in a post-processing script instead of as part of Spike. (Of course, that only works if you have a correspondence between PC and static instruction, which means you are running only a single program... but I think that is a common case.)

Do you think it is suggested to move all `HISTOGRAM` to that post-processing script? As I understand, the PC based stat may have no difference compared to the instr name.

Or you may have a better suggestion about it from the expert's perspective?

riscv/processor.cc
Incarnation-p-lee: Updated.

aswaterman:
> Do you think it is suggested to move all `HISTOGRAM` to that post-processing script?

For the specific case you mention, I think it can be done with no modification to Spike.  You can combine the current histogram output with the ELF objdump to produce the instruction opcode histogram.

More generally, I do think Spike needs to do _some_ of the work; the question is, how much.  @sifiverobert may have some thoughts to share on this topic.

Incarnation-p-lee:
> > Do you think it is suggested to move all `HISTOGRAM` to that post-processing script?
> 
> For the specific case you mention, I think it can be done with no modification to Spike. You can combine the current histogram output with the ELF objdump to produce the instruction opcode histogram.
> 
> More generally, I do think Spike needs to do _some_ of the work; the question is, how much. @sifiverobert may have some thoughts to share on this topic.

I see, combine the objdump and PC-based histogram can get similar data, but as I understanding this cannot cover JIT or self-modified code?

sifiverobert:
@Incarnation-p-lee 

I really wonder why you want to see how many instructions with particular 'name' were executed? Do you want to remove some instructions from HW assuming you do NOT see them in a code?

Histogram of PC (every individual PC has 'count') is useful as it shows 'hot-spots' in a program. 

Thy only reasonable other count could be counts by types of instructions (loads/stores/calls/returns/jumps/branches/...), which can provide you with 'profile' of a program (data intensive, call intensive, jumps a lot etc.). In general less data-transfers is done by a program, it may run faster. For example if you re-compile program with different options (or link with different libraries ...) you can see if this is going into 'better-direction'.

I think doing any sort of 'analysis' as post-process is better as it can be shared with many other simulators or even PC trace produced from real SoC. All you really need is a stream of PCs (which Spike is already produced).

It also makes simulator a simulator and not 'analyzer of execution' (in the same time ...). Adding different types of 'analysis' will expand the code, which will be harder to maintain and will run slower.

And I am not sure if JIT is supported by Spike ... Self modifying code is very tricky, makes caches useless and it has limited use-case ...

Extensive analysis of opcodes and 'sequences of instructions' (frequency etc.) was done as part of 'Code Size Reduction TG: https://github.com/riscv/riscv-code-size-reduction/tree/main/benchmarks

This is processing entire ELF file (all instructions ...), while you want to histogram of instructions actually executed.

Incarnation-p-lee:
@sifiverobert 

Thank you for sharing so much useful information. The goal of hit count by instruction name is to provide the instruction stat by workload. For example, I would like to learn the top 10 hottest instructions by TVM and/or MLPerf. With such kind of information, the HW may pay more attention to these hot instructions.

In the previous, I leveraged something combination of both the PC-based and `objdump` as you mentioned. Considering there may be JIT workload from ML/DL, I file this PR to get the instruction(s) name(s) stat when processing.

sifiverobert:
Now I see - you want to learn what instructions are 'hot' in typical workload and focus (in HW) on them. For example if 'div' is used only 3 times, it does not make sense to speed-it up. Or maybe not implement at all in HW and 'emulate' in exception handler. When processor will run only one, fixed program at the end, this kind of 'HW trimming' makes sense. But SW is evolving and ML/DL may be different tomorrow ...

I also understand why you mentioned JIT/self-modifying code - I was not aware what kind of 'trickery' ML/DL does ...

Again - I still think post-processing (with minimal changes in simulator, which is sending what is needed) is better.

For this case, best would be to output <PC, opcode> records. It is rather trivial to produce it from ANY simulator. Even if code is dynamic (JIT or self-modifying ...), you will know ALL what is executed. RISC-V ISA is not so complex, so it is easy to detect instruction type/group looking at few opcode bytes.

I think <PC,opcode> output allows to implement many counting/analysis ideas even Without looking at ELF/OBJDUMP.

Incarnation-p-lee:
Yes, you are right.

For the <PC, opcode>, we have a `PC => hit count` map already in spike currently. To double confirm, do you suggest enhancing this map with `opcode`? For example, a map `PC-opcode => hit count` or `PC-opcode-instr_name => hit count`.

Then I can reconcile the PR for this.

aswaterman:
I do not promise we will accept such a PR.  The existing histogram functionality is very hacky, IMO, and extending it further would just seem to increase the amount of technical debt.  Until a more general solution is discussed and proposed, you might just want to maintain your patches on your own branch.

Incarnation-p-lee:
> I do not promise we will accept such a PR. The existing histogram functionality is very hacky, IMO, and extending it further would just seem to increase the amount of technical debt. Until a more general solution is discussed and proposed, you might just want to maintain your patches on your own branch.

Thank you for this important information. Please feel free to let me know if a more general solution is proposed.

@sifiverobert It looks not the right way to do this as `The existing histogram functionality is very hacky`. We may need a better design/solution for performance related stat.

sifiverobert:
I was even not aware (until I noticed this PR ...) that spike has any kind of histogram.
IMO best way for any 'post-analysis' is <PC, opcode>pairs EMITTED from simulator.
I am NOT suggesting to enhance histogram with <opcode>, but rather REMOVE histogram ('...very hacky') from Spike.
As I said sequence of logical <PC, opcode> records (as an output ...) is very much needed and 'trivial' to do.
It can be even binary file with (for faster 'export' and consumption) - it can be piped to another process for live consumption if needed.
Having <PC, opcode> stream, existing 'histogram' can be created (few lines of code if you do not mind top-notch efficiency ...). And any other analysis tools can be built on this. Without looking at ELF file and being 'universal' as other simulators may produce <PC, opcode> records as well.

scottj97:
> Having <PC, opcode> stream, existing 'histogram' can be created (few lines of code if you do not mind top-notch efficiency ...). And any other analysis tools can be built on this. Without looking at ELF file and being 'universal' as other simulators may produce <PC, opcode> records as well.

Isn't the commit log sufficient for that?

```
core   0: 0x0000000080002016 (0x00026117) auipc   sp, 0x26
core   0: 3 0x0000000080002016 (0x00026117) x2  0x0000000080028016
core   0: 0x000000008000201a (0x6da10113) addi    sp, sp, 1754
core   0: 3 0x000000008000201a (0x6da10113) x2  0x00000000800286f0
```

sifiverobert:
@scottj97: Yes. It is exactly what is needed (and even more ...). I am not 'fluent' in running Spike. It even has disassembly, so @Incarnation-p-lee can do all the 'counting' outside today (without looking at ELF file at all).

Incarnation-p-lee:
Thank you all, I think the commit log with some post-processing script is good enough.

BTW, @sifiverobert and @aswaterman  do you think we need to remove the existing `histogram` staff? Consider it is `'...very hacky`. If so I may help to make it happen.

aswaterman:
Although I personally would be OK with removing it, I know it has existing users, so there might be some consternation if we remove it.

sifiverobert:
No problem to keep it (from 'legacy reasons' and most probably execution penalty is very small if not enabled. But I would try to more 'analysis' things outside of spike.

Incarnation-p-lee:
@aswaterman and @sifiverobert Thank you all, if no more questions I will close this PR by the end of today.

Incarnation-p-lee:
Closed this PR as discussed.

PR #1141, opened 2022-11-04
Add support for Zc* extension

liweiwei90:
This patchset implements RISC-V Zc* extension v1.0.0.RC5.7 version instructions. It can pass the basic zc tests from https://github.com/yulong-plct/zc-test.


aswaterman:
Thanks.  I'll try to review this next week.

riscv/isa_parser.cc
```diff
@@ -155,6 +155,20 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
     } else if (ext_str == "zhinxmin") {
       extension_table[EXT_ZFINX] = true;
       extension_table[EXT_ZHINXMIN] = true;
+    } else if (ext_str == "zca") {
+      extension_table[EXT_ZCA] = true;
+    } else if (ext_str == "zcf") {
+      if (max_xlen != 32)
+        bad_isa_string(str, "'Zcf' is only relevant to RV32");
```
aswaterman: "'Zcf' requires RV32" would be more consistent with the other error messages.

riscv/isa_parser.cc
```diff
@@ -260,6 +274,17 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
     bad_isa_string(str, ("Zfinx/ZDinx/Zhinx{min} extensions conflict with Base 'F/D/Q/Zfh{min}' extensions"));
   }
 
+  if ((extension_table[EXT_ZCMP] || extension_table[EXT_ZCMT]) &&
+      (extension_table[EXT_ZCD] || (extension_table['C'] && extension_table['D']))) {
```
aswaterman: I think we should do something like

```
  if (extension_table['C'] && extension_table['D'])
    extension_table[EXT_ZCD] = true;
```

In addition to simplifying this check, this simplifies other code later in this PR: e.g. you won't need to use `require_either_extension` inside of instructions like `c.fsd`.

Same for Zca and Zcf.

riscv/isa_parser.cc
```diff
@@ -260,6 +274,17 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
...
+    bad_isa_string(str, "'Zcmp/Zcmt' are incompatible with Zcd which is included "
+                        "when 'C' and 'D' extension are both present");
```
aswaterman: "'Zcmp' and 'Zcmt' are incompatible with 'Zcd', and are incompatible with 'C' when 'D' is present"

riscv/processor.h
```diff
@@ -283,7 +283,7 @@ class processor_t : public abstract_device_t
     return impl_table[impl];
   }
   reg_t pc_alignment_mask() {
-    return ~(reg_t)(extension_enabled('C') ? 0 : 2);
+    return ~(reg_t)(extension_enabled('C') || extension_enabled(EXT_ZCA)  ? 0 : 2);
```
aswaterman: After making my suggested change to implicitly enable Zca when C is present, this should be simplified to check only Zca.

riscv/processor.cc
```diff
@@ -1073,6 +1073,7 @@ void processor_t::register_base_instructions()
   #undef DECLARE_INSN
 
   #define DECLARE_OVERLAP_INSN(name, ext) { name##_supported &= isa->extension_enabled(ext); }
+  #define DECLARE_OVERLAP_INSN_OR(name, ext) { name##_supported |= isa->extension_enabled(ext); }
```
aswaterman: The name "DECLARE_OVERLAP_INSN_OR" doesn't clearly indicate its purpose to me.

riscv/insns/c_sext_b.h
```diff
@@ -0,0 +1,3 @@
+require_extension(EXT_ZCB);
+require_extension(EXT_ZBB);
+WRITE_RVC_RS1S(sext8(RVC_RS1S));
```
aswaterman: It is strange that the definition of `c.sext.b` looks different from the definition of `sext.b`.  Please either change `sext.b` to use this new macro, or change `c.sext.b` to look like `sext.b`.  Your choice.  Same for all of the other instructions.

riscv/decode.h
```diff
@@ -226,6 +236,14 @@ class regfile_t
 #define RVC_FRS2S READ_FREG(insn.rvc_rs2s())
 #define RVC_SP READ_REG(X_SP)
 
+// ZCE macros
+#define S0_1(n) (X_S0 + n)
+#define Sn(n)   (X_Sn + n)
```
aswaterman: Why not define `Sn` to be `((n) < 2 ? X_S0 + (n) : X_Sn + (n))`?  Then you can delete the `S0_1` macro and simplify other logic downstream.

riscv/decode.h
```diff
@@ -472,6 +490,123 @@ inline long double to_f(float128_t f) { long double r; memcpy(&r, &f, sizeof(r))
   reg_t s##field = get_field(STATE.senvcfg->read(), SENVCFG_##field); \
   reg_t h##field = get_field(STATE.henvcfg->read(), HENVCFG_##field)
 
+#define XREG_LIST { \
```
aswaterman: It would be much cleaner to have this return a value rather than assume the existence of a variable named `xreg_list`.

I suggest deleting this and instead add a method to `insn_t` that returns this as a mask.  Something like:


```
uint64_t zcmt_regmask()
{
  unsigned mask = 0;

  if (rvc_rlist() >= 4)
    mask |= 1U << X_RA;

  for (int i = 5; i < 16; i++)
    if (rvc_rlist() >= i)
      mask |= 1U << Sn(i);

  return mask;
} 
```

riscv/decode.h
```diff
@@ -472,6 +490,123 @@ inline long double to_f(float128_t f) { long double r; memcpy(&r, &f, sizeof(r))
...
+    xreg_list[X_S1] = 1; \
+  case 5: \
+    xreg_list[X_S0] = 1; \
+  case 4: \
+    xreg_list[X_RA] = 1; \
+    break; \
+  } \
+}
+
+#define STACK_ADJ {\
```
aswaterman: Similarly, delete this and add a method to `insn_t` called `zcmt_stack_adjustment(int xlen)` that returns the stack adjustment.

riscv/decode.h
```diff
@@ -472,6 +490,123 @@ inline long double to_f(float128_t f) { long double r; memcpy(&r, &f, sizeof(r))
...
+      stack_adj_base += 16; \
+  case 5: \
+  case 4: \
+    stack_adj_base += 16; \
+    break; \
+  } \
+  stack_adjust = stack_adj_base + spimm; \
+}
+
+#define ZCMP_PUSH(bytes, rlist, spimm) \
```
aswaterman: Can you try to rewrite this as template function instead of a macro?  I know most of this file uses macros, but we are trying to get away from that style.

riscv/decode.h
```diff
@@ -472,6 +490,123 @@ inline long double to_f(float128_t f) { long double r; memcpy(&r, &f, sizeof(r))
...
+      MMU.store<uint64_t>(addr, data); \
+      break; \
+    default: \
+      break; \
+    } \
+    addr -= bytes; \
+} \
+WRITE_REG(X_SP, SP - stack_adjust)
+
+#define ZCMP_POP(bytes, rlist, spimm, ret_val, ret) \
```
aswaterman: Same comment as for ZCMP_PUSH.

riscv/mmu.h
```diff
@@ -225,6 +225,18 @@ class mmu_t
     return (addr / PC_ALIGN) % ICACHE_ENTRIES;
   }
 
+  inline reg_t fetch_jump_table(reg_t addr, int length)
+  {
+    auto tlb_entry = translate_insn_addr(addr, true);
```
aswaterman: I think all of this will change soon; there is discussion about making fetch table accesses perform read permission checks instead of fetch permission checks.  I recommend we not merge this until that change has been made, because it will dramatically change this code.

After that change is made, I think you can simply call `load_slow_path_intrapage` here.  I don't want to create special cases for `is_jvt`.

disasm/disasm.cc
```diff
@@ -1203,7 +1277,7 @@ void disassembler_t::add_instructions(const isa_parser_t* isa)
   }
 
   // ext-c
-  if (isa->extension_enabled('C')) {
+  if (isa->extension_enabled('C') || isa->extension_enabled(EXT_ZCA)) {
```
aswaterman: This can be simplified to only checking for Zca if you follow my advice to implicitly enable Zca when C is present.

aswaterman (changes requested):
This looks pretty good overall, but I would like a few minor changes.  And I think we should wait to merge until https://github.com/riscv/riscv-code-size-reduction/issues/195 is resolved.

riscv/isa_parser.cc
liweiwei90: OK. I'll update it.

riscv/isa_parser.cc
liweiwei90: This way can simplify the check. There is a question here: "Zcd requires Zca". If we also set Zca = true when C is enabled, it have another question, whether rv32imac_zcb is legal? I didn't find the explicit description about whether C imply Zca and  Zcf.

riscv/isa_parser.cc
liweiwei90: OK. I'll update this. By the way, the upper is the original description from the spec.

riscv/isa_parser.cc
aswaterman: C implies Zca, and so C_Zca, C_Zcb, C_Zcd, and RV32C_Zcf should all be legal.

riscv/isa_parser.cc
aswaterman: Both are correct; this version is just a bit easier to read.

riscv/processor.cc
liweiwei90: This is for Zcd or C && D. If the upper question is solved. This can be deleted.

aswaterman: Great!

riscv/decode.h
liweiwei90: OK.

riscv/insns/c_sext_b.h
liweiwei90: OK. I'll change this to the way of sext.b by type conversion.

riscv/decode.h
liweiwei90: OK. I'll update this.

riscv/decode.h
liweiwei90: OK.

riscv/decode.h
liweiwei90: OK. I'll update this.

riscv/mmu.h
liweiwei90: OK. I'll update this if spec changes.

aswaterman: BTW, if the change is not made, and we instead continue to use execute permissions, then I have an alternate suggestion.

The spec says "It is recommended that the second fetch be ignored for hardware triggers and breakpoints."  That is, it is still legal to check the triggers.  So, I recommend that we take advantage of this property, and continue to check the triggers, to avoid adding a special case to this complicated code.

riscv/isa_parser.cc
liweiwei90: If C implies Zca, whether it implies Zcf and Zcd?

liweiwei90: That is to say whether C_Zcmt(without D) is legal?

aswaterman: C implies Zcd only if D is specified.

C implies Zcf only if F is specified and the base ISA is RV32.

C_Zcmt is legal if neither D nor Zcd is specified.

liweiwei90: OK. Thanks for your patient explanation.  By the way,  whether "require_extension('F') can be deleted for Zcf instructions? that is to say they can be legal when zcf and zfinx are specified without  F.

aswaterman: As long as you don't break backwards compatibility, that sounds OK.

scottj97:
I've run 11934c5256362932c5d7d4ebb1523a8ef9196774 through my usual battery of tests and found no regressions, but I don't plan to review the code.

aswaterman (approved):
Thanks for the edits.  Looks good now.

I decided I didn't like the Zcmp logic being added to mmu_t, so I'll find a way to rewrite it, then I'll manually merge this.

aswaterman:
@liweiwei90 I rewrote the Zcmp patch to disentangle it from the MMU.  Can you sanity check 1160ea7f1b1026884bef3b8ac2caf613c8e9a475?

liweiwei90:
Currently, there are only several basic tests for push/pop get from our gnu-toolchain team, you can get it from https://github.com/yulong-plct/zc-test.

> @liweiwei90 I rewrote the Zcmp patch to disentangle it from the MMU. Can you sanity check [1160ea7](https://github.com/riscv-software-src/riscv-isa-sim/commit/1160ea7f1b1026884bef3b8ac2caf613c8e9a475)?

PR #1145, opened 2022-11-16
Document implemented trigger configuration

scottj97:
Adds an explanation to the README about what trigger types are supported.

I want this in place because it's about to change with #1128 and this README will change as needed, to document the trigger changes as they're made.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Spike supports the following RISC-V ISA features:
   - Svpbmt extension, v1.0
   - Svinval extension, v1.0
   - Debug v0.14
+    - 4 triggers fixed to type=2 (mcontrol)
   - Smepmp extension v1.0
   - Smstateen extension, v1.0
   - Sscofpmf v0.5.2


timsifive (approved):
Looks good.

A very minor nit is that spike only supports timing=before for execute triggers. Probably not worth mentioning here, though.

aswaterman (approved):

sifiverobert:
I think 'before' is desired for execute triggers (as 'after' jump/branch may be very hard to harness ...). If someone wants 'after' they may always single-step.

timsifive:
Yes, but spike also plays the role of a reference platform, so from that point of view it makes sense to implement as many features as possible.

sifiverobert:
Good angle. But even not practical ones? I do not know any debugger which will tolerate 'after' on PC-match (=HW breakpoint). IMO this should be not allowed to implement PC-match AFTER without BEFORE. If someone will do it, they will have nothing but problems.
Skid on data is acceptable (and common ...).

PR #1150, opened 2022-11-21
Reimplement memory merging routines

aap-sc:
This commit simplifies the codebase by factoring out memory merging and intersection checking routines into separate functions. This allows us to check for corner cases more easily.

### DIFF
--- a/spike_main/spike.cc
+++ b/spike_main/spike.cc
@@ -123,31 +123,49 @@ bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
     return (a.base < b.base);
 }
 
-void merge_overlapping_memory_regions(std::vector<mem_cfg_t> &mems)
+static bool check_mem_overlap(const mem_cfg_t& L, const mem_cfg_t& R)
 {
-  // check the user specified memory regions and merge the overlapping or
-  // eliminate the containing parts
-  assert(!mems.empty());
+  const reg_t L_end   = L.base + L.size - 1;
+  const reg_t R_end   = R.base + R.size - 1;
+
+  return std::max(L.base, R.base) <= std::min(L_end, R_end);
+}
+
+static mem_cfg_t merge_mem_regions(const mem_cfg_t& L, const mem_cfg_t& R)
+{
+  // one can merge only intersecting regions
+  assert(check_mem_overlap(L, R));
+
+  const reg_t merged_base = std::min(L.base, R.base);
+  const reg_t merged_end_incl = std::max(L.base + L.size - 1, R.base + R.size - 1);
+  const reg_t merged_size = merged_end_incl - merged_base + 1;
+
+  return mem_cfg_t(merged_base, merged_size);
+}
+
+// check the user specified memory regions and merge the overlapping or
+// eliminate the containing parts
+static std::vector<mem_cfg_t>
+merge_overlapping_memory_regions(std::vector<mem_cfg_t> mems)
+{
+  if (mems.empty())
+    return {};
 
   std::sort(mems.begin(), mems.end(), sort_mem_region);
-  for (auto it = mems.begin() + 1; it != mems.end(); ) {
-    reg_t start = prev(it)->base;
-    reg_t end = prev(it)->base + prev(it)->size;
-    reg_t start2 = it->base;
-    reg_t end2 = it->base + it->size;
 
-    //contains -> remove
-    if (start2 >= start && end2 <= end) {
-      it = mems.erase(it);
-    //partial overlapped -> extend
-    } else if (start2 >= start && start2 < end) {
-      prev(it)->size = std::max(end, end2) - start;
-      it = mems.erase(it);
-    // no overlapping -> keep it
-    } else {
-      it++;
+  std::vector<mem_cfg_t> merged_mem;
+  merged_mem.push_back(mems.front());
+
+  for (auto mem_it = std::next(mems.begin()); mem_it != mems.end(); ++mem_it) {
+    const auto& mem_int = *mem_it;
+    if (!check_mem_overlap(merged_mem.back(), mem_int)) {
+      merged_mem.push_back(mem_int);
+      continue;
     }
+    merged_mem.back() = merge_mem_regions(merged_mem.back(), mem_int);
   }
+
+  return merged_mem;
 }
 
 static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
@@ -217,9 +235,10 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
     arg = p + 1;
   }
 
-  merge_overlapping_memory_regions(res);
+  auto merged_mem = merge_overlapping_memory_regions(res);
 
-  return res;
+  assert(!merged_mem.empty());
+  return merged_mem;
 }
 
 static std::vector<std::pair<reg_t, mem_t*>> make_mems(const std::vector<mem_cfg_t> &layout)


aap-sc:
This commit is continuation of discussion from https://github.com/riscv-software-src/riscv-isa-sim/pull/1100 . This is NFT change to prepare the codebase for corner-case checks.

spike_main/spike.cc
```diff
@@ -123,31 +123,48 @@ bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
...
-  // check the user specified memory regions and merge the overlapping or
-  // eliminate the containing parts
-  assert(!mems.empty());
+  reg_t L_start = L.base;
+  reg_t L_end   = L.base + L.size - 1;
+
+  reg_t R_start = R.base;
+  reg_t R_end   = R.base + R.size - 1;
+
+  return std::max(L_start, R_start) <= std::min(L_end, R_end);
```
aswaterman: Minor stylistic thing, but I'd just use `L.base` here instead of declaring and using `L_start`.

spike_main/spike.cc
```diff
@@ -123,31 +123,48 @@ bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
...
 }
 
-void merge_overlapping_memory_regions(std::vector<mem_cfg_t> &mems)
+static bool check_mem_overlap(const mem_cfg_t& L, const mem_cfg_t& R)
 {
-  // check the user specified memory regions and merge the overlapping or
-  // eliminate the containing parts
-  assert(!mems.empty());
+  reg_t L_start = L.base;
+  reg_t L_end   = L.base + L.size - 1;
```
aswaterman: We've been trying to use `const` more consistently on local variable declarations that are meant to be immutable

spike_main/spike.cc
```diff
@@ -123,31 +123,48 @@ bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
     return (a.base < b.base);
 }
 
-void merge_overlapping_memory_regions(std::vector<mem_cfg_t> &mems)
+static bool check_mem_overlap(const mem_cfg_t& L, const mem_cfg_t& R)
```
aswaterman, line 126: Maybe make this a member function on `mem_cfg_t`, so that below you can write something like `assert(!L.overlaps(R))`

spike_main/spike.cc

spike_main/spike.cc

spike_main/spike.cc
aap-sc: I'm somewhat reluctant to do that, since this is the only place where we require this functionality. I can certainly do that if necessary - no big deal. My concern is that the interface of mem_cfg_t may become somewhat bloated (given that I'd like to introduce `get_base` and `get_inclusive_end` getters to mem_cfg_t). Could you please take a look once again?

aswaterman: Leaving it as-is is OK with me.

spike_main/spike.cc
```diff
@@ -123,31 +123,46 @@ bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
...
+  const reg_t merged_end_incl = std::max(L.base + L.size - 1, R.base + R.size - 1);
+  const reg_t merged_size = merged_end_incl - merged_base + 1;
+
+  return mem_cfg_t(merged_base, merged_size);
+}
+
+static void merge_overlapping_memory_regions(std::vector<mem_cfg_t> &mems)
+{
+  if (mems.empty())
+    return;
```
scottj97: This looks like a behavior change, since previously this was an assertion instead of a nop. Intentional?

EccoTheDolphin: yes and no :).

"Yes" in the sense that  `merge_overlapping_memory_regions` does not need an assertion - the function itself  does not care of any invariant, and is only concerned about the merge.
"No" in the sense that probably I should preserve the assertion. To me it seems that if we need one - it's better to add it at the end `parse_mem_layout` function (right before we return the result).

scottj97: Agree that it doesn't belong in this function, but if it's to be removed it should go in its own commit with an explanation of why.

So, assuming it's okay to keep it, I'd just amend this commit to put the assertion in `parse_mem_layout`.

EccoTheDolphin: makes sense, thank you! addressed.

scottj97: The amended commit seems to put the assertion back into `merge_overlapping_memory_regions` which I thought we just agreed is not where it belongs? The old code had this assumption, but the new code does not, so the assertion doesn't really belong there anymore.

EccoTheDolphin: > I thought we just agreed is not where it belongs?

My bad... Fixed.

spike_main/spike.cc
```diff
@@ -123,31 +123,45 @@ bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
...
-    //partial overlapped -> extend
-    } else if (start2 >= start && start2 < end) {
-      prev(it)->size = std::max(end, end2) - start;
-      it = mems.erase(it);
-    // no overlapping -> keep it
-    } else {
-      it++;
+
+  std::vector<mem_cfg_t> merged_mem;
+  merged_mem.push_back(mems.front());
```
scottj97, line 157: I'm still confused by this. What if `mems` is empty? Before, didn't you have a check here to return immediately if that's the case? Don't you still need that check?

I thought what you were saying the other day is that the assertion `assert(!mems.empty());` was no longer needed because this function would still work correctly (doing nothing) even if `mems` was empty, so that's why you removed it. But if I'm reading this right, it doesn't look like this function will work correctly in that case any more.

spike_main/spike.cc
```diff
@@ -123,31 +123,45 @@ bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
...
+  for (auto mem_it = std::next(mems.begin()); mem_it != mems.end(); ++mem_it) {
+    const auto& mem_int = *mem_it;
+    if (!check_mem_overlap(merged_mem.back(), mem_int)) {
+      merged_mem.push_back(mem_int);
+      continue;
     }
+    merged_mem.back() = merge_mem_regions(merged_mem.back(), mem_int);
   }
+
+  mems.swap(merged_mem);
```
scottj97: While we're here making these changes, instead of modifying the function's argument, why not return the new vector and let the caller decide what to do with it?

scottj97 (changes requested):

spike_main/spike.cc
EccoTheDolphin:
```
While we're here making these changes, instead of modifying the function's argument, why not return the new vector and let the caller decide what to do with it?
```
I've just tried to preserve the old behavior, but I agree that probably it would be more readable if I were to return the new value. Will adjust.

spike_main/spike.cc
EccoTheDolphin: > What if mems is empty? 

Yes you are right, this is yet another mistake on my part :(. Thanks for catching this.

EccoTheDolphin:
Nevermind this iteration. It's incorrect.

spike_main/spike.cc
EccoTheDolphin: fixed

spike_main/spike.cc
EccoTheDolphin: adjusted

EccoTheDolphin:
@scottj97 , could you take a look once again? I believe I've addressed recent comments.

spike_main/spike.cc
```diff
@@ -123,31 +123,49 @@ bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
...
+  const reg_t merged_end_incl = std::max(L.base + L.size - 1, R.base + R.size - 1);
+  const reg_t merged_size = merged_end_incl - merged_base + 1;
+
+  return mem_cfg_t(merged_base, merged_size);
+}
+
+// check the user specified memory regions and merge the overlapping or
+// eliminate the containing parts
+static std::vector<mem_cfg_t>
+merge_overlapping_memory_regions(std::vector<mem_cfg_t> mems)
```
scottj97, line 149: The param should be passed by const ref so it doesn't have to be copied

scottj97 (changes requested):

spike_main/spike.cc
EccoTheDolphin: but we have to copy these elements since we want to sort them.

scottj97: Oh. I see what you mean. `mems` is being modified-in-place inside the function for sort.

scottj97 (approved):

aswaterman (approved):

PR #1151, opened 2022-11-22
v_ext_macros: Drop index from `VI_ELEMENT_SKIP`

dgreid:
If a variable other than `i` is passed as `inx`, `VI_LOOP_ELEMENT_SKIP` will still check `i` in the `else` case, ignoring `inx`.

Since `i` is the only value ever passed, drop the argument to avoid any future confusion.

Signed-off-by: Dylan Reid <dgreid@rivosinc.com>

### DIFF
--- a/riscv/v_ext_macros.h
+++ b/riscv/v_ext_macros.h
@@ -20,10 +20,10 @@
     } \
   }
 
-#define VI_ELEMENT_SKIP(inx) \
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
Yet another reason I wish we had tried using templates instead of macros for this stuff.  Anyway, thanks.

PR #1153, opened 2022-11-22
Add tinfo register.

timsifive:
Not very interesting while spike only supports one trigger type, but #1128 is about to change that. Without tinfo, it can become quite slow for a debugger to discover which types are supported.


aswaterman (approved):
Fine by me, but I’ll give @scottj97 a chance to review the CSR-related changes.

scottj97 (approved):
@YenHaoChen I think @timsifive has an even better idea here than my README idea from last week. Can you please rebase #1128 onto this and keep `tinfo` updated along the way?

PR #1154, opened 2022-11-22
use extension_enabled() for 'H'

kwalker27:
The H bit in MISA isn't constant, so don't use
extension_enabled_const().

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -241,7 +241,7 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
     const reg_t mevent_mask = proc->extension_enabled_const(EXT_SSCOFPMF) ? MHPMEVENT_OF | MHPMEVENT_MINH
       | (proc->extension_enabled_const('U') ? MHPMEVENT_UINH : 0)
       | (proc->extension_enabled_const('S') ? MHPMEVENT_SINH : 0)
-      | (proc->extension_enabled_const('H') ? MHPMEVENT_VUINH | MHPMEVENT_VSINH : 0) : 0;
+      | (proc->extension_enabled('H') ? MHPMEVENT_VUINH | MHPMEVENT_VSINH : 0) : 0;
     mevent[i - 3] = std::make_shared<masked_csr_t>(proc, which_mevent, mevent_mask, 0);
     auto mcounter = std::make_shared<const_csr_t>(proc, which_mcounter, 0);
     csrmap[which_mcounter] = mcounter;


kwalker27:
Since I had to look it up, I'll paste the statement from the privilege spec:

```
The hypervisor extension is enabled by setting bit 7 in the misa CSR, which corresponds to the letter H.
RISC-V harts that implement the hypervisor extension are encouraged not to hardwire misa[7], so that
the extension may be disabled.
```

And from `csrs.cc` the `misa_csr_t.write_mask` set up confirms Spike's implementation choice:

```
  write_mask(max_isa & (0  // allow MAFDQCHV bits in MISA to be modified
                        | (1L << ('M' - 'A'))
                        | (1L << ('A' - 'A'))
                        | (1L << ('F' - 'A'))
                        | (1L << ('D' - 'A'))
                        | (1L << ('Q' - 'A'))
                        | (1L << ('C' - 'A'))
                        | (1L << ('H' - 'A'))
                        | (1L << ('V' - 'A'))
                        )
             ) {
```

scottj97 (changes requested):
The purpose of `extension_enabled_const` is to document assumptions about hard-wired bits in `misa`, so that if those assumptions ever become invalid, the code will fail, alerting us to the no-longer-valid assumption.

In this case, there **is** an assumption here that `misa.H` is hard-wired, since according to #1133 these two bits should be 0 if `misa.H==0`. Since `misa.H` is not hard-wired, this looks broken to me.

1. I don't know why `extension_enabled_const()` did not immediately fail when called, since [this line](https://github.com/riscv-software-src/riscv-isa-sim/blob/b8ed4e9ade7d8c56e215ddbb33c719d42bcaefa2/riscv/processor.h#L280) should have called [this line](https://github.com/riscv-software-src/riscv-isa-sim/blob/b8ed4e9ade7d8c56e215ddbb33c719d42bcaefa2/riscv/csrs.cc#L659) and fired that assertion.
2. It appears that `mevent` will need to use a custom CSR class to properly WARL these bits when `misa.H==0`, and when `misa.H` is cleared, [this code](https://github.com/riscv-software-src/riscv-isa-sim/blob/b8ed4e9ade7d8c56e215ddbb33c719d42bcaefa2/riscv/csrs.cc#L646-L652) should also clear those two bits of `mevent`.

@kwalker27 can you look into the first? And @YenHaoChen can you look into the second?

kwalker27:
Thanks @scottj97 - the assertion does fire when the ISA string includes both 'H' and 'Sscofpmf'. I see that my naive change is not the solution, so it'll be great if @YenHaoChen can help clean this up.

```
work/riscv-isa-sim/build$ ./spike --isa rv64gchv /rivos/riscv-pk/riscv64-unknown-linux-gnu/bin/pk hello
bbl loader
Hello world!

work/riscv-isa-sim/build$ ./spike --isa rv64gchv_Sscofpmf /rivos/riscv-pk/riscv64-unknown-linux-gnu/bin/pk hello
spike: ../riscv/csrs.cc:659: bool misa_csr_t::extension_enabled_const(unsigned char) const: Assertion `!(1 & (write_mask >> (ext - 'A')))' failed.
```

YenHaoChen:
@kwalker27 I create a branch for this issue (https://github.com/YenHaoChen/riscv-isa-sim/tree/pr-h-not-const). 
Would you give it a check?

kwalker27:
> @kwalker27 I create a branch for this issue (https://github.com/YenHaoChen/riscv-isa-sim/tree/pr-h-not-const). Would you give it a check?

It resolves my issue, thanks! I'm not testing the Sscofpmf behavior though so others may weigh in on that. I'll close this one.

PR #1155, opened 2022-11-23
hypervisor extension does not hardwire misa.H

YenHaoChen:
This PR corrects the wrong assumption that misa.H is read-only (hardwired).

Reported by @kwalker27 (https://github.com/riscv-software-src/riscv-isa-sim/pull/1154)

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -650,6 +650,10 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
     state->mie->write_with_mask(MIP_HS_MASK, 0);  // also takes care of hie, sie
     state->mip->write_with_mask(MIP_HS_MASK, 0);  // also takes care of hip, sip, hvip
     state->hstatus->write(0);
+    for (reg_t i = 3; i < N_HPMCOUNTERS + 3; ++i) {
+      const reg_t new_mevent = state->mevent[i - 3]->read() & ~(MHPMEVENT_VUINH | MHPMEVENT_VSINH);
+      state->mevent[i - 3]->write(new_mevent);
+    }
   }
 
   return basic_csr_t::unlogged_write(new_misa);
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1062,7 +1062,7 @@ bool mevent_csr_t::unlogged_write(const reg_t val) noexcept {
   const reg_t mask = proc->extension_enabled(EXT_SSCOFPMF) ? MHPMEVENT_OF | MHPMEVENT_MINH
     | (proc->extension_enabled_const('U') ? MHPMEVENT_UINH : 0)
     | (proc->extension_enabled_const('S') ? MHPMEVENT_SINH : 0)
-    | (proc->extension_enabled_const('H') ? MHPMEVENT_VUINH | MHPMEVENT_VSINH : 0) : 0;
+    | (proc->extension_enabled('H') ? MHPMEVENT_VUINH | MHPMEVENT_VSINH : 0) : 0;
   return basic_csr_t::unlogged_write((read() & ~mask) | (val & mask));
 }
 
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1054,6 +1054,18 @@ void counter_proxy_csr_t::verify_permissions(insn_t insn, bool write) const {
   }
 }
 
+mevent_csr_t::mevent_csr_t(processor_t* const proc, const reg_t addr):
+  basic_csr_t(proc, addr, 0) {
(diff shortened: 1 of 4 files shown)


riscv/csrs.cc
```diff
@@ -1054,6 +1054,18 @@ void counter_proxy_csr_t::verify_permissions(insn_t insn, bool write) const {
...
 }
 
+mevent_csr_t::mevent_csr_t(processor_t* const proc, const reg_t addr):
+  basic_csr_t(proc, addr, 0) {
+}
+
+bool mevent_csr_t::unlogged_write(const reg_t val) noexcept {
+  const reg_t mask = proc->extension_enabled(EXT_SSCOFPMF) ? MHPMEVENT_OF | MHPMEVENT_MINH
+    | (proc->extension_enabled('U') ? MHPMEVENT_UINH : 0)
+    | (proc->extension_enabled('S') ? MHPMEVENT_SINH : 0)
```
scottj97: Because you're not checking for `misa` writes that toggle these bits (as the second commit here does for `H`), there's still an assumption here that these `misa` bits are const. So I would use `extension_enabled_const()` for `U` and `S` here.

riscv/csrs.cc
```diff
@@ -650,6 +650,10 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
     state->mie->write_with_mask(MIP_HS_MASK, 0);  // also takes care of hie, sie
     state->mip->write_with_mask(MIP_HS_MASK, 0);  // also takes care of hip, sip, hvip
     state->hstatus->write(0);
+    for (reg_t i = 3; i <= 31; ++i) {
```
scottj97: There are now two places that assume 32 performance counters. (Here and line 234 of processor.cc.) We should first make that into a `#define`.

riscv/csrs.cc
```diff
@@ -650,6 +650,10 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
...
+      const reg_t new_mevent = state->mevent[i - 3]->read() & ~(MHPMEVENT_VUINH | MHPMEVENT_VSINH);
```
scottj97, line 654: This masking is unnecessary because the `write()` method will do the same masking now that `misa.H==0`.

scottj97 (changes requested):
I wrote some comments here when this PR was new, but forgot to submit my review, so they were never published. Sorry for the delay.

riscv/csrs.cc
YenHaoChen: The misa_csr_t::unlogged_write() updates ```misa.H``` at the end of the function before updating ```mevent```. Thus, masking is necessary.

riscv/csrs.cc
YenHaoChen: Add a macro ```#define N_HPMCOUNTERS 29``` in ```processor.h```.

YenHaoChen:
@scottj97 no worries. Thanks for the comments.

riscv/csrs.cc
```diff
@@ -1054,6 +1058,18 @@ void counter_proxy_csr_t::verify_permissions(insn_t insn, bool write) const {
...
 
+mevent_csr_t::mevent_csr_t(processor_t* const proc, const reg_t addr):
+  basic_csr_t(proc, addr, 0) {
+}
+
+bool mevent_csr_t::unlogged_write(const reg_t val) noexcept {
+  const reg_t mask = proc->extension_enabled(EXT_SSCOFPMF) ? MHPMEVENT_OF | MHPMEVENT_MINH
+    | (proc->extension_enabled_const('U') ? MHPMEVENT_UINH : 0)
+    | (proc->extension_enabled_const('S') ? MHPMEVENT_SINH : 0)
+    | (proc->extension_enabled('H') ? MHPMEVENT_VUINH | MHPMEVENT_VSINH : 0) : 0;
```
scottj97, line 1069: Can you please split bba915f852d01b78daf14062b24c28ec81120e07 to clearly distinguish the bug fix from the refactoring that enables the bug fix? (i.e. the bug fix commit should simply change `extension_enabled_const` to `extension_enabled`)

scottj97 (changes requested):

scottj97 (changes requested):
Commits now look good, but the bug fix should be reported in the commit message for d4e753925ed693f119a9c047257491ac62a00ac3 instead of 41186bc95e65ef6438cc51b71f2c3abf6c224936

YenHaoChen:
I copy the report and keep one in the refactor commit because the purpose of the refactor is for the bug fixing.

scottj97 (approved):

PR #1157, opened 2022-11-30
Allow construction of memory regions that span the whole 64-bit address space

aap-sc:
This PR improves robustness of memory merging routine and allows construction of memory regions that touch the border of 64-bit address space.

### DIFF
--- a/spike_main/spike.cc
+++ b/spike_main/spike.cc
@@ -128,10 +128,7 @@ bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
 
 static bool check_mem_overlap(const mem_cfg_t& L, const mem_cfg_t& R)
 {
-  const auto L_end   = L.get_base() + L.get_size() - 1;
-  const auto R_end   = R.get_base() + R.get_size() - 1;
-
-  return std::max(L.get_base(), R.get_base()) <= std::min(L_end, R_end);
+  return std::max(L.get_base(), R.get_base()) <= std::min(L.get_inclusive_end(), R.get_inclusive_end());
 }
 
 static mem_cfg_t merge_mem_regions(const mem_cfg_t& L, const mem_cfg_t& R)
--- a/spike_main/spike.cc
+++ b/spike_main/spike.cc
@@ -121,15 +121,15 @@ static void read_file_bytes(const char *filename,size_t fileoff,
 bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
 {
   if (a.get_base() == b.get_base())
-    return (a.size < b.size);
+    return (a.get_size() < b.get_size());
   else
     return (a.get_base() < b.get_base());
 }
 
 static bool check_mem_overlap(const mem_cfg_t& L, const mem_cfg_t& R)
 {
-  const reg_t L_end   = L.get_base() + L.size - 1;
-  const reg_t R_end   = R.get_base() + R.size - 1;
+  const auto L_end   = L.get_base() + L.get_size() - 1;
+  const auto R_end   = R.get_base() + R.get_size() - 1;
 
   return std::max(L.get_base(), R.get_base()) <= std::min(L_end, R_end);
 }
@@ -140,7 +140,7 @@ static mem_cfg_t merge_mem_regions(const mem_cfg_t& L, const mem_cfg_t& R)
   assert(check_mem_overlap(L, R));
 
   const reg_t merged_base = std::min(L.get_base(), R.get_base());
-  const reg_t merged_end_incl = std::max(L.get_base() + L.size - 1, R.get_base() + R.size - 1);
+  const reg_t merged_end_incl = std::max(L.get_base() + L.get_size() - 1, R.get_base() + R.get_size() - 1);
   const reg_t merged_size = merged_end_incl - merged_base + 1;
 
   return mem_cfg_t(merged_base, merged_size);
@@ -216,7 +216,7 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
     const unsigned long long max_allowed_pa = (1ull << MAX_PADDR_BITS) - 1ull;
     assert(max_allowed_pa <= std::numeric_limits<reg_t>::max());
     mem_cfg_t mem_region(base, size);
-    auto last_pa_region = mem_region.get_base() + mem_region.size - 1;
+    auto last_pa_region = mem_region.get_base() + mem_region.get_size() - 1;
     if (last_pa_region > max_allowed_pa) {
       int bits_required = 64 - clz(last_pa_region);
       fprintf(stderr, "Unsupported memory region "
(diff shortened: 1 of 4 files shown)


aap-sc:
This commit is continuation of discussion from https://github.com/riscv-software-src/riscv-isa-sim/pull/1100 . This commit fixes a weird corner case in the memory merging routine. It should not  affect  existing configurations, but may help us in future if spike will chose to support more memory configurations.

spike_main/spike.cc
```diff
@@ -246,7 +242,8 @@ static std::vector<std::pair<reg_t, mem_t*>> make_mems(const std::vector<mem_cfg
   std::vector<std::pair<reg_t, mem_t*>> mems;
   mems.reserve(layout.size());
   for (const auto &cfg : layout) {
-    mems.push_back(std::make_pair(cfg.base, new mem_t(cfg.size)));
+    size_t size = cfg.get_inclusive_end() - cfg.get_base() + 1;
```
scottj97: `size()` should be a method on `mem_cfg_t`. I know you want to discourage use of `size` directly but there are legit uses of it, and this smells of Feature Envy.

spike_main/spike.cc
```diff
@@ -117,27 +117,24 @@ static void read_file_bytes(const char *filename,size_t fileoff,
 
 bool sort_mem_region(const mem_cfg_t &a, const mem_cfg_t &b)
 {
-  if (a.base == b.base)
-    return (a.size < b.size);
+  if (a.get_base() == b.get_base())
+    return (a.get_inclusive_end() < b.get_inclusive_end());
```
scottj97: Is this a functional change if one of these memories wraps around the 64-bit address space? Or have we already forbidden such memories before we get to this code?

spike_main/spike.cc
```diff
@@ -128,6 +128,23 @@ static bool check_mem_overlap(const mem_cfg_t& L, const mem_cfg_t& R)
...
+{
+  if (!check_mem_overlap(L, R))
+    return false;
+
+  reg_t start = std::min(L.get_base(), R.get_base());
+  reg_t end = std::max(L.get_inclusive_end(), R.get_inclusive_end());
+
+  // if [base == 0] and [inclusive_end == (2 ^ 64 - 1)] - we can't merge such
+  // regions, since the mem_cfg_t does not support the resulting size
+  return (start == 0ull) && (end == std::numeric_limits<uint64_t>::max());
```
scottj97, line 143: `mem_cfg_t` uses `reg_t`, not `uint64_t`.

In C++, is there a way to get the type of a method at compile time? I.e. instead of `uint64_t` or `reg_t` here, you use the return type of `get_inclusive_end()`.

scottj97 (changes requested):
19a044dcbe8c6252a30a566a4178907741b6a988 does way too much, and is not only a refactoring as the commit message says.

Especially because we don't have a good regression system for testing these kind of changes, any regression is likely to go undetected for months. If and when someone discovers a regression, we need to be able to bisect it down to one tiny commit.

This commit should be several:

1. Change error message to report [start, end] instead of [base, size]
2. Add `get_base()` and use that instead of `base`
3. Add `size()` and use that instead of `size` (see comment below)
4. Add `get_inclusive_end()` and use that in one place since this is not just a simple substitution
5. Use `get_inclusive_end()` in one more new place; repeat with additional commits until all unnecessary uses of `size` removed
6. Functional change to allow `base + size == 0`

Edit: you could also leave direct uses of `size` in place until they are mostly replaced with `get_inclusive_end()`, and then privatize `size`. (Versus replacing those with `size()` and then soon after replacing them again with `get_inclusive_end()`).

aap-sc:
@scottj97 , thanks for the feedback, will try to address.

spike_main/spike.cc

spike_main/spike.cc
aap-sc: > In C++, is there a way to get the type of a method at compile time? 
I've replaced  explicit usage of `reg_t` type with an auto placeholder - this should be enough for `start` / `end` calculations.

I still use uint64_t type explicitly, since this type is exactly what we need if we want to get the maximum value for  **64** bit number.

spike_main/spike.cc
aap-sc: this was not a functional change (since we've already forbidden such regions because of assertions). However I've updated the PR and left this place without significant changes (except for the usage of getters)

aap-sc:
> This commit should be several:
> 1. Change error message to report [start, end] instead of [base, size]
> 2. Add get_base() and use that instead of base
> 3. Add size() and use that instead of size (see comment below)
> 4. Add get_inclusive_end() and use that in one place since this is not just a simple substitution
> 5. Use get_inclusive_end() in one more new place; repeat with additional commits until all unnecessary uses of size removed
> 6. Functional change to allow base + size == 0

@scottj97 , sorry for the long delay... I believe I've addressed all the bullets. Few notes:
* I decided to drop the change that modifies error message. The original motivation was deletion of the `size` method. But now it is back and there is no reason to touch the message - it is good as is.
* I feel that it would be better to combine commits that introduce `get_base()` and `get_size()` methods. These changes are easy to understand since and they are NFC.

spike_main/spike.cc
```diff
@@ -118,28 +118,42 @@ static void read_file_bytes(const char *filename,size_t fileoff,
...
+{
+  if (!check_mem_overlap(L, R))
+    return false;
 
-  return std::max(L.base, R.base) <= std::min(L_end, R_end);
+  auto start = std::min(L.get_base(), R.get_base());
+  auto end = std::max(L.get_inclusive_end(), R.get_inclusive_end());
+
+  // if [base == 0] and [inclusive_end == (2 ^ 64 - 1)] - we can't merge such
+  // regions, since the mem_cfg_t does not support the resulting size
```
scottj97: I think this comment better belongs down on line 181, along with an explanation of the workaround, such as: "So create two smaller memories that total 2^64 bytes."

scottj97 (changes requested):
I think 2d5111910d293bbe82cb2ced09561b2e613ddbd8 is making two different functional changes? (1) allow memory to end at 0xffffffff_ffffffff; (2) allow memory of size 2^64. Those could & should be two separate commits.

spike_main/spike.cc

aap-sc:
> I think [2d51119](https://github.com/riscv-software-src/riscv-isa-sim/commit/2d5111910d293bbe82cb2ced09561b2e613ddbd8) is making two different functional changes? (1) allow memory to end at 0xffffffff_ffffffff; (2) allow memory of size 2^64. Those could & should be two separate commits.

yes, these are effectively two functional changes. addressed.

scottj97 (approved):
My last and final gripe is that the commit message for e335af16175d070cf46bc9e6b05aba1f2a51a9d5 uses the verb "refactor" to describe a functional change/bug fix. A refactoring by definition does not change the behavior of the code, so this is not a refactor.

aswaterman (approved):
Out of appreciation for working with us on this, I'll fix the "refactor" commit title myself and merge this :-)

aswaterman:
GitHub didn't pick that up, unfortunately.  Manually merged via 3b2e8d53b39b8baba8d2e4c31408d92776b32e86

PR #1158, opened 2022-12-01
Minor trigger cleanup

scottj97:
Cleans up a few things in the trigger code. See individual commits for details.

The biggest change is to eliminate dont-care fields in `match_result_t` so that nobody can accidentally use `timing` and `action` when they contain garbage.

cc @YenHaoChen

### DIFF
--- a/riscv/triggers.h
+++ b/riscv/triggers.h
@@ -187,7 +187,7 @@ public:
   unsigned count() const { return triggers.size(); }
 
   std::optional<match_result_t> detect_memory_access_match(operation_t operation, reg_t address, std::optional<reg_t> data) noexcept;
-  match_result_t detect_trap_match(const trap_t& t) noexcept;
+  std::optional<match_result_t> detect_trap_match(const trap_t& t) noexcept;
 
   processor_t *proc;
 private:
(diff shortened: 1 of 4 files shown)


riscv/triggers.h
```diff
@@ -185,8 +185,8 @@ class module_t {
 
   unsigned count() const { return triggers.size(); }
 
-  match_result_t detect_memory_access_match(operation_t operation, reg_t address, std::optional<reg_t> data);
-  match_result_t detect_trap_match(const trap_t& t);
+  match_result_t detect_memory_access_match(operation_t operation, reg_t address, std::optional<reg_t> data) noexcept;
+  match_result_t detect_trap_match(const trap_t& t) noexcept;
```
aswaterman: Longer-term, what should be our discipline using `noexcept`?  Seems like your strategy is to use it in places where it isn't otherwise obvious, as a form of self-documentation, which makes sense to me.

aswaterman (approved):
LGTM, great use of `optional`.

riscv/triggers.h
scottj97: Yes, that is my strategy, and to enforce the caller's expectations as to who will be throwing traps. I try to add it on methods whose name suggests they might involve throwing (but don't, and shouldn't).

For example in `csr_t`, methods `read()` and `write()` are `noexcept` to document and enforce that permission checking must be done elsewhere.

I added `noexcept` here because triggers involve throwing, and I want to make it clear that the throwing should not be done from these match detection methods.

PR #1159, opened 2022-12-02
Use proper include for MCONTROL_ACTION_*

scottj97:
The include of decode.h was added in
08442c1aad70be858f177e277b02a340a22346ba presumably to get MCONTROL_ACTION_* defines, but those are defined in encoding.h, not decode.h.

### DIFF
--- a/riscv/triggers.h
+++ b/riscv/triggers.h
@@ -4,7 +4,7 @@
 #include <vector>
 #include <optional>
 
-#include "decode.h"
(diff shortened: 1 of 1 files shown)


timsifive (approved):
That sounds right.

scottj97:
Turns out, this is incorrect. We need `decode.h` (which includes `encoding.h`) for things like `reg_t`.

PR #1160, opened 2022-12-05
Implement textra (tdata3) of triggers 

YenHaoChen:
This PR implements the CSR textra (tdata3) in Section 5.5.17 and 5.5.18, debug spec.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1153,6 +1153,18 @@ bool tdata2_csr_t::unlogged_write(const reg_t val) noexcept {
   return proc->TM.tdata2_write(state->tselect->read(), val);
 }
 
+tdata3_csr_t::tdata3_csr_t(processor_t* const proc, const reg_t addr):
+  csr_t(proc, addr) {
+}
+
+reg_t tdata3_csr_t::read() const noexcept {
+  return 0;
+}
+
+bool tdata3_csr_t::unlogged_write(const reg_t val) noexcept {
+  return false;
+}
+
 tinfo_csr_t::tinfo_csr_t(processor_t* const proc, const reg_t addr) :
   csr_t(proc, addr) {
 }
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1134,11 +1134,11 @@ tdata1_csr_t::tdata1_csr_t(processor_t* const proc, const reg_t addr):
 }
 
 reg_t tdata1_csr_t::read() const noexcept {
-  return proc->TM.tdata1_read(proc, state->tselect->read());
+  return proc->TM.tdata1_read(state->tselect->read());
 }
 
 bool tdata1_csr_t::unlogged_write(const reg_t val) noexcept {
-  return proc->TM.tdata1_write(proc, state->tselect->read(), val);
+  return proc->TM.tdata1_write(state->tselect->read(), val);
 }
 
 tdata2_csr_t::tdata2_csr_t(processor_t* const proc, const reg_t addr):
@@ -1146,11 +1146,11 @@ tdata2_csr_t::tdata2_csr_t(processor_t* const proc, const reg_t addr):
 }
 
 reg_t tdata2_csr_t::read() const noexcept {
-  return proc->TM.tdata2_read(proc, state->tselect->read());
+  return proc->TM.tdata2_read(state->tselect->read());
 }
 
 bool tdata2_csr_t::unlogged_write(const reg_t val) noexcept {
-  return proc->TM.tdata2_write(proc, state->tselect->read(), val);
+  return proc->TM.tdata2_write(state->tselect->read(), val);
 }
 
 tinfo_csr_t::tinfo_csr_t(processor_t* const proc, const reg_t addr) :
@@ -1158,7 +1158,7 @@ tinfo_csr_t::tinfo_csr_t(processor_t* const proc, const reg_t addr) :
 }
 
 reg_t tinfo_csr_t::read() const noexcept {
-  return proc->TM.tinfo_read(proc, state->tselect->read());
+  return proc->TM.tinfo_read(state->tselect->read());
 }
 
 debug_mode_csr_t::debug_mode_csr_t(processor_t* const proc, const reg_t addr):
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1158,11 +1158,11 @@ tdata3_csr_t::tdata3_csr_t(processor_t* const proc, const reg_t addr):
 }
 
 reg_t tdata3_csr_t::read() const noexcept {
-  return 0;
+  return proc->TM.tdata3_read(state->tselect->read());
 }
 
 bool tdata3_csr_t::unlogged_write(const reg_t val) noexcept {
-  return false;
+  return proc->TM.tdata3_write(state->tselect->read(), val);
 }
 
 tinfo_csr_t::tinfo_csr_t(processor_t* const proc, const reg_t addr) :
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -394,6 +394,9 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   csrmap[CSR_TINFO] = std::make_shared<tinfo_csr_t>(proc, CSR_TINFO);
   unsigned scontext_length = (xlen == 32 ? 16 : 34); // debug spec suggests 16-bit for RV32 and 34-bit for RV64
   csrmap[CSR_SCONTEXT] = scontext = std::make_shared<masked_csr_t>(proc, CSR_SCONTEXT, (reg_t(1) << scontext_length) - 1, 0);
+  unsigned hcontext_length = (xlen == 32 ? 6 : 13) + (proc->extension_enabled('H') ? 1 : 0); // debug spec suggest 7-bit (6-bit) for RV32 and 14-bit (13-bit) for RV64 with (without) H extension
+  csrmap[CSR_HCONTEXT] = std::make_shared<masked_csr_t>(proc, CSR_HCONTEXT, (reg_t(1) << hcontext_length) - 1, 0);
+  csrmap[CSR_MCONTEXT] = mcontext = std::make_shared<proxy_csr_t>(proc, CSR_MCONTEXT, csrmap[CSR_HCONTEXT]);
   debug_mode = false;
   single_step = STEP_NONE;
 
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -390,7 +390,7 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
 
   csrmap[CSR_TDATA1] = std::make_shared<tdata1_csr_t>(proc, CSR_TDATA1);
   csrmap[CSR_TDATA2] = tdata2 = std::make_shared<tdata2_csr_t>(proc, CSR_TDATA2);
-  csrmap[CSR_TDATA3] = std::make_shared<const_csr_t>(proc, CSR_TDATA3, 0);
+  csrmap[CSR_TDATA3] = std::make_shared<tdata3_csr_t>(proc, CSR_TDATA3);
   csrmap[CSR_TINFO] = std::make_shared<tinfo_csr_t>(proc, CSR_TINFO);
   debug_mode = false;
   single_step = STEP_NONE;
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -392,6 +392,8 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   csrmap[CSR_TDATA2] = tdata2 = std::make_shared<tdata2_csr_t>(proc, CSR_TDATA2);
   csrmap[CSR_TDATA3] = std::make_shared<tdata3_csr_t>(proc, CSR_TDATA3);
   csrmap[CSR_TINFO] = std::make_shared<tinfo_csr_t>(proc, CSR_TINFO);
+  unsigned scontext_length = (xlen == 32 ? 16 : 34); // debug spec suggests 16-bit for RV32 and 34-bit for RV64
+  csrmap[CSR_SCONTEXT] = scontext = std::make_shared<masked_csr_t>(proc, CSR_SCONTEXT, (reg_t(1) << scontext_length) - 1, 0);
   debug_mode = false;
   single_step = STEP_NONE;
 
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -178,6 +178,7 @@ struct state_t
   csr_t_p tselect;
   tdata2_csr_t_p tdata2;
   csr_t_p scontext;
+  csr_t_p mcontext;
 
   csr_t_p jvt;
 
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -177,6 +177,7 @@ struct state_t
   dcsr_csr_t_p dcsr;
   csr_t_p tselect;
   tdata2_csr_t_p tdata2;
+  csr_t_p scontext;
 
   csr_t_p jvt;
 
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -8,6 +8,7 @@
 #define VMIDMAX(HSXLEN) (HSXLEN == 32 ? 7 : 14)
 #define HGATP_VMID(HSXLEN) (HSXLEN == 32 ? HGATP32_VMID : HGATP64_VMID)
 
+#define CSR_TEXTRA_MHVALUE_LENGTH(XLEN) (XLEN == 32 ? CSR_TEXTRA32_MHVALUE_LENGTH : CSR_TEXTRA64_MHVALUE_LENGTH)
 #define CSR_TEXTRA_MHVALUE(XLEN)   (XLEN == 32 ? CSR_TEXTRA32_MHVALUE : CSR_TEXTRA64_MHVALUE)
 #define CSR_TEXTRA_MHSELECT(XLEN)  (XLEN == 32 ? CSR_TEXTRA32_MHSELECT : CSR_TEXTRA64_MHSELECT)
 #define CSR_TEXTRA_SBYTEMASK(XLEN) (XLEN == 32 ? CSR_TEXTRA32_SBYTEMASK : CSR_TEXTRA64_SBYTEMASK)
@@ -29,16 +30,15 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
 }
 
 unsigned trigger_t::legalize_mhselect(bool h_enabled) const noexcept {
-  // consider a mask with lowest bit 0
   unsigned convert[8] = {
     (unsigned)                 0 ,  // 0
-    (unsigned)                 0 ,  // 1
+    (unsigned)(h_enabled ? 1 : 0),  // 1
     (unsigned)(h_enabled ? 2 : 0),  // 2
     (unsigned)                 0 ,  // 3
-    (unsigned)                 0 ,  // 4
-    (unsigned)                 0 ,  // 5
-    (unsigned)(h_enabled ? 6 : 0),  // 6
-    (unsigned)                 0    // 7
+    (unsigned)                 4 ,  // 4
+    (unsigned)(h_enabled ? 5 : 4),  // 5
+    (unsigned)(h_enabled ? 6 : 4),  // 6
+    (unsigned)                 4    // 7
   };
   assert(mhselect < 8);
   return convert[mhselect];
@@ -102,7 +102,11 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
   }
 
   mhselect_mode_t mode = mhselect_mode(proc->extension_enabled('H'));
-  if (mode == MHSELECT_MODE_VMID) { // 2 and 6 are vmid
+  if (mode == MHSELECT_MODE_MCONTEXT) { // 4, 1, and 5 are mcontext
+    reg_t mask = (1 << (CSR_TEXTRA_MHVALUE_LENGTH(xlen) + 1)) - 1;
+    if ((state->mcontext->read() & mask) != mhselect_compare(proc->extension_enabled('H')))
+      return false;
+  } else if (mode == MHSELECT_MODE_VMID) { // 2 and 6 are vmid
     const reg_t vmid = get_field(state->hgatp->read(), HGATP_VMID(hsxlen));
     if (vmid != (mhselect_compare(proc->extension_enabled('H')) & ((1 << VMIDMAX(hsxlen)) - 1)))
       return false;
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -34,21 +34,6 @@ unsigned trigger_t::legalize_mhselect(bool h_enabled) const noexcept {
   return interp.mhselect;
 }
 
-mhselect_mode_t trigger_t::mhselect_mode(bool h_enabled) const noexcept {
-  switch (legalize_mhselect(h_enabled)) {
-    case 0: // ignore mhvalue
-      return MHSELECT_MODE_IGNORE;
-    case 4: // match only if mcontext equal mhvalue
-    case 1: // match only if mcontext equal {mhvalue, 1b'0}
-    case 5: // match only if mcontext equal {mhvalue, 1b'1}
-      return MHSELECT_MODE_MCONTEXT;
-    case 2: // match only if hgapt.VMID equal {mhvalue, 1b'0}
-    case 6: // match only if hgapt.VMID equal {mhvalue, 1b'1}
-      return MHSELECT_MODE_VMID;
-    default: assert(false);
-  }
-}
-
 reg_t trigger_t::tdata3_read(const processor_t * const proc) const noexcept {
   auto xlen = proc->get_xlen();
   reg_t tdata3 = 0;
@@ -91,7 +76,8 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
       return false;
   }
 
-  mhselect_mode_t mode = mhselect_mode(proc->extension_enabled('H'));
+  const auto mhselect_interp = interpret_mhselect(proc->extension_enabled('H'));
+  const mhselect_mode_t mode = mhselect_interp.mode;
   if (mode == MHSELECT_MODE_MCONTEXT) { // 4, 1, and 5 are mcontext
     reg_t mask = (1 << (CSR_TEXTRA_MHVALUE_LENGTH(xlen) + 1)) - 1;
     if ((state->mcontext->read() & mask) != mhselect_compare(proc->extension_enabled('H')))
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -270,12 +270,12 @@ module_t::~module_t() {
   }
 }
 
-reg_t module_t::tdata1_read(const processor_t * const proc, unsigned index) const noexcept
+reg_t module_t::tdata1_read(unsigned index) const noexcept
 {
   return triggers[index]->tdata1_read(proc);
 }
 
-bool module_t::tdata1_write(processor_t * const proc, unsigned index, const reg_t val) noexcept
+bool module_t::tdata1_write(unsigned index, const reg_t val) noexcept
 {
   if (triggers[index]->get_dmode() && !proc->get_state()->debug_mode) {
     return false;
@@ -316,12 +316,12 @@ bool module_t::tdata1_write(processor_t * const proc, unsigned index, const reg_
   return true;
 }
 
-reg_t module_t::tdata2_read(const processor_t * const proc, unsigned index) const noexcept
+reg_t module_t::tdata2_read(unsigned index) const noexcept
 {
   return triggers[index]->tdata2_read(proc);
 }
 
-bool module_t::tdata2_write(processor_t * const proc, unsigned index, const reg_t val) noexcept
+bool module_t::tdata2_write(unsigned index, const reg_t val) noexcept
 {
   if (triggers[index]->get_dmode() && !proc->get_state()->debug_mode) {
     return false;
@@ -374,7 +374,7 @@ std::optional<match_result_t> module_t::detect_trap_match(const trap_t& t) noexc
   return std::nullopt;
 }
 
-reg_t module_t::tinfo_read(UNUSED const processor_t * const proc, unsigned UNUSED index) const noexcept
+reg_t module_t::tinfo_read(unsigned UNUSED index) const noexcept
 {
   /* In spike, every trigger supports the same types. */
   return (1 << CSR_TDATA1_TYPE_MCONTROL) |
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -76,7 +76,7 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
   mhselect = get_field(val, CSR_TEXTRA_MHSELECT(xlen));
   sbytemask = get_field(val, CSR_TEXTRA_SBYTEMASK(xlen));
   svalue = proc->extension_enabled_const('S') ? get_field(val, CSR_TEXTRA_SVALUE(xlen)) : 0;
-  sselect = (sselect_t)((proc->extension_enabled_const('S') && get_field(val, CSR_TEXTRA_SSELECT(xlen)) == SSELECT_ASID) ? SSELECT_ASID : SSELECT_IGNORE);
+  sselect = (sselect_t)((proc->extension_enabled_const('S') && get_field(val, CSR_TEXTRA_SSELECT(xlen)) <= SSELECT_MAXVAL) ? get_field(val, CSR_TEXTRA_SSELECT(xlen)) : SSELECT_IGNORE);
 }
 
 bool trigger_t::textra_match(processor_t * const proc) const noexcept
@@ -86,7 +86,15 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
   state_t * const state = proc->get_state();
 
   assert(sselect <= SSELECT_MAXVAL);
-  if (sselect == SSELECT_ASID) {
+  if (sselect == SSELECT_SCONTEXT) {
+    reg_t mask = (reg_t(1) << ((xlen == 32) ? CSR_TEXTRA32_SVALUE_LENGTH : CSR_TEXTRA64_SVALUE_LENGTH)) - 1;
+    assert(CSR_TEXTRA32_SBYTEMASK_LENGTH < CSR_TEXTRA64_SBYTEMASK_LENGTH);
+    for (int i = 0; i < CSR_TEXTRA64_SBYTEMASK_LENGTH; i++)
+      if (sbytemask & (1 << i))
+        mask &= 0xff << (i * 8);
+    if ((state->scontext->read() & mask) != (svalue & mask))
+      return false;
+  } else if (sselect == SSELECT_ASID) {
     const reg_t satp = state->satp->read();
     const reg_t asid = get_field(satp,  SATP_ASID(xlen));
     if (asid != (svalue & ((1 << ASIDMAX(xlen)) - 1)))
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -3,6 +3,11 @@
 #include "processor.h"
 #include "triggers.h"
 
+#define ASIDMAX(SXLEN) (SXLEN == 32 ? 9 : 16)
+#define SATP_ASID(SXLEN) (SXLEN == 32 ? SATP32_ASID : SATP64_ASID)
+#define VMIDMAX(HSXLEN) (HSXLEN == 32 ? 7 : 14)
+#define HGATP_VMID(HSXLEN) (HSXLEN == 32 ? HGATP32_VMID : HGATP64_VMID)
+
 #define CSR_TEXTRA_MHVALUE(XLEN)   (XLEN == 32 ? CSR_TEXTRA32_MHVALUE : CSR_TEXTRA64_MHVALUE)
 #define CSR_TEXTRA_MHSELECT(XLEN)  (XLEN == 32 ? CSR_TEXTRA32_MHSELECT : CSR_TEXTRA64_MHSELECT)
 #define CSR_TEXTRA_SBYTEMASK(XLEN) (XLEN == 32 ? CSR_TEXTRA32_SBYTEMASK : CSR_TEXTRA64_SBYTEMASK)
@@ -39,6 +44,21 @@ unsigned trigger_t::legalize_mhselect(bool h_enabled) const noexcept {
   return convert[mhselect];
 }
 
+mhselect_mode_t trigger_t::mhselect_mode(bool h_enabled) const noexcept {
+  switch (legalize_mhselect(h_enabled)) {
+    case 0: // ignore mhvalue
+      return MHSELECT_MODE_IGNORE;
+    case 4: // match only if mcontext equal mhvalue
+    case 1: // match only if mcontext equal {mhvalue, 1b'0}
+    case 5: // match only if mcontext equal {mhvalue, 1b'1}
+      return MHSELECT_MODE_MCONTEXT;
+    case 2: // match only if hgapt.VMID equal {mhvalue, 1b'0}
+    case 6: // match only if hgapt.VMID equal {mhvalue, 1b'1}
+      return MHSELECT_MODE_VMID;
+    default: assert(false);
+  }
+}
+
 reg_t trigger_t::tdata3_read(const processor_t * const proc) const noexcept {
   auto xlen = proc->get_xlen();
   reg_t tdata3 = 0;
@@ -57,6 +77,30 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
   sbytemask = get_field(val, CSR_TEXTRA_SBYTEMASK(xlen));
   svalue = proc->extension_enabled_const('S') ? get_field(val, CSR_TEXTRA_SVALUE(xlen)) : 0;
   sselect = (sselect_t)((proc->extension_enabled_const('S') && get_field(val, CSR_TEXTRA_SSELECT(xlen)) == SSELECT_ASID) ? SSELECT_ASID : SSELECT_IGNORE);
+}
+
+bool trigger_t::textra_match(processor_t * const proc) const noexcept
+{
+  auto xlen = proc->get_xlen();
+  auto hsxlen = proc->get_xlen(); // use xlen since no hsxlen
+  state_t * const state = proc->get_state();
+
+  assert(sselect <= SSELECT_MAXVAL);
+  if (sselect == SSELECT_ASID) {
+    const reg_t satp = state->satp->read();
+    const reg_t asid = get_field(satp,  SATP_ASID(xlen));
+    if (asid != (svalue & ((1 << ASIDMAX(xlen)) - 1)))
+      return false;
+  }
+
+  mhselect_mode_t mode = mhselect_mode(proc->extension_enabled('H'));
+  if (mode == MHSELECT_MODE_VMID) { // 2 and 6 are vmid
+    const reg_t vmid = get_field(state->hgatp->read(), HGATP_VMID(hsxlen));
+    if (vmid != (mhselect_compare(proc->extension_enabled('H')) & ((1 << VMIDMAX(hsxlen)) - 1)))
+      return false;
+  }
+
+  return true;
 }
 
 reg_t disabled_trigger_t::tdata1_read(const processor_t * const proc) const noexcept
@@ -410,7 +454,7 @@ std::optional<match_result_t> module_t::detect_memory_access_match(operation_t o
      * entire chain did not match. This is allowed by the spec, because the final
      * trigger in the chain will never get `hit` set unless the entire chain
      * matches. */
-    auto result = trigger->detect_memory_access_match(proc, operation, address, data);
+    auto result = trigger->textra_match(proc) ? trigger->detect_memory_access_match(proc, operation, address, data) : std::nullopt;
     if (result.has_value() && !trigger->get_chain())
       return result;
 
@@ -426,7 +470,7 @@ std::optional<match_result_t> module_t::detect_trap_match(const trap_t& t) noexc
     return std::nullopt;
 
   for (auto trigger: triggers) {
-    auto result = trigger->detect_trap_match(proc, t);
+    auto result = trigger->textra_match(proc) ? trigger->detect_trap_match(proc, t) : std::nullopt;
     if (result.has_value())
       return result;
   }
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -30,18 +30,8 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
 }
 
 unsigned trigger_t::legalize_mhselect(bool h_enabled) const noexcept {
-  unsigned convert[8] = {
-    (unsigned)                 0 ,  // 0
-    (unsigned)(h_enabled ? 1 : 0),  // 1
-    (unsigned)(h_enabled ? 2 : 0),  // 2
-    (unsigned)                 0 ,  // 3
-    (unsigned)                 4 ,  // 4
-    (unsigned)(h_enabled ? 5 : 4),  // 5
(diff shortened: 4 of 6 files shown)


riscv/triggers.cc
```diff
@@ -17,6 +17,94 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+    tdata3 = set_field(tdata3, CSR_TEXTRA32_SBYTEMASK, sbytemask);
+    tdata3 = set_field(tdata3, CSR_TEXTRA32_SVALUE, svalue);
+    tdata3 = set_field(tdata3, CSR_TEXTRA32_SSELECT, sselect);
+  } else {
+    tdata3 = set_field(tdata3, CSR_TEXTRA64_MHVALUE, mhvalue);
+    tdata3 = set_field(tdata3, CSR_TEXTRA64_MHSELECT, mhselect);
+    tdata3 = set_field(tdata3, CSR_TEXTRA64_SBYTEMASK, sbytemask);
+    tdata3 = set_field(tdata3, CSR_TEXTRA64_SVALUE, svalue);
+    tdata3 = set_field(tdata3, CSR_TEXTRA64_SSELECT, sselect);
+  }
```
scottj97: This is unfortunate. Can we do macros like we do for `MCONTROL_DMODE(xlen)` e.g. `CSR_TEXTRA_MHVALUE(xlen)`?

riscv/triggers.cc
```diff
@@ -17,6 +17,94 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+    sselect = get_field(val, CSR_TEXTRA32_SSELECT);
+  } else {
+    mhvalue = get_field(val, CSR_TEXTRA64_MHVALUE);
+    mhselect = get_field(val, CSR_TEXTRA64_MHSELECT);
+    sbytemask = get_field(val, CSR_TEXTRA64_SBYTEMASK);
+    svalue = get_field(val, CSR_TEXTRA64_SVALUE);
+    sselect = get_field(val, CSR_TEXTRA64_SSELECT);
+  }
+  if (!proc->extension_enabled('H'))
+    mhselect = mhselect & MHSELECT_MCONTEXT; // the only legal value is 4, otherwise 0 (ignore)
```
scottj97: Wouldn't you need some special behavior in `misa_csr_t` to deal with this when `misa.H` is toggled from 1 to 0?

riscv/triggers.cc
```diff
@@ -17,6 +17,94 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+  if ((mhselect & MHSELECT_RESERVED) == MHSELECT_RESERVED) // 3 and 7 are reserved for mhselect
+    mhselect = 0;
+  if (!proc->extension_enabled('S'))
+    svalue = 0;
+  if (!proc->extension_enabled('S') || sselect > SSELECT_MAXVAL)
+    sselect = SSELECT_IGNORE;
```
scottj97: These should both be `extension_enabled_const` since `misa_csr_t` does not modify this field when `misa.S` is changed.

riscv/triggers.cc
```diff
@@ -17,6 +17,94 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+  state_t * const state = proc->get_state();
+
+  assert(sselect <= SSELECT_MAXVAL);
+  if (sselect == SSELECT_SCONTEXT) {
+    reg_t mask = (reg_t(1) << ((xlen == 32) ? CSR_TEXTRA32_SVALUE_LENGTH : CSR_TEXTRA64_SVALUE_LENGTH)) - 1;
+    assert(CSR_TEXTRA32_SBYTEMASK_LENGTH < CSR_TEXTRA64_SBYTEMASK_LENGTH);
+    for (int i = 0; i < CSR_TEXTRA64_SBYTEMASK_LENGTH; i++)
+      if (sbytemask & (1 << i))
+        mask &= 0xff << (i * 8);
+    if ((state->csrmap[CSR_SCONTEXT]->read() ^ svalue) & mask)
```
scottj97: I can't find any implementation of `scontext` in Spike. How does this work?

riscv/triggers.cc
```diff
@@ -17,6 +17,94 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+      return false;
+  } else if (sselect == SSELECT_ASID) {
+    reg_t satp = state->v ? state->vsatp->read() : state->satp->read();
```
scottj97: Unnecessary ternary. `satp` derives from type `virtualized_csr_t` so calling its `read()` will read `vsatp` or `satp` as appropriate.

riscv/triggers.cc
```diff
@@ -17,6 +17,94 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+    reg_t satp_asid = xlen == 32 ? SATP32_ASID : SATP64_ASID;
+    reg_t asid = get_field(satp,  satp_asid);
+    reg_t asidmax = xlen == 32 ? 9 : 16;
```
scottj97: The ASID width can vary by implementation. Does this assume the implementation has maximum width?

Also it would be better to get these values from the existing #defines somehow, instead of magic numbers.

riscv/triggers.cc
```diff
@@ -17,6 +17,94 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+  if (mhselect == MHSELECT_MCONTEXT) {
+    if ((state->csrmap[CSR_MCONTEXT]->read() ^ mhvalue) & ((1 << mhvalue_length) - 1))
+      return false;
+  } else if (mhselect & MHSELECT_MCONTEXT_SELECT) { // 1 and 5 are mcontext_select
+    if ((state->csrmap[CSR_MCONTEXT]->read() ^ ((mhvalue << 1) + (mhselect >> 2))) & ((1 << (mhvalue_length + 1)) - 1))
+      return false;
+  } else if (mhselect & MHSELECT_VMID_SELECT) { // 2 and 6 are vmid_select
+    reg_t hgatp_vmid = xlen == 32 ? HGATP32_VMID : HGATP64_VMID;
+    reg_t vmid = get_field(state->hgatp->read(), hgatp_vmid);
+    reg_t vmidmax = hsxlen == 32 ? 7 : 14;
```
scottj97: Same comments about ASID above also apply here to VMID

riscv/triggers.cc
```diff
@@ -17,6 +17,94 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+    mhvalue = get_field(val, CSR_TEXTRA64_MHVALUE);
+    mhselect = get_field(val, CSR_TEXTRA64_MHSELECT);
+    sbytemask = get_field(val, CSR_TEXTRA64_SBYTEMASK);
+    svalue = get_field(val, CSR_TEXTRA64_SVALUE);
+    sselect = get_field(val, CSR_TEXTRA64_SSELECT);
+  }
+  if (!proc->extension_enabled('H'))
+    mhselect = mhselect & MHSELECT_MCONTEXT; // the only legal value is 4, otherwise 0 (ignore)
+  if ((mhselect & MHSELECT_RESERVED) == MHSELECT_RESERVED) // 3 and 7 are reserved for mhselect
+    mhselect = 0;
```
scottj97: The WARLization logic here is pretty hard to understand and there are some embedded assumptions about the values of these enums (or else the bitwise logic wouldn't work). Since it's only a 3-bit field, can you express it as a table mapping the 8 possible inputs to 8 outputs?

scottj97 (changes requested):

riscv/triggers.cc
YenHaoChen: The spec uses the term ```ASIDMAX``` instead of ```ASIDLEN```. I believe that means it is not an implementation-defined value.
image
I support your point that there should be a macro ```#define ASIDMAX(XLEN) (XLEN==32?9:16)```. However, I need help deciding the file defining the macro.

riscv/triggers.cc
YenHaoChen: Add additonal macros to ```triggers.h```.
@timsifive Should those macros be somewhere else? ```debug_defines.h```? or?

riscv/triggers.cc
YenHaoChen: Add a legalization function ```legalize_mhselect()```.

riscv/triggers.cc
scottj97: That kind of thing usually goes in `encoding.h`, which is created from [another repo](https://github.com/riscv/riscv-opcodes).

My question, though, is about implementations where ASIDLEN < ASIDMAX. Will this code still work as-is?

scottj97 (changes requested):
I still need to review the code today, but I've noticed that as of af09ac156c88282e47ba41a80ad51baaaeb6d93a it appears that `tdata3` can never take any value other than 0? In the previous patch series (0b20d1c8b771f3e3581a11dcad2624adf6784ec7) it looked like it could be written with any value.

riscv/triggers.h
```diff
@@ -71,7 +95,15 @@ class trigger_t {
...
 protected:
   action_t legalize_action(reg_t val) const noexcept;
+  unsigned legalize_mhselect(reg_t val, bool h_enabled) const noexcept;
   reg_t tdata2;
+
+private:
+  unsigned sselect;
+  unsigned svalue;
+  unsigned sbytemask;
+  unsigned mhselect;
```
scottj97, line 126: Why isn't this (and method `legalize_mhselect()` of type `mhselect_t`?

riscv/triggers.h
```diff
@@ -71,7 +95,15 @@ class trigger_t {
 
 protected:
   action_t legalize_action(reg_t val) const noexcept;
+  unsigned legalize_mhselect(reg_t val, bool h_enabled) const noexcept;
   reg_t tdata2;
+
+private:
+  unsigned sselect;
```
scottj97: Why isn't this type `sselect_t`?

riscv/triggers.h
```diff
@@ -29,6 +36,21 @@ typedef enum {
...
+  SSELECT_ASID = 2,
+  SSELECT_MAXVAL = 2
+} sselect_t;
+
+typedef enum {
+  MHSELECT_IGNORE = 0,
+  MHSELECT_MCONTEXT = 4,
+  MHSELECT_MCONTEXT_SELECT = 1 /* and 5 */,
+  MHSELECT_VMID_SELECT = 2 /* and 6 */,
+  MHSELECT_RESERVED = 3 /* and 7 */
```
scottj97: This is pretty ugly. I guess this is why you can't use type `mhselect_t` for the `mhselect` attribute. Is there any reason not to have the enum cover all possible cases? There are assumptions within `legalize_mhselect()` about the values of these enums, making it messy.

riscv/processor.cc
```diff
@@ -392,6 +392,9 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   csrmap[CSR_TDATA2] = tdata2 = std::make_shared<tdata2_csr_t>(proc, CSR_TDATA2);
   csrmap[CSR_TDATA3] = std::make_shared<tdata3_csr_t>(proc, CSR_TDATA3);
   csrmap[CSR_TINFO] = std::make_shared<tinfo_csr_t>(proc, CSR_TINFO);
+  csrmap[CSR_SCONTEXT] = std::make_shared<const_csr_t>(proc, CSR_SCONTEXT, 0);
+  csrmap[CSR_HCONTEXT] = std::make_shared<const_csr_t>(proc, CSR_HCONTEXT, 0);
+  csrmap[CSR_MCONTEXT] = std::make_shared<proxy_csr_t>(proc, CSR_MCONTEXT, csrmap[CSR_HCONTEXT]);
```
scottj97: The `csrmap` is intended for `csrrw` instructions that have only a CSR number and not a name. For internal code that accesses CSRs, prefer keeping a handle to it and referring to it by that handle.

The handle can be either local to this `reset()` method, such as `minstreth`, or if wider access is needed, in `state_t`, such as `mideleg`. It can be generic type `csr_t_p`, or if access to special methods are needed, a child class pointer.

riscv/triggers.cc
```diff
@@ -17,6 +17,51 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+  mhvalue = get_field(val, CSR_TEXTRA_MHVALUE(xlen));
+  mhselect = legalize_mhselect(get_field(val, CSR_TEXTRA_MHSELECT(xlen)), proc->extension_enabled('H'));
+  sbytemask = get_field(val, CSR_TEXTRA_SBYTEMASK(xlen));
+  svalue = get_field(val, CSR_TEXTRA_SVALUE(xlen));
+  sselect = get_field(val, CSR_TEXTRA_SSELECT(xlen));
+  if (!proc->extension_enabled_const('S'))
+    svalue = 0;
+  if (!proc->extension_enabled_const('S') || sselect > SSELECT_MAXVAL)
+    sselect = SSELECT_IGNORE;
+}
```
scottj97: I would prefer to set these two in one expression each, instead of setting them and then modifying them in a later line of code, since a future maintainer could easily miss that later modification.

riscv/triggers.cc
```diff
@@ -62,6 +65,46 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
...
+    reg_t satp = state->satp->read();
+    reg_t satp_asid = xlen == 32 ? SATP32_ASID : SATP64_ASID;
+    reg_t asid = get_field(satp,  satp_asid);
+    if ((asid ^ svalue) & ((1 << ASIDMAX(xlen)) - 1))
+      return false;
+  }
+
+  reg_t mhvalue_length = xlen == 32 ? CSR_TEXTRA32_MHVALUE_LENGTH : CSR_TEXTRA64_MHVALUE_LENGTH;
+  if (mhselect == MHSELECT_MCONTEXT) {
+    if ((state->csrmap[CSR_MCONTEXT]->read() ^ mhvalue) & ((1 << mhvalue_length) - 1))
```
scottj97: With `scontext` and `mcontext` both being hardwired to 0, will this code ever do anything useful?

riscv/triggers.cc
```diff
@@ -62,6 +65,46 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
...
+      return false;
+  } else if (mhselect & MHSELECT_MCONTEXT_SELECT) { // 1 and 5 are mcontext_select
+    if ((state->csrmap[CSR_MCONTEXT]->read() ^ ((mhvalue << 1) + (mhselect >> 2))) & ((1 << (mhvalue_length + 1)) - 1))
```
scottj97: Put `scontext` and `mcontext` into `state_t` so you don't need to use `csrmap` in this method.

riscv/csrs.cc
```diff
@@ -654,6 +654,13 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
       const reg_t new_mevent = state->mevent[i - 3]->read() & ~(MHPMEVENT_VUINH | MHPMEVENT_VSINH);
       state->mevent[i - 3]->write(new_mevent);
     }
+    for (unsigned i = 0; i < proc->TM.count(); i++) {
```
scottj97: Why not `for (trig& : proc->TM.triggers)` and then call `trig->tdata3_read()` directly?

riscv/triggers.cc
```diff
@@ -62,6 +65,46 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
...
+      return false;
+  } else if (mhselect & MHSELECT_VMID_SELECT) { // 2 and 6 are vmid_select
```
scottj97: I haven't studied the spec for `tdata3` but this is an awfully messy way to handle the `mhselect` field. There are subtle assumptions embedded in this method about the encoded values in `mhselect`. Maybe you could store it as multiple variables?

riscv/triggers.cc
```diff
@@ -17,6 +17,51 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+      return val;
+    case 3: return MHSELECT_IGNORE;
+    case 5:
+    case 6:
+      if (!h_enabled)
+        return MHSELECT_MCONTEXT;
+      return val;
+    case 7: return MHSELECT_IGNORE;
+    default: assert(false);
+  }
```
scottj97: This is still messy and still embeds assumptions about the values of the enums. I was thinking something more like:
```
  unsigned convert[8] = {
    MHSELECT_IGNORE,  // 0                                                                                                                          
    h_enabled ? MHSELECT_MCONTEXT_SELECT : MHSELECT_IGNORE,  // 1                                                                                   
    h_enabled ? MHSELECT_VMID_SELECT : MHSELECT_IGNORE,  // 2                                                                                       
    MHSELECT_IGNORE,  // 3                                                                                                                          
    MHSELECT_MCONTEXT,  // 4                                                                                                                        
    h_enabled ? 5 : MHSELECT_MCONTEXT,  // 5                                                                                                        
    h_enabled ? 6 : MHSELECT_MCONTEXT,  // 6                                                                                                        
    MHSELECT_IGNORE  // 7                                                                                                                           
  };
  assert(val < 8);
  return convert[val];
```

But I still don't like having 5 & 6 in there instead of always returning an enum value.

scottj97 (changes requested):

scottj97:
I've read the spec on `mhvalue` and `mhselect`. Crazy, but it's needed for backward compatibility with the 0.13.2 debug spec.

I think the best implementation would be to store `mhvalue_raw` and `mhselect_raw` and use those for `tdata3_read()` only. Inside `tdata3_write()`, those two fields are reinterpreted and stored as two values: `mhvalue_compare` which is either `mhvalue` or `{mhvalue, mhselect[2]}` depending on `mhselect`, and `mhselect_mode` which is one of `MHSELECT_IGNORE`, `MHSELECT_MCONTEXT`, or `MHSELECT_VMID`. (Note that mcontext and mcontext_select are the same mode but different `mhvalue_compare` values.)

That should simplify a lot of the interpretation of all the `mhselect` values.

This will complicate disabling `misa.H`, which is already getting unwieldy.

scottj97:
> Inside `tdata3_write()`, those two fields are reinterpreted and stored as two values: `mhvalue_compare` which is either `mhvalue` or `{mhvalue, mhselect[2]}` depending on `mhselect`, and `mhselect_mode` which is one of `MHSELECT_IGNORE`, `MHSELECT_MCONTEXT`, or `MHSELECT_VMID`. (Note that mcontext and mcontext_select are the same mode but different `mhvalue_compare` values.)

Or, instead of storing these reinterpreted values, methods could do the interpretation of the raw fields and return the mode and comparison values.

riscv/csrs.cc
YenHaoChen: The ```proc->TM.triggers``` is private in ```triggers::module_t```. Should we make it public?

scottj97: Ugh, this just gets messier.

I am thinking of making a PR to change the behavior of toggling `misa.H` from 1 to 0. Instead of masking on writes based on the current `misa.H` and writing all the relevant CSRs when it's disabled, each CSR would instead mask on reads based on the current `misa.H`. Then there's no need to update anything here in `misa_csr_t`.

This would be a functional change (in the case where `misa.H` goes from 1 to 0 then back to 1), but the changed behavior is explicitly undefined in the spec, so I don't think it would hurt, and it would simplify all this mess.

You could take the same approach here with the triggers: mask the bits on read instead of writing them here in `misa_csr_t`.

riscv/triggers.cc
YenHaoChen: Is the following better?
```
  mask = h_enabled ? (MHSELECT_MCONTEXT | MHSELECT_MCONTEXT_SELECT | MHSELECT_VMID_SELECT) : (MHSELECT_MCONTEXT);
  val = val & mask;
  return val == MHSELECT_RESERVED ? MHSELECT_IGNORE : val;
```

YenHaoChen: Is the following better?
```
  mask = h_enabled ? (MHSELECT_MCONTEXT | MHSELECT_MCONTEXT_SELECT | MHSELECT_VMID_SELECT) : (MHSELECT_MCONTEXT);
  val = val & mask;
  return val == MHSELECT_RESERVED ? MHSELECT_IGNORE : val;
```

riscv/triggers.cc
YenHaoChen: The debug spec does not consider the ASIDLEN value here. For ASIDLEN < ASIDMAX, the higher bits are 0. The trigger compares the value as-is and can only fire on setting the corresponding bits 0.
image
Does this behavior make sense to you?

riscv/triggers.h
YenHaoChen: Change the ```mhselect_t``` to ```mhselect_mode_t``` and add a function maps the variable ```mhselect``` to ```mhselect_mode_t```.

riscv/triggers.h
YenHaoChen: Change the ```mhselect_t``` to ```mhselect_mode_t```, which does not directly represent the ```mhselect```.

riscv/triggers.cc
YenHaoChen: I choose to add two additonal functions: mhselect_mode() and mhselect_compare().

YenHaoChen:
1. Use ```masked_csr_t``` for scontext and hcontext instead of ```const_csr_t```.
2. Breakdown commit for texta_match(). Incrementally include checking of scontext and mcontext. Because the CSRs are optional, each commit is still a valid implementation.
3. Add xlen-related macro definitions on the top of ```triggers.cc``` and ```triggers.h```. The macros in ```encoding.h``` does not have a parameter. Should we make a PR to https://github.com/riscv/riscv-opcodes?
4. Store the raw mhselect value and add mhselect_mode() and mhselect_compare() functions for operation.
5. Does not change the underline mhselect value when toggling misa.H. Always use legalize_mhselect() function to obtain the effective mhselect value.

riscv/triggers.cc
scottj97: Yes, I think this is good.

riscv/processor.cc
```diff
@@ -390,8 +390,13 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
 
   csrmap[CSR_TDATA1] = std::make_shared<tdata1_csr_t>(proc, CSR_TDATA1);
   csrmap[CSR_TDATA2] = tdata2 = std::make_shared<tdata2_csr_t>(proc, CSR_TDATA2);
-  csrmap[CSR_TDATA3] = std::make_shared<const_csr_t>(proc, CSR_TDATA3, 0);
+  csrmap[CSR_TDATA3] = std::make_shared<tdata3_csr_t>(proc, CSR_TDATA3);
   csrmap[CSR_TINFO] = std::make_shared<tinfo_csr_t>(proc, CSR_TINFO);
+  unsigned scontext_length = (xlen == 32 ? 16 : 34); // debug spec suggests 16-bit for RV32 and 34-bit for RV64
+  csrmap[CSR_SCONTEXT] = scontext = std::make_shared<masked_csr_t>(proc, CSR_SCONTEXT, (reg_t(1) << scontext_length) - 1, 0);
+  unsigned hcontext_length = (xlen == 32 ? 6 : 13) + (proc->extension_enabled_const('H') ? 1 : 0); // debug spec suggest 7-bit (6-bit) for RV32 and 14-bit (13-bit) for RV64 with (without) H extension
```
scottj97: Every test is now failing with an assertion because `misa.H` is not const.

scottj97 (changes requested):

scottj97:
As of 55f375413942cc01182a2667d8d56d4e7b3f614f I still do not see `tdata3` being writable to any value other than 0.

riscv/triggers.cc
```diff
@@ -3,6 +3,12 @@
 #include "processor.h"
 #include "triggers.h"
 
+#define ASIDMAX(XLEN) (XLEN == 32 ? 9 : 16)
+#define VMIDMAX(HSXLEN) (HSXLEN == 32 ? 7 : 14)
+
+#define SATP_ASID(XLEN) (XLEN == 32 ? SATP32_ASID : SATP64_ASID)
+#define HGATP_VMID(XLEN) (XLEN == 32 ? HGATP32_VMID : HGATP64_VMID)
```
scottj97: `hgatp` is also `HSXLEN` width

riscv/triggers.cc
```diff
@@ -52,6 +58,30 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
...
+    reg_t satp = state->satp->read();
+    reg_t asid = get_field(satp,  SATP_ASID(xlen));
+    if ((asid ^ svalue) & ((1 << ASIDMAX(xlen)) - 1))
+      return false;
+  }
+
+  mhselect_mode_t mode = mhselect_mode(mhselect, proc->extension_enabled('H'));
+  if (mode == MHSELECT_VMID_SELECT) { // 2 and 6 are vmid_select
+    reg_t vmid = get_field(state->hgatp->read(), HGATP_VMID(xlen));
+    if ((vmid ^ mhselect_compare()) & ((1 << VMIDMAX(hsxlen)) - 1))
```
scottj97: Using xor obfuscates what's happening here. The spec says:

> This trigger will only match if VMID in hgatp equals the lower VMIDMAX (defined in the Privileged Spec) bits of {mhvalue, mhselect[2]}.

The code should read as closely to the prose as possible:
```
if (vmid == mhselect_compare() & ((1 << VMIDMAX(hsxlen)) - 1)) ...
```

riscv/triggers.h
```diff
@@ -96,9 +96,15 @@ class trigger_t {
 protected:
   action_t legalize_action(reg_t val) const noexcept;
   unsigned legalize_mhselect(reg_t val, bool h_enabled) const noexcept;
+  bool textra_match(processor_t * const proc) const noexcept;
   reg_t tdata2;
 
 private:
+  mhselect_mode_t mhselect_mode(reg_t val, bool h_enabled) const noexcept {
+    val = legalize_mhselect(val, h_enabled);
```
scottj97: Prefer to create a new [const] variable instead of mutating a parameter.

riscv/triggers.h
```diff
@@ -96,9 +96,15 @@ class trigger_t {
...
+    return (mhselect_mode_t)(val == MHSELECT_MCONTEXT ?  (MHSELECT_MCONTEXT) : (val & (MHSELECT_IGNORE | MHSELECT_MCONTEXT_SELECT | MHSELECT_VMID_SELECT | MHSELECT_RESERVED)));
+  }
+  unsigned mhselect_compare() const noexcept { return (mhvalue << 1) + (mhselect >> 2); } // {mhvalue, mhselect[2]}
```
scottj97: This is not what I was suggesting, and still relies on the specific integer values of the enums (via bitwise manipulation of their values) which is not great.

What I meant was that (mhvalue=4, mhselect=4) has the exact same semantics as (mhvalue=2, mhselect=1) and therefore `mhselect_mode()` and `mhselect_compare()` should return the exact same things for those two scenarios.

scottj97 (changes requested):
> Breakdown commit for texta_match(). Incrementally include checking of scontext and mcontext. Because the CSRs are optional, each commit is still a valid implementation.

I don't think it's a valid implementation to allow writing `sselect=SSELECT_CONTEXT` when `scontext` doesn't exist, or `mhselect=MHSELECT_MCONTEXT` when `mcontext` doesn't exist.

riscv/triggers.cc
timsifive: Everything in debug_defines.h is auto-generated from the same XML that generates the debug spec registers itself. That mechanism supports some basic operations based on XLEN, but using `? :` like is needed here won't work well. (At the very least, it will be hard to comprehend the document, which is the primary goal.)

I suggest putting it in `triggers.h`. These values won't change in the future.

riscv/triggers.h
YenHaoChen: Are you suggesting to group mhselect={4,1,5} into one mode?
```
typedef enum {
  MHSELECT_IGNORE,
  MHSELECT_MCONTEXT,
  MHSELECT_VMID,
  MHSELECT_RESERVED
} mhselect_mode_t;
```
With that, the mhselect_compare() returns mhvalue or {mhvalue, mhselect[2]}.
```
unsigned mhselect_compare() { return legalize_mhselect(mhselect)==4 ? mhvalue : (mhvalue << 1) + (mhselect >> 2) }
```

scottj97: Yes, exactly, because they work the same, the only difference being an extra bit in the compared value.

I don't think you need RESERVED in the enum because that doesn't describe a compare mode.

YenHaoChen:
> As of [55f3754](https://github.com/riscv-software-src/riscv-isa-sim/commit/55f375413942cc01182a2667d8d56d4e7b3f614f) I still do not see `tdata3` being writable to any value other than 0.

The https://github.com/riscv-software-src/riscv-isa-sim/pull/1160/commits/546fd4ba0e0538dc4ed97fa16f189562e99a1995 passes CSR instructions to tdata3_csr_t::unlogged_write(). The https://github.com/riscv-software-src/riscv-isa-sim/pull/1160/commits/5f7cbbdce8e8308a45a47f50496c7e3ba3d85fb9 gives the value to trigger_t::tdata3_write(). 
I observed non-zero values of tdata3 in my tests. 
Have I messed up or misunderstood anything?

scottj97:
> > As of [55f3754](https://github.com/riscv-software-src/riscv-isa-sim/commit/55f375413942cc01182a2667d8d56d4e7b3f614f) I still do not see `tdata3` being writable to any value other than 0.
> 
> The [546fd4b](https://github.com/riscv-software-src/riscv-isa-sim/commit/546fd4ba0e0538dc4ed97fa16f189562e99a1995) passes CSR instructions to dedicate tdata3_csr_t::unlogged_write(). The [5f7cbbd](https://github.com/riscv-software-src/riscv-isa-sim/commit/5f7cbbdce8e8308a45a47f50496c7e3ba3d85fb9) gives the value to trigger_t::tdata3_write(). I observed non-zero values of tdata3 in my tests. Have I messed up or misunderstood anything?

My mistake. I'd screwed up my test coverage of tdata3. It looks correct.

riscv/triggers.cc
```diff
@@ -17,6 +23,42 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+  tdata3 = set_field(tdata3, CSR_TEXTRA_SBYTEMASK(xlen), sbytemask);
+  tdata3 = set_field(tdata3, CSR_TEXTRA_SVALUE(xlen), svalue);
+  tdata3 = set_field(tdata3, CSR_TEXTRA_SSELECT(xlen), sselect);
+  return tdata3;
+}
+
+void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept {
+  auto xlen = proc->get_xlen();
+  mhvalue = get_field(val, CSR_TEXTRA_MHVALUE(xlen));
+  mhselect = legalize_mhselect(get_field(val, CSR_TEXTRA_MHSELECT(xlen)), proc->extension_enabled('H'));
```
scottj97: Is there any reason to legalize on write, now that it's being legalized on read?

riscv/triggers.h
```diff
@@ -29,6 +29,19 @@ typedef enum {
...
+  SSELECT_ASID = 2,
+  SSELECT_MAXVAL = 2
+} sselect_t;
+
+typedef enum {
+  MHSELECT_MODE_IGNORE,
+  MHSELECT_MODE_MCONTEXT,
+  MHSELECT_MODE_VMID,
+} mhselect_mode_t;
+
```
scottj97, line 44: This is unused in 5f7cbbdce8e8308a45a47f50496c7e3ba3d85fb9; introduce it later, where it is first used

riscv/triggers.h
```diff
@@ -87,9 +87,29 @@ class trigger_t {
 protected:
   action_t legalize_action(reg_t val) const noexcept;
   unsigned legalize_mhselect(reg_t val, bool h_enabled) const noexcept;
+  bool textra_match(processor_t * const proc) const noexcept;
   reg_t tdata2;
 
 private:
+  mhselect_mode_t mhselect_mode(reg_t val, bool h_enabled) const noexcept {
+    switch (legalize_mhselect(val, h_enabled)) {
+      case 0:
```
scottj97: Add comments on spec's interpretation of each value

riscv/triggers.h
```diff
@@ -87,9 +87,29 @@ class trigger_t {
...
+      case 3:
+      case 7:
```
scottj97: Reserved mhselect should not be ignored. This code should assume these have been legalized to something else, and fall through to the `assert(false)` if not.

riscv/triggers.cc
```diff
@@ -59,6 +64,30 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
...
+
+  assert(sselect <= SSELECT_MAXVAL);
+  if (sselect == SSELECT_ASID) {
+    reg_t satp = state->satp->read();
+    reg_t asid = get_field(satp,  SATP_ASID(xlen));
+    if (asid != (svalue & ((1 << ASIDMAX(xlen)) - 1)))
+      return false;
+  }
+
+  mhselect_mode_t mode = mhselect_mode(mhselect, proc->extension_enabled('H'));
```
scottj97: The method already has direct access to `mhselect`; no need to pass it as a param

riscv/triggers.cc
```diff
@@ -59,6 +64,30 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
...
+  if (mode == MHSELECT_MODE_VMID) { // 2 and 6 are vmid_select
+    reg_t vmid = get_field(state->hgatp->read(), HGATP_VMID(hsxlen));
+    if (vmid != (mhselect_compare(mhselect, proc->extension_enabled('H')) & ((1 << VMIDMAX(hsxlen)) - 1)))
```
scottj97: No need to pass `mhselect` to `mhselect_compare()`

riscv/triggers.cc
```diff
@@ -29,16 +30,16 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
-    (unsigned)(h_enabled ? 2 : 0),  // 3
-    (unsigned)                 0 ,  // 4
-    (unsigned)                 0 ,  // 5
-    (unsigned)(h_enabled ? 6 : 0),  // 6
-    (unsigned)(h_enabled ? 6 : 0)   // 7
+    (unsigned)                 0 ,  // 3
+    (unsigned)                 4 ,  // 4
+    (unsigned)(h_enabled ? 5 : 4),  // 5
+    (unsigned)(h_enabled ? 6 : 4),  // 6
+    (unsigned)(h_enabled ? 0 : 4)   // 7
```
scottj97: Why does 16ed9be0b03bc9bf6d8ebaa2f5699bc490db8eb9 change the WARLization of some of the unsupported values? It obscures the important change here, which is enabling 1, 4, and 5.

riscv/triggers.cc
```diff
@@ -71,15 +72,29 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
...
   }
 
   mhselect_mode_t mode = mhselect_mode(mhselect, proc->extension_enabled('H'));
-  if (mode == MHSELECT_MODE_VMID) { // 2 and 6 are vmid_select
+  if (mode == MHSELECT_MODE_MCONTEXT) {
+    if ((state->mcontext->read() ^ mhvalue) & ((1 << CSR_TEXTRA_MHVALUE_LENGTH(xlen)) - 1))
+      return false;
+  } else if (mode == MHSELECT_MODE_MCONTEXT) { // 1 and 5 are mcontext_select
+    if ((state->mcontext->read() ^ mhselect_compare(mhselect, proc->extension_enabled('H'))) & ((1 << (CSR_TEXTRA_MHVALUE_LENGTH(xlen) + 1)) - 1))
+      return false;
```
scottj97: Logic is incorrect. It's checking for the same mode twice in a row, and the first block is broken for mhselect of 1 or 5. The second block of code should work for 1/4/5 (which was the whole point of method `mhselect_compare()`).

Also using XOR unnecessarily again.

riscv/triggers.cc
```diff
@@ -71,15 +72,29 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
...
 
   assert(sselect <= SSELECT_MAXVAL);
-  if (sselect == SSELECT_ASID) {
+  if (sselect == SSELECT_SCONTEXT) {
+    reg_t mask = (reg_t(1) << ((xlen == 32) ? CSR_TEXTRA32_SVALUE_LENGTH : CSR_TEXTRA64_SVALUE_LENGTH)) - 1;
+    assert(CSR_TEXTRA32_SBYTEMASK_LENGTH < CSR_TEXTRA64_SBYTEMASK_LENGTH);
+    for (int i = 0; i < CSR_TEXTRA64_SBYTEMASK_LENGTH; i++)
+      if (sbytemask & (1 << i))
+        mask &= 0xff << (i * 8);
+    if ((state->scontext->read() ^ svalue) & mask)
```
scottj97: Unnecessary use of XOR again

riscv/processor.cc
```diff
@@ -392,6 +392,11 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   csrmap[CSR_TDATA2] = tdata2 = std::make_shared<tdata2_csr_t>(proc, CSR_TDATA2);
   csrmap[CSR_TDATA3] = std::make_shared<tdata3_csr_t>(proc, CSR_TDATA3);
   csrmap[CSR_TINFO] = std::make_shared<tinfo_csr_t>(proc, CSR_TINFO);
+  unsigned scontext_length = (xlen == 32 ? 16 : 34); // debug spec suggests 16-bit for RV32 and 34-bit for RV64
+  csrmap[CSR_SCONTEXT] = scontext = std::make_shared<masked_csr_t>(proc, CSR_SCONTEXT, (reg_t(1) << scontext_length) - 1, 0);
+  unsigned hcontext_length = (xlen == 32 ? 6 : 13) + (proc->extension_enabled('H') ? 1 : 0); // debug spec suggest 7-bit (6-bit) for RV32 and 14-bit (13-bit) for RV64 with (without) H extension
+  csrmap[CSR_HCONTEXT] = hcontext = std::make_shared<masked_csr_t>(proc, CSR_HCONTEXT, (reg_t(1) << hcontext_length) - 1, 0);
+  csrmap[CSR_MCONTEXT] = mcontext = std::make_shared<proxy_csr_t>(proc, CSR_MCONTEXT, csrmap[CSR_HCONTEXT]);
```
scottj97: The commit message for 16ed9be0b03bc9bf6d8ebaa2f5699bc490db8eb9 should mention all 3 CSRs being added

riscv/processor.h
```diff
@@ -177,6 +177,9 @@ struct state_t
   dcsr_csr_t_p dcsr;
   csr_t_p tselect;
   tdata2_csr_t_p tdata2;
+  csr_t_p scontext;
+  csr_t_p hcontext;
```
scottj97: No need to keep a handle to `hcontext` here.

riscv/triggers.cc
```diff
@@ -17,6 +29,80 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
   return (val > ACTION_MAXVAL || (val == ACTION_DEBUG_MODE && get_dmode() == 0)) ? ACTION_DEBUG_EXCEPTION : (action_t)val;
 }
 
+unsigned trigger_t::legalize_mhselect(reg_t val, bool h_enabled) const noexcept {
+  // Consider a mask 0x4 if without h extension
```
scottj97: I don't understand this comment?

riscv/triggers.cc
```diff
@@ -17,6 +29,80 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+  unsigned convert[8] = {
+    (unsigned)                 0 ,  // 0
+    (unsigned)(h_enabled ? 1 : 0),  // 1
+    (unsigned)(h_enabled ? 2 : 0),  // 2
+    (unsigned)                 0 ,  // 3
```
scottj97: Does this really need an `(unsigned)` cast on each line?

riscv/triggers.cc
```diff
@@ -185,7 +214,7 @@ std::optional<match_result_t> mcontrol_t::detect_memory_access_match(processor_t
     value &= 0xffffffff;
   }
 
-  if (simple_match(xlen, value)) {
+  if (simple_match(xlen, value) && textra_match(proc)) {
```
scottj97: Rather than each trigger type having to consider `textra`, can this be done centrally inside `module_t::detect_memory_access_match` and `module_t::detect_trap_match`?

riscv/triggers.h
```diff
@@ -71,7 +86,15 @@ class trigger_t {
 
 protected:
   action_t legalize_action(reg_t val) const noexcept;
+  unsigned legalize_mhselect(reg_t val, bool h_enabled) const noexcept;
```
scottj97: Does this need to be `protected`? Why not `private`?

riscv/triggers.h
```diff
@@ -87,9 +87,29 @@ class trigger_t {
 protected:
   action_t legalize_action(reg_t val) const noexcept;
   unsigned legalize_mhselect(reg_t val, bool h_enabled) const noexcept;
+  bool textra_match(processor_t * const proc) const noexcept;
   reg_t tdata2;
 
 private:
+  mhselect_mode_t mhselect_mode(reg_t val, bool h_enabled) const noexcept {
```
scottj97: Put this in `triggers.cc` with the others

riscv/triggers.cc
```diff
@@ -17,6 +29,80 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+    reg_t mask = (reg_t(1) << ((xlen == 32) ? CSR_TEXTRA32_SVALUE_LENGTH : CSR_TEXTRA64_SVALUE_LENGTH)) - 1;
+    assert(CSR_TEXTRA32_SBYTEMASK_LENGTH < CSR_TEXTRA64_SBYTEMASK_LENGTH);
+    for (int i = 0; i < CSR_TEXTRA64_SBYTEMASK_LENGTH; i++)
+      if (sbytemask & (1 << i))
+        mask &= 0xff << (i * 8);
+    if ((state->scontext->read() ^ svalue) & mask)
+      return false;
+  } else if (sselect == SSELECT_ASID) {
+    reg_t satp = state->satp->read();
+    reg_t asid = get_field(satp,  SATP_ASID(xlen));
```
scottj97: Use `const` for temps where possible

riscv/triggers.cc
```diff
@@ -17,6 +29,80 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+
+  mhselect_mode_t mode = mhselect_mode(mhselect, proc->extension_enabled('H'));
+  if (mode == MHSELECT_MODE_MCONTEXT) {
+    if ((state->mcontext->read() ^ mhvalue) & ((1 << CSR_TEXTRA_MHVALUE_LENGTH(xlen)) - 1))
+      return false;
+  } else if (mode == MHSELECT_MODE_MCONTEXT) { // 1 and 5 are mcontext_select
+    if ((state->mcontext->read() ^ mhselect_compare(mhselect, proc->extension_enabled('H'))) & ((1 << (CSR_TEXTRA_MHVALUE_LENGTH(xlen) + 1)) - 1))
+      return false;
+  } else if (mode == MHSELECT_MODE_VMID) { // 2 and 6 are vmid_select
+    reg_t vmid = get_field(state->hgatp->read(), HGATP_VMID(hsxlen));
```
scottj97: `const`

riscv/triggers.h
```diff
@@ -180,6 +223,8 @@ class module_t {
   bool tdata1_write(processor_t * const proc, unsigned index, const reg_t val) noexcept;
   reg_t tdata2_read(const processor_t * const proc, unsigned index) const noexcept;
   bool tdata2_write(processor_t * const proc, unsigned index, const reg_t val) noexcept;
+  reg_t tdata3_read(const processor_t * const proc, unsigned index) const noexcept;
+  bool tdata3_write(processor_t * const proc, unsigned index, const reg_t val) noexcept;
```
scottj97: `module_t` already has a handle to `proc`; no need to pass it in again. (And we could remove it from the other methods above, too.)

scottj97 (changes requested):
This is much improved. I'm still a little concerned that there are three separate places where the mhselect raw value gets interpreted. This is an opportunity for bugs when the spec changes.

What if we consolidated the interpretation into one table so it was obvious where it all lived in the code? I've implemented this in a [branch](https://github.com/scottj97/riscv-isa-sim/tree/mhselect_interpret). See the [table here](https://github.com/scottj97/riscv-isa-sim/blob/73f07f146b8c47362611ec8988f6f1458163665b/riscv/triggers.h#L104-L115)

Feel free to rework this into your own commits however you think best.

riscv/triggers.cc
YenHaoChen: I want to provide a reasonable WARLization from a hardware perspective. From a software perspective, any mapping to legal values works likewise; however, practical WARLization hardware aims to minimize area and latency and usually uses a constant mask to legalize.
Nevertheless, the WARLization is implementation-dependent. How do you think about the following WARLization?
```
 unsigned convert[8] = { // without mcontext/hcontext
                     0 ,  // 0
                     0 ,  // 1
    (h_enabled ? 2 : 0),  // 2
                     0 ,  // 3
                     0 ,  // 4
                     0 ,  // 5
    (h_enabled ? 6 : 0),  // 6
                     0    // 7
 };
```
```
  unsigned convert[8] = { // with mcontext/hcontext
                     0 ,  // 0
    (h_enabled ? 1 : 0),  // 1
    (h_enabled ? 2 : 0),  // 2
                     0 ,  // 3
                     4 ,  // 4
    (h_enabled ? 5 : 4),  // 5
    (h_enabled ? 6 : 4),  // 6
                     4    // 7
  };
```

riscv/triggers.cc
YenHaoChen: Change to "the only valid value is 4, otherwise 0"

YenHaoChen: Remove the comment. Update the WALization, which aims to be self-explanatory.

riscv/triggers.cc
YenHaoChen: I get warnings about conversion from "int" to "unsigned" without the cast.
```
warning: narrowing conversion of '(h_enabled ? 2 : 0)' from 'int' to 'unsigned int' [-Wnarrowing]
```

riscv/processor.h
YenHaoChen: Remove hcontext since it is not used at all.

riscv/triggers.cc
```diff
@@ -3,6 +3,18 @@
 #include "processor.h"
 #include "triggers.h"
 
+#define ASIDMAX(XLEN) (XLEN == 32 ? 9 : 16)
+#define SATP_ASID(XLEN) (XLEN == 32 ? SATP32_ASID : SATP64_ASID)
```
scottj97: Params should be `SXLEN`, no?

riscv/triggers.cc
```diff
@@ -17,6 +29,68 @@ action_t trigger_t::legalize_action(reg_t val) const noexcept {
...
+    const reg_t asid = get_field(satp,  SATP_ASID(xlen));
+    if (asid != (svalue & ((1 << ASIDMAX(xlen)) - 1)))
+      return false;
+  }
+
+  const auto mhselect_interp = interpret_mhselect(proc->extension_enabled('H'));
+  const mhselect_mode_t mode = mhselect_interp.mode;
+  if (mode == MHSELECT_MODE_MCONTEXT) { // 4, 1, and 5 are mcontext
+    reg_t mask = (1 << (CSR_TEXTRA_MHVALUE_LENGTH(xlen) + 1)) - 1;
+    if ((state->mcontext->read() & mask) != (mhselect_interp.compare_val(mhvalue) & mask))
```
scottj97: No need to mask `mhvalue`; spec says:
> This trigger will only match if the low bits of mcontext/hcontext equal {mhvalue, mhselect[2]}.

scottj97 (changes requested):

scottj97 (approved):
Excellent. Thank you for your contribution and your patience with the review process.

aswaterman:
Thanks to both of you for bringing this to fruition!

PR #1161, opened 2022-12-05
Avoid use of sprintf

aswaterman:
clang 14 declares `sprintf` unsafe, which causes the build to fail.  We should just fall in line and avoid using `sprintf`.

None of our uses were perf-critical, so I switched over to using STL strings when appropriate.  In the cases that were harder to refactor, I switched to `snprintf`.

### DIFF
--- a/disasm/disasm.cc
+++ b/disasm/disasm.cc
@@ -1289,14 +1289,12 @@ void disassembler_t::add_instructions(const isa_parser_t* isa)
                          0x10000000, 0x10005000, 0x10006000, 0x10007000};
 
       for (unsigned nf = 0; nf <= 7; ++nf) {
-        char seg_str[8] = "";
-        if (nf)
-          sprintf(seg_str, "seg%u", nf + 1);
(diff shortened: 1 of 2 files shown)


scottj97 (changes requested):
59570a5d9cf9ef24b21ba84fa14182e125296a64 changes the disassembly of one of the vector instructions:
```
-core   0: 0x00000000800021ba (0x02056507) vle32.v v10, (a0)
+core   0: 0x00000000800021ba (0x02056507) vlseg1e32.v v10, (a0)
```
(I'm not sure I have comprehensive coverage of all vector instructions, so it's possible this is not the only one that changed.)

bef261851cca6d86634b24a98c5300454e8c13d2 (or a subsequent commit) should also remove the now-unused `_name` attribute

aswaterman:
Thanks for catching that screwup.  I missed the `if` statement guarding the `nf`=0 case when "refactoring" the code.

scottj97 (approved):

PR #1162, opened 2022-12-05
SFENCE.W.INVAL and SFENCE.INVAL.IR should check privilege mode

aswaterman:
See discussion on https://lists.riscv.org/g/tech-privileged/topic/95461195#1214

To avoid boilerplate, I did some light refactoring and defined a couple new macros.

cc @ingallsj


scottj97 (changes requested):
I would squash 6090e82d7285edee6efac161cb862f4f32e74c4b and 192addeaef3b020ec92bd132c2afa0adcf164a3c since the former is only introducing dead code.

aswaterman:
Ack.

scottj97 (approved):

PR #1164, opened 2022-12-06
add an AMO operation to mcontrol trigger

YenHaoChen:
This commit fixes the following two issues for address trigger on AMO:
1. The address trigger has a higher priority than the misaligned fault and page fault. 
    [riscv-software-src/riscv-isa-sim/issues/971] 
    [riscv-software-src/riscv-isa-sim/pull/1105]
2. AMO can fire on a chain with a load-only trigger and a store-only trigger. 
    [riscv/riscv-debug-spec/issues/780]

This commit does NOT fix the priority for data trigger on AMO.

The fix lies on the critical path of AMO. Is this OK?

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -146,6 +146,7 @@ public:
       store_slow_path(addr, sizeof(T), nullptr, 0, false, true);
       auto lhs = load<T>(addr);
       store<T>(addr, f(lhs));
+      check_triggers(triggers::OPERATION_AMO, addr, f(lhs));
       return lhs;
     })
   }
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -141,6 +141,7 @@ public:
   // template for functions that perform an atomic memory operation
   template<typename T, typename op>
   T amo(reg_t addr, op f) {
+    check_triggers(triggers::OPERATION_AMO, addr);
     convert_load_traps_to_store_traps({
       store_slow_path(addr, sizeof(T), nullptr, 0, false, true);
       auto lhs = load<T>(addr);
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -120,6 +120,7 @@ std::optional<match_result_t> mcontrol_t::detect_memory_access_match(processor_t
   if ((operation == triggers::OPERATION_EXECUTE && !execute) ||
       (operation == triggers::OPERATION_STORE && !store) ||
       (operation == triggers::OPERATION_LOAD && !load) ||
+      (operation == triggers::OPERATION_AMO && (!store && !load)) ||
       (state->prv == PRV_M && !m) ||
       (state->prv == PRV_S && !s) ||
       (state->prv == PRV_U && !u) ||
--- a/riscv/triggers.h
+++ b/riscv/triggers.h
@@ -12,6 +12,7 @@ typedef enum {
   OPERATION_EXECUTE,
   OPERATION_STORE,
   OPERATION_LOAD,
+  OPERATION_AMO // both LOAD and STORE
 } operation_t;
 
 typedef enum


aswaterman (commented):
I think unconditionally checking the triggers on AMOs is acceptable, provided you have brainstormed for other solutions but haven't found any.

However, I wonder if fixing the data trigger priority will change the nature of the ideal solution.  Maybe you should prototype the complete fix and see if this approach is still ideal?

YenHaoChen:
@aswaterman Thanks for the comments. I add a commit for data trigger.

aswaterman:
This seems wrong.  What if the AMO is to a region where loads have side effects?  Then the side effect will occur even though the AMO didn't actually complete, which is an atomicity violation.

It would seem to me that there are two valid approaches that preserve the atomicity property on regions with side-effecting loads: either disallow data triggers on AMOs, or perform the store piece of the AMO before acting on the data trigger.

Then, once we agree for how to do this on regions with side-effecting loads, we should probably do the same thing for all regions for simplicity.

YenHaoChen:
Yes, you are right. The data trigger can only fire "after" on load and AMO instructions. The data trigger can only check after the data return, i.e., after the execution.

Move the data trigger checking after the store piece of the AMO.

riscv/mmu.h
```diff
@@ -141,10 +141,12 @@ class mmu_t
   // template for functions that perform an atomic memory operation
   template<typename T, typename op>
   T amo(reg_t addr, op f) {
+    check_triggers(triggers::OPERATION_AMO, addr);
     convert_load_traps_to_store_traps({
       store_slow_path(addr, sizeof(T), nullptr, 0, false, true);
       auto lhs = load<T>(addr);
```
aswaterman, line 147: Isn't it possible that this `load` will accidentally hit the data trigger, therefore causing an unintended side effect without performing the store?

riscv/mmu.h
```diff
@@ -141,10 +141,12 @@ class mmu_t
...
       store<T>(addr, f(lhs));
```
aswaterman, line 148: Similarly, isn't it possible that `store` will accidentally hit the data trigger, therefore allowing the load side effect to occur without actually performing the store?

YenHaoChen:
Yes, the data trigger inside load/store breaks the atomicity of AMO.
Should we add a flag to disable the data trigger to the load/store functions?
For example:
```T ALWAYS_INLINE load(reg_t addr, uint32_t xlate_flags = 0, bool no_trigger = false)```
```void ALWAYS_INLINE store(reg_t addr, T val, uint32_t xlate_flags = 0, bool no_trigger = false)```

aswaterman:
It's probably simpler to just write a dedicated `slow_path`-style function for AMOs.

It will be simpler than load/store `slow_path` because AMOs are always aligned.

YenHaoChen:
> However, I wonder if fixing the data trigger priority will change the nature of the ideal solution. Maybe you should prototype the complete fix and see if this approach is still ideal?

After the discussion, I believe the AMO data trigger has the same priority as load data trigger (the lowest one in the table).
image

YenHaoChen:
> It's probably simpler to just write a dedicated `slow_path`-style function for AMOs.
> 
> It will be simpler than load/store `slow_path` because AMOs are always aligned.

I like the idea. But that will be out of this PR's scope.

aswaterman:
> But that will be out of this PR's scope.

Not necessarily.  The stated goal of this PR is to fix these various bugs, but as proposed, it doesn't actually fix them.  And your recently proposed fix (adding new flags to load/store functions) adds technical debt.  So perhaps the only clean way of fixing the problem is to fix it directly, by adding the new function.

YenHaoChen:
Agree

YenHaoChen:
I don't want to be demanding. But I want to show my high interest in the dedicated ```slow_path```-style function for AMOs.

aswaterman:
I was hoping you'd prototype it.  I won't have time for a few weeks, but I can at least provide feedback on a proposal.

YenHaoChen:
OK, will do.

PR #1166, opened 2022-12-08
Fix segfault with non-contiguous hartid

GuoShibo-cn:
when using --hartid with non-continuous hartids ,spike returns segmentation fault,and finds out that debug module confused hartids with "i" in "procs[i]". when processor resets itself, it won't know which proc it is defined by simulator, it only know it's hartid. therefore adding a logic in sim.cc to lookup "i" according to hartid will fix this bug.

### DIFF
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -449,5 +449,12 @@ memif_endianness_t sim_t::get_target_endianness() const
 
 void sim_t::proc_reset(unsigned id)
 {
-  debug_module.proc_reset(id);
+  int i;
+  for (int j = 0; j < cfg->nprocs(); j++)
+  {
+    if(cfg->hartids()[j] == id)
+	  i = j;
+  }
+// Each proccessor only knows and could provide it's hartid,so simulator looks up which proc it is and pass it to debug module
+  debug_module.proc_reset(i);
 }


aswaterman:
Is this actually a complete fix?  If the debug module assumes hart IDs are contiguous, then this will fix the segfault, but will still be incorrect.  I'm also wondering if there are more subtle problems outside of the debug module.

I'm wondering if we should totally rework the way we handle sparse hart IDs.  We could instead densely construct all harts, through the largest hart ID, but and then just avoid executing the ones that don't exist.  This would perform acceptably after some startup overhead, and would avoid special cases throughout the code.  Alternatively, we could rework the `procs` data structure to be a map of hart ID to hart.  What do you think, @scottj97?

scottj97:
Definite NACK as of [cb0b8ca](https://github.com/riscv-software-src/riscv-isa-sim/pull/1166/commits/cb0b8caf57df319c92d4edd9c5b753b5575ff864) just based on the coding style and failing CI.

I suspect @aswaterman is right about lots of problems. Like what about CLINT's array of `mtimecmp` values? Is that indexed by hartid? A cursory reading of `clint.cc` suggests no, but should it be?

There's probably a half-dozen other questions like that, each of which would need to be investigated.

I would frown on densely constructing. A map sounds much better, though it's not clear to me how many places ever need to look up a processor by hartid instead of simply looping through all.

aswaterman:
I suspect the runtime cost of densely constructing the harts will be preferable to the SW eng costs of extra indirection. We’d need to audit a large amount of the code base to get this right—index vs. ID, size vs. max ID, etc., including in devices like the PLIC and CLINT. By contrast, the lazy thing will just work.

Frowning on the lazy scheme is understandable, but the lazy scheme still seems more pragmatic to me.

scottj97:
Do we have an idea of what the max hartid would be limited to with the lazy scheme?

aswaterman:
Spike can already handle up to 1024 harts, at which point the debug module craps out. So, I guess that’s also the limit on hart ID in this scheme, without additional work.

There is some noticeable startup overhead when constructing 1K harts (several hundred ms), but I suspect there’s low-hanging fruit there, including constructing 1K identical disassemblers.

scottj97:
If max hartid of 1K is acceptable then I'm okay with the lazy way.

scottj97:
@aswaterman what should we do with this? At the least, we could detect non-contiguous hartids and bomb with a clean error instead of segfault.

aswaterman:
Let me try to prototype my scheme today.  I got some bonus free time because that lovely Austin weather cancelled my flight.

aswaterman:
Superseded by #1246.

PR #1167, opened 2022-12-08
[RFC] make XLENs changable

liweiwei90:
This patchset add the support for changable XLENs:
- Instruction functions use different XLENs for different privilege modes
- Make *XL field of *status  CSRs writable
- The length of CSRs  is changed to the XLEN of its related privilege mode
- Limit the *XL to legal value(1 or 2 or unchanged)
- Ensure MXLEN >= SXLEN >=VSXLEN/VUXLEN/UXLEN, and VSXLEN >= VUXLEN

Tests for changable XLENs are lacked currently.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -714,9 +714,25 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
 
   const reg_t old_misa = read();
   const bool prev_h = old_misa & (1L << ('H' - 'A'));
-  const reg_t new_misa = (adjusted_val & write_mask) | (old_misa & ~write_mask);
+  reg_t new_misa = (adjusted_val & write_mask) | (old_misa & ~write_mask);
   const bool new_h = new_misa & (1L << ('H' - 'A'));
   unsigned mxlen = proc->get_xlen((priv_mode_t){ PRV_M, false });
+
+  /* MXL is writable when MAX_XLEN > 32 */
+  if (((max_isa >> 62) & 3) > 1) {
+    const uint8_t old_mxl = (old_misa >> (mxlen - 2)) & 3;
+    const uint8_t new_mxl = adjusted_val >> (mxlen - 2);
+    if (is_legal_xl(new_mxl) && old_mxl != new_mxl) {
+      unsigned new_mxlen = xl_to_xlen(new_mxl);
+      proc->update_mxlen(new_mxlen);
+      /*
+       * If a write to misa causes MXLEN to change, the position of
+       * MXL moves to the most-significant two bits of misa at the new width
+       */
+      new_misa = (new_misa & ~(3 << (mxlen - 2))) | (new_mxl << (new_mxlen - 2));
+      mxlen = new_mxlen;
+    }
+  }
 
   // update the hypervisor-only bits in MEDELEG and other CSRs
   if (!new_h && prev_h) {
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -387,11 +387,10 @@ cause_csr_t::cause_csr_t(processor_t* const proc, const reg_t addr, priv_mode_t 
 
 reg_t cause_csr_t::read() const noexcept {
   reg_t val = basic_csr_t::read();
-  // When reading, the interrupt bit needs to adjust to xlen. Spike does
-  // not generally support dynamic xlen, but this code was (partly)
-  // there since at least 2015 (ea58df8 and c4350ef).
-  if (proc->get_isa().get_max_xlen() > proc->get_xlen()) // Move interrupt bit to top of xlen
-    return val | ((val >> (proc->get_isa().get_max_xlen()-1)) << (proc->get_xlen()-1));
+
+  unsigned xlen = proc->get_xlen(prv);
+  if (proc->get_isa().get_max_xlen() > xlen) // Move interrupt bit to top of xlen
+    return val | ((val >> (proc->get_isa().get_max_xlen() - 1)) << (xlen - 1));
   return val;
 }
 
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -872,8 +872,9 @@ henvcfg_csr_t::henvcfg_csr_t(processor_t* const proc, const reg_t addr, const re
 }
 
 // implement class base_atp_csr_t and family
-base_atp_csr_t::base_atp_csr_t(processor_t* const proc, const reg_t addr):
-  basic_csr_t(proc, addr, 0) {
+base_atp_csr_t::base_atp_csr_t(processor_t* const proc, const reg_t addr, priv_mode_t prv):
+  basic_csr_t(proc, addr, 0),
+  prv(prv) {
 }
 
 bool base_atp_csr_t::unlogged_write(const reg_t val) noexcept {
@@ -884,7 +885,7 @@ bool base_atp_csr_t::unlogged_write(const reg_t val) noexcept {
 }
 
 bool base_atp_csr_t::satp_valid(reg_t val) const noexcept {
-  if (proc->get_xlen() == 32) {
+  if (proc->get_xlen(prv) == 32) {
     switch (get_field(val, SATP32_MODE)) {
       case SATP_MODE_SV32: return proc->supports_impl(IMPL_MMU_SV32);
       case SATP_MODE_OFF: return true;
@@ -915,7 +916,7 @@ reg_t base_atp_csr_t::compute_new_satp(reg_t val) const noexcept {
 }
 
 satp_csr_t::satp_csr_t(processor_t* const proc, const reg_t addr):
-  base_atp_csr_t(proc, addr) {
+  base_atp_csr_t(proc, addr, (priv_mode_t){ PRV_S, false }) {
 }
 
 void satp_csr_t::verify_permissions(insn_t insn, bool write) const {
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -905,10 +905,11 @@ bool base_atp_csr_t::satp_valid(reg_t val) const noexcept {
 reg_t base_atp_csr_t::compute_new_satp(reg_t val) const noexcept {
   reg_t rv64_ppn_mask = (reg_t(1) << (MAX_PADDR_BITS - PGSHIFT)) - 1;
 
-  reg_t mode_mask = proc->get_xlen() == 32 ? SATP32_MODE : SATP64_MODE;
-  reg_t asid_mask_if_enabled = proc->get_xlen() == 32 ? SATP32_ASID : SATP64_ASID;
+  reg_t xlen = proc->get_xlen(prv);
+  reg_t mode_mask = xlen == 32 ? SATP32_MODE : SATP64_MODE;
+  reg_t asid_mask_if_enabled = xlen == 32 ? SATP32_ASID : SATP64_ASID;
   reg_t asid_mask = proc->supports_impl(IMPL_MMU_ASID) ? asid_mask_if_enabled : 0;
-  reg_t ppn_mask = proc->get_xlen() == 32 ? SATP32_PPN : SATP64_PPN & rv64_ppn_mask;
+  reg_t ppn_mask = xlen == 32 ? SATP32_PPN : SATP64_PPN & rv64_ppn_mask;
   reg_t new_mask = (satp_valid(val) ? mode_mask : 0) | asid_mask | ppn_mask;
   reg_t old_mask = satp_valid(val) ? 0 : mode_mask;
 
@@ -1102,7 +1103,7 @@ bool hgatp_csr_t::unlogged_write(const reg_t val) noexcept {
   proc->get_mmu()->flush_tlb();
 
   reg_t mask;
-  if (proc->get_const_xlen() == 32) {
+  if (proc->get_xlen((priv_mode_t){ PRV_S, false }) == 32) {
     mask = HGATP32_PPN |
         HGATP32_MODE |
         (proc->supports_impl(IMPL_MMU_VMID) ? HGATP32_VMID : 0);
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -531,10 +531,40 @@ bool mstatus_csr_t::unlogged_write(const reg_t val) noexcept {
 
   const reg_t requested_mpp = proc->legalize_privilege(get_field(val, MSTATUS_MPP));
   const reg_t adjusted_val = set_field(val, MSTATUS_MPP, requested_mpp);
-  const reg_t new_mstatus = (read() & ~mask) | (adjusted_val & mask);
+  const reg_t new_mstatus = (this->val & ~mask) | (adjusted_val & mask);
+
   maybe_flush_tlb(new_mstatus);
-  this->val = adjust_sd(new_mstatus, proc->get_xlen((priv_mode_t){ PRV_M, false }));
+
+  const unsigned mxlen = proc->get_xlen((priv_mode_t) { PRV_M, false });
+
+  if (mxlen > 32) {
+    int new_sxl = get_field(val, MSTATUS_SXL);
+    int new_uxl = get_field(val, MSTATUS_UXL);
+
+    if (is_legal_xl(new_sxl))
+      proc->update_sxlen(xl_to_xlen(new_sxl));
+
+    if (is_legal_xl(new_uxl))
+      proc->update_uxlen(xl_to_xlen(new_uxl));
+  }
+
+  this->val = new_mstatus;
+
   return true;
+}
+
+reg_t mstatus_csr_t::read() const noexcept {
+  const unsigned mxlen = proc->get_xlen((priv_mode_t) { PRV_M, false });
+  reg_t mstatus_val = val;
+
+  if (mxlen != 32) {
+    const unsigned sxlen = proc->get_xlen((priv_mode_t) { PRV_S, false });
+    const unsigned uxlen = proc->get_xlen((priv_mode_t) { PRV_U, false });
+    mstatus_val = set_field(mstatus_val, MSTATUS_SXL, xlen_to_xl(sxlen));
+    mstatus_val = set_field(mstatus_val, MSTATUS_UXL, xlen_to_xl(uxlen));
+  }
+
+  return adjust_sd(mstatus_val, mxlen);
 }
 
 reg_t mstatus_csr_t::compute_mstatus_initial_value() const noexcept {
@@ -542,8 +572,6 @@ reg_t mstatus_csr_t::compute_mstatus_initial_value() const noexcept {
                               | (proc->extension_enabled_const('S') ? MSTATUS_SBE : 0)
                               | MSTATUS_MBE;
   return 0
-         | (proc->extension_enabled_const('U') && (proc->get_const_xlen() != 32) ? set_field((reg_t)0, MSTATUS_UXL, xlen_to_xl(proc->get_const_xlen())) : 0)
-         | (proc->extension_enabled_const('S') && (proc->get_const_xlen() != 32) ? set_field((reg_t)0, MSTATUS_SXL, xlen_to_xl(proc->get_const_xlen())) : 0)
          | (proc->get_mmu()->is_target_big_endian() ? big_endian_bits : 0)
          | 0;  // initial value for mstatus
 }
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -634,6 +634,7 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   const bool prev_h = old_misa & (1L << ('H' - 'A'));
   const reg_t new_misa = (adjusted_val & write_mask) | (old_misa & ~write_mask);
   const bool new_h = new_misa & (1L << ('H' - 'A'));
+  unsigned mxlen = proc->get_xlen((priv_mode_t){ PRV_M, false });
 
   // update the hypervisor-only bits in MEDELEG and other CSRs
   if (!new_h && prev_h) {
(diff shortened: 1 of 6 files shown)


aswaterman:
I think we should take a step back and determine whether Spike really needs to support this feature. It adds a lot of complexity, but it isn’t clearly of high value.

More generally, we should look at the roadmap of RISC-V features and decide what Spike needs to support and what can be omitted.

liweiwei90:
This seems to be a very interesting feature supported by RISC-V. As spike is the golden model, I think adding some complexity to support this feature seems acceptable. It introduces minor change to the main stream of simulation. Most of the additional complexity is for CSRs access.

riscv/csrs.cc
```diff
@@ -378,30 +381,31 @@ bool tvec_csr_t::unlogged_write(const reg_t val) noexcept {
...
-  basic_csr_t(proc, addr, 0) {
+cause_csr_t::cause_csr_t(processor_t* const proc, const reg_t addr, uint8_t prv_index):
+  basic_csr_t(proc, addr, 0),
+  prv_index(prv_index) {
 }
 
 reg_t cause_csr_t::read() const noexcept {
   reg_t val = basic_csr_t::read();
   // When reading, the interrupt bit needs to adjust to xlen. Spike does
   // not generally support dynamic xlen, but this code was (partly)
```
scottj97: Comment is now out of date

riscv/csrs.cc
```diff
@@ -378,30 +381,31 @@ bool tvec_csr_t::unlogged_write(const reg_t val) noexcept {
...
   // there since at least 2015 (ea58df8 and c4350ef).
-  if (proc->get_isa().get_max_xlen() > proc->get_xlen()) // Move interrupt bit to top of xlen
-    return val | ((val >> (proc->get_isa().get_max_xlen()-1)) << (proc->get_xlen()-1));
+  unsigned xlen = proc->get_xlen(prv_index & 3, prv_index & 4);
```
scottj97: Don't use bit fields when C++ has much better data structures. Make a new struct and pass that to `get_xlen()`.

riscv/csrs.cc
```diff
@@ -500,19 +545,51 @@ bool mstatus_csr_t::unlogged_write(const reg_t val) noexcept {
 
   const reg_t requested_mpp = proc->legalize_privilege(get_field(val, MSTATUS_MPP));
   const reg_t adjusted_val = set_field(val, MSTATUS_MPP, requested_mpp);
-  const reg_t new_mstatus = (read() & ~mask) | (adjusted_val & mask);
+  reg_t new_mstatus = (this->val & ~mask) | (adjusted_val & mask);
```
scottj97: Why can't this still be `const`? I don't see it being modified below...

riscv/csrs.cc
```diff
@@ -500,19 +545,51 @@ bool mstatus_csr_t::unlogged_write(const reg_t val) noexcept {
...
   maybe_flush_tlb(new_mstatus);
-  this->val = adjust_sd(new_mstatus);
+
+  this->val = new_mstatus;
+
+  if (proc->get_xlen() > 32) {
+    int orig_uxl = get_field(new_mstatus, MSTATUS_UXL);
+    int new_uxl = legalize_xl(get_field(val, MSTATUS_UXL), orig_uxl);
+    int orig_sxl = get_field(new_mstatus, MSTATUS_SXL);
+    int new_sxl = legalize_xl(get_field(val, MSTATUS_SXL), orig_sxl);
```
scottj97: Why can't you do these adjustments up above where all the other fields (e.g. MPP) are legalized? Then write `this->val` only once with the final value in the end.

riscv/processor.h
```diff
@@ -252,17 +252,25 @@ class processor_t : public abstract_device_t
   mmu_t* get_mmu() { return mmu; }
   state_t* get_state() { return &state; }
   unsigned get_xlen() const { return xlen; }
```
scottj97, line 202: Having the `xlen` stored here in `processor_t` and simultaneously in the `mstatus` etc CSRs is a source of potential bugs when they get out of sync. Can you have this dynamically refer to `mstatus` and friends to query the XLEN on-demand?

riscv/csrs.cc
```diff
@@ -1239,7 +1239,7 @@ bool hgatp_csr_t::unlogged_write(const reg_t val) noexcept {
   proc->get_mmu()->flush_tlb();
 
   reg_t mask;
-  if (proc->get_const_xlen() == 32) {
+  if (proc->get_xlen(PRV_S) == 32) {
```
scottj97: Not a safe change, since this CSR's implementation still assumes a const xlen. (Because it doesn't dynamically adjust its shape when xlen changes.)

riscv/csrs.cc
```diff
@@ -447,38 +452,78 @@ namespace {
...
+}
+
+reg_t vsstatus_csr_t::read() const noexcept {
+  unsigned vsxlen = proc->get_xlen(PRV_S, true);
+  return adjust_sd(val, vsxlen);
 }
 
 bool vsstatus_csr_t::unlogged_write(const reg_t val) noexcept {
-  const reg_t newval = (this->val & ~sstatus_write_mask) | (val & sstatus_write_mask);
+  unsigned vslen = proc->get_xlen(PRV_S, true);
```
scottj97: Use consistent names. In the previous method this was `vsxlen`

scottj97 (changes requested):
Commit a4f6efe12d38ee90bacda187eca401d85abc092e does way more than what its commit message claims.

Don't make XLEN fields writable before everything else is fixed. We don't want Spike to be broken at any commit point. You should be able to change all the CSRs to support dynamic XLEN first (one at a time!), and everything should still work with fixed XLEN. The last step is to enable writable XLEN fields.

No tests == doesn't work.

I concur with @aswaterman that this might not be desirable to have. Is there any demand for this support? I'm not aware of any RISC-V implementation that supports this capability.

scottj97:
f8e150932838730c25b57a84cd074ab3fb7458a9 does not compile

aswaterman:
Historically, other architectures had the feature to change XLEN to facilitate an ISA transition. When x86-64 processors were deployed, IA-32 software was still the norm, and so the processors wouldn’t have been commercially successful without the ability to run IA-32 code. RISC-V is not in the same boat: RV32 and RV64 were created simultaneously, and so there was no such transition. RV64 is already the norm for applications-class processors. For this reason, I suspect few vendors will choose to implement this feature, and so there probably won’t be much demand to support it in Spike.

If I’m proven wrong about the demand, then obviously I’d change my opinion and would support adding this feature to Spike.

I imagine RV128 processors might need this feature (but RV128 support will be a major feature for other reasons).

riscv/csrs.cc
liweiwei90: OK. I'll modify it.

riscv/csrs.cc
liweiwei90: OK. I'll adjust it.

riscv/csrs.cc
liweiwei90: Yeah. It's not necessary currently. I'll restore it.

riscv/csrs.cc
liweiwei90: In current implementation, update_sxl may modify uxl field of mstatus. Maybe I can add this to the legalize_xl function.

riscv/processor.h
liweiwei90: 'xlen' is for current privilege mode. And it's updated if privilege mode changes. It's used by every instruction function, so I think it can get better performance to keep it alone instead of getting it from CSRs every time. Maybe the other *xlen's are not very necessary.

riscv/csrs.cc
liweiwei90: Why the CSR's implementation is const xlen? Hgatp is stated as HSXLEN bits in spec . So I think its length is changable.

riscv/csrs.cc
liweiwei90: OK. I'll modify it.

liweiwei90:
Thanks for your comments.
> Commit [a4f6efe](https://github.com/riscv-software-src/riscv-isa-sim/commit/a4f6efe12d38ee90bacda187eca401d85abc092e) does way more than what its commit message claims.
> 
> Don't make XLEN fields writable before everything else is fixed. We don't want Spike to be broken at any commit point. You should be able to change all the CSRs to support dynamic XLEN first (one at a time!), and everything should still work with fixed XLEN. The last step is to enable writable XLEN fields.

OK. I'll reorganized the commits.
> 
> No tests == doesn't work.

Yeah. I'll try to adjust the riscv-tests to test it(any other seggestion?). The main reason I sent this RFC is to discuss whether there are problems in my understanding of how XLENs work.

> 
> I concur with @aswaterman that this might not be desirable to have. Is there any demand for this support? I'm not aware of any RISC-V implementation that supports this capability.

I search the RISC-V implementations, and find Xuantie C908 seems to support partial of this feature:
"Support for RV32 COMPAT mode which allows for 64-bit RISC-V CPUs to run 32-bit binary code" (https://www.cnx-software.com/2022/11/04/t-head-xuantie-c908-risc-v-core-targets-aiot-applications/)

liweiwei90:
> Historically, other architectures had the feature to change XLEN to facilitate an ISA transition. When x86-64 processors were deployed, IA-32 software was still the norm, and so the processors wouldn’t have been commercially successful without the ability to run IA-32 code. RISC-V is not in the same boat: RV32 and RV64 were created simultaneously, and so there was no such transition. RV64 is already the norm for applications-class processors. For this reason, I suspect few vendors will choose to implement this feature, and so there probably won’t be much demand to support it in Spike.

Yeah. I didn't find an implementation to support all of this feature. Xuantie C908 seems to support partial of it.
> 
> If I’m proven wrong about the demand, then obviously I’d change my opinion and would support adding this feature to Spike.
> 
> I imagine RV128 processors might need this feature (but RV128 support will be a major feature for other reasons).

I have a question of RV128: there is description about the xlen changing from 32 to wider, how about 64 to 128?

aswaterman:
> I have a question of RV128: there is description about the xlen changing from 32 to wider, how about 64 to 128?

RV128 isn’t defined yet, so this isn’t fully defined, either. The presumption is that this feature will support RV128 in the obvious way.

aswaterman:
> Yeah. I didn't find an implementation to support all of this feature. Xuantie C908 seems to support partial of it.

Yeah, this isn’t compelling evidence of demand (especially since it sounds like a nonconforming implementation).

liweiwei90:
> Yeah, this isn’t compelling evidence of demand (especially since it sounds like a nonconforming implementation).

Sorry, I don't know how much it supports currently. Why I said 'partial' is that it doesn't support 'H' extension which means VSXLEN/VUXLEN will not work.

liweiwei90:
> [f8e1509](https://github.com/riscv-software-src/riscv-isa-sim/commit/f8e150932838730c25b57a84cd074ab3fb7458a9) does not compile

rv64_rv32_low_csr_t is used in the next commit.

riscv/csrs.cc
scottj97: If I do CSRW when HSXLEN=64 and then read it back when HSXLEN=32 what happens?

I'm not even sure what's supposed to happen in that case, since the MODE field changes interpretation. But at least the VMID field needs to move. And there's no code to do that in this implementation.

scottj97:
> > [f8e1509](https://github.com/riscv-software-src/riscv-isa-sim/commit/f8e150932838730c25b57a84cd074ab3fb7458a9) does not compile
> 
> rv64_rv32_low_csr_t is used in the next commit.

If it doesn't compile, that's a problem. Here's the beginning of the giant list of errors that I get when doing a clean compile from f8e150932838730c25b57a84cd074ab3fb7458a9:
```
g++ -MMD -MP  -DPREFIX=\"/usr/local\" -Wall -Wno-unused -Wno-nonportable-include-path -g -O2 -fPIC -g -O2 -DPREFIX=\"/usr/local\" -Wall -Wno-unused -Wno-nonportable-include-path -g -O2 -fPIC -std=c++17 -g -O2   -I. -I/local_home/sjohnson/spike-regress/riscv-isa-sim -I/local_home/sjohnson/spike-regress/riscv-isa-sim/fesvr -I/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv -I/local_home/sjohnson/spike-regress/riscv-isa-sim/disasm -I/local_home/sjohnson/spike-regress/riscv-isa-sim/customext -I/local_home/sjohnson/spike-regress/riscv-isa-sim/fdt -I/local_home/sjohnson/spike-regress/riscv-isa-sim/softfloat -I/local_home/sjohnson/spike-regress/riscv-isa-sim/spike_main -I/local_home/sjohnson/spike-regress/riscv-isa-sim/spike_dasm -I/include/boost-0 -x c++-header /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/insn_template.h -c -o insn_template.h.gch
In file included from /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/processor.h:18,
                 from /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/mmu.h:11,
                 from /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/insn_template.h:4:
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/triggers.h:9:1: error: expected unqualified-id before ‘namespace’
 namespace triggers {
 ^~~~~~~~~
In file included from /local_home/sjohnson/spike-regress/riscv-isa-sim/fesvr/memif.h:9,
                 from /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/processor.h:19,
                 from /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/mmu.h:11,
                 from /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/insn_template.h:4:
/local_home/sjohnson/spike-regress/riscv-isa-sim/fesvr/byteorder.h:68:88: error: invalid use of qualified-name ‘rv32_high_csr_t::target_endian<T>::zero’
 template<typename T> const target_endian<T> target_endian<T>::zero = target_endian(T(0));
                                                                                        ^
/local_home/sjohnson/spike-regress/riscv-isa-sim/fesvr/byteorder.h:69:93: error: invalid use of qualified-name ‘rv32_high_csr_t::target_endian<T>::all_ones’
 template<typename T> const target_endian<T> target_endian<T>::all_ones = target_endian(~T(0));
                                                                                             ^
/local_home/sjohnson/spike-regress/riscv-isa-sim/fesvr/byteorder.h:74:10: error: explicit specialization in non-namespace scope ‘class rv32_high_csr_t’
 template<> class target_endian<uint8_t> : public base_endian<uint8_t> {
          ^
```

liweiwei90:
OK, there is bug in the code(Maybe not the unused rv64_rv32_low_csr_t). I'll fix it later.

riscv/csrs.cc
liweiwei90: The fields for HSXLEN=64 cannot totally map to fields for HSXLEN=32, even though they have the same field names. I think this may be the work of software.

scottj97: Yes, what's supposed to happen here when XLEN changes doesn't seem to be well-specified. But at least we could make some effort not to leave total garbage in there -- move VMID, reset Mode, etc?

I think you're pioneering a lot of architectural work here, and there will be many more questions like this that we don't have good answers for.

aswaterman: “Complete garbage” might be what’s required by the CSR width modulation section in chapter 2. I think it requires a bitwise copy of the whole register, even if bits end up in the wrong fields.

liweiwei90: The current implementation is to use bitwise copy for most CSRs, by leaving them alone, and truncating their length to their related XLEN when read and write. 
However, the CSR value may be unchanged if their XLEN is back-to-back changed(64->32->64). This seems a little different from the description from the CSR width modulation section in chapter 2：
_`“If the new width is wider than the previous width, the temporary register is zero-extended to the wider width”`_

scottj97: > “Complete garbage” might be what’s required by the CSR width modulation section in chapter 2. I think it requires a bitwise copy of the whole register, even if bits end up in the wrong fields.

The opening paragraph there says: "fields and bits of the new-width CSR are, unless specified otherwise, determined..."

I think the two different figures for `hgatp` for HSXLEN of 32/64 count as "specified otherwise." VMID and MODE move and it wouldn't be correct to jam the upper bits of PPN into the VMID field when switching from 64 to 32.

This kind of per-field relocation might not be well explained in the spec, but is necessary in order for changing `misa.MXL` back and forth between 2 (xlen=64) and 1 (xlen=32) to work.

But it's because of architectural questions like this that I am very leery of this being the right time & place to pioneer this feature.

scottj97 (changes requested):
ee8c63f8ed6eb79052e69f91c21511eb16e623de is way too big and does way too much at once. It also changes behavior of `csrw minstret`, changing the value written to the CSR by 1. It also changes logging of `mstatush` sometimes, though that behavior is then reverted back by ffe656a963f10f5ebd43ff54ed84398855af67e5.

dbfbc9ffceaba28406b9c641b32833a1217e50be somehow changes the values in `mstatush` bits `MBE` and `SBE`.

Though I still don't think this is the right time for this feature.

liweiwei90:
> 



> [ee8c63f](https://github.com/riscv-software-src/riscv-isa-sim/commit/ee8c63f8ed6eb79052e69f91c21511eb16e623de) is way too big and does way too much at once. It also changes behavior of `csrw minstret`, changing the value written to the CSR by 1. It also changes logging of `mstatush` sometimes, though that behavior is then reverted back by [ffe656a](https://github.com/riscv-software-src/riscv-isa-sim/commit/ffe656a963f10f5ebd43ff54ed84398855af67e5).

OK.  I'll reorganize them later.
> 
> [dbfbc9f](https://github.com/riscv-software-src/riscv-isa-sim/commit/dbfbc9ffceaba28406b9c641b32833a1217e50be) somehow changes the values in `mstatush` bits `MBE` and `SBE`.

Sorry. I didn't find the way to change `MBE` and `SBE` in this commit. Can you tell me how it works?

> 
> Though I still don't think this is the right time for this feature.

OK. Maybe we can keep it in development here.

scottj97:
> > [dbfbc9f](https://github.com/riscv-software-src/riscv-isa-sim/commit/dbfbc9ffceaba28406b9c641b32833a1217e50be) somehow changes the values in `mstatush` bits `MBE` and `SBE`.
> 
> Sorry. I didn't find the way to change `MBE` and `SBE` in this commit. Can you tell me how it works?

My test is proprietary (sorry) but here's all it does with `mstatus` or `mstatush`:

```
core   0: 0x80000160 (0x30005073) csrwi   mstatus, 0
core   0: 3 0x80000160 (0x30005073) c768_mstatus 0x00000000
core   0: 0x80000164 (0x00002537) lui     a0, 0x2
core   0: 3 0x80000164 (0x00002537) x10 0x00002000
core   0: 0x80000168 (0x80050513) addi    a0, a0, -2048
core   0: 3 0x80000168 (0x80050513) x10 0x00001800
core   0: 0x8000016c (0x30052073) csrs    mstatus, a0
core   0: 3 0x8000016c (0x30052073) c768_mstatus 0x00001800
core   0: 0x80000170 (0x00000297) auipc   t0, 0x0
core   0: 3 0x80000170 (0x00000297) x5  0x80000170
core   0: 0x80000174 (0x01428293) addi    t0, t0, 20
core   0: 3 0x80000174 (0x01428293) x5  0x80000184
core   0: 0x80000178 (0x34129073) csrw    mepc, t0
core   0: 3 0x80000178 (0x34129073) c833_mepc 0x80000184
core   0: 0x8000017c (0xf1402573) csrr    a0, mhartid
core   0: 3 0x8000017c (0xf1402573) x10 0x00000000
core   0: 0x80000180 (0x30200073) mret
core   0: 3 0x80000180 (0x30200073) c768_mstatus 0x00000080
```
... some time later ...
```
core   0: 0x80000218 (0x31002273) csrr    tp, mstatush
core   0: 3 0x80000218 (0x31002273) x4  0x00000005
```

Before dbfbc9ffceaba28406b9c641b32833a1217e50be, that last `csrr` wrote 0 to `x4`.

It would help if you broke up dbfbc9ffceaba28406b9c641b32833a1217e50be into several steps: first keep sxl/uxl fixed, but move the computation of those fields into `mstatus_csr_t::read()`. Then make them writable in a second commit.

riscv/csrs.cc
```diff
@@ -538,17 +542,46 @@ reg_t rv32_low_csr_t::written_value() const noexcept {
...
+  orig->verify_permissions(insn, write);
+}
+
+bool rv64_rv32_low_csr_t::unlogged_write(const reg_t val) noexcept {
+  unsigned xlen = proc->get_xlen(prv);
+  if (xlen == 32)
+    return low_half->unlogged_write(val);
+  else
+    return orig->unlogged_write(val);
+}
```
scottj97, line 625: This doesn't sound right. What about a CSR with S-mode privilege, such as `stimecmp`, which can be read from either M-mode or S-mode? If mxlen==64 and sxlen==32 then `csrw` should behave differently when executed from M-mode vs S-mode.

It appears to me that with this implementation, it will write only 32-bits even when run from M-mode where mxlen=64.

scottj97 (changes requested):

liweiwei90:
> 



> ... some time later ...
> 
> ```
> core   0: 0x80000218 (0x31002273) csrr    tp, mstatush
> core   0: 3 0x80000218 (0x31002273) x4  0x00000005
> ```
> 
> Before [dbfbc9f](https://github.com/riscv-software-src/riscv-isa-sim/commit/dbfbc9ffceaba28406b9c641b32833a1217e50be), that last `csrr` wrote 0 to `x4`.
> 
> It would help if you broke up [dbfbc9f](https://github.com/riscv-software-src/riscv-isa-sim/commit/dbfbc9ffceaba28406b9c641b32833a1217e50be) into several steps: first keep sxl/uxl fixed, but move the computation of those fields into `mstatus_csr_t::read()`. Then make them writable in a second commit.

Yeah. This is the UXL and SXL fields problem. I didn't clear it when MXLEN = 32. I'll fix it later.

riscv/csrs.cc
liweiwei90: Yes. From spec, stimecmp is SXLEN-bits length. So I think  it will only read/write 32-bits if sxlen = 32, no matter mxlen is 32 or 64 (in S mode or M mode).

scottj97:
> Yeah. This is the UXL and SXL fields problem. I didn't clear it when MXLEN = 32. I'll fix it later.

Right. My mistake, that wasn't MBE/SBE bits.

Issue #1168, opened 2022-12-10
Avoiding dependencies on config.h


jerryz123:
Currently many spike-installed headers depend on the autoconf-generated `config.h`, which prevents usage of spike-as-a-library when including those headers.

I would like to remove these dependencies so I can just include `sim.h` in an external project and link with the spike library.
What is the canonical way to do this?

scottj97:
Linking Spike with outside code was enabled by a recent addition to this repo but is not well specified or well supported.

I link Spike with python but I use only a very small & simple C interface (that I added to my local fork of Spike) so I don't need any header files.

Spike compilation uses several configure-time choices such as `--enable-commitlog` and I assume this is why `config.h` is included. But perhaps the header files themselves don't need to care?

In the case of `sim.h` the interface changes if Boost libraries are available (`HAVE_BOOST_ASIO`). This has been a source of [issues](https://github.com/riscv-software-src/riscv-isa-sim/issues/834#issuecomment-944425960) in the past.

What if we moved all such `#ifdef` statements into the `*.cc` files and removed the include of `config.h` from the headers. Would that help you?

jerryz123:
> What if we moved all such #ifdef statements into the *.cc files and removed the include of config.h from the headers. Would that help you?

This would be great, and I'd be happy to submit some PRs towards this.

aswaterman:
If we can nuke all header dependencies on config.h, that would be the best outcome. If any headers truly must depend on config.h, they shouldn’t be installed. Thanks for helping improve this.

PR #1169, opened 2022-12-11
Move boost asio socket interface to socketif_t

jerryz123:
This PR reduces the complexity of linking against spike, addressing some of the issues discussed in #1168.

All boost-asio objects are moved to a `socketif_t` member of `sim_t`, moving spike towards a world where the interface of `sim_t` does not depend on `configure`-time settings.

### DIFF
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -1,5 +1,6 @@
 // See LICENSE for license details.
 
+#include "config.h"
 #include "sim.h"
 #include "mmu.h"
 #include "dts.h"
@@ -7,6 +8,7 @@
 #include "byteorder.h"
 #include "platform.h"
 #include "libfdt.h"
+#include "socketif.h"
 #include <fstream>
 #include <map>
 #include <iostream>
@@ -35,9 +37,7 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
              const debug_module_config_t &dm_config,
              const char *log_path,
              bool dtb_enabled, const char *dtb_file,
-#ifdef HAVE_BOOST_ASIO
-             boost::asio::io_service *io_service_ptr, boost::asio::ip::tcp::acceptor *acceptor_ptr, // option -s
-#endif
+             bool socket_enabled,
              FILE *cmd_file) // needed for command line option --cmd
   : htif_t(args),
     isa(cfg->isa(), cfg->priv()),
@@ -49,10 +49,6 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
     dtb_enabled(dtb_enabled),
     log_file(log_path),
     cmd_file(cmd_file),
-#ifdef HAVE_BOOST_ASIO
-    io_service_ptr(io_service_ptr), // socket interface
-    acceptor_ptr(acceptor_ptr),
-#endif
     sout_(nullptr),
     current_step(0),
     current_proc(0),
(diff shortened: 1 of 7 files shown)


riscv/socketif.cc
```diff
@@ -0,0 +1,77 @@
...
+      io_service_ptr = new boost::asio::io_service;
+      acceptor_ptr = new tcp::acceptor(*io_service_ptr, tcp::endpoint(tcp::v4(), 0));
+      // acceptor is created passing argument port=0, so O.S. will choose a free port
+      std::string name = boost::asio::ip::host_name();
+      std::cout << "Listening for debug commands on " << name.substr(0,name.find('.'))
+                << " port " << acceptor_ptr->local_endpoint().port() << " ." << std::endl;
+      // at the end, add space and some other character for convenience of javascript .split(" ")
+    }
+  catch (std::exception& e)
+    {
```
aswaterman: Looks like the brace formatting in these blocks got screwed up.  Should change it to match the subsequent blocks in this file, which conform to our house style.

riscv/sim.cc
```diff
@@ -74,9 +70,23 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
...
+         "please rebuild the riscv-isa-sim project without --without-boost-asio\n",
+         stderr);
+    abort();
+  }
+#endif
+
 #ifndef RISCV_ENABLE_DUAL_ENDIAN
   if (cfg->endianness != memif_endianness_little) {
-    fputs("Big-endian support has not been prroperly enabled; "
+    fputs("Big-endian support has not been properly enabled; "
```
aswaterman: This sort of change should be in its own commit; refactoring commits should not have functional changes (even trivial ones) when possible.

aswaterman (commented):
This certainly seems like a step in the right direction.  I'll give @scottj97 a chance to review and merge if he feels it's appropriate.

aswaterman:
BTW, the CI flow only tests without Boost.  I assume you tested with Boost, as well?

jerryz123:
> BTW, the CI flow only tests without Boost. I assume you tested with Boost, as well?

I ran the ci-tests script on my machine, which does have boost asio. I also stepped through a few instructions in `-d` mode with boost enabled, and confirmed that `spike -s` does go to `Listening for debug commands on ...`. I didn't actually test connecting to the socket though, I don't know what the socket client is supposed to be.

aswaterman:
Cool.

riscv/sim.cc
```diff
@@ -74,6 +70,20 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
...
 
+  socketif = NULL;
+#ifdef HAVE_BOOST_ASIO
+  if (socket_enabled) {
+    socketif = new socketif_t();
+  }
+#else
+  if (socket_enabled) {
+    fputs("Socket support requires compilation with boost asio; "
+          "please rebuild the riscv-isa-sim project without --without-boost-asio\n",
```
scottj97: This could be confusing to a user who doesn't have Boost installed, and did not use that option to `configure`.

Would it be correct to say `rebuild with --with-boost-asio`? Then `configure` will tell them that the Boost package is missing. (Right?)

scottj97 (changes requested):

scottj97 (approved):

aswaterman (approved):

jerryz123:
Rebased, should be good to go now.

PR #1170, opened 2022-12-12
Split decode.h into public decode.h and internal decode_macros.h

jerryz123:
- `decode.h` should contain generally useful macros/typedefs/classes, but should not depend on `config.h`
- `decode_macros.h` contains internally used macros, especially those used in the instruction definitions.

I tried to preserve git blame history for both `decode.h` and `decode_macros.h`. Unfortunately, this makes the history messier, and also introduces a non-functioning commit.

Is there a way to maintain a linear, bisectable commit history, while also preserving blame history?


scottj97:
I don't understand the point of the embedded merge here, though I can see that the git blame is maintained, which is nice. I don't like the non-functioning commit, and I don't know of a better way, but I'm not sure this messiness in the history is worth it just for the blame functionality.

The commit message in 8fb9e7b25697297ad8d453cf2fb524f6660633bd erroneously refers to `riscv_decode.h`

jerryz123:
I'm following the steps here: https://devblogs.microsoft.com/oldnewthing/20190916-00/?p=102892, but all the approaches seem to require using a `git mv` to preserve the history in the copy,.I don't think there's a way to do it without a broken commit.

In that case, I'll squash all these into a single commit.

aswaterman (approved):

PR #1174, opened 2022-12-13
Make the processor_t interface independent of configure'd variables

jerryz123:
This is more work towards #1168

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -594,7 +594,15 @@ void processor_t::set_histogram(bool value)
 
 void processor_t::enable_log_commits()
 {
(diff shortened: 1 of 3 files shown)


aswaterman (approved):

scottj97:
@aswaterman please don't squash. We need commits to be as tiny as possible to best support bisect in the event of a regression discovered later.

You can disable squashing in the project settings.

aswaterman:
@scottj97 Yes, it was an accident.  I didn't see fit to force-push to undo it.

aswaterman:
I've disabled the option on this repo.

PR #1173, opened 2022-12-13
Split vectorUnit into separate source/header files

jerryz123:
This pulls `vectorUnit_t` out of the declaration of `processor_t`. This is part of work towards making spike more usable as a library.

PR #1175, opened 2022-12-13
Fix build on M1

aswaterman:
https://github.com/riscv-software-src/riscv-isa-sim/pull/1173 introduced a link error on M1.  Fix it.


aswaterman:
cc @jerryz123

PR #1176, opened 2022-12-14
Fix Zc* support for writable misa

### LINKED ISSUE #1172: Alignment  when 'C'  in misa CSR is cleared by software
Whether jump to 0x10082 triggers misaligned exception when  'C' in misa CSR is disabled by software? 
The answer is yes before Zc* is supported. 
However,  Zca is implied by 'C', but Zca can work alone without 'C' enabled.  So this becomes a new problem:
Whether Zca is also disabled when 'C' is disabled?

liweiwei90:
Fix https://github.com/riscv-software-src/riscv-isa-sim/issues/1172
- Disable Zca/Zcf/Zcd when misa.C is cleared if Zcb/Zcmt/Zcmp is not supported
- Enable Zca/Zcf/Zcd when misa.C is enabled
- Make Zcf/Zcd only work when misa.F/D enabled

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -619,9 +619,10 @@ reg_t misa_csr_t::dependency(const reg_t val, const char feature, const char dep
 }
 
 bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
+  const reg_t old_misa = read();
+
   // the write is ignored if increasing IALIGN would misalign the PC
-  if (!(val & (1L << ('C' - 'A'))) && (state->pc & 2) &&
-      proc->extension_changeable(EXT_ZCA))
+  if (!(val & (1L << ('C' - 'A'))) && (old_misa & (1L << ('C' - 'A'))) && (state->pc & 2))
     return false;
 
   reg_t adjusted_val = val;
@@ -629,16 +630,16 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   adjusted_val = dependency(adjusted_val, 'Q', 'D');
   adjusted_val = dependency(adjusted_val, 'V', 'D');
 
-  const reg_t old_misa = read();
   const bool prev_h = old_misa & (1L << ('H' - 'A'));
   const reg_t new_misa = (adjusted_val & write_mask) | (old_misa & ~write_mask);
   const bool new_h = new_misa & (1L << ('H' - 'A'));
 
-  proc->set_extension_enable(EXT_ZCA, (new_misa & (1L << ('C' - 'A'))));
-  proc->set_extension_enable(EXT_ZCF, (new_misa & (1L << ('F' - 'A'))) &&
-                                      proc->extension_enabled(EXT_ZCA));
-  proc->set_extension_enable(EXT_ZCD, (new_misa & (1L << ('D' - 'A'))) &&
-                                      proc->extension_enabled(EXT_ZCA));
+  proc->set_extension_enable(EXT_ZCA, (new_misa & (1L << ('C' - 'A'))) || !proc->get_isa().extension_enabled('C'));
+  proc->set_extension_enable(EXT_ZCF, (new_misa & (1L << ('F' - 'A'))) && proc->extension_enabled(EXT_ZCA));
+  proc->set_extension_enable(EXT_ZCD, (new_misa & (1L << ('D' - 'A'))) && proc->extension_enabled(EXT_ZCA));
+  proc->set_extension_enable(EXT_ZCB, proc->extension_enabled(EXT_ZCA));
+  proc->set_extension_enable(EXT_ZCMP, proc->extension_enabled(EXT_ZCA));
+  proc->set_extension_enable(EXT_ZCMT, proc->extension_enabled(EXT_ZCA));
 
   // update the hypervisor-only bits in MEDELEG and other CSRs
   if (!new_h && prev_h) {
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -628,6 +628,11 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   const reg_t new_misa = (adjusted_val & write_mask) | (old_misa & ~write_mask);
   const bool new_h = new_misa & (1L << ('H' - 'A'));
 
+  proc->set_extension_enable(EXT_ZCF, (new_misa & (1L << ('F' - 'A'))) &&
+                                      proc->extension_enabled(EXT_ZCA));
+  proc->set_extension_enable(EXT_ZCD, (new_misa & (1L << ('D' - 'A'))) &&
+                                      proc->extension_enabled(EXT_ZCA));
+
   // update the hypervisor-only bits in MEDELEG and other CSRs
   if (!new_h && prev_h) {
     reg_t hypervisor_exceptions = 0
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -601,12 +601,13 @@ bool sstatus_csr_t::enabled(const reg_t which) {
 misa_csr_t::misa_csr_t(processor_t* const proc, const reg_t addr, const reg_t max_isa):
   basic_csr_t(proc, addr, max_isa),
   max_isa(max_isa),
-  write_mask(max_isa & (0  // allow MAFDQHV bits in MISA to be modified
+  write_mask(max_isa & (0  // allow MAFDQCHV bits in MISA to be modified
                         | (1L << ('M' - 'A'))
                         | (1L << ('A' - 'A'))
                         | (1L << ('F' - 'A'))
                         | (1L << ('D' - 'A'))
                         | (1L << ('Q' - 'A'))
+                        | (1L << ('C' - 'A'))
                         | (1L << ('H' - 'A'))
                         | (1L << ('V' - 'A'))
                         )
@@ -618,6 +619,11 @@ reg_t misa_csr_t::dependency(const reg_t val, const char feature, const char dep
 }
 
 bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
+  // the write is ignored if increasing IALIGN would misalign the PC
+  if (!(val & (1L << ('C' - 'A'))) && (state->pc & 2) &&
+      proc->extension_changeable(EXT_ZCA))
+    return false;
+
   reg_t adjusted_val = val;
   adjusted_val = dependency(adjusted_val, 'D', 'F');
   adjusted_val = dependency(adjusted_val, 'Q', 'D');
@@ -628,6 +634,7 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   const reg_t new_misa = (adjusted_val & write_mask) | (old_misa & ~write_mask);
   const bool new_h = new_misa & (1L << ('H' - 'A'));
 
+  proc->set_extension_enable(EXT_ZCA, (new_misa & (1L << ('C' - 'A'))));
   proc->set_extension_enable(EXT_ZCF, (new_misa & (1L << ('F' - 'A'))) &&
                                       proc->extension_enabled(EXT_ZCA));
   proc->set_extension_enable(EXT_ZCD, (new_misa & (1L << ('D' - 'A'))) &&
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -223,7 +223,7 @@ public:
     if (ext >= 'A' && ext <= 'Z')
       return state.misa->extension_enabled(ext);
     else
-      return isa->extension_enabled(ext);
+      return extension_enable_table[ext];
   }
   // Is this extension enabled? and abort if this extension can
   // possibly be disabled dynamically. Useful for documenting
@@ -303,6 +303,8 @@ private:
   bool in_wfi;
   bool check_triggers_icount;
   std::vector<bool> impl_table;
+
+  std::bitset<NUM_ISA_EXTENSIONS> extension_enable_table;
 
   std::vector<insn_desc_t> instructions;
   std::unordered_map<reg_t,uint64_t> pc_histogram;
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -218,7 +218,12 @@ public:
   }
   // Only the extensions supported in isa string can be changeable
   bool extension_changeable(unsigned char ext) const {
-    return isa->extension_enabled(ext) && (ext == EXT_ZCF || ext == EXT_ZCD);
+    return isa->extension_enabled(ext) &&
+           (ext == EXT_ZCF || ext == EXT_ZCD ||
+            (ext == EXT_ZCA && isa->extension_enabled('C') &&
+             !(isa->extension_enabled(EXT_ZCB) ||
+               isa->extension_enabled(EXT_ZCMP) ||
+               isa->extension_enabled(EXT_ZCMT))));
   }
   bool extension_enabled(unsigned char ext) const {
     return extension_enabled(isa_extension_t(ext));
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -216,15 +216,6 @@ public:
   bool any_custom_extensions() const {
     return !custom_extensions.empty();
   }
-  // Only the extensions supported in isa string can be changeable
-  bool extension_changeable(unsigned char ext) const {
-    return isa->extension_enabled(ext) &&
-           (ext == EXT_ZCF || ext == EXT_ZCD ||
-            (ext == EXT_ZCA && isa->extension_enabled('C') &&
-             !(isa->extension_enabled(EXT_ZCB) ||
-               isa->extension_enabled(EXT_ZCMP) ||
-               isa->extension_enabled(EXT_ZCMT))));
-  }
   bool extension_enabled(unsigned char ext) const {
     return extension_enabled(isa_extension_t(ext));
   }
@@ -244,14 +235,11 @@ public:
     if (ext >= 'A' && ext <= 'Z')
       return state.misa->extension_enabled_const(ext);
     else {
-      assert(!extension_changeable(ext));
       return isa->extension_enabled(ext);
     }
   }
-  // Only the changeable extensions can be disabled/enabled
   void set_extension_enable(unsigned char ext, bool enable) {
-    if (extension_changeable(ext))
-      extension_enable_table[ext] = enable;
+    extension_enable_table[ext] = enable && isa->extension_enabled(ext);
   }
   void set_impl(uint8_t impl, bool val) { impl_table[impl] = val; }
   bool supports_impl(uint8_t impl) const {
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -216,6 +216,10 @@ public:
   bool any_custom_extensions() const {
     return !custom_extensions.empty();
   }
+  // Only the extensions supported in isa string can be changeable
+  bool extension_changeable(unsigned char ext) const {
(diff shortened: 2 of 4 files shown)


riscv/processor.h
```diff
@@ -267,7 +276,7 @@ class processor_t : public abstract_device_t
   const char* get_symbol(uint64_t addr);
 
 private:
-  const isa_parser_t * const isa;
+  isa_parser_t * const isa;
```
aswaterman: This doesn't seem right.  I think the parsed ISA string should be immutable; it is the processor whose state changes when `misa` changes.

riscv/insns/c_fld.h
```diff
@@ -1,3 +1,4 @@
 require_extension(EXT_ZCD);
+require_extension('D');
```
aswaterman: The load/store changes should be a separate commit.

riscv/processor.h
liweiwei90: OK. Maybe we can maintain a disable list in processor.

riscv/insns/c_fld.h
liweiwei90: We can not simply enable Zcd when we set misa.D if we disable it when we clear misa.D in current implementation. There are two different cases: D_Zca_Zcd and D_Zca. So I do nothing for Zcd when misa.D changes and add these checks.

riscv/csrs.cc
```diff
@@ -620,20 +620,49 @@ reg_t misa_csr_t::dependency(const reg_t val, const char feature, const char dep
...
+      // the write is ignored if increasing IALIGN would misalign the PC
+      if (state->pc & 2)
+        return false;
+      proc->disable_extension(EXT_ZCA);
+      proc->disable_extension(EXT_ZCF);
+      proc->disable_extension(EXT_ZCD);
+    }
+  } else if (new_c && !prev_c) {
+    proc->enable_extension(EXT_ZCA);
+  }
```
scottj97: Does it make sense to disable 3 extensions but only re-enable one of the three?

riscv/csrs.cc
```diff
@@ -620,20 +620,49 @@ reg_t misa_csr_t::dependency(const reg_t val, const char feature, const char dep
...
+
+  const bool prev_f = old_misa & (1L << ('F' - 'A'));
+  const bool new_f = new_misa & (1L << ('F' - 'A'));
+  if (!new_f && prev_f) {
+    proc->disable_extension(EXT_ZCF);
+  } else if (new_f && !prev_f && proc->extension_enabled(EXT_ZCA)) {
+    proc->enable_extension(EXT_ZCF);
+  }
```
scottj97: I assume Zca requires Zcf? This is not the best place to express that restriction, and I suspect it's a duplicate of the ISA parser which also checks that restriction.

If the ISA object will not allow Zcf without Zca then I don't see the harm in calling `enable_extension` here, since `isa->extension_enabled(EXT_ZCF)` will still be false.

This entire `if` should be one function call: `set_extension_enable(EXT_ZCF, new_f)`

riscv/processor.h
```diff
@@ -209,8 +209,17 @@ class processor_t : public abstract_device_t
...
+      return isa->extension_enabled(ext) && !extension_disable_table[ext];
   }
+
+  void enable_extension(unsigned char ext) {
+    extension_disable_table[ext] = false;
+  }
+
+  void disable_extension(unsigned char ext) {
+    extension_disable_table[ext] = true;
+  }
```
scottj97: Please move these below (or above) `extension_enabled_const()` so we can keep that next to its kin `extension_enabled()`.

Also, `extension_enabled_const()` is now subtly broken. See the comment about "assume this can't change" which is no longer a valid assumption.

There needs to be somewhere a list of extensions that are allowed to be dynamically disabled. `disable_extension()` (and the `set_extension_enable()` I suggested) should check that list and `abort()` if any code tries to disable something not in that list. Then `extension_enabled_const()` can check that same list. Only then can assumptions about unchanging extensions be validated, and caught immediately if that should ever change.

riscv/processor.h
```diff
@@ -283,6 +292,8 @@ class processor_t : public abstract_device_t
   bool halt_on_reset;
   std::vector<bool> impl_table;
 
+  std::vector<bool> extension_disable_table;
```
scottj97: `vector<bool>` is weird and slow. Simply:
```
bool extension_disable_table[256] = {};
```

scottj97 (changes requested):

riscv/csrs.cc
liweiwei90: Yeah. I attempted re-enable them in following code. But something wrong in their logic  I'll fix it later.

riscv/csrs.cc
liweiwei90: Zcf requires Zca. We cannot enable Zcf without Zca enabled. The corner case is  isa=CF, if we disable C and F firstly, and enable F without C later.

riscv/processor.h
liweiwei90: OK. I'll update it.

riscv/processor.h
```diff
@@ -209,16 +209,24 @@ class processor_t : public abstract_device_t
...
   }
   // Is this extension enabled? and abort if this extension can
   // possibly be disabled dynamically. Useful for documenting
   // assumptions about writable misa bits.
   bool extension_enabled_const(unsigned char ext) const {
     if (ext >= 'A' && ext <= 'Z')
       return state.misa->extension_enabled_const(ext);
-    else
+    else {
+      if (ext == EXT_ZCA || ext == EXT_ZCF || ext == EXT_ZCD) {
```
scottj97: This is duplicate code. There must be one definition of this list. Otherwise someone will change one without realizing there's two.

riscv/csrs.cc
scottj97: This is a really confusing block of code. We need a way to express these dependencies succinctly instead of through jumbled logic like this. For example we have `misa_csr_t::dependency()` which enables understandable adjustments of the `misa` value like so:
```
  adjusted_val = dependency(adjusted_val, 'D', 'F');
  adjusted_val = dependency(adjusted_val, 'Q', 'D');
  adjusted_val = dependency(adjusted_val, 'V', 'D');
```

scottj97 (changes requested):

scottj97:
There should not be foxtrot merges like [ffbd5bb](https://github.com/riscv-software-src/riscv-isa-sim/pull/1176/commits/ffbd5bb3d013cd9dd804985043a65213b8a221b3) in PRs. Use rebase if you want to include the latest master.

riscv/processor.h
liweiwei90: OK. I'll try to update this.

riscv/processor.h
```diff
@@ -205,20 +205,33 @@ class processor_t : public abstract_device_t
...
   // Is this extension enabled? and abort if this extension can
   // possibly be disabled dynamically. Useful for documenting
   // assumptions about writable misa bits.
   bool extension_enabled_const(unsigned char ext) const {
     if (ext >= 'A' && ext <= 'Z')
       return state.misa->extension_enabled_const(ext);
-    else
+    else {
+      assert(!extension_changable(ext));
       return isa->extension_enabled(ext);  // assume this can't change
```
scottj97: Remove now-obsolete comment

scottj97 (approved):
This is much improved.

One little nit to clean up and this is good.

riscv/processor.h
liweiwei90: OK. I'll remove the "assume ..."

scottj97:
@liweiwei90 please rebase onto the latest master. @aswaterman any reason not to merge this as soon as that's complete?

riscv/csrs.cc
```diff
@@ -620,8 +620,10 @@ reg_t misa_csr_t::dependency(const reg_t val, const char feature, const char dep
 }
 
 bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
+  bool zca_changable = proc->extension_changable(EXT_ZCA);
```
aswaterman: "changeable" is misspelled here and elsewhere; let's fix that before merging.

riscv/processor.h
```diff
@@ -283,6 +296,8 @@ class processor_t : public abstract_device_t
   bool halt_on_reset;
   std::vector<bool> impl_table;
 
+  bool extension_disable_table[256] = {};
```
aswaterman: To match recent changes to `isa_parser_t`, and to support more than 256 extensions, please use the data structure `std::bitset<NUM_ISA_EXTENSIONS>` instead of `bool[256]`.

riscv/processor.h
```diff
@@ -205,20 +205,33 @@ class processor_t : public abstract_device_t
...
+              isa->extension_enabled(EXT_ZCMP) ||
+              isa->extension_enabled(EXT_ZCMT))) ||
+           ext == EXT_ZCF || ext == EXT_ZCD;
+  }
   bool extension_enabled(unsigned char ext) const {
     if (ext >= 'A' && ext <= 'Z')
       return state.misa->extension_enabled(ext);
     else
-      return isa->extension_enabled(ext);
+      return isa->extension_enabled(ext) && !extension_disable_table[ext];
```
aswaterman: I would prefer that we check only the data structure `extension_disable_table` here, rather than checking two things.  This is better for performance and IMO is easier to reason about.

The implication would be that `extension_enable_table[ext]` always be false when `isa->extension_enabled(ext)` is false.

riscv/processor.h
aswaterman: I think it would be easier to understand if we invert the polarity of this table (`extension_enable_table`).  This makes more sense when combined with my comment above https://github.com/riscv-software-src/riscv-isa-sim/pull/1176/files#r1092287066

aswaterman:
@scottj97 I think the premise is fine, but I'd like to re-review after @liweiwei90 has rebased it and has addressed my suggestions.

riscv/csrs.cc
liweiwei90: OK. I'll fix it later.

riscv/processor.h
liweiwei90: OK. I'll try to update these.

riscv/processor.h
```diff
@@ -234,8 +241,14 @@ class processor_t : public abstract_device_t
...
       return state.misa->extension_enabled_const(ext);
-    else
-      return isa->extension_enabled(ext);  // assume this can't change
+    else {
+      assert(!extension_changeable(ext));
+      return isa->extension_enabled(ext);
+    }
+  }
+  void set_extension_enable(unsigned char ext, bool enable) {
+    assert(enable || extension_changeable(ext));
```
scottj97: This doesn't sound right. If the extension is disabled and not changeable, this will incorrectly allow it to be enabled.

riscv/processor.h
```diff
@@ -216,14 +216,21 @@ class processor_t : public abstract_device_t
   bool any_custom_extensions() const {
     return !custom_extensions.empty();
   }
+  bool extension_changeable(unsigned char ext) const {
+    return (ext == EXT_ZCA &&
+            !(isa->extension_enabled(EXT_ZCB) ||
+              isa->extension_enabled(EXT_ZCMP) ||
+              isa->extension_enabled(EXT_ZCMT))) ||
```
scottj97: Should these be `extension_enabled_const()`?

I know that method calls `extension_changeable()` so there is a potential for endless recursion, but I think it should be okay as long as `ext != EXT_ZCA` on those recursive calls.

scottj97:
This PR needs to discuss what happened in #1203 because it seems to undo that change.

scottj97 (changes requested):
I know I approved this before, but either I'm paying more attention today, or something has changed. This is now doing quite a bit of refactoring, and also making behavioral changes (misa.C is now writable). This needs to be split up into multiple commits, to separate the refactoring from the behavior changes, and so the commit messages can clearly explain exactly what's changing at each step.

riscv/processor.h
liweiwei90: Yeah. I didn't take other extensions into consideration. I'll fix this later.

riscv/processor.h
liweiwei90: This changeable function is only for extensions > 'Z' . So `extension_enabled_const()` will finally call` isa->extension_enabled() ` too.

liweiwei90:
> This PR needs to discuss what happened in #1203 because it seems to undo that change.

Yeah. In previous Zc* implementation, disable C will not disable Zca. So it doesn't works and makes the ma_fetch test fail. So #1203 make 'C' read-only to bypass this question. This patch is also to fix this problem from another way: make 'C' in misa still writable, and conditionally disable Zca when clear C in misa.

scottj97 (approved):
I think the ISA is quite confusing here, and I have no opinion about the correctness of the ISA or its implementation here, but otherwise this PR ~looks good~.

Edit: except for the two problems I found below!

riscv/processor.h
scottj97: I was also confused about `proc->extension_enabled()` versus `isa->extension_enabled()`.

riscv/csrs.cc
```diff
@@ -607,6 +607,7 @@ misa_csr_t::misa_csr_t(processor_t* const proc, const reg_t addr, const reg_t ma
                         | (1L << ('F' - 'A'))
                         | (1L << ('D' - 'A'))
                         | (1L << ('Q' - 'A'))
+                        | (1L << ('C' - 'A'))
```
scottj97, line 610: The comment on line 604 above needs updating

scottj97 (changes requested):

riscv/csrs.cc
```diff
@@ -618,6 +619,11 @@ reg_t misa_csr_t::dependency(const reg_t val, const char feature, const char dep
 }
 
 bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
+  // the write is ignored if increasing IALIGN would misalign the PC
+  if (!(val & (1L << ('C' - 'A'))) && (state->pc & 2) &&
+      proc->extension_changeable(EXT_ZCA))
+    return false;
```
scottj97: Will this work correctly with `--isa=rv64imafdc`? It seems that `extension_changeable(EXT_ZCA)` will be false in that case, so this would incorrectly allow C to be disabled for that ISA even when the PC is misaligned.

scottj97 (changes requested):

riscv/csrs.cc
liweiwei90: OK.

riscv/csrs.cc
liweiwei90: `extension_changeable(EXT_ZCA) `will be true in this case. 'C' will imply Zca, so`isa->extension_enabled(EXT_ZCA)` and `isa->extension_enabled('C') `will be true. And `isa->extension_enabled(EXT_ZCB) || isa->extension_enabled(EXT_ZCMP) || isa->extension_enabled(EXT_ZCMT)` will be false in this case.

scottj97 (approved):

aswaterman (changes requested):
My concern is that we have complicated logic spread across multiple places.  I tried to build on this PR to keep the logic isolated in the misa-write logic, and if I got it correct, it's definitely simpler.  Can y'all take a look, and if it seems correct (or can be made correct), then can you incorporate it, @liweiwei90?

aswaterman:
Here's my attempted simplification: https://github.com/riscv-software-src/riscv-isa-sim/commit/a883a5839dc1f09e5fc8aa80e563f4c841c4a4bc

aswaterman:
@liweiwei90 @scottj97 I believe I've fixed the problems with my proposed version and have pushed to the https://github.com/riscv-software-src/riscv-isa-sim/tree/plctlab-plct-zce-fix2 branch again; see https://github.com/riscv-software-src/riscv-isa-sim/commit/c790cfd3b7e87f69b6c43018f1ccdbf188188d49

This version fixes the other Zc* extensions.  The logic is that they can only be enabled if they are statically enabled, Zca is dynamically enabled, and any other extension they depend upon is dynamically enabled.

This version also keeps the logic in one place, and simplifies usage of processor_t:: set_extension_enable by having its implementation mask the argument with isa_t::extension_enabled.

Despite the annoyance of the specification and the somewhat limited benefit, I'm getting comfortable with this version being simple enough that it isn't too much of a burden to continue supporting.

aswaterman (approved):

aswaterman:
@scottj97 can you re-review?

riscv/processor.h
```diff
@@ -234,8 +234,12 @@ class processor_t : public abstract_device_t
   bool extension_enabled_const(isa_extension_t ext) const {
     if (ext >= 'A' && ext <= 'Z')
       return state.misa->extension_enabled_const(ext);
-    else
-      return isa->extension_enabled(ext);  // assume this can't change
+    else {
+      return isa->extension_enabled(ext);
+    }
```
scottj97, line 239: The goal of `extension_enabled_const()` (versus plain `extension_enabled()`) is to document and enforce an assumption that an extension will not be dynamically changing during the life of the target program. A future change (such as this one, which allows some extensions to be newly disabled) could invalidate those assumptions, and without `extension_enabled_const()`, could be subtly broken.

When `--isa=rv64imafdc`, following a `csrw misa.C=0` it looks like `extension_enabled_const(EXT_ZCA)` will return `true`, even though this extension is not actually enabled anymore. It should always either fire an assertion or return the same value as `extension_enabled()`.

After this change, since `extension_enabled(EXT_ZCA)` can change during the program, we want `extension_enabled_const(EXT_ZCA)` to fire an assertion -- at least when C is enabled in the ISA (and I think it's okay to even when it's not).

Unfortunately I don't see a robust and easy way to do this without duplicating some of the logic in `misa_csr_t::unlogged_write()` about which extensions can be dynamically disabled.

We could keep track of which extensions have been assumed const by recording that in a new `bitset`, and keep track of which extensions have been dynamically disabled by recording that in another new `bitset`, and then assert in both `extension_enabled_const()` and `set_extension_enable()` by checking those new variables. That would be robust but it seems overly complicated.

scottj97 (changes requested):
I think this should be squashed/rebased into two commits:

1. Enable dynamic disabling of extensions (beyond single-letter ones already supported via `misa`) -- adds `extension_enable_table` and `set_extension_enable()`
2. Enable `misa.C` toggling

Edit: also please update the description of this PR to match the new behavior

riscv/processor.h
scottj97: I've implemented this in [c9448044](https://github.com/scottj97/riscv-isa-sim/commit/c9448044bd96fe6a5666d68f90137fcbceb90a57) and it's not as bad as I thought.

The new assertions fire because in processor.cc we have this line:
```
  if (proc->extension_enabled_const(EXT_ZCMT))
    csrmap[CSR_JVT] = jvt = std::make_shared<jvt_csr_t>(proc, CSR_JVT, 0);
```

Which needs to be addressed first. If Zcmt is disabled (because `misa.C==0`) should the JVT CSR still be accessible? If yes, this should be changed to `extension_enabled()`. If not, then in addition to that change, `jvt_csr_t::verify_permissions()` needs to consider `extension_enabled(EXT_ZCMT)`.

aswaterman: Nice job with c9448044bd96fe6a5666d68f90137fcbceb90a57: the dynamic approach is actually quite clean.

Re: `Zcmt`, our only choices are to make `misa.C` read-only when Zcmt is present, or make the change you suggested.  It's more in keeping with this PR's approach to do the latter.  I'll open a new PR since I don't have write permissions on this one.

liweiwei90: OK. I'll close this PR.

PR #1177, opened 2022-12-14
Support mcontrol6 trigger

YenHaoChen:
This PR implements mcontrol6 trigger in Section 5.5.12, debug spec.

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -234,6 +234,50 @@ mcontrol_common_t::match_t mcontrol_common_t::legalize_match(reg_t val) const no
   }
 }
 
+reg_t mcontrol6_t::tdata1_read(const processor_t * const proc) const noexcept {
+  unsigned xlen = proc->get_const_xlen();
+  reg_t tdata1 = 0;
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_TYPE(xlen), 6);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_DMODE(xlen), dmode);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_VS, proc->extension_enabled('H') ? vs : 0);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_VU, proc->extension_enabled('H') ? vu : 0);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_HIT, hit);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_SELECT, select);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_TIMING, timing);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_ACTION, action);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_CHAIN, chain);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_MATCH, match);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_M, m);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_S, s);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_U, u);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_EXECUTE, execute);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_STORE, store);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_LOAD, load);
+  return tdata1;
+}
+
+void mcontrol6_t::tdata1_write(processor_t * const proc, const reg_t val, const bool allow_chain) noexcept {
+  auto xlen = proc->get_const_xlen();
+  assert(get_field(val, CSR_MCONTROL6_TYPE(xlen)) == CSR_TDATA1_TYPE_MCONTROL6);
+  dmode = get_field(val, CSR_MCONTROL6_DMODE(xlen));
+  vs = get_field(val, CSR_MCONTROL6_VS);
+  vu = get_field(val, CSR_MCONTROL6_VU);
+  hit = get_field(val, CSR_MCONTROL6_HIT);
+  select = get_field(val, CSR_MCONTROL6_SELECT);
+  timing = get_field(val, CSR_MCONTROL6_TIMING);
+  action = legalize_action(get_field(val, CSR_MCONTROL6_ACTION));
+  chain = allow_chain ? get_field(val, CSR_MCONTROL6_CHAIN) : 0;
+  match = legalize_match(get_field(val, CSR_MCONTROL6_MATCH));
+  m = get_field(val, CSR_MCONTROL6_M);
+  s = proc->extension_enabled_const('S') ? get_field(val, CSR_MCONTROL6_S) : 0;
+  u = proc->extension_enabled_const('U') ? get_field(val, CSR_MCONTROL6_U) : 0;
+  execute = get_field(val, CSR_MCONTROL6_EXECUTE);
+  store = get_field(val, CSR_MCONTROL6_STORE);
+  load = get_field(val, CSR_MCONTROL6_LOAD);
+  if (execute)
+    timing = 0;
+}
+
 reg_t itrigger_t::tdata1_read(const processor_t * const proc) const noexcept
 {
   auto xlen = proc->get_xlen();
@@ -378,6 +422,7 @@ bool module_t::tdata1_write(unsigned index, const reg_t val) noexcept
     case CSR_TDATA1_TYPE_MCONTROL: triggers[index] = new mcontrol_t(); break;
     case CSR_TDATA1_TYPE_ITRIGGER: triggers[index] = new itrigger_t(); break;
     case CSR_TDATA1_TYPE_ETRIGGER: triggers[index] = new etrigger_t(); break;
+    case CSR_TDATA1_TYPE_MCONTROL6: triggers[index] = new mcontrol6_t(); break;
     default: triggers[index] = new disabled_trigger_t(); break;
   }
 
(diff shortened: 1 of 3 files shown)


riscv/triggers.cc
```diff
@@ -195,9 +195,10 @@ std::optional<match_result_t> mcontrol_t::detect_memory_access_match(processor_t
       (operation == triggers::OPERATION_STORE && !store) ||
       (operation == triggers::OPERATION_LOAD && !load) ||
       (state->prv == PRV_M && !m) ||
-      (state->prv == PRV_S && !s) ||
-      (state->prv == PRV_U && !u) ||
-      (state->v)) {
+      (!state->v && state->prv == PRV_S && !s) ||
+      (!state->v && state->prv == PRV_U && !u) ||
+      (state->v && state->prv == PRV_S && !vs) ||
+      (state->v && state->prv == PRV_U && !vu)) {
```
aswaterman: Maybe cleaner as:

```
      (state->prv == PRV_S && !(state->v ? vs : s)) ||
      (state->prv == PRV_U && !(state->v ? vu : u)) {
```

riscv/triggers.h
```diff
@@ -224,6 +224,12 @@ class mcontrol_t : public trigger_t {
   bool load = false;
 };
 
+class mcontrol6_t : public mcontrol_t {
```
scottj97: This is suspicious. Inheritance should define an _isa_ relationship. Is an mcontrol6 object also an mcontrol object? No. This violates the Liskov substitution principle.

How about first extracting an `mcontrol_common_t` with the fields and logic common to both, and then adding a new subclass for `mcontrol6_t`?

riscv/triggers.cc
```diff
@@ -314,10 +314,8 @@ std::optional<match_result_t> itrigger_t::detect_trap_match(processor_t * const
...
   state_t * const state = proc->get_state();
   if ((state->prv == PRV_M && !m) ||
-      (!state->v && state->prv == PRV_S && !s) ||
-      (!state->v && state->prv == PRV_U && !u) ||
-      (state->v && state->prv == PRV_S && !vs) ||
-      (state->v && state->prv == PRV_U && !vu)) {
+      (state->prv == PRV_S && !(state->v ? vs : s)) ||
+      (state->prv == PRV_U && !(state->v ? vu : u))) {
     return std::nullopt;
   }
```
scottj97: Please DRY up this duplicate code before changing it in two places at once. I'm embarassed that I let this copypasta get past my review of #1128. `itrigger_t::detect_trap_match()` and `etrigger_t::detect_trap_match()` differ by only one `if()` expression.

It might be best to move m/s/u/vs/vu all up to `trigger_t`, and for those trigger types that don't use those bits, their constructors would set them to appropriate fixed values.

scottj97 (changes requested):

riscv/triggers.h
YenHaoChen: Agree

riscv/triggers.cc
YenHaoChen: Add a trigger_t::mode_match() function for privilege mode checking.

riscv/triggers.cc
```diff
@@ -151,6 +151,8 @@ void mcontrol_t::tdata1_write(processor_t * const proc, const reg_t val, const b
       match = MATCH_EQUAL;
       break;
   }
+  vs = 0;
+  vu = 0;
```
scottj97: No need for these two lines; they are both initialized by the constructor and will never change.

riscv/triggers.h
```diff
@@ -89,6 +89,12 @@ class trigger_t {
   action_t legalize_action(reg_t val) const noexcept;
   reg_t tdata2;
 
+  bool vs;
+  bool vu;
+  bool m;
+  bool s;
+  bool u;
```
scottj97: These should be explicitly initialized as `= false;` like the `mcontrol_t` ones were

riscv/triggers.h
```diff
@@ -83,6 +83,7 @@ class trigger_t {
   virtual std::optional<match_result_t> detect_memory_access_match(processor_t UNUSED * const proc,
       operation_t UNUSED operation, reg_t UNUSED address, std::optional<reg_t> UNUSED data) noexcept { return std::nullopt; }
   virtual std::optional<match_result_t> detect_trap_match(processor_t UNUSED * const proc, const trap_t UNUSED & t) noexcept { return std::nullopt; }
+  bool mode_match(state_t * const state) const noexcept;
```
scottj97: This can be `protected`, yes?

riscv/triggers.cc
```diff
@@ -234,6 +234,62 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
...
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_VU, proc->extension_enabled('H') ? vu : 0);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_HIT, hit);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_SELECT, select);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_TIMING, timing);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_ACTION, action);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_CHAIN, chain);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_MATCH, match);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_M, m);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_S, s);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_U, u);
```
scottj97, line 252: Don't these need to consider `proc->extension_enabled_const('S')`, like `mcontrol_t` does?

riscv/triggers.cc
```diff
@@ -234,6 +234,62 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
...
+    case MATCH_NAPOT:
+    case MATCH_GE:
+    case MATCH_LT:
+    case MATCH_MASK_LOW:
+    case MATCH_MASK_HIGH:
+      match = (match_t)match_value;
+      break;
+    default:
+      match = MATCH_EQUAL;
+  }
```
scottj97: Create a `legalize_match()` in `mcontrol_common_t` instead of copy/pasting this paragraph from `mcontrol_t`

riscv/triggers.cc
```diff
@@ -155,7 +155,7 @@ void mcontrol_t::tdata1_write(processor_t * const proc, const reg_t val, const b
     case MATCH_LT:
     case MATCH_MASK_LOW:
     case MATCH_MASK_HIGH:
-      match = (triggers::mcontrol_t::match_t) match_value;
+      match = (match_t) match_value;
```
scottj97: In 05de1277500b770a9fb2f83896f00b596f4b51d0 I don't know what "indicator of variables" means...how about "remove redundant namespace qualifiers"

riscv/triggers.cc
```diff
@@ -261,12 +268,8 @@ void itrigger_t::tdata1_write(processor_t * const proc, const reg_t val, const b
 
 std::optional<match_result_t> itrigger_t::detect_trap_match(processor_t * const proc, const trap_t& t) noexcept
```
scottj97, line 313: Please file another PR to remove the copypasta between `itrigger_t::detect_trap_match()` and `etrigger_t::detect_trap_match()`

scottj97 (changes requested):

riscv/triggers.h
YenHaoChen: Should we move the mode_match() checking to module_t along with textra_match()?
```auto result = trigger->textra_match(proc) ? trigger->detect_trap_match(proc, t) : std::nullopt;```
->
```auto result = (trigger->mode_match(proc->get_state()) && trigger->textra_match(proc)) ? trigger->detect_trap_match(proc, t) : std::nullopt;```

File: riscv/triggers.cc (hunk 7 above)
Comment (YenHaoChen) [reply riscv/triggers.cc:252]: We check the ```proc->extension_enabled_const('S')``` in tdata1_write() but not in tdata1_read().

File: riscv/triggers.cc (hunk 10 above)
Comment (YenHaoChen) [reply riscv/triggers.cc:313]: We can move those copypasta to ```module_t::detect_trap_match()``` and update the function of ```itrigger_t``` and ```etrigger_t``` to 
```
... itrigger_t::detect_trap_match(bool interrupt, reg_t bit)
{
  assert(bit < xlen);
  if (interrupt && ((bit == 0 && nmi) || ((tdata2 >> bit) & 1))) {
    hit = true;
    return match_result_t(TIMING_AFTER, action);
  }
  return std::nullopt;
}
... etrigger_t::detect_trap_match(bool interrupt, reg_t bit)
{
  assert(bit < xlen);
  if (!interrupt && ((tdata2 >> bit) & 1)) {
    hit = true;
    return match_result_t(TIMING_AFTER, action);
  }
  return std::nullopt;
}
```
How do you feel about this?

riscv/triggers.h
scottj97: You could even put it into `textra_match()` (and rename it to `mode_and_textra_match()`) since it's needed in both places that `textra_match()` is called.

But let's do that as a new PR.

riscv/triggers.cc
scottj97: Still too much copypasta. The only difference here is the condition in the `if()`, so that should be a virtual function returning a bool. A parent class (`trap_common_t`?) has the `detect_trap_match()` method that calls this virtual function.

But again, new PR.

scottj97 (approved):

riscv/triggers.h
YenHaoChen: OK will do

riscv/triggers.cc
YenHaoChen: OK will do

riscv/triggers.h
YenHaoChen: After adding trap_common_t, I think move textra_match to detect_??_match() functions makes more sense.
(https://github.com/riscv-software-src/riscv-isa-sim/pull/1207)

riscv/triggers.cc
YenHaoChen: Add trap_common_t in PR (https://github.com/riscv-software-src/riscv-isa-sim/pull/1207).

PR #1180, opened 2022-12-15
Pull mmu.h out of cfg.h

jerryz123:
My reasoning is that `cfg.h` should not pull in headers defining simulator components, it should just contain definitions of config config-related classes. 

Now cfg.h contains endianness_t, cfg_arg_t, mem_cfg_t, and cfg_t

### DIFF
--- a/fesvr/dtm.cc
+++ b/fesvr/dtm.cc
@@ -44,9 +44,6 @@
 #define CLEAR 3
 #define CSRRx(type, dst, csr, src) (0x73 | ((type) << 12) | ((dst) << 7) | ((src) << 15) | (uint32_t)((csr) << 20))
 
-#define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
-#define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
-
 #define RUN_AC_OR_DIE(a, b, c, d, e) { \
     uint32_t cmderr = run_abstract_command(a, b, c, d, e);      \
     if (cmderr) {                                               \
--- a/fesvr/elfloader.cc
+++ b/fesvr/elfloader.cc
@@ -98,7 +98,7 @@ std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* 
   } while (0)
 
   if (IS_ELFLE(*eh64)) {
-    if (memif->get_target_endianness() != memif_endianness_little) {
+    if (memif->get_target_endianness() != endianness_little) {
       throw std::invalid_argument("Specified ELF is little endian, but system uses a big-endian memory system. Rerun without --big-endian");
     }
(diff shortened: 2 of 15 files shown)


fesvr/dtm.cc
```diff
@@ -44,9 +44,6 @@
 #define CLEAR 3
 #define CSRRx(type, dst, csr, src) (0x73 | ((type) << 12) | ((dst) << 7) | ((src) << 15) | (uint32_t)((csr) << 20))
 
-#define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
-#define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
-
```
aswaterman, line 49: convenient we can get rid of these

aswaterman (approved):

aswaterman:
Uh oh, this breaks on Mac.  I'll diagnose it and PR a fix.

scottj97:
I wish I'd had time to review this. 4d4159e76d61980e300cea7bdbdab873f07db12a fails compilation,

And so does 9f93b98c87cfdbcbeb59b9da298344c840f747a4! How did this pass CI?

```
In file included from /local_home/sjohnson/spike-regress/riscv-isa-sim/fesvr/memif.h:10,                                                                                   
                 from /local_home/sjohnson/spike-regress/riscv-isa-sim/fesvr/htif.h:6,                                                                                     
                 from /local_home/sjohnson/spike-regress/riscv-isa-sim/fesvr/htif.cc:3:                                                                                    
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/cfg.h:56:20: error: ‘vector’ in namespace ‘std’ does not name a template type                                       
         const std::vector<mem_cfg_t> &default_mem_layout,                                                                                                                 
                    ^~~~~~                                                                                                                                                 
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/cfg.h:56:15: note: ‘std::vector’ is defined in header ‘<vector>’; did you forget to ‘#include <vector>’?            
```

jerryz123:
> I wish I'd had time to review this. https://github.com/riscv-software-src/riscv-isa-sim/commit/4d4159e76d61980e300cea7bdbdab873f07db12a fails compilation,

Sorry about that, I checked that the intermediate commits compiled on my system (Ubuntu 20.04). 

Although I agree its surprising the `#include <vector>` bug didn't produce an error

aswaterman:
In the past, I've encountered situations where different C++ STL implementations will fail to expose a missing `#include`, because the STL internals are causing it to be included.  Changing GCC versions, for example, might expose this.

I haven't been manually trying to compile the individual commits before merging.  If CI says it's OK, and I don't spot a glaring problem in an intermediate commit, I've just been running with it.  If we want to strictly enforce this policy, we should find a way to automate it.  But the intermediate commits seem to be a red herring in this case.

PR #1181, opened 2022-12-15
Print better error message when requesting >1024 harts

aswaterman:
The debug module imposes this restriction and was enforcing it with an assertion.  But assertions should be used for internal checks only, not for input validation.


aswaterman:
Yikes, I mismanaged my tabs and clicked the merge button on the wrong PR.  To prevent this in the future, I've enabled the feature to require PR reviews before merging (administrator override is still allowed, though).

PR #1182, opened 2022-12-15
Fix build broken by #1170 and #1180

aswaterman:
It looks like #1170 and #1180 conflict, so they individually pass CI but collectively don't.  Trivially fix by #including vector.


jerryz123 (approved):
Oops

PR #1183, opened 2022-12-15
Add github actions test on macos

jerryz123:
Maybe one day this will run on aarch64.


aswaterman (approved):
Thanks.  Even though it's still x86 for now, it's still a useful sanity check.

PR #1184, opened 2022-12-16
Support using sim.h/processor.h as public interfaces

### LINKED ISSUE #1168: Avoiding dependencies on config.h

jerryz123:
This is the final set of patches towards enabling usage of sim.h and processor.h as public interfaces. Resolves #1168.

The primary danger with this change is dropping `config.h` off the include list in a file which references one of those variables, as the bug might only appear when a specific feature is enabled or disabled. To avoid this, I first add `#include "config.h"` directly to any source file which references any variables in `config.h`

The rest of these patches fix the installed headers list, change to relative include paths, and finally remove the config.h dependencies out of processor.h and sim.h. A new test in the `test-spike` script is also added.

### DIFF
--- a/ci-tests/test-spike
+++ b/ci-tests/test-spike
@@ -20,3 +20,7 @@ cd run
 wget https://github.com/riscv-software-src/riscv-isa-sim/releases/download/dummy-tag-for-ci-storage/spike-ci.tar
 tar xf spike-ci.tar
 time ../install/bin/spike --isa=rv64gc pk hello | grep "Hello, world!  Pi is approximately 3.141588."
+
+# check that including sim.h in an external project works
+g++ -std=c++17 -I../install/include -L../install/lib $DIR/testlib.c -lriscv -o test-libriscv
+LD_LIBRARY_PATH=../install/lib ./test-libriscv | grep "Hello, world!  Pi is approximately 3.141588."
--- a/ci-tests/testlib.c
+++ b/ci-tests/testlib.c
@@ -1,0 +1,54 @@
+#include <riscv/sim.h>
+
+// Copied from spike main.
+// TODO: This should really be provided in libriscv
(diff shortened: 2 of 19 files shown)


scottj97 (changes requested):
Commit 1885ad54dfb7a70c32e753ab3784d713ae71ead6 fails compilation:
```
In file included from /local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/jtag_dtm.cc:5:                                                                               
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/debug_module.h:159:10: error: ‘vector’ in namespace ‘std’ does not name a template type                            
     std::vector<hart_debug_state_t> hart_state;                                                                                                                          
          ^~~~~~                                                                                                                                                          
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/debug_module.h:159:5: note: ‘std::vector’ is defined in header ‘<vector>’; did you forget to ‘#include <vector>’?  
/local_home/sjohnson/spike-regress/riscv-isa-sim/riscv/debug_module.h:8:1:                                                                                                
+#include <vector>                                                                                                                                                        
                                                                                                                                                                          ```

jerryz123:
> Commit [1885ad5](https://github.com/riscv-software-src/riscv-isa-sim/commit/1885ad54dfb7a70c32e753ab3784d713ae71ead6) fails compilation:
> 

Thanks, fixed

scottj97 (approved):
I have not examined the changes thoroughly but I have tested for regressions and found none.

aswaterman (approved):
LGTM.  Thank you for adding a CI test.

jerryz123:
Thanks @scottj97 and @aswaterman for the quick reviews on these.

PR #1187, opened 2022-12-19
Allow reads/writes to reserved CLINT regions

jerryz123:
This more closely matches actual hardware implementations


scottj97:
Can you describe the behavior change in a little more detail?

I guess it was taking an access fault previously on these reserved addresses, and now it will simply drop all writes and return 0 on reads. Yes?

jerryz123:
Yes, that is correct

scottj97:
> Yes, that is correct

Please explain that in the commit message.

This is good to go after that.

scottj97 (approved):

aswaterman (approved):
Thanks.  This closely matches a similar change I recently made to the PLIC model.  I probably should've handled the CLINT at the same time.

PR #1191, opened 2022-12-21
Support histogram feature without configure option

aswaterman:
This builds upon #1189 in the obvious way.

There's no perf impact when not histogramming.  Initially, there was a modest slowdown when histogramming, because we moved it to the slow path.  7831b3e5 removes some of this overhead by speeding up the slow path.

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -152,9 +152,7 @@ static void commit_log_print_insn(processor_t *p, reg_t pc, insn_t insn)
 
 inline void processor_t::update_histogram(reg_t UNUSED pc)
 {
-#ifdef RISCV_ENABLE_HISTOGRAM
   pc_histogram[pc]++;
-#endif
 }
 
 // These two functions are expected to be inlined by the compiler separately in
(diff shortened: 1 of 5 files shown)


riscv/execute.cc
```diff
@@ -261,7 +261,7 @@ void processor_t::step(size_t n)
             state.single_step = state.STEP_STEPPED;
           }
 
-          insn_fetch_t fetch = mmu->load_insn(pc);
+          insn_fetch_t fetch = _mmu->access_icache(pc)->data;
```
jerryz123: I think this breaks if you enable/disable logs or histogramming partway through execution, since the icache would end up containing a mix of `fast`/`logged` funcs.

jerryz123: Unrelated, perhaps: What the point of `_mmu` here?

aswaterman: I realize it's a risky functional change for other reasons, so let's just drop it.

aswaterman: Reading `_mmu` at the top of this function allows it to be register-allocated, rather than re-loading it after calls across which the compiler can't prove it hasn't been modified.  It's just one load per iteration, but it makes a difference on the fast path (and probably only a slight difference here).

scottj97 (approved):
👍

jerryz123 (approved):

scottj97 (changes requested):
Please hold off on merging. I've hit some kind of incremental build issue that might be related to this PR. I'm investigating.

aswaterman:
Ack. Feel free to merge if you determine this one is innocent.

scottj97 (approved):
Incremental build issue solved and fixed in #1193

PR #1192, opened 2022-12-21
Slightly improve histogram feature

aswaterman:
In creating #1191, I noticed the histogram prints in order of PC, rather than order of frequency.  This isn't terribly useful, so swap that around.  After doing so, we can switch to `unordered_map` without a functional change, for a significant speedup.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -75,8 +75,12 @@ processor_t::~processor_t()
 #ifdef RISCV_ENABLE_HISTOGRAM
   if (histogram_enabled)
   {
-    fprintf(stderr, "PC Histogram size:%zu\n", pc_histogram.size());
-    for (auto it : pc_histogram)
+    std::vector<std::pair<reg_t, uint64_t>> ordered_histo(pc_histogram.begin(), pc_histogram.end());
+    std::sort(ordered_histo.begin(), ordered_histo.end(),
+              [](auto& lhs, auto& rhs) { return lhs.second < rhs.second; });
+
+    fprintf(stderr, "PC Histogram size:%zu\n", ordered_histo.size());
+    for (auto it : ordered_histo)
       fprintf(stderr, "%0" PRIx64 " %" PRIu64 "\n", it.first, it.second);
   }
 #endif
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -295,7 +295,7 @@ private:
   std::vector<bool> impl_table;
 
   std::vector<insn_desc_t> instructions;
-  std::map<reg_t,uint64_t> pc_histogram;
+  std::unordered_map<reg_t,uint64_t> pc_histogram;
 
   static const size_t OPCODE_CACHE_SIZE = 8191;
   insn_desc_t opcode_cache[OPCODE_CACHE_SIZE];


scottj97 (approved):
👍

jerryz123 (approved):

scottj97 (changes requested):
Please hold off on merging. I've hit some kind of incremental build issue that might be related to this PR. I'm investigating.

aswaterman:
I could imagine the churn related to the instruction template changes in #1189 and #1190 might be to blame. Please merge if you determine this one is innocent.

scottj97:
Maybe you can make some sense of what I've found.

| Clean build first on:  | Then incremental build on: | Outcome |
| ---------------- | --------------------- | ------- |
| #1189 (3cb3d10f058a9ceec6ec93f2aba4c49e2eaa58c1) | #1190 (edcf2d59b7019798f5c2ab303734883936c0b947) | ok      |
| #1189 (3cb3d10f058a9ceec6ec93f2aba4c49e2eaa58c1) | #1191 (b5f68dc55c2b310d24db35a3c7df1ad86509ab61) | ok      |
| #1189 (3cb3d10f058a9ceec6ec93f2aba4c49e2eaa58c1) | #1192 (d6e5a0ed89f1eb918f7a01e9fc7cdf7e71d6d791) | BROKEN! |
| #1190 (edcf2d59b7019798f5c2ab303734883936c0b947) | #1189 (3cb3d10f058a9ceec6ec93f2aba4c49e2eaa58c1) | ok      |
| #1190 (edcf2d59b7019798f5c2ab303734883936c0b947) | #1191 (b5f68dc55c2b310d24db35a3c7df1ad86509ab61) | ok      |
| #1190 (edcf2d59b7019798f5c2ab303734883936c0b947) | #1192 (d6e5a0ed89f1eb918f7a01e9fc7cdf7e71d6d791) | ok      |
| #1191 (b5f68dc55c2b310d24db35a3c7df1ad86509ab61) | #1189 (3cb3d10f058a9ceec6ec93f2aba4c49e2eaa58c1) | ok      |
| #1191 (b5f68dc55c2b310d24db35a3c7df1ad86509ab61) | #1192 (d6e5a0ed89f1eb918f7a01e9fc7cdf7e71d6d791) | ok      |
| #1192 (d6e5a0ed89f1eb918f7a01e9fc7cdf7e71d6d791) | #1189 (3cb3d10f058a9ceec6ec93f2aba4c49e2eaa58c1) | BROKEN! |
| #1192 (d6e5a0ed89f1eb918f7a01e9fc7cdf7e71d6d791) | #1190 (edcf2d59b7019798f5c2ab303734883936c0b947) | ok      |

"Broken" means that `vsetvli` (and possibly any vector instruction?) core dumps.

I haven't tried going to/from any older points before #1189 was merged.

scottj97:
Another interesting data point: starting with a clean build from #1189, then incremental to #1190, then incremental to #1192, it's broken only once I reach #1192.

aswaterman:
I can't make sense of it from that data.  I'll try to replicate your results so I can play around with the problem.

aswaterman:
OK, at least I can confirm that I can replicate it.  I'll dig into it after my next meeting.

aswaterman:
@scottj97 Figured it out.  It's a bit esoteric.  We're using `ar rcs` to build the archives prior to linking, which preserves and adds to the archive if it already existed.  This usually works by coincidence because our object files don't change names frequently, so they'll still overwrite their predecessors within the archive.  But because #1189 made new names for some of the objects, we now might end up with multiple objects in the archive with functions of the same name.

In the event of multiple objects satisfying a dependence at link time, I'm not sure which one gets chosen (alphabetical? order added to archive? arbitrary?) but the point is, the wrong version of the function is getting linked in.  From there, it's obvious how the result might be a segfault.

The fix is to delete the archive before building it.  Unfortunately, we're stuck with this broken intermediate state in our tree, but so be it.

I'll PR that fix before merging any other PRs.

scottj97:
> frequently, so they'll still overwrite their predecessors within the archive. But because #1189 made new names for some of the objects, we now might end up with multiple objects in the archive with functions of the same name.

But #1190 changes the file names back to the same as before, yes?

If so, as long as I always include both #1189 and #1190, or neither #1189 nor #1190, then I shouldn't have an issue?

aswaterman:
@scottj97 Right, that passes the smell test.

scottj97 (approved):
Incremental build issue solved and fixed in #1193

PR #1193, opened 2022-12-22
Fix build nondeterminism by deleting archives before recreating

aswaterman:
Otherwise, `ar rcs` will add to the previous archive, creating the possibility of multiple functions with the same name in the archive. The linker might not choose the most recent version, resulting in undefined behavior.


aswaterman:
See discussion on #1192

aswaterman:
FYI @jerryz123 - you didn't break this; you just revealed a latent bug in the build system.

scottj97 (approved):

PR #1194, opened 2022-12-22
Remove dead code in Makefiles

aswaterman:
The primary change is that headers that aren't installed don't need to be explicitly listed.  This was already just dead code, because dependences are auto-generated as `.d` files.


scottj97 (approved):
I don't really understand it, but it doesn't seem to hurt anything.

aswaterman:
Legacy cruft is all it is.

PR #1195, opened 2022-12-23
Expose access type in simif_t mmio_load interface

jerryz123:
This is useful for alternative implementations of simif_t, which may want to provide different behaviors for fetch/load/store

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -89,7 +89,7 @@ tlb_entry_t mmu_t::fetch_slow_path(reg_t vaddr)
     if (auto host_addr = sim->addr_to_mem(paddr)) {
       result = refill_tlb(vaddr, paddr, host_addr, FETCH);
     } else {
-      if (!mmio_load(paddr, sizeof fetch_temp, (uint8_t*)&fetch_temp))
+      if (!mmio_fetch(paddr, sizeof fetch_temp, (uint8_t*)&fetch_temp))
         throw trap_instruction_access_fault(proc->state.v, vaddr, 0, 0);
       result = {(char*)&fetch_temp - vaddr, paddr - vaddr};
     }
@@ -135,6 +135,14 @@ bool mmu_t::mmio_ok(reg_t addr, access_type UNUSED type)
     return false;
 
   return true;
+}
+
+bool mmu_t::mmio_fetch(reg_t addr, size_t len, uint8_t* bytes)
+{
+  if (!mmio_ok(addr, FETCH))
+    return false;
+
+  return sim->mmio_fetch(addr, len, bytes);
 }
 
 bool mmu_t::mmio_load(reg_t addr, size_t len, uint8_t* bytes)
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -347,6 +347,7 @@ private:
   void load_slow_path_intrapage(reg_t addr, reg_t len, uint8_t* bytes, uint32_t xlate_flags);
   void store_slow_path(reg_t addr, reg_t len, const uint8_t* bytes, uint32_t xlate_flags, bool actually_store, bool require_alignment);
   void store_slow_path_intrapage(reg_t addr, reg_t len, const uint8_t* bytes, uint32_t xlate_flags, bool actually_store);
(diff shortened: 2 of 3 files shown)


aswaterman (commented):
I haven't carefully considered the merits, but instead of adding a parameter, might it instead make sense to define `mmu_t::mmio_fetch` and `simif_t::mmio_fetch`?  And `sim_t`'s implementation of `mmio_fetch` could simply be a call to `mmio_load`.

scottj97:
Is there really any use case for this? Would any real memory-mapped IO device wish to do something different with a fetch than a load? Until 10 years ago, most bus protocols didn't even distinguish the two.

Based on the principle of YAGNI, I'd say this is undesirable.

jerryz123:
The use case is interfacing `processor_t` with a RTL simulation of the uncore. In order to avoid sending every load/store through the uncore, I'm experimenting with an alternative `simif_t` that implements an actual cache (not just a cache model). In this implementation, `addr_to_mem` always returns null, so everything goes through `mmio_load/store`, which is where it interfaces with the cache.

I prefer to implement a separate icache and dcache, so I'd like simif_t to differentiate loads from fetches.

scottj97:
An interface for "memory-mapped I/O" is not the right place to implement caches. You might hack something up that meets your purposes but at the very least it would be misnamed then.

Are you planning to PR the change to `addr_to_mem` here also? Otherwise, unless there's a use case for someone implementing an actual memory-mapped IO device, this PR adds no value.

(Unrelated to this PR, are you planning to have your new icache & dcache models exactly match the RTL behavior? If yes, you're gonna have a bad time. If no, I don't see the point? Edit: I think I get it...you want to model the uncore in RTL and send realistic memory requests to it based on the target program running in Spike.)

jerryz123:
Typed all this up before I saw your edit... I'll leave it so we can be on the same page:

I want spike's processor_t to interact with a RTL simulation of a SoC via DPI calls. Thus, instead of `sim_t : simif_t`, I have in a separate repo `soc_interface_t : simif_t`. In this use case, the mechanism for accessing memory is fundamentally different from `sim_t`, as `soc_interface_t` interfaces with a RTL simulation of a SoC, instead of modelling any memory or bus devices itself.

The way I understand the simif_t interface, is that `addr_to_mem` determines which addresses can be accessed directly via `memcpy`, and which addresses should be accessed through an alternative channel (`mmio_load/mmio_store`). I can't memcpy between `processor_t` and the RTL simulation world, so I want all memory accesses to go through an alternative channel. I abuse `addr_to_mem` in my `soc_interface_t` to always return nullptr, so `mmio_load/mmio_store` interacts with the SoC via DPI.

While this works (and does not require any changes to Spike), it incurs an extreme performance penalty, as every fetch/load/mem must go out through the RTL simulation. The solution is the same for an actual core: caches.
So now I have implemented caches in `soc_interface_t` such that cache hits don't need to interact with the RTL simulation. This leads to the original proposal.

I agree that the naming of some methods doesn't really make sense in my use case, and that this might be abusing the functionality in simif_t. I don't think I need changes beyond this make what I'm doing work; I already have some of the caching working correctly and interfacing with the RTL.

jerryz123:
> (Unrelated to this PR, are you planning to have your new icache & dcache models exactly match the RTL behavior? If yes, you're gonna have a bad time. If no, I don't see the point? Edit: I think I get it...you want to model the uncore in RTL and send realistic memory requests to it based on the target program running in Spike.)

No, the spike-interface is not meant to match performance of any RTL, It should be able to generate reasonable realistic memory traffic.

scottj97:
OK, I see the utility. I like @aswaterman's idea, but how about having a default `virtual mmio_fetch()` in `simif_t`? Then clients who don't care about the distinction won't need to change.

aswaterman:
Just catching up on this thread.  I also think @jerryz123's use case is defensible, and given the minor complexity cost, I support allowing this.

@scottj97 when I wrote my earlier comment, I waffled on whether to define the method in `simif_t` or `sim_t`, but I agree with your reasoning.

scottj97 (approved):

aswaterman (approved):

PR #1197, opened 2022-12-27
Prevent processor_t from retiring instructions after a WFI

jerryz123:
Previously, after a WFI, the next call to `proc->step(n)` would always retire another instruction. With this PR, no instructions will retire until a pending interrupt is seen, regardless of how many times `proc->step(n)` is called.

The primary motivation for this change is for use-cases where `proc->step(n)` is called with small `n`.

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -242,6 +242,10 @@ void processor_t::step(size_t n)
     try
     {
       take_pending_interrupt();
+      if (unlikely(in_wfi)) {
+        // No unmasked pending interrupt, so remain in wfi
+        throw wait_for_interrupt_t();
+      }
 
       if (unlikely(slow_path()))
       {
@@ -321,6 +325,7 @@ void processor_t::step(size_t n)
       // allows us to switch to other threads only once per idle loop in case
       // there is activity.
       n = ++instret;
+      in_wfi = true;
     }
 
     state.minstret->bump(instret);
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -35,6 +35,7 @@ processor_t::processor_t(const isa_parser_t *isa, const char* varch,
   : debug(false), halt_request(HR_NONE), isa(isa), sim(sim), id(id), xlen(0),
   histogram_enabled(false), log_commits_enabled(false),
   log_file(log_file), sout_(sout_.rdbuf()), halt_on_reset(halt_on_reset),
+  in_wfi(false),
   impl_table(256, false), last_pc(1), executions(1), TM(4)
 {
   VU.p = this;
@@ -626,6 +627,9 @@ void processor_t::take_interrupt(reg_t pending_interrupts)
   if (!pending_interrupts) {
     return;
   }
+
+  // Exit WFI if there are any pending interrupts
+  in_wfi = false;
 
   // M-ints have higher priority over HS-ints and VS-ints
   const reg_t mie = get_field(state.mstatus->read(), MSTATUS_MIE);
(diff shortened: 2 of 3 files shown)


aswaterman (changes requested):
I don’t think this is right, since you must exit WFI-state if any locally unmasked interrupt is pending, even if the interrupt is globally masked, preventing the interrupt from actually being taken. On systems without AIA, that means the exit condition is (mip & mie) != 0.

The old scheme was correct using a very conservative algorithm: don’t ever truly halt on WFI. I like this for its simplicity and would prefer to keep it unless you have a compelling reason to do the stricter thing.

jerryz123:
Ok. If its preferable to keep the `step()` loop simpler, then I have an alternative approach that addresses my use case here: https://github.com/riscv-software-src/riscv-isa-sim/pull/1198

sifiverobert:
@aswaterman If I am not mistaken debug spec assures WFI is 'terminated' on halt request and STEP is always working.
As far as simulating WFI, correct simulation should IMO suspend execution of a program until next interrupt (what means forever if there is no interrupts). From other side legal implementation of WFI may be just NOP. So probably it is OK for Spike to treat WFI as NOP (for now ...). Full simulation of WFI may not be so easy as we think.

aswaterman:
@jerryz123 Let’s see what @scottj97 thinks about a corrected version of this PR vs. #1198 (which is obviously harmless but might be useful to fewer people). I’m coming around to this version, but I’d like another opinion.

@sifiverobert it isn’t even clear what “full simulation” would be, since there’s such a breadth of legal behaviors. If most HW developers are implementing WFI in the strictest way, then perhaps that argues for doing so in Spike, even though the NOP-like approach is fully legal.

sifiverobert:
Agree about 'full simulation' .... I am not sure how WFI (or simulation of interrupts ...) was handled before, but it seems this PR changed it. It means that some older program may produce different PC-flow what may create problems ...

jerryz123:
> On systems without AIA, that means the exit condition is (mip & mie) != 0.

In this implementation, nonzero `mip & mie` would set `in_wfi = false` in `take_interrupt`, so I believe the behavior should still respect the exit condition.

jerryz123:
Another advantage of this approach is that it avoids inconsistent behavior between
`step(2)` and `step(1); step(1);`

aswaterman (approved):
Yes, you’re right; I misread the condition under which you were clearing in_wfi. I now believe this is correct, and I think the implementation is clean enough. Let’s still wait for @scottj97 on this one.

sifiverobert:
> Another advantage of this approach is that it avoids inconsistent behavior between `step(2)` and `step(1); step(1);`

This is certainly good thing! 'Step(a); Step(b);' should work the same as 'Step(a+b)'.

riscv/execute.cc
```diff
@@ -242,6 +242,10 @@ void processor_t::step(size_t n)
     try
     {
       take_pending_interrupt();
+      if (unlikely(in_wfi)) {
+        // No pending interrupt is taken, so remain in wfi
```
aswaterman: Change comment to “no unmasked pending interrupt, so remain in wfi” or similar?

scottj97:
I don't have a strong opinion on this.

I won't be able to test any of these PRs until next week, but no need to wait for me.

aswaterman (approved):
In that case, let's run with it.

timsifive:
This change breaks external debug through OpenOCD. Looks like spike hangs the first time it executes the `wfi` at https://github.com/riscv-software-src/riscv-isa-sim/blob/master/debug_rom/debug_rom.S#L42.

jerryz123:
Sorry, Tim. Does #1209 fix it?

PR #1198, opened 2022-12-27
Add method to determine if processor is wfi

jerryz123:
Alternative to https://github.com/riscv-software-src/riscv-isa-sim/pull/1197, avoids changing simulator behavior

With this method, an alternative platform can determine when it is safe to not call `proc->step()` by checking `is_waiting_for_interrupt()` along with that processor's `mip` and `mie` CSRs.


aswaterman:
Tentatively closing in favor of #1197

PR #1200, opened 2022-12-28
Support accessing PTEs through mmio_load/mmio_store

jerryz123:
Previously, PTE loads/stores were always performed through direct access to host memory regions. This PR enables accessing PTEs in regions through `mmio_load/mmio_store`

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -42,7 +42,7 @@ void mmu_t::flush_tlb()
   flush_icache();
 }
 
-static void throw_access_exception(bool virt, reg_t addr, access_type type)
+void throw_access_exception(bool virt, reg_t addr, access_type type)
 {
   switch (type) {
     case FETCH: throw trap_instruction_access_fault(virt, addr, 0, 0);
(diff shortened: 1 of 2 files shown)


aswaterman:
Note that the command-line handling for A/D bits will have to change next year, when we add the extension that dynamically controls availability this feature.  But that's OK--making this dynamic is a step in the right direction anyway.

riscv/mmu.cc
```diff
@@ -548,3 +535,50 @@ void mmu_t::register_memtracer(memtracer_t* t)
   flush_tlb();
   tracer.hook(t);
 }
+
+reg_t mmu_t::pte_load(reg_t pte_paddr, reg_t addr, bool virt, access_type trap_type, int ptesize) {
+  if (!pmp_ok(pte_paddr, ptesize, LOAD, PRV_S))
+    throw_access_exception(virt, addr, trap_type);
+
+  void* host_pte_paddr = sim->addr_to_mem(pte_paddr);
+  if (ptesize == 4) {
```
aswaterman: Would it reduce code duplication here to make `pte_load` type-parameterized on the PTE data type, and going back to the old scheme where `mmu_t::walk` writes something like `vm.ptesize == 4 ? pte_load<uint32_t>(...) : pte_load<uint64_t>(...)`?

jerryz123: :+1: The templated approach avoids some repetition

aswaterman: cool

scottj97 (changes requested):
4cbd76957e6c2ecc23c5161cf80daa2f0574859d should be a separate PR, and should be several commits -- especially the (apparently?) whitespace changes around line 20 of `testlib.c` which obscure the substantive changes.

jerryz123:
Sure. I'll move those changes to a follow-up PR.

scottj97 (approved):

aswaterman (approved):

PR #1201, opened 2022-12-29
Use 'C' in misa for PC-alignment mask calculation

jerryz123:
https://github.com/riscv/riscv-isa-manual/blob/cb3b9d1dcdacefbde6602ada7a0050f5c723ddee/src/machine.tex#L1969-L1973 states that misa.C should control the PC-alignment mask.

The regression was originally introduced in 0adf9307eaef62402c4368d33e88bbb5e1211653

This fixes the behavior on the `rv64mi-p-ma_fetch` test.

### DIFF
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -236,7 +236,7 @@ public:
     return impl_table[impl];
   }
   reg_t pc_alignment_mask() {
-    return ~(reg_t)(extension_enabled(EXT_ZCA) ? 0 : 2);
+    return ~(reg_t)(extension_enabled('C') ? 0 : 2);
   }
   void check_pc_alignment(reg_t pc) {
     if (unlikely(pc & ~pc_alignment_mask()))


aswaterman:
I agree there's at least one bug, but it's more complex.  IALIGN can be influenced by more than just misa.C.  For example, an RV32I_Zca implementation has IALIGN=16 even though it has misa.C=0.  That's what 0adf9307eaef62402c4368d33e88bbb5e1211653 was driving at.

Furthermore, there might be implementations where misa.C is writable but IALIGN=16 regardless of the setting of misa.C.

So, here's where we are now:
- Spike allows `misa.C` to be written, but writing `misa.C` doesn't do anything useful in Spike.  It's not clear to me this is incorrect, since the spec doesn't mandate that disabling C increase IALIGN, nor does it mandate disabling C disable the sub-extensions like Zca.  But I think we all agree it violates the principle of least surprise.
- The test erroneously assumes that clearing `misa.C` increases IALIGN, which isn't guaranteed.  The test contains a comment to this effect, so it wasn't unanticipated: https://github.com/riscv-software-src/riscv-tests/commit/8f9f628b84826b7fbd1d91b24c56f634cef91cf4#diff-df63643b133702ae723a9d7fc4d91c9f59a8309b2ca86aeeb73f3b492f06206dR111-R112

I propose we fix the test by deleting all of the subtests that assume misa.C=0 implies IALIGN=32.

Although Spike is possibly correct as-is, I propose we make its behavior cleaner by making misa.C read-only.  (At this point, I'd support making Spike's misa completely read-only, but that is a topic for another day.) This change, if we can agree on it, would supersede this PR.

What do you think @jerryz123?

jerryz123:
Disallowing writable misa.C would make spike's behavior more intuitive if disabling C does not disable Zca/Zcf/Zcd. 

We could avoid losing the test cases if we make the test check first whether or not disabling C does indeed set IALIGN=32.

aswaterman:
@scottj97 Your thoughts?

scottj97:
IIUC @aswaterman you're proposing that `misa.C` is not writable, and is 1 iff C is configured in the ISA string; IALIGN is independent of `misa.C` and is 16 whenever any of C/Zca/Zcf/Zcd are configured.

I'm good with this, but let's make the code in `pc_alignment_mask()` a little more explicit about it, by first setting a variable called `ialign` to 16 or 32.

aswaterman:
@scottj97 Exactly.  I'll prototype it.

PR #1203, opened 2022-12-29
Make misa.C read-only

aswaterman:
This resolves the issue discussed in https://github.com/riscv-software-src/riscv-isa-sim/pull/1201.

Prior to https://github.com/riscv-software-src/riscv-isa-sim/commit/0adf9307eaef62402c4368d33e88bbb5e1211653, clearing misa.C would disable compressed instructions and increase IALIGN to 32.  Afterwards, clearing misa.C had essentially no effect because Zca and friends would stay enabled.  While AFAICS this isn't technically incorrect, it certainly doesn't follow the principle of least surprise.

Instead, remove the feature to toggle misa.C.  The effect is that misa.C is 1 iff C is included in the ISA string, and IALIGN is independent of misa.C: specifically, IALIGN is 16 iff Zca is present. (And of course C implies Zca.)

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -601,13 +601,12 @@ bool sstatus_csr_t::enabled(const reg_t which) {
 misa_csr_t::misa_csr_t(processor_t* const proc, const reg_t addr, const reg_t max_isa):
   basic_csr_t(proc, addr, max_isa),
   max_isa(max_isa),
-  write_mask(max_isa & (0  // allow MAFDQCHV bits in MISA to be modified
+  write_mask(max_isa & (0  // allow MAFDQHV bits in MISA to be modified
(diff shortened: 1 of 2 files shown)


riscv/csrs.cc
```diff
@@ -607,7 +607,6 @@ misa_csr_t::misa_csr_t(processor_t* const proc, const reg_t addr, const reg_t ma
                         | (1L << ('F' - 'A'))
                         | (1L << ('D' - 'A'))
                         | (1L << ('Q' - 'A'))
-                        | (1L << ('C' - 'A'))
```
scottj97, line 610: The comment a few lines up needs updating too

aswaterman: Ack.

scottj97 (approved):

PR #1204, opened 2022-12-29
Support processor_t::step without incrementing mcycle

jerryz123:
The use case is in an environment where I want mcycle to reflect uncore simulation cycles, rather than instructions retired.


aswaterman:
Any chance that you could get away without modifying `processor_t::step` and instead, in your environment, call `step(1)` then overwrite `mcycle` to the value you want?

jerryz123:
Yeah that approach works too.

PR #1205, opened 2022-12-29
Add method to probe which memory regions are reservable

jerryz123:
This PR adds `simif_t::reservable`, enabling specification of custom reservable regions by alternate implementations of `simif_t`.

The behavior for spike-main is unchanged.

### DIFF
--- a/riscv/simif.h
+++ b/riscv/simif.h
@@ -10,15 +10,15 @@ class simif_t
 {
 public:
   // should return NULL for MMIO addresses
-  virtual char* addr_to_mem(reg_t addr) = 0;
+  virtual char* addr_to_mem(reg_t paddr) = 0;
   // used for MMIO addresses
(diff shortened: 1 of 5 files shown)


aswaterman (approved):
Seems reasonable to me.

riscv/simif.h
```diff
@@ -11,6 +11,7 @@ class simif_t
 public:
   // should return NULL for MMIO addresses
   virtual char* addr_to_mem(reg_t addr) = 0;
+  virtual bool reservable(reg_t addr) { return addr_to_mem(addr); };
```
scottj97: The param should be named `paddr`, and remove extraneous semicolon at end of line

scottj97 (approved):

riscv/simif.h
jerryz123: To maintain consistency, I've changed all of the simif_t member functions to use `paddr` instead of `addr`.

PR #1206, opened 2022-12-30
Control whether misaligned accesses are supported via command line

aswaterman:
Command-line option `--misaligned` now controls whether misaligned accesses raise exceptions.

The configure argument `--enable-misaligned` has been removed.

### DIFF
--- a/ci-tests/testlib.c
+++ b/ci-tests/testlib.c
@@ -21,6 +21,7 @@ int main()
 	    "rv64gcv",
 	    "MSU",
 	    "vlen:128,elen:64",
+	    false,
 	    endianness_little,
 	    16,
 	    mem_cfg,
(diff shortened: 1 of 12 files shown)


scottj97 (commented):
I'm not sure about keeping the configure option. We didn't do that for the recent changes that removed `--enable-histogram` and `--enable-commitlog`. And I think the help for configure is pretty misleading now:
```
  --enable-misaligned     Enable hardware support for misaligned loads and                                                              
                          stores                                                                                                        
```
It's not clear this is only changing a runtime flag's default value now.

I think it just adds extra complexity for minimal benefit.

If there is benefit to controlling the default, how about making it a `--with-` type option, such as the existing `--with-isa` and `--with-priv` flags?

aswaterman:
I don't object to getting rid of it.

aswaterman:
OK, done.

scottj97 (approved):

PR #1207, opened 2022-12-30
triggers: refactor: add trap_common_t

YenHaoChen:
This PR depends on https://github.com/riscv-software-src/riscv-isa-sim/pull/1177.
I will rebase this PR after merging https://github.com/riscv-software-src/riscv-isa-sim/pull/1177

This PR adds trap_common_t to clean up itrigger and etrigger.

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -370,11 +370,16 @@ std::optional<match_result_t> etrigger_t::detect_trap_match(processor_t * const 
   bool interrupt = (t.cause() & ((reg_t)1 << (xlen - 1))) != 0;
   reg_t bit = t.cause() & ~((reg_t)1 << (xlen - 1));
   assert(bit < xlen);
-  if (!interrupt && ((tdata2 >> bit) & 1)) {
+  if (simple_match(interrupt, bit)) {
     hit = true;
     return match_result_t(TIMING_AFTER, action);
   }
   return std::nullopt;
+}
+
+bool etrigger_t::simple_match(bool interrupt, reg_t bit) const
+{
+  return !interrupt && ((tdata2 >> bit) & 1);
 }
 
 module_t::module_t(unsigned count) : triggers(count) {
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -310,7 +310,7 @@ void itrigger_t::tdata1_write(processor_t * const proc, const reg_t val, const b
   action = legalize_action(get_field(val, CSR_ITRIGGER_ACTION));
 }
 
-std::optional<match_result_t> itrigger_t::detect_trap_match(processor_t * const proc, const trap_t& t) noexcept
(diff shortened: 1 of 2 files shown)


riscv/triggers.cc
```diff
@@ -190,7 +190,8 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
   if ((operation == triggers::OPERATION_EXECUTE && !execute) ||
       (operation == triggers::OPERATION_STORE && !store) ||
       (operation == triggers::OPERATION_LOAD && !load) ||
-      !mode_match(proc->get_state())) {
+      !mode_match(proc->get_state()) ||
+      !textra_match(proc)) {
```
scottj97: Why not combine these two methods? They are always called together.

scottj97 (changes requested):
Please break up 21b8a8f3286cc213a22b6eaaee913fa0d98fcb59 into smaller commits. First create an empty parent `trap_common_t` and then one-by-one move code/vars into the parent.

The commit message for b7a5882da2b533388128dc79d025b5d7012ae53f has an incorrect method name.

riscv/triggers.cc
YenHaoChen: I prefer to separate the ```mode_match()``` and ```textra_match()``` for the following reasons.
1. The ```(!mode_match(proc->get_state()) || !textra_match(proc))``` clarifies two independent checkings and, thus, has a higher readability than ```(!mode_and_textra_match(proc))```.
2. The current separated version of ```mode_match()``` is more elegant than ```mode_and_textra_match()```.
> bool trigger_t::mode_match(state_t * const state) const noexcept
> {
>   switch (state->prv) {
>     case PRV_M: return m;
>     case PRV_S: return state->v ? vs : s;
>     case PRV_U: return state->v ? vu : u;
>     default: assert(false);
>   }
> }

The ```mode_and_textra_match()``` will go with something like:

>   switch (state->prv) {
>     case PRV_M: if (!m) return false; break;
>     case PRV_S: if (!(state->v ? vs : s)) return false; break;
>     case PRV_U: if (!(state->v ? vu : u)) return false; break;
>     default: assert(false);
>   }
>   ...

3. The ```textra_match()``` is already a structurally complicated function. I want to keep it conceptually simple, i.e., checking textra CSR.

With the above concerns, how about adding a new function ```trigger_t::common_match()```, which performs common checks across all triggers? (prototyped in https://github.com/riscv-software-src/riscv-isa-sim/pull/1207/commits/590a36e487dc9bbb8d2c13bc4fd7e9af40cff9ad)
```
bool trigger_t::common_match(processor_t * const proc) const noexcept 
{ return mode_match(proc->get_state()) && textra_match(proc); }
```

scottj97: I'm impressed with your reasoning. Let's leave it like it is.

scottj97 (approved):

PR #1208, opened 2023-01-02
Data trigger breaks AMO atomicity

YenHaoChen:
The data trigger breaks AMO's atomicity. (https://github.com/riscv-software-src/riscv-isa-sim/pull/1164)

This PR aims to separate the trigger checking and load/store operation to guarantee the atomicity.

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -39,6 +39,7 @@ struct tlb_entry_t {
 };
 
 void throw_access_exception(bool virt, reg_t addr, access_type type);
+reg_t reg_from_bytes(size_t len, const uint8_t* bytes);
 
 // this class implements a processor's port into the virtual memory system.
 // an MMU and instruction cache are maintained for simulator performance.
@@ -138,7 +139,9 @@ public:
       if (likely(tlb_load_hit)) {
         res = *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr);
       } else {
-        load_slow_path(addr, sizeof(T), (uint8_t*)&res, 0);
+        check_triggers(triggers::OPERATION_LOAD, addr);
+        load_slow_path_intrapage(addr, sizeof(T), (uint8_t*)&res, 0);
+        check_triggers(triggers::OPERATION_LOAD, addr, reg_from_bytes(sizeof(T), (uint8_t*)&res));
       }
 
       if (unlikely(proc && proc->get_log_commits_enabled()))
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -143,9 +143,6 @@ public:
         load_slow_path_intrapage(addr, sizeof(T), (uint8_t*)&res, 0);
       }
 
-      if (unlikely(proc && proc->get_log_commits_enabled()))
-        proc->state.log_mem_read.push_back(std::make_tuple(addr, 0, sizeof(T)));
-
       auto lhs = from_target(res);
       bool tlb_store_hit = tlb_store_tag[vpn % TLB_ENTRIES] == vpn;
 
@@ -156,8 +153,10 @@ public:
         store_slow_path_intrapage(addr, sizeof(T), (const uint8_t*)&target_val, 0, true);
       }
 
-      if (unlikely(proc && proc->get_log_commits_enabled()))
+      if (unlikely(proc && proc->get_log_commits_enabled())) {
+        proc->state.log_mem_read.push_back(std::make_tuple(addr, 0, sizeof(T)));
         proc->state.log_mem_write.push_back(std::make_tuple(addr, (T)f(lhs), sizeof(T)));
+      }
 
       if (unlikely(!tlb_load_hit))
         check_triggers(triggers::OPERATION_LOAD, addr, reg_from_bytes(sizeof(T), (uint8_t*)&res));
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -148,7 +148,18 @@ public:
         proc->state.log_mem_read.push_back(std::make_tuple(addr, 0, sizeof(T)));
 
       auto lhs = from_target(res);
-      store<T>(addr, f(lhs));
+      bool tlb_store_hit = tlb_store_tag[vpn % TLB_ENTRIES] == vpn;
+
+      if (likely(tlb_store_hit)) {
+        *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr) = to_target((T)f(lhs));
+      } else {
+        target_endian<T> target_val = to_target((T)f(lhs));
+        store_slow_path(addr, sizeof(T), (const uint8_t*)&target_val, 0, true, false);
+      }
+
+      if (unlikely(proc && proc->get_log_commits_enabled()))
+        proc->state.log_mem_write.push_back(std::make_tuple(addr, (T)f(lhs), sizeof(T)));
+
       return lhs;
     })
   }
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -135,6 +135,7 @@ public:
       target_endian<T> res;
       reg_t vpn = addr >> PGSHIFT;
       bool tlb_load_hit = tlb_load_tag[vpn % TLB_ENTRIES] == vpn;
+      bool tlb_store_hit = tlb_store_tag[vpn % TLB_ENTRIES] == vpn;
 
       if (likely(tlb_load_hit)) {
         res = *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr);
@@ -144,7 +145,6 @@ public:
       }
 
       auto lhs = from_target(res);
-      bool tlb_store_hit = tlb_store_tag[vpn % TLB_ENTRIES] == vpn;
 
       if (likely(tlb_store_hit)) {
         *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr) = to_target((T)f(lhs));
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -131,21 +131,21 @@ public:
   template<typename T, typename op>
   T amo(reg_t addr, op f) {
     convert_load_traps_to_store_traps({
-      store_slow_path(addr, sizeof(T), nullptr, 0, false, true);
-      target_endian<T> res;
+      store_slow_path(addr, sizeof(T), nullptr, 0, false, true); // check misaligned, access fault, and page fault
       reg_t vpn = addr >> PGSHIFT;
       bool tlb_load_hit = tlb_load_tag[vpn % TLB_ENTRIES] == vpn;
       bool tlb_store_hit = tlb_store_tag[vpn % TLB_ENTRIES] == vpn;
+      if (unlikely(!tlb_load_hit))
+        check_triggers(triggers::OPERATION_LOAD, addr);
 
+      // Atomic load and store. Should not throw exception nor trigger action.
+      target_endian<T> res;
       if (likely(tlb_load_hit)) {
         res = *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr);
       } else {
-        check_triggers(triggers::OPERATION_LOAD, addr);
         load_slow_path_intrapage(addr, sizeof(T), (uint8_t*)&res, 0);
       }
-
       auto lhs = from_target(res);
-
       if (likely(tlb_store_hit)) {
         *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr) = to_target((T)f(lhs));
       } else {
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -141,7 +141,6 @@ public:
       } else {
         check_triggers(triggers::OPERATION_LOAD, addr);
         load_slow_path_intrapage(addr, sizeof(T), (uint8_t*)&res, 0);
-        check_triggers(triggers::OPERATION_LOAD, addr, reg_from_bytes(sizeof(T), (uint8_t*)&res));
       }
 
       if (unlikely(proc && proc->get_log_commits_enabled()))
@@ -154,12 +153,18 @@ public:
         *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr) = to_target((T)f(lhs));
       } else {
         target_endian<T> target_val = to_target((T)f(lhs));
-        check_triggers(triggers::OPERATION_STORE, addr, reg_from_bytes(sizeof(T), (const uint8_t*)&target_val));
         store_slow_path_intrapage(addr, sizeof(T), (const uint8_t*)&target_val, 0, true);
       }
 
       if (unlikely(proc && proc->get_log_commits_enabled()))
         proc->state.log_mem_write.push_back(std::make_tuple(addr, (T)f(lhs), sizeof(T)));
+
+      if (unlikely(!tlb_load_hit))
+        check_triggers(triggers::OPERATION_LOAD, addr, reg_from_bytes(sizeof(T), (uint8_t*)&res));
+      if (unlikely(!tlb_store_hit)) {
+        target_endian<T> target_val = to_target((T)f(lhs));
+        check_triggers(triggers::OPERATION_STORE, addr, reg_from_bytes(sizeof(T), (const uint8_t*)&target_val));
+      }
 
       return lhs;
     })
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -131,7 +131,20 @@ public:
   T amo(reg_t addr, op f) {
     convert_load_traps_to_store_traps({
       store_slow_path(addr, sizeof(T), nullptr, 0, false, true);
-      auto lhs = load<T>(addr);
+      target_endian<T> res;
+      reg_t vpn = addr >> PGSHIFT;
+      bool tlb_load_hit = tlb_load_tag[vpn % TLB_ENTRIES] == vpn;
+
+      if (likely(tlb_load_hit)) {
+        res = *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr);
+      } else {
+        load_slow_path(addr, sizeof(T), (uint8_t*)&res, 0);
+      }
+
+      if (unlikely(proc && proc->get_log_commits_enabled()))
+        proc->state.log_mem_read.push_back(std::make_tuple(addr, 0, sizeof(T)));
+
+      auto lhs = from_target(res);
       store<T>(addr, f(lhs));
       return lhs;
     })
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -158,12 +158,16 @@ public:
         proc->state.log_mem_write.push_back(std::make_tuple(addr, (T)f(lhs), sizeof(T)));
       }
 
-      if (unlikely(!tlb_load_hit))
-        check_triggers(triggers::OPERATION_LOAD, addr, reg_from_bytes(sizeof(T), (uint8_t*)&res));
-      if (unlikely(!tlb_store_hit)) {
-        target_endian<T> target_val = to_target((T)f(lhs));
-        check_triggers(triggers::OPERATION_STORE, addr, reg_from_bytes(sizeof(T), (const uint8_t*)&target_val));
-      }
+      try {
+        if (unlikely(!tlb_load_hit))
+          check_triggers(triggers::OPERATION_LOAD, addr, reg_from_bytes(sizeof(T), (uint8_t*)&res));
+      } catch (triggers::matched_t& t) { /* suppress data trigger firing before AMO to avoid breaking atomicity */ }
+      try {
+        if (unlikely(!tlb_store_hit)) {
+          target_endian<T> target_val = to_target((T)f(lhs));
+          check_triggers(triggers::OPERATION_STORE, addr, reg_from_bytes(sizeof(T), (const uint8_t*)&target_val));
+        }
+      } catch (triggers::matched_t& t) { /* suppress data trigger firing before AMO to avoid breaking atomicity */ }
 
       return lhs;
     })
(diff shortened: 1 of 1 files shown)


YenHaoChen:
I tried to give a prototype for dedicated ```slow_path```-style function for AMOs (discussed in https://github.com/riscv-software-src/riscv-isa-sim/pull/1164).
The actual fix is the conver from trigger before to trigger after in https://github.com/riscv-software-src/riscv-isa-sim/pull/1208/commits/ce3a688f012eaae592b78ed39a517c3094c695cd. 

I am not sure whether to simply convert the trigger in original load/store functions. Or, we can merge the ```convert_load_traps_to_store_traps(BODY)``` and ```convert_trigger_before_to_trigger_after(BODY)``` into ```convert_amo_traps_and_triggers(BODY)```.

scottj97 (commented):
Is a6927724c126dac31f52eb7284c8281437d253dc really only a refactor, as its commit message says? It looks to me like it will change the behavior, in the case of a load data trigger hit. But I don't have good test coverage of these kinds of triggers so I'm not sure.

riscv/mmu.h
```diff
@@ -126,6 +126,16 @@ class mmu_t
       throw trap_store_guest_page_fault(t.get_tval(), t.get_tval2(), t.get_tinst()); \
     }
 
+  #define convert_trigger_before_to_trigger_after(BODY) \
```
scottj97: I see no benefit to the macro here. It's only used once, and BODY is only used once.

riscv/mmu.h
```diff
@@ -157,12 +167,14 @@ class mmu_t
         proc->state.log_mem_write.push_back(std::make_tuple(addr, (T)f(lhs), sizeof(T)));
       }
 
-      if (unlikely(!tlb_load_hit))
-        check_triggers(triggers::OPERATION_LOAD, addr, reg_from_bytes(sizeof(T), (uint8_t*)&res));
-      if (unlikely(!tlb_store_hit)) {
-        target_endian<T> target_val = to_target((T)f(lhs));
-        check_triggers(triggers::OPERATION_STORE, addr, reg_from_bytes(sizeof(T), (const uint8_t*)&target_val));
-      }
+      convert_trigger_before_to_trigger_after({
```
scottj97: I don't understand why all data triggers on AMOs should be converted to `timing=1 (after)`. This does not seem compatible with the debug spec to me. The spec warns about this case for loads:

> Data load triggers with timing of 0 will result in the same load happening again when the debugger lets the hart run. For data load triggers, debuggers must first attempt to set the breakpoint with timing of 1.

It also says that `timing` is expected to be fixed by the hardware:

> Most hardware will only implement one timing or the other, possibly dependent on select, execute, load, and store. This bit primarily exists for the hardware to communicate to the debugger what will happen. Hardware may implement the bit fully writable, in which case the debugger has a little more control.

So it seems to me that if someone sets a data breakpoint with `timing=0 (before)`, they are taking a risk, and if it results in repeating the AMO, that's their problem.

If we wanted to avoid this footgun, we should have mcontrol triggers force `timing=1 (after)` when `select=1 (data)`.

scottj97 (changes requested):

scottj97:
I wonder if there's an easier way to fix this using the `actually_store` parameter to `store_slow_path`. AMOs call `store_slow_path(actually_store=false)` before they do any load, in order to check for store permissions (see #976).

Currently `store_slow_path` does not check any triggers if `actually_store=false`. Would it be correct for it to check only address triggers in that case? I suppose that would change the priority between load address triggers and store address triggers, so it's probably incorrect. But it seems 90% of the way there with only a very small change.

riscv/mmu.h
YenHaoChen: Agree that legalizing tdata1.timing works better.
I will try to implement a legalize_timing() function based on Table 5.10.
image
(Clarifying https://github.com/riscv/riscv-debug-spec/issues/792)

riscv/mmu.h
YenHaoChen: Attempt to legalize tdata1.timing for load data trigger instead of converting behavior on firing.

YenHaoChen:
There are four issues with the trigger on AMO.
1. The data trigger breaks the AMO atomicity of load and store operations.
2. The address trigger has a wrong priority to address misaligned and page faults.
3. AMO cannot fire on a chain with a load-only address trigger and a store-only address trigger.
4. AMO cannot fire on a chain with a load-only data trigger and a store-only data trigger.

Checking triggers if ```actually_store=false``` in ```store_slow_path``` can only correct point 2. The method in https://github.com/riscv-software-src/riscv-isa-sim/pull/1164 can correct point 2, 3, and 4.

This PR aims at point 1, where the data trigger checking in load<T>()/store<T>() breaks the AMO atomicity. The https://github.com/riscv-software-src/riscv-isa-sim/pull/1164 discussed using a dedicated ```slow_path```-style function for AMOs, and this PR aims to prototype the idea. To avoid breaking AMO atomicity, data trigger in store() must fire to after the AMO instruction because the load() has already been performed. Additionally, although data trigger in load() should fire after the instruction by the specification, the current implementation allows firing data trigger before the instruction, which also breaks AMO atomicity.

I propose to make the following changes to resolve point 1 before reopening https://github.com/riscv-software-src/riscv-isa-sim/pull/1164 for the others.
1. Legalize tdata1.timing=1 (after) for load data trigger. 
2. Catch store data trigger with timing=0 (before) and conver to timing=1 (after) in amo().

@scottj97 How do you feel about this?

scottj97:
> I propose to make the following changes to resolve point 1 before reopening #1164 for the others.
> 
> 1. Legalize tdata1.timing=1 (after) for load data trigger.
> 2. Catch store data trigger with timing=0 (before) and conver to timing=1 (after) in amo().

I don't think it's right to convert timing inside amo(). The debugger will read `tdata1` and see that timing=0 and it will expect the exception to be before.

I'm not sure what is wrong with an AMO with a store data trigger with timing=0? As long as the store doesn't happen, the AMO is still correct, yes? When the breakpoint trap handler disables the trigger and returns, it will execute correctly.

YenHaoChen:
The comment (https://github.com/riscv-software-src/riscv-isa-sim/pull/1164#issuecomment-1338868058) worried about the side effects of the load. Spike allows users to add plugin devices and perform AMO on the devices, and users may implement a device with side effects on load operations. So, if AMO completes the load operation, AMO should also fulfill the store operation; otherwise, it violates atomicity.
@aswaterman, do I catch your concern correctly?

scottj97:
> The comment ([#1164 (comment)](https://github.com/riscv-software-src/riscv-isa-sim/pull/1164#issuecomment-1338868058)) worried about the side effects of the load. Spike allows users to add plugin devices and perform AMO on the devices, and users may implement a device with side effects on load operations. So, if AMO completes the load operation, AMO should also fulfill the store operation; otherwise, it violates atomicity.

I think you correctly understand his concern. But I don't agree. If someone sets a load data trigger with timing=0 (before) on a region with load side-effects, then they're asking for trouble, AMO or not. If someone sets a store data trigger with timing=0 (before) on a region with load side-effects, then AMOs will be broken.

@aswaterman do you want the AMO in that case to complete the store and commit, as if timing=1, and then take the breakpoint trap? @timsifive is it okay for a trigger set with timing=0 to occasionally (i.e. for all AMOs) behave as though timing=1?

aswaterman:
We need to address the spec question before tackling the Spike problem.

My initial reaction is that @scottj97 accurately summarized the situation.  A side-effecting load accessing a load-triggered region with `timing`=before is impossible to implement in practice--you won't have the data to action the trigger without having already actioned the side effect.  (I raised this concern in the early debug-spec discussions.)  In my mind, AMOs being broken by store-triggered regions with `timing`=before is something of a corollary.

If we are forced to choose between strict adherence to trigger semantics or accurate implementation of the unprivileged ISA, then the answer is obvious: the act of debugging mustn't alter the unprivileged behavior.  So the AMO must remain atomic, even if the trigger timing isn't respected.

Does that mean we must disallow `timing`=before for systems with side-effecting AMOs?  No, `timing`=before is obviously too useful for plain stores; banning that to rectify the spec with respect to AMOs would be throwing the baby out with the bathwater.  Therefore, the compromises would seem to be weakening `timing` with respect to AMOs (as @scottj97 suggested) or even further ("before" means "before when possible, otherwise after").

scottj97:
> If we are forced to choose between strict adherence to trigger semantics or accurate implementation of the unprivileged ISA, then the answer is obvious: the act of debugging mustn't alter the unprivileged behavior. So the AMO must remain atomic, even if the trigger timing isn't respected.

To me, "atomic" means that no other access can squeeze between the load & store. But if the store never happens, is that a violation of atomicity? I wasn't thinking of it that way, but maybe you're right. However I can't find any kind of promise about this in the "A" standard extension spec.

Is there any other case in the ISA where an AMO could perform a load but not a store? IIRC, Spike before #976 would do so if the store caused an exception but the load did not.

sifiverobert:
I fully agree with @aswaterman here. Debug (and trace ...) MUST NOT slow down execution at all.

It is indeed rare to have 'before' data breakpoints implemented. For OoO cores doing it correctly may be even impossible (=very hard ...) to be done without affecting how pipelines are advancing.

I have however a problem accepting fact, that 'before' may (sometimes ...) work as 'after' and debugger to NOT be aware about it. If implementation CANNOT guarantee 'before', it should NOT allow to set 'timing=0' and revert to 'timing=1'. This would require debugger to check this bit - it may be even behave differently for different memory regions. For example 'before' for MMIO region may be possible, while for RAM it may not be. It is already said, that Atomics are only working/allowed for RAM memories.

IMO Spike should 'model' such a (common sense ...) behavior. So when A is enabled, then RAM areas should NOT allow to choose 'before'.

I am not sure if this is clearly defined, but 'after' may even be implemented with a 'skid' (if my memory serves well, it was the case in early Arm7/9/11).

I am a bit concerned about this statement (especially 'will' word):

Data load triggers with timing of 0 will result in the same load happening again when the debugger lets the hart run.

I do not see a reason to execute a read when there is a read-watchpoint on data address (and not on data value!).
Some devices (like UART RX register ...) may not like second read ...

Description of 'timing' field is speculating different scenarios, but in general 'before' should be (more ...) discouraged.

sifiverobert:
@scottj97 - I think AMO must be AMO ... Reading several times is allowed, but final 'read+modify+write' must be performed together. So I think breaking 'before' AMO (after a read ...) is OK, but only when 'read+modify+write' is re-executed after resume.

PS. I think write must/should be performed even if 'modify' does not change the value ...

scottj97:
> I fully agree with @aswaterman here. Debug (and trace ...) MUST NOT slow down execution at all.

I don't see where anybody here is talking about slowing down execution.

> I have however a problem accepting fact, that 'before' may (sometimes ...) work as 'after' and debugger to NOT be aware about it. If implementation CANNOT guarantee 'before', it should NOT allow to set 'timing=0' and revert to 'timing=1'.

But that is what Andrew is recommending. It sounds like you don't fully agree with him after all?

> It is already said, that Atomics are only working/allowed for RAM memories.

But Andrew is concerned about what happens to AMOs when there are load side effects?

> IMO Spike should 'model' such a (common sense ...) behavior. So when A is enabled, then RAM areas should NOT allow to choose 'before'.

If it's RAM, why is timing=before a problem? Did you mean to say non-RAM areas? But you just said that atomics are only working/allowed for RAM memories? Why are you worried about non-RAM?

I am very confused.

> I am a bit concerned about this statement (especially 'will' word):
> 
> Data load triggers with timing of 0 will result in the same load happening again when the debugger lets the hart run.
> 
> I do not see a reason to execute a read when there is a read-watchpoint on data address (and not on data value!). Some devices (like UART RX register ...) may not like second read ...
> 
> Description of 'timing' field is speculating different scenarios, but in general 'before' should be (more ...) discouraged.

That's why the following sentence in the spec does exactly that! "For data load triggers, debuggers must first attempt to set the breakpoint with timing of 1."

sifiverobert:
@scottj97 Sorry for confusions ... IMO this 'before' is not very clearly defined, that why we have this "broken expectations" and discussions ...

"For data load triggers, debuggers must first attempt to set the breakpoint with timing of 1"

if debugger will follow this and 'after' is implemented, it may never set 'before' (which may also be implemented without side-effects ...). Again I am NOT talking on triggers on VALUES (being read ...), but addresses only!

IMPORTANT: Triggers may be also used to start and stop trace. I may want to start tracing at every read of UART RX and I want to see value being read in trace! For this I may want 'before' trigger, so I can SEE value from UART in trace ...

I also may want to stop trace AFTER write to UART TX registers, so I see value being written in trace ...

So, certainly there is a need for BOTH "before and after" and saying "must first try timing=1" is not exactly what is needed ...

YenHaoChen:
Everyone here agrees that the load data trigger can only be ```timing=1(after)```, and implementation should legalize the ```timing``` if invalid, unsupported, or impossible to achieve. I fired another PR to legalize the timing of the load data trigger in https://github.com/riscv-software-src/riscv-isa-sim/pull/1214.

This PR's remaining issue is about the store data trigger on AMOs. If a store data trigger with ```timing=1(after)``` matches an AMO, the trigger can fire a breakpoint exception without breaking the AMO load/store atomicity. However, if a store data trigger with ```timing=0(before)``` matches an AMO, the trigger may attempt to stop AMO store after the AMO load and breaks the atomicity.

@scottj97 suggested treating this misbehavior as a software bug, and the spec allows an implementation breaking atomicity in this scenario.
@aswaterman preferred to keep AMO atomicity (unprivileged spec) over trigger semantics (debug spec) but still wanted to have the data trigger benefits for debugging AMOs.
@sifiverobert addressed converting ```timing=0(before)``` to ```timing=1(after)``` does not seem right, and @scottj97 also opposed on the convertion.

How about letting store data trigger with ```timing=1(after)``` can fire on AMOs but suppressing store data trigger with ```timing=0(before)``` on AMOs? (The Spike implementation will be converting to no trigger matched.)

riscv/mmu.h
YenHaoChen: PR https://github.com/riscv-software-src/riscv-isa-sim/pull/1214 adds a legalize_timing() for legalizing tdata1.timing of load data trigger.

aswaterman:
> To me, "atomic" means that no other access can squeeze between the load & store. But if the store never happens, is that a violation of atomicity? I wasn't thinking of it that way, but maybe you're right. However I can't find any kind of promise about this in the "A" standard extension spec.

By "atomic" I meant "indivisible", which is perhaps an abuse of terminology in this context.  It isn't a memory-ordering matter; it's a precise-exception matter.  Triggering an AMO's load side effect but not actually completing the AMO is an imprecise exception.  This event is architecturally visible even on a uniprocessor.

My broader point is that a debugger silently causing semantic changes because load side effects might be actioned multiple times is an anti-feature.  This is presumably the part where @sifiverobert was agreeing with me, wrt. his comment about popping the RX FIFO.

aswaterman:
> How about letting store data trigger with `timing=1(after)` can fire on AMOs but suppressing store data trigger with `timing=0(before)` on AMOs? (The Spike implementation will be converting to no trigger matched.)

I am leaning towards going further and redefining `timing`=before to be a request, not a guarantee; and that implementations should not honor the request when doing so would result in additional memory accesses to non-idempotent memory regions.  (There might be other situations where implementations might not honor the request, e.g., because it's too difficult to implement for some operation.)

Obviously, either my proposal or yours is a spec change.  But it's hard to see how we can make forward progress without making _some_ change here.

The recommendation @scottj97 mentioned that "For data load triggers, debuggers must first attempt to set the breakpoint with timing of 1" arguably means @YenHaoChen's suggestion would suffice and mine isn't necessary, but mine has the benefit of allowing implementations to provide "before" semantics only when convenient, whereas otherwise some implementations might just decide to give up and never offer "before" semantics.  The relative merits of these options are obviously debatable.

YenHaoChen:
For load data triggers, we can keep AMO indivisible by legalizing ```timing``` without loss of generality. If an implementation can achieve trace actions without breaking AMO nor additional access to non-idempotent regions, we legalize ```timing=1(after)``` only if ```select=1(data), load=1, action=0,1(breakpoint,debug)```. (No load data trigger can achieve ```timing=0(before)``` and ```action=0,1(breakpoint,debug)``` for AMOs or regular loads.) @sifiverobert Will this work for the trace?

The store data triggers with ```timing=before``` are practically possible and valuable for regular stores. However, when matching an AMO, we have three possible architecturally visible results (candidates).
1. Break AMO to non-idempotent regions. (current implementation)
2. Trigger fire as if ```timing=after```. (prototyped in https://github.com/riscv-software-src/riscv-isa-sim/pull/1208/commits/e12df3767427a2a1ff837e8181609194488524d9)
3. Suppress the trigger from firing.

Is there any other possibility or concerns that I overlook?

aswaterman:
My take: triggering after is more useful than not triggering at all, and breaking the AMO semantics is worse than either alternative.

timsifive:
I agree with @aswaterman. For now spike can simply only implement store timing=after.

I'll see about clarifying in the debug spec what should happen with store timing=before (which is pretty useful) and AMO. Likely it'll go one of the ways suggested here, and then spike can follow suit (or just stick with store timing=after).

pdonahue-ventana:
Copy of [what I just sent](https://lists.riscv.org/g/tech-debug/message/929) to tech-debug:

Perhaps this is why another popular architecture only has address breakpoints, not store value breakpoints.

I wonder if this half-completed AMO to non-idempotent memory is really a problem in practice.  AMOs to I/O space are either not supported at all in an implementation or they are going to be far atomics where the hart doesn't even see the store data (because some I/O device is doing the whole AMO operation and not communicating intermediate values back to the hart).  The only other option that I can think of is the really old-school bus-locking thing which nobody does anymore.

Note that the spec already says "Each AMO instruction is a store for the write portion of the operation. The address is always available to trigger on, although the value stored might not be, depending on the hardware implementation."  That's to accommodate far atomics.  I think that if you have near atomics which are idempotent then you allow an implementation to do the load portion, detect the trigger match, and trap/halt before doing the store.  That doesn't break atomicity.  If you have far atomics (or near atomics to non-idempotent memory via bus locking) then you just suppress firing of the trigger as you're already allowed to do.  Or you just suppress firing of the trigger on AMOs altogether (near or far) since that's also allowed.

The only thing to change in the debug spec is maybe to provide non-normative guidance along those lines.

YenHaoChen:
This issue focuses on the store data trigger to non-idempotent memory, although most accesses are to idempotent memory. The privileged spec does not allow AMO to break on non-idempotent memory; thus, debug spec should not allow so in all scenarios. Unfortunately, AMOs matching the store data trigger of the current debug spec could violate this primary rule.

It is not obvious to me that the current debug spec allows suppressing store data triggers on AMOs. The following sentence expresses that store data triggers may sometimes not get store values. However, the sentence does not mention the allowance of suppression. Thus, the allowance of suppression in the next debug spec version is also a change from my reading.
```Each AMO instruction is a store for the write portion of the operation. The address is always available to trigger on, although the value stored might not be, depending on the hardware implementation.```

Converting to ```timing=after``` has an advantage over suppressing the store data trigger with ```timing=before```. The conversion at least lets the store data trigger with ```timing=before``` to react on matching AMOs for debugging, while the suppression completely silences the store data trigger on AMOs. In other words, suppression is more restricted than conversion.

@aswaterman Do I correctly catch your reasoning logic for the conversion?
@pdonahue-ventana Am I missing anything in the debug spec for the allowance of suppression?

aswaterman:
I believe you did capture my reasoning.

I'll also again suggest that we revisit the definition of the `timing` setting with respect to data triggers.  I think we'd be better off defining `timing` as a preference, rather than a requirement; and we'd be better off recommending that implementations should, whenever there is a conflict, maintain precise exceptions rather than respecting `timing`.

As has been pointed out a few times, Allegedly RISC Machines have abandoned data triggers altogether, and so perhaps we are just spinning our wheels on this pedantic detail.  I'd rather move on to something more important.

pdonahue-ventana:
> I'll also again suggest that we revisit the definition of the timing setting with respect to data triggers. I think we'd be better off defining timing as a preference, rather than a requirement; and we'd be better off recommending that implementations should, whenever there is a conflict, maintain precise exceptions rather than respecting timing.

I don't think that's necessary.  As was pointed out, the spec explicitly says that store data triggers may not be available on AMOs.  See my comments in https://lists.riscv.org/g/tech-debug/message/955 which include (among other things) these three cases about triggers that compare the store data value:
> 1. If you have far atomics then it is impossible for data triggers to fire at all.  So there's no atomicity problem.  There might be a "why didn't the debugger stop when I wrote to this memory location" problem but the hardware simply cannot support that (without somehow transmitting trigger information to an entity that's outside the hart).  The sentence that I quoted from the spec currently allows this behavior.
> 2. If you have near atomics and the access is to idempotent memory then I don't see a problem.  You speculatively read the value from your local cache, you notice the match, and you trap instead of performing the write.  No architectural state is updated so there's no atomicity problem and no surprising behavior since there's no visible effect.
> 3. The only remaining case is if you have near atomics and the access is to non-idempotent memory.  This is like bus locking which I haven't seen since before the turn of the century.  But if an implementation does do this then my suggestion is that the implementation treats this like a far atomic and doesn't fire the trigger (even though it has enough information that it could do so).

Maybe we need a non-normative comment to that effect.  But I don't think that the spec is nonsensical, that it doesn't comply with the A extension, that it breaks atomicity, etc.  As I said, it might not provide the best debug experience but I don't think there's anything to be done about that.  Is a 99% solution to data value triggers better than not having them at all?  If so then live with the inability to trigger on far atomic AMOs.  If not then don't implement or use those triggers.

(One possible problem that I see with case 2 is that if hart A has a reservation and hart B does an AMO then hart A may lose the reservation even if hart B does not actually perform the AMO due to the trigger.  That's architecturally visible and could cause forward progress problems in theory.)

> As has been pointed out a few times, Allegedly RISC Machines have abandoned data triggers altogether, and so perhaps we are just spinning our wheels on this pedantic detail. I'd rather move on to something more important.

And now I understand why they only have address triggers.

aswaterman:
I can get on board with that line of reasoning.

As pertains to Spike, then, what should our plan be?  Don't check data triggers for AMOs to regions other than main memory?

pdonahue-ventana:
Yeah, that's what I'd do.

YenHaoChen:
The debug spec (https://github.com/riscv/riscv-debug-spec/pull/800) decides to suppress data triggers from firing on AMO to avoid breaking atomicity (without implementing the mcontrol6.{uncertain,uncertainen} bits). The commit (https://github.com/riscv-software-src/riscv-isa-sim/pull/1208/commits/4f4b0d6f5679168c2d5ac4beee1fd970ad5a8c6e) provides a prototype for this behavior.

PR #1209, opened 2023-01-03
Fix debug-mode regression introduced by 20e7f53

jerryz123:
Debug mode should break the processor out of wfi

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -203,7 +203,7 @@ static inline reg_t execute_insn_logged(processor_t* p, reg_t pc, insn_fetch_t f
 bool processor_t::slow_path()
 {
   return debug || state.single_step != state.STEP_NONE || state.debug_mode ||
-         log_commits_enabled || histogram_enabled;
+         log_commits_enabled || histogram_enabled || in_wfi;
 }
 
 // fetch/decode/execute loop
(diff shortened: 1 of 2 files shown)


timsifive:
This change improves things from all tests failing to almost all tests failing. It's conceivable that #1177 caused these failures.
I'm going to dig into this more, but if you have time you should be able to run these tests by following the instructions at https://github.com/riscv-software-src/riscv-tests/tree/master/debug.

timsifive:
Just tested some more, and #1177 is *not* the cause of these failures.

aswaterman (approved):

aswaterman:
It's not clear from Tim's feedback whether or not I should merge this, but feel free to merge when you think it's appropriate @jerryz123

jerryz123:
Still investigating.

jerryz123:
@timsifive I believe this should work now.

The use of `wfi` in debug mode as a performance optimization threw me off. Those should be executed as nops.

aswaterman:
Thanks, I’ll wait for Tim’s confirmation before merging.

scottj97 (approved):
I don't have any test coverage of this but the code looks clean and it doesn't seem to break anything else.

aswaterman:
@jerryz123 / @timsifive please merge once this is deemed sufficient.

timsifive (approved):
This makes debug work again. Thanks!

PR #1210, opened 2023-01-04
Control mmu-dirtying via command line

jerryz123:
Command-line option `--mmu-dirty` now controls whether the mmu writes PTE a/d bits.

The configure argument `--enable-dirty` has been removed.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -413,16 +413,17 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
         break;
       } else {
         reg_t ad = PTE_A | ((type == STORE) * PTE_D);
-#ifdef RISCV_ENABLE_DIRTY
-        // set accessed and possibly dirty bits.
+
         if ((pte & ad) != ad) {
-          pte_store(pte_paddr, pte | ad, gva, virt, type, vm.ptesize);
+          if (proc->cfg->dirty_enabled) {
+            // set accessed and possibly dirty bits
+            pte_store(pte_paddr, pte | ad, gva, virt, type, vm.ptesize);
+          } else {
+            // take exception if access or possibly dirty bit is not set.
+            break;
+          }
         }
-#else
-        // take exception if access or possibly dirty bit is not set.
-        if ((pte & ad) != ad)
-          break;
-#endif
+
         reg_t vpn = gpa >> PGSHIFT;
         reg_t page_mask = (reg_t(1) << PGSHIFT) - 1;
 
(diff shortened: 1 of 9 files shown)


riscv/mmu.cc
```diff
@@ -413,16 +415,17 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
         break;
       } else {
         reg_t ad = PTE_A | ((type == STORE) * PTE_D);
-#ifdef RISCV_ENABLE_DIRTY
-        // set accessed and possibly dirty bits.
+
         if ((pte & ad) != ad) {
-          pte_store(pte_paddr, pte | ad, gva, virt, type, vm.ptesize);
+          if (dirty_enabled) {
```
aswaterman: Rather than adding the new `dirty_enabled` member, I think you could just check `proc->cfg->dirty_enabled` here.  (At this point in the code, we know that `proc` is non-null.)

riscv/mmu.h
```diff
@@ -274,14 +274,7 @@ class mmu_t
...
 
-  int is_dirty_enabled()
-  {
-#ifdef RISCV_ENABLE_DIRTY
-    return 1;
-#else
-    return 0;
-#endif
-  }
+  int is_dirty_enabled() { return dirty_enabled; }
```
aswaterman: (and then I guess this would become `proc && proc->cfg->dirty_enabled`)

scottj97 (changes requested):
Can you please do this in steps, like Andrew did for #1206? Otherwise clients will have to make local changes simultaneous with Spike changes. (I.e. pull in new Spike, add spike `--mmu-dirty`, remove configure `--enable-dirty`, all in one atomic commit.) If instead you made this change in three steps:
1. Add `--mmu-dirty` option (that's ignored)
2. Control dirty with `--mmu-dirty` instead of configure-time option
3. Remove configure option

Then clients could update in a safer way in 5 steps:
1. Pull in Spike change (1) above
2. Add `--mmu-dirty` to their `spike` run command line
3. Pull in Spike change (2) above
4. Remove no-longer-needed configure option `--enable-dirty`
5. Pull in Spike change (3) above

And everything is fully working at every step of the way.

This is how I handled #1206 in my own company-internal usage of Spike.

jerryz123:
Makes sense. Removing the configure option is a separate commit now.

scottj97 (changes requested):
f3c5cc76edb400d86d73243c7dd4da3441b5223e does both steps 1 and 2

jerryz123:
Fixed

scottj97 (approved):

aswaterman (approved):

PR #1211, opened 2023-01-04
Only update histogram when histogramming

aswaterman:
This is worth a 1.4x speedup on the slow path (when not histogramming).

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -152,7 +152,8 @@ static void commit_log_print_insn(processor_t *p, reg_t pc, insn_t insn)
 
 inline void processor_t::update_histogram(reg_t UNUSED pc)
 {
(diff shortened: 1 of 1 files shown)


scottj97:
Please rebase...this appears to be based off #1206 and will make a bit of a messy merge history as it is currently.

Edit: nevermind; since your commits are not PGP-signed, I can rebase them myself with no ill consequences.

Edit edit: nope! Since I have github set to require signatures on my commits, letting Github rebase this results in "Unverified" warnings on the new commits created.

So @aswaterman please rebase :)

aswaterman:
Hah, sorry you had to go through that chicanery.  This is one of the situations where GitHub's "rebase and merge" strategy would be helpful, though I do realize it has its downsides.

scottj97 (approved):

PR #1212, opened 2023-01-04
Automatically test that external debug still works.

timsifive:


.github/workflows/debug-smoke.yml
```diff
@@ -0,0 +1,30 @@
...
+    name: Test debug (Ubuntu)
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Dependencies
+        run: sudo xargs apt-get install -y < .github/workflows/apt-packages.txt
+
+      - run: |
+          git clone --depth 1 https://github.com/riscv-software-src/riscv-tests.git
```
aswaterman: We might want to check out a known-working commit (which we can periodically update), else changes to riscv-tests risk preventing Spike PRs from being merged.

timsifive: Sorry if this is getting you email spam. I had hoped that marking this as a draft would avoid that.

Let me see if I can get it working first, and then I'll worry about making it nice.

aswaterman: no worries!

timsifive: Fixed.

aswaterman (approved):
I'm leery of increasing the CI time, but it's better than spending human time tracking down regressions.  We can try this out for a while and see how people feel about it.

timsifive:
What's the cost of CI time to us? I honestly don't know how/where this runs, or who pays for it.

timsifive:
PS
We could save a lot of time by doing these tests as part of the Ubuntu build. Half the time is spent building spike.

aswaterman:
GitHub Actions is free for public repositories (for now, at least...), so the only cost to us is a human one (a few minutes' extra latency to know whether a PR passes CI).  So, I think it's OK to keep these as separate CI tasks for now.

PR #1213, opened 2023-01-05
build: replace `-Wl,--export-dynamic` with `-rdynamic`

compnerd:
Rely on the linker driver to map the flags appropriately for the linker.
This largely should allow us to avoid the `AX_LINKER_FLAGS` check as
this is a driver level flag that should be supported by most compilers
capable of building SPIKE at this point.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -33,6 +33,9 @@
 /* define if the Boost::ASIO library is available */
 #undef HAVE_BOOST_ASIO
 
+/* define if the Boost::Regex library is available */
+#undef HAVE_BOOST_REGEX
+
 /* Dynamic library loading is supported */
 #undef HAVE_DLOPEN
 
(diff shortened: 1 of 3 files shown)


scottj97:
Please provide some kind of explanation or justification for this change, either here in the PR or in the commit messages.

Edit: or maybe this is the entire reason? "This largely should allow us to avoid the `AX_LINKER_FLAGS` check as this is a driver level flag that should be supported by most compilers capable of building SPIKE at this point." Who is "us" and where are we able to avoid this `AX_LINKER_FLAGS` check?

compnerd:
Yes, that is the reason: `-rdynamic` is a driver level flag and allows the linker driver (clang/gcc, cl unfortunately does not serve as a linker driver) to translate the flag to the appropriate flag for the linker in use.  `-Wl,--export-dynamic` is directly passing the flag to the linker bypassing the driver.  I suppose I could've split the change into two PRs, but seemed unnecessary since the first one was just preparatory to reduce the diff in the second change itself.

Us would be the Spike project itself.

scottj97:
You didn't answer the last part of my question: where are we able to avoid this `AX_LINKER_FLAGS` check? Do you have a follow-on PR that will make some change related to that?

compnerd:
Sorry, missed that part.  I need to look up the autoconf macro, but the idea is that this doesn't require a linker check, this is something that we can directly check that the driver supports, but this sets up everything so that we can do that.

PR #1214, opened 2023-01-06
Add legalize_timing() for tdata1.timing

YenHaoChen:
The load data trigger can only be after executing the instruction. This PR adds legalize_timing() to enforce the legal timing value of the load data trigger.

The legalize_timing() implicitly depends on select, execution, load, and store. So, reordering the statement in the tdata1_write() may break the functionality. I need help finding a good way to avoid this and seeking suggestions.

Additionally, I cannot find the reference in debug spec for legalizing timing=0(before) of execution triggers. Although this implementation conforms to the debug spec, should we consider removing the legalization of execution triggers' timing for a more general implementation? (The test suite https://github.com/riscv-software-src/riscv-tests/tree/master/debug is still passed after removing the legalization.)

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -25,8 +25,9 @@ void trigger_t::tdata2_write(processor_t UNUSED * const proc, const reg_t UNUSED
   tdata2 = val;
 }
 
-action_t trigger_t::legalize_action(reg_t val) const noexcept {
-  return (val > ACTION_MAXVAL || (val == ACTION_DEBUG_MODE && get_dmode() == 0)) ? ACTION_DEBUG_EXCEPTION : (action_t)val;
+action_t trigger_t::legalize_action(reg_t val, reg_t action_mask, reg_t dmode_mask) const noexcept {
+  reg_t act = get_field(val, action_mask);
+  return (act > ACTION_MAXVAL || (act == ACTION_DEBUG_MODE && get_field(val, dmode_mask) == 0)) ? ACTION_DEBUG_EXCEPTION : (action_t)act;
 }
 
 unsigned trigger_t::legalize_mhselect(bool h_enabled) const noexcept {
@@ -149,7 +150,7 @@ void mcontrol_t::tdata1_write(processor_t * const proc, const reg_t val, const b
   hit = get_field(val, CSR_MCONTROL_HIT);
   select = get_field(val, MCONTROL_SELECT);
   timing = legalize_timing(val, MCONTROL_TIMING, MCONTROL_SELECT, MCONTROL_EXECUTE, MCONTROL_LOAD);
-  action = legalize_action(get_field(val, MCONTROL_ACTION));
+  action = legalize_action(val, MCONTROL_ACTION, CSR_MCONTROL_DMODE(xlen));
   chain = allow_chain ? get_field(val, MCONTROL_CHAIN) : 0;
   match = legalize_match(get_field(val, MCONTROL_MATCH));
   m = get_field(val, MCONTROL_M);
@@ -275,7 +276,7 @@ void mcontrol6_t::tdata1_write(processor_t * const proc, const reg_t val, const 
   hit = get_field(val, CSR_MCONTROL6_HIT);
   select = get_field(val, CSR_MCONTROL6_SELECT);
   timing = legalize_timing(val, CSR_MCONTROL6_TIMING, CSR_MCONTROL6_SELECT, CSR_MCONTROL6_EXECUTE, CSR_MCONTROL6_LOAD);
-  action = legalize_action(get_field(val, CSR_MCONTROL6_ACTION));
+  action = legalize_action(val, CSR_MCONTROL6_ACTION, CSR_MCONTROL6_DMODE(xlen));
   chain = allow_chain ? get_field(val, CSR_MCONTROL6_CHAIN) : 0;
   match = legalize_match(get_field(val, CSR_MCONTROL6_MATCH));
   m = get_field(val, CSR_MCONTROL6_M);
@@ -315,7 +316,7 @@ void itrigger_t::tdata1_write(processor_t * const proc, const reg_t val, const b
   m = get_field(val, CSR_ITRIGGER_M);
   s = proc->extension_enabled_const('S') ? get_field(val, CSR_ITRIGGER_S) : 0;
   u = proc->extension_enabled_const('U') ? get_field(val, CSR_ITRIGGER_U) : 0;
-  action = legalize_action(get_field(val, CSR_ITRIGGER_ACTION));
+  action = legalize_action(val, CSR_ITRIGGER_ACTION, CSR_ITRIGGER_DMODE(xlen));
 }
 
 std::optional<match_result_t> trap_common_t::detect_trap_match(processor_t * const proc, const trap_t& t) noexcept
@@ -366,7 +367,7 @@ void etrigger_t::tdata1_write(processor_t * const proc, const reg_t val, const b
   m = get_field(val, CSR_ETRIGGER_M);
   s = proc->extension_enabled_const('S') ? get_field(val, CSR_ETRIGGER_S) : 0;
   u = proc->extension_enabled_const('U') ? get_field(val, CSR_ETRIGGER_U) : 0;
-  action = legalize_action(get_field(val, CSR_ETRIGGER_ACTION));
+  action = legalize_action(val, CSR_ETRIGGER_ACTION, CSR_ETRIGGER_DMODE(xlen));
 }
 
 bool etrigger_t::simple_match(bool interrupt, reg_t bit) const
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -221,7 +221,7 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
   return std::nullopt;
 }
 
-mcontrol_common_t::match_t mcontrol_common_t::legalize_match(reg_t val) const noexcept
+mcontrol_common_t::match_t mcontrol_common_t::legalize_match(reg_t val) noexcept
 {
   switch (val) {
     case MATCH_EQUAL:
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -148,7 +148,7 @@ void mcontrol_t::tdata1_write(processor_t * const proc, const reg_t val, const b
   dmode = get_field(val, CSR_MCONTROL_DMODE(xlen));
   hit = get_field(val, CSR_MCONTROL_HIT);
   select = get_field(val, MCONTROL_SELECT);
-  timing = get_field(val, MCONTROL_TIMING);
+  timing = legalize_timing(val, MCONTROL_TIMING, MCONTROL_EXECUTE);
   action = legalize_action(get_field(val, MCONTROL_ACTION));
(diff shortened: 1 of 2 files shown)


scottj97:
~~I don't think there is any agreement here. In fact the debug spec recommends the opposite, right after Table 5.10:~~

> A chain of triggers that don’t all have the same timing value will never fire. That means to implement the suggestions in Table 5.10, both timings should be supported on load address triggers.

Edit: sorry, my mistake. That quote is for load **address** triggers, and this PR is about load **data** triggers.

Interesting that Table 5.10 doesn't include "Load Data + Store Data". Wouldn't that be an interesting combination (so it hits on AMO)?

riscv/triggers.cc
```diff
@@ -157,7 +157,6 @@ void mcontrol_t::tdata1_write(processor_t * const proc, const reg_t val, const b
   execute = get_field(val, MCONTROL_EXECUTE);
   store = get_field(val, MCONTROL_STORE);
   load = get_field(val, MCONTROL_LOAD);
-  // Assume we're here because of csrw.
```
scottj97, line 161: It seems this comment goes all the way back to the [original trigger implementation](https://github.com/riscv-software-src/riscv-isa-sim/commit/d6dae451821656d7676fec3ec515691d4c66e065#diff-96f446fa4c024af925fb82998baa4145356afb5a6940d639fa705162029c8a78R404-R405) and probably has something to do with calling `trigger_updated()` -- though I don't understand the comment even in its original context.

Good riddance!

YenHaoChen:
> Interesting that Table 5.10 doesn't include "Load Data + Store Data". Wouldn't that be an interesting combination (so it hits on AMO)?

Yes. In this case "Load Data + Store Data", legalizing ```timing=1(after)``` seems reasonable for both loads and AMOs. Issue https://github.com/riscv/riscv-debug-spec/issues/792 clarifies there is no suggested timing for "Load Data + Store Data" in current debug spec.

riscv/triggers.cc
```diff
@@ -236,6 +236,8 @@ mcontrol_common_t::match_t mcontrol_common_t::legalize_match(reg_t val) const no
 }
 
 bool mcontrol_common_t::legalize_timing(bool val) const noexcept { // dependent on select, execute, load, and store
+  if (select && load)
+    return TIMING_AFTER;
```
scottj97: This deserves a comment explaining it. "For load data triggers, force timing=after to avoid debugger having to repeat loads which may have side effects."

riscv/triggers.cc
```diff
@@ -238,6 +235,14 @@ mcontrol_common_t::match_t mcontrol_common_t::legalize_match(reg_t val) const no
   }
 }
 
+bool mcontrol_common_t::legalize_timing(bool val) const noexcept { // dependent on select, execute, load, and store
```
scottj97: Comment seems incorrect -- I don't see how this is dependent on `store`? (same for line 160 above and 285 below).

I would prefer to just remove all three comments, since they are likely to diverge over time and the method is simple enough to read.

scottj97 (changes requested):

scottj97:
> The legalize_timing() implicitly depends on select, execution, load, and store. So, reordering the statement in the tdata1_write() may break the functionality. I need help finding a good way to avoid this and seeking suggestions.

You're right, that's a little dangerous. Perhaps pass select/execute/load as parameters to `legalize_timing()` so the dependency is explicit in the code? That still doesn't really solve the problem because reordering the statements would still compile but would be subtly broken.

You could do something like this:
```
  bool new_select = get_field(val, MCONTROL_SELECT);
  bool new_execute = get_field(val, MCONTROL_EXECUTE);
  bool new_load = get_field(val, MCONTROL_LOAD);
  timing = legalize_timing(get_field(val, MCONTROL_TIMING), new_select, new_execute, new_load);
  select = new_select;
  execute = new_execute;
  load = new_load;
```

But that's a mess and I don't think it's an improvement.

You could also implement `legalize_timing` so that it took `val` and all the field masks (which differ between mcontrol and mcontrol6). Such as:
```
  timing = legalize_timing(val, MCONTROL_TIMING, MCONTROL_SELECT, MCONTROL_EXECUTE, MCONTROL_LOAD);
```
And then in `legalize_timing()` you look only at `val` and not at the instance attributes `select`, `execute`, `load`.

YenHaoChen:
The https://github.com/riscv-software-src/riscv-isa-sim/commit/e464ab8efb36fc089303f11ddf78e32a90530c43 introduced the timing legalization of the execution trigger, which also legalized the load trigger. The https://github.com/riscv-software-src/riscv-isa-sim/commit/db6051f2001243e5132f224ece06753aff85748d removed the legalization of the load (address) trigger for a general behavior but kept the one of the execution trigger.

Legalizing the load data trigger is necessary, and removing the legalization of the execution trigger could provide a more general behavior. 
@timsifive, would you share some opinions on this?

riscv/triggers.h
```diff
@@ -208,6 +208,7 @@ class mcontrol_common_t : public trigger_t {
 
 protected:
   match_t legalize_match(reg_t val) const noexcept;
+  bool legalize_timing(reg_t val, reg_t timing_mask, reg_t select_mask, reg_t execute_mask, reg_t load_mask) const noexcept;
```
scottj97: This could be `static` now. (And same for `legalize_match()` while you're here.)

`legalize_action()` has a similar issue to your previous implementation of `legalize_timing()` because it looks at `dmode`.

scottj97 (changes requested):

scottj97:
> The [e464ab8](https://github.com/riscv-software-src/riscv-isa-sim/commit/e464ab8efb36fc089303f11ddf78e32a90530c43) introduced the timing legalization of the execution trigger, which also legalized the load trigger. The [db6051f](https://github.com/riscv-software-src/riscv-isa-sim/commit/db6051f2001243e5132f224ece06753aff85748d) removed the legalization of the load (address) trigger for a general behavior but kept the one of the execution trigger.
> 
> Legalizing the load data trigger is necessary, and removing the legalization of the execution trigger could provide a more general behavior. @timsifive, would you share some opinions on this?

This is a very interesting find. Apparently forcing load data triggers to timing=after was not desirable in 2016.

timsifive:
> The e464ab8 introduced the timing legalization of the execution trigger, which also legalized the load trigger. The db6051f removed the legalization of the load (address) trigger for a general behavior but kept the one of the execution trigger.

I think e464ab8 made it so that the timing bit reflected the actual spike implementation at the time. Before the timing bit was just ignored entirely.

I'm not sure why db6051f removed the timing legalization for loads (except that it apparently makes a test pass). In general, I think it makes sense to only support timing=before for execute triggers.

> Interesting that Table 5.10 doesn't include "Load Data + Store Data". Wouldn't that be an interesting combination (so it hits on AMO)?

I'm pretty sure I didn't consider AMO when I wrote that. The rule of thumb in that table is that for stores we want timing=before because then the user can read the value that's about to be overwritten. (In general before is preferable for this reason.) But when loads are involved, you can't do timing=before without reading the value first. Then the value would be read again when resuming from the trigger. Now you've read twice, and reading might have side effects. Putting that all together, Load+Store should probably be after, especially since in AMO operations don't know what the write is going to be until the load has completed.

scottj97:
> Putting that all together, Load+Store should probably be after, especially since in AMO operations don't know what the write is going to be until the load has completed.

With this PR, load data triggers are forced to After. A chain must all have the same timing in order to fire. Therefore a chain of Load+Store is already effectively forced to After. So, no further legalization is needed.

riscv/triggers.h
YenHaoChen: The gcc complained that the "static member function cannot have cv-qualifier," which means the "static" cannot be with "const noexcept." I prefer keeping the cv-qualifiers ("const noexcept") because we don't need to call the functions with no object.

Referene: https://stackoverflow.com/questions/19812786/why-cant-a-static-member-function-have-a-cv-qualifier

YenHaoChen: Commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1214/commits/798e0daf667ccff0f9bc8d68ad6bf91d0c351dad removes legalize_action()'s dependency on ```dmode```.

YenHaoChen:
Both execute and store triggers are preferred to be timing=before, which informs users before overwriting. The store trigger can be timing=after to chain with a load data trigger, which may be interesting for AMO. I don't have a reason to legalize timing differently for the execute and store triggers. A chain with the execute trigger and a load data trigger is also interesting for detecting a specific instruction of a particular value.

riscv/triggers.h
YenHaoChen: @scottj97 Should we keep the "const noexcept"?? Or, we can move ```legalize_timing()``` and ```legalize_match()``` out of the classes and make them available in triggers.cc only.

scottj97: `const` has no meaning (I guess?) when it's a static method. (Although what about static attributes? Hmm. But this class has no such attributes, so we don't care.)

_cv-qualifier_ does not include `noexcept`, so you should be able to keep that.

So, better to make it `static` and remove `const`.

riscv/triggers.h
```diff
@@ -207,7 +207,7 @@ class mcontrol_common_t : public trigger_t {
   bool simple_match(unsigned xlen, reg_t value) const;
 
 protected:
-  match_t legalize_match(reg_t val) const noexcept;
+  static match_t legalize_match(reg_t val) noexcept;
```
scottj97, line 210: Please add a justification to the commit message: "Since this method no longer uses `this`."

riscv/triggers.cc
```diff
@@ -25,8 +25,9 @@ void trigger_t::tdata2_write(processor_t UNUSED * const proc, const reg_t UNUSED
   tdata2 = val;
 }
 
-action_t trigger_t::legalize_action(reg_t val) const noexcept {
-  return (val > ACTION_MAXVAL || (val == ACTION_DEBUG_MODE && get_dmode() == 0)) ? ACTION_DEBUG_EXCEPTION : (action_t)val;
+action_t trigger_t::legalize_action(reg_t val, reg_t action_mask, reg_t dmode_mask) const noexcept {
```
scottj97: Might as well add another commit making this one `static` too.

scottj97 (changes requested):
In the commit message for dba22b0c3a26d595687209486be2c1da1271b2b7 please add a link to the recent discussions on the topic, so people will understand why this change is being made now.

scottj97 (changes requested):
> In the commit message for [dba22b0](https://github.com/riscv-software-src/riscv-isa-sim/commit/dba22b0c3a26d595687209486be2c1da1271b2b7) please add a link to the recent discussions on the topic, so people will understand why this change is being made now.

I was referring not to the implementation details (passing the masks as params) but to the forcing of load data triggers to be timing=before. Link to https://github.com/riscv-software-src/riscv-isa-sim/pull/1208#issuecomment-1373035906 in 28b60d9b1143a99e658d1e98e88c664bf6b9e049

scottj97 (changes requested):
> The load data trigger can only be timing=1(after) since load data is only available "after" the load instruction.

This is confusing and untrue, as timing=before worked before this commit. This needs a better justification. How about:

As recommended in the debug spec table "Suggested Trigger Timings", to avoid the footgun of replaying a load (which may have side effects) when the breakpoint trap handler returns.

scottj97 (approved):

PR #1219, opened 2023-01-10
Add --triggers=n to control the number of supported triggers

jerryz123:


scottj97:
Could/should this be part of the device tree?

Spike doesn't have a consistent/clean configuration system today. There's a haphazard mix of command-line options, device tree strings, and source-code constants. There was some movement to clean this up (see #938 and #974) but that contributor apparently lost interest.

riscv/triggers.cc
```diff
@@ -384,11 +384,13 @@ module_t::~module_t() {
 
 reg_t module_t::tdata1_read(unsigned index) const noexcept
 {
+  if (triggers.empty()) return 0;
```
scottj97: This doesn't seem necessary. If no triggers exist, should the `tdata1` CSR exist at all? The debug spec says about `tdata1` (and a bunch of other trigger-related CSRs):

> This register is optional if no triggers are implemented.

scottj97 (changes requested):

riscv/triggers.cc
jerryz123: I have a very slight preference towards providing the CSRs in the optional case, since the existing riscv-tests breakpoint tests assumes so. Those should still be fixed though.

I think these checks should be preserved either way, however, since a future libriscv user or spike fork may choose to provide the CSRs in a legal implementation.

jerryz123:
Yeah clearly there should be a more unified configuration system. I like your proposal in https://github.com/riscv-software-src/riscv-isa-sim/pull/974 of "unifying" the DTS and command line flags in a `cfg_t`.

My approach has been to continue shoving options for now into `cfg_t`, with the intent that a future refactor of the flag-based and DTS-based configuration paths could be based around that class.

> Could/should this be part of the device tree?

I don't know the answer to this, but I couldn't find any existing examples of this information embedded in a DTS.

riscv/triggers.cc
scottj97: How about instantiating all those CSRs as `const_csr_t` when no triggers exist? In `state_t::reset()`.

scottj97:
> I don't know the answer to this, but I couldn't find any existing examples of this information embedded in a DTS.

In that case, I guess command-line is the best we can do. Sigh.

riscv/csrs.h
```diff
@@ -637,7 +637,7 @@ class debug_mode_csr_t: public basic_csr_t {
   virtual void verify_permissions(insn_t insn, bool write) const override;
 };
 
-typedef std::shared_ptr<tdata2_csr_t> tdata2_csr_t_p;
+typedef std::shared_ptr<csr_t> tdata2_csr_t_p;
```
scottj97: This typedef is no longer useful and is now misleading because it doesn't point to a `tdata2_csr_t`. Instead simply change any uses of `tdata2_csr_t_p` to use `csr_t_p` instead and delete this typedef.

scottj97 (changes requested):

riscv/csrs.h
jerryz123: Fixed

scottj97 (changes requested):
No reason to leave aeae9f0781f1f628061e9c3f8c1cc77edcc258fc in the patch series anymore

scottj97 (approved):

aswaterman (approved):

PR #1220, opened 2023-01-11
Add icount trigger

YenHaoChen:
This PR adds icount trigger (debug spec 5.5.13).

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -287,6 +287,27 @@ void mcontrol6_t::tdata1_write(processor_t * const proc, const reg_t val, const 
   load = get_field(val, CSR_MCONTROL6_LOAD);
 }
 
+std::optional<match_result_t> icount_t::detect_icount_match(processor_t * const proc) noexcept
+{
+  if (!common_match(proc))
+    return std::nullopt;
+
+  std::optional<match_result_t> ret = std::nullopt;
+  if (pending) {
+    pending = 0;
+    hit = true;
+    ret = match_result_t(TIMING_BEFORE, action);
+  }
+
+  if (count >= 1) {
+    if (count == 1)
+      pending = 1;
+    count = count - 1;
+  }
+
+  return ret;
+}
+
 reg_t icount_t::tdata1_read(const processor_t * const proc) const noexcept
 {
   auto xlen = proc->get_xlen();
@@ -296,9 +317,9 @@ reg_t icount_t::tdata1_read(const processor_t * const proc) const noexcept
   tdata1 = set_field(tdata1, CSR_ICOUNT_VS, proc->extension_enabled('H') ? vs : 0);
   tdata1 = set_field(tdata1, CSR_ICOUNT_VU, proc->extension_enabled('H') ? vu : 0);
   tdata1 = set_field(tdata1, CSR_ICOUNT_HIT, hit);
-  tdata1 = set_field(tdata1, CSR_ICOUNT_COUNT, count);
+  tdata1 = set_field(tdata1, CSR_ICOUNT_COUNT, count_read_value);
   tdata1 = set_field(tdata1, CSR_ICOUNT_M, m);
-  tdata1 = set_field(tdata1, CSR_ICOUNT_PENDING, pending);
+  tdata1 = set_field(tdata1, CSR_ICOUNT_PENDING, pending_read_value);
   tdata1 = set_field(tdata1, CSR_ICOUNT_S, s);
   tdata1 = set_field(tdata1, CSR_ICOUNT_U, u);
   tdata1 = set_field(tdata1, CSR_ICOUNT_ACTION, action);
@@ -313,12 +334,18 @@ void icount_t::tdata1_write(processor_t * const proc, const reg_t val, const boo
   vs = get_field(val, CSR_ICOUNT_VS);
   vu = get_field(val, CSR_ICOUNT_VU);
   hit = get_field(val, CSR_ICOUNT_HIT);
-  count = get_field(val, CSR_ICOUNT_COUNT);
+  count = count_read_value = get_field(val, CSR_ICOUNT_COUNT);
   m = get_field(val, CSR_ICOUNT_M);
-  pending = get_field(val, CSR_ICOUNT_PENDING);
+  pending = pending_read_value = get_field(val, CSR_ICOUNT_PENDING);
   s = proc->extension_enabled_const('S') ? get_field(val, CSR_ICOUNT_S) : 0;
   u = proc->extension_enabled_const('U') ? get_field(val, CSR_ICOUNT_U) : 0;
   action = legalize_action(val, CSR_ICOUNT_ACTION, CSR_ICOUNT_DMODE(xlen));
+}
+
+void icount_t::stash_read_values()
+{
+  count_read_value = count;
+  pending_read_value = pending;
 }
 
 reg_t itrigger_t::tdata1_read(const processor_t * const proc) const noexcept
@@ -527,6 +554,24 @@ std::optional<match_result_t> module_t::detect_memory_access_match(operation_t o
       ret = result;
 
     chain_ok = result.has_value() || !trigger->get_chain();
+  }
+  return ret;
+}
+
+std::optional<match_result_t> module_t::detect_icount_match() noexcept
+{
+  for (auto trigger: triggers)
+    trigger->stash_read_values();
+
+  state_t * const state = proc->get_state();
+  if (state->debug_mode)
+    return std::nullopt;
+
+  std::optional<match_result_t> ret = std::nullopt;
+  for (auto trigger: triggers) {
+    auto result = trigger->detect_icount_match(proc);
+    if (result.has_value() && (!ret.has_value() || ret->action < result->action))
+      ret = result;
   }
   return ret;
 }
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -474,6 +474,7 @@ std::optional<match_result_t> module_t::detect_memory_access_match(operation_t o
 
   bool chain_ok = true;
 
+  std::optional<match_result_t> ret = std::nullopt;
   for (auto trigger: triggers) {
     if (!chain_ok) {
       chain_ok = !trigger->get_chain();
@@ -487,12 +488,12 @@ std::optional<match_result_t> module_t::detect_memory_access_match(operation_t o
      * trigger in the chain will never get `hit` set unless the entire chain
      * matches. */
     auto result = trigger->detect_memory_access_match(proc, operation, address, data);
-    if (result.has_value() && !trigger->get_chain())
-      return result;
+    if (result.has_value() && !trigger->get_chain() && (!ret.has_value() || ret->action < result->action))
+      ret = result;
 
     chain_ok = result.has_value() || !trigger->get_chain();
   }
-  return std::nullopt;
+  return ret;
 }
 
 std::optional<match_result_t> module_t::detect_trap_match(const trap_t& t) noexcept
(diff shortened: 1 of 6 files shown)


riscv/triggers.cc
```diff
@@ -461,6 +461,16 @@ bool module_t::tdata3_write(unsigned index, const reg_t val) noexcept
   return true;
 }
 
+bool module_t::icount_enabled() const
```
aswaterman: This function is fairly expensive to compute, which slows down the computation of whether the slow path should be taken. Consider precomputing it so that the triggers don’t need to be scanned every time. It’s OK to be conservative and occasionally take the slow path when unnecessary, if that helps.

YenHaoChen: https://github.com/riscv-software-src/riscv-isa-sim/pull/1220/commits/f52de86dc168222264a401e59a7c5e4eef252670 adds a ```check_icount_enabled``` variable to reduce the computation amount of the ```icount_enabled()```. For general cases, the ```check_icount_enabled``` will be false, and the ```icount_enabled()``` will return false directly. The only scenario that requires performing the complete check is when writing tdata1. In this case, the ```check_icount_enabled``` will be set to true, and ```icount_enabled()``` will perform the full check. If the checking is false, we can set the ```check_icount_enabled``` back to false and skip the full check next time.

@aswaterman Is this computation overhead acceptable?

aswaterman: Yeah, SGTM.

scottj97: @YenHaoChen there is already a `processor_t::trigger_updated()` method that calculates similar things for fetch/load/store triggers. Why not do the computation and set a variable `check_triggers_icount` (name chosen to be similar to others that already exist) in that method?

riscv/triggers.cc
```diff
@@ -282,6 +282,40 @@ void mcontrol6_t::tdata1_write(processor_t * const proc, const reg_t val, const
...
+  assert(get_field(val, CSR_ICOUNT_TYPE(xlen)) == CSR_TDATA1_TYPE_ICOUNT);
+  dmode = proc->get_state()->debug_mode ? get_field(val, CSR_ICOUNT_DMODE(xlen)) : 0;
+  vs = get_field(val, CSR_ICOUNT_VS);
+  vu = get_field(val, CSR_ICOUNT_VU);
+  hit = get_field(val, CSR_ICOUNT_HIT);
+  count = get_field(val, CSR_ICOUNT_COUNT) & 1; // icount.count is either 0 or 1 (single step)
+  m = get_field(val, CSR_ICOUNT_M);
+  pending = get_field(val, CSR_ICOUNT_PENDING);
+  s = proc->extension_enabled('S') ? get_field(val, CSR_ICOUNT_S) : 0;
+  u = proc->extension_enabled('U') ? get_field(val, CSR_ICOUNT_U) : 0;
```
scottj97: These two should be `extension_enabled_const()` since they don't get cleared if `misa.S` is written to 0, like `vs` and `vu` do.

riscv/triggers.cc
```diff
@@ -592,6 +592,7 @@ reg_t module_t::tinfo_read(unsigned UNUSED index) const noexcept
 {
   /* In spike, every trigger supports the same types. */
   return (1 << CSR_TDATA1_TYPE_MCONTROL) |
+         (1 << CSR_TDATA1_TYPE_ICOUNT) |
```
scottj97, line 598: My tests use `tinfo` to attempt to cover icount triggers, so because this is the last commit in the series, they won't even attempt to test any of the earlier commits. Can you combine this with c081a020630a50fdc85bbcfe4b176822fd7abf29 so it will be visible as soon as the CSRs are available?

riscv/triggers.cc
```diff
@@ -282,6 +282,40 @@ void mcontrol6_t::tdata1_write(processor_t * const proc, const reg_t val, const
...
+
+void icount_t::tdata1_write(processor_t * const proc, const reg_t val, const bool UNUSED allow_chain) noexcept
+{
+  auto xlen = proc->get_xlen();
+  assert(get_field(val, CSR_ICOUNT_TYPE(xlen)) == CSR_TDATA1_TYPE_ICOUNT);
+  dmode = proc->get_state()->debug_mode ? get_field(val, CSR_ICOUNT_DMODE(xlen)) : 0;
+  vs = get_field(val, CSR_ICOUNT_VS);
+  vu = get_field(val, CSR_ICOUNT_VU);
+  hit = get_field(val, CSR_ICOUNT_HIT);
+  count = get_field(val, CSR_ICOUNT_COUNT) & 1; // icount.count is either 0 or 1 (single step)
```
scottj97: Why limit the count to 0 or 1? Based on the rest of the new logic, it appears there is no limit beyond the 14-bit width of the field.

riscv/triggers.cc
```diff
@@ -461,6 +518,19 @@ bool module_t::tdata3_write(unsigned index, const reg_t val) noexcept
   return true;
 }
 
+bool module_t::icount_enabled()
+{
+  if (check_icount_enabled) {
+    auto xlen = proc->get_xlen();
+    for (auto trigger: triggers) {
+      if (get_field(trigger->tdata1_read(proc), CSR_TDATA1_TYPE(xlen)) == CSR_TDATA1_TYPE_ICOUNT)
```
scottj97: Shouldn't this also consider other fields (such as `count > 0`)? If `type` is 3 but there's no `pending` and zero `count` then the trigger can't possibly fire and doesn't need to be checked.

riscv/triggers.cc
```diff
@@ -282,6 +282,61 @@ void mcontrol6_t::tdata1_write(processor_t * const proc, const reg_t val, const
...
+  auto xlen = proc->get_xlen();
+  reg_t tdata1 = 0;
+  tdata1 = set_field(tdata1, CSR_ICOUNT_TYPE(xlen), CSR_TDATA1_TYPE_ICOUNT);
+  tdata1 = set_field(tdata1, CSR_ICOUNT_DMODE(xlen), dmode);
+  tdata1 = set_field(tdata1, CSR_ICOUNT_VS, proc->extension_enabled('H') ? vs : 0);
+  tdata1 = set_field(tdata1, CSR_ICOUNT_VU, proc->extension_enabled('H') ? vu : 0);
+  tdata1 = set_field(tdata1, CSR_ICOUNT_HIT, hit);
+  tdata1 = set_field(tdata1, CSR_ICOUNT_COUNT, count);
+  tdata1 = set_field(tdata1, CSR_ICOUNT_M, m);
+  tdata1 = set_field(tdata1, CSR_ICOUNT_PENDING, pending);
```
scottj97: I think `csrr` will be a problem here, because it is supposed to read the pre-decremented value. From the Zicsr spec:

> Some CSRs, such as the instructions-retired counter, instret, may be modified as side effects of instruction execution. In these cases, if a CSR access instruction reads a CSR, it reads the value prior to the execution of the instruction.

It seems to me that `count` and `pending` have already been updated by the time this function is called, so it will read the updated value.

scottj97 (changes requested):

riscv/triggers.cc
scottj97: Actually I think this would best be done as a method on `trigger_t::icount_check_needed()` and then you can check if any trigger responds `true` to this method.

riscv/triggers.cc
scottj97: You can see in `wide_counter_csr_t::unlogged_write()` how we handle this for `minstret` and friends. I think you'll need to do something like that, though it will be more complicated.

riscv/triggers.cc
aswaterman: Good idea, computing `processor_t::trigger_updated()` could compute `processor_t::check_triggers_icount` precisely, superseding the need for `module_t::icount_enabled()`.  That seems a little cleaner.

riscv/triggers.cc
YenHaoChen: The ```wide_counter_csr_t``` assumes each instruction will bump the counter unconditionally, which is correct for the ```minstret``` and ```mcycle```. Unfortunately, the ```icount.count``` countdowns conditionally. For instance, when writing n to ```minstret```, the ```wide_counter_csr_t``` updates the CSR to (n-1), and the next immediate read will get n after bumping. However, when writing n to ```icount.count```, if we update the CSR to (n+1), the next read should be n but can be (n+1) if it does not match. Thus, the technique of ```minstret``` does not work for ```icount.count```. (This issue exists on the Sscofpmf's mhpmcounters with conditional bumping.)

How about adding new variables to track the reading value? https://github.com/riscv-software-src/riscv-isa-sim/pull/1220/commits/a277f69f078b51332e131ece43e947e8e1c5bbe6

scottj97: Yes, that's what I meant by "more complicated". :)

I think your solution will work but I don't (yet) have a good way to test it. I'm working on that but it might take a while, so I won't hold up this PR for that.

Of course a277f69 should be squashed back.

riscv/triggers.cc
```diff
@@ -490,6 +552,23 @@ std::optional<match_result_t> module_t::detect_memory_access_match(operation_t o
...
+    trigger->stash_read_values();
+
+  state_t * const state = proc->get_state();
+  if (state->debug_mode)
+    return std::nullopt;
+
+  for (auto trigger: triggers) {
+    auto result = trigger->detect_icount_match(proc);
+    if (result.has_value())
+      return result;
```
scottj97: What if multiple icount triggers are running simultaneously? They should all be decremented, but this looks like it will stop at the first match.

scottj97 (changes requested):

scottj97 (changes requested):
Because the `icount_check_needed()` is only a performance optimization, it would be best to add that feature only after the icount trigger is working. That way, if there's a bug in the performance optimization, it will show up as a behavioral difference on that commit.

Or at least, `icount_t::icount_check_needed()` should always return `true` at first. Only once everything else is working, then optimize this expression.

riscv/triggers.cc
YenHaoChen: OK, if something goes wrong later, I will provide another PR.

riscv/triggers.cc
```diff
@@ -525,6 +552,24 @@ std::optional<match_result_t> module_t::detect_memory_access_match(operation_t o
...
+  if (state->debug_mode)
+    return std::nullopt;
+
+  std::optional<match_result_t> ret = std::nullopt;
+  for (auto trigger: triggers) {
+    auto result = trigger->detect_icount_match(proc);
+    if (result.has_value())
+      ret = result;
+  }
+  return ret;
```
scottj97: There is one other consideration that I just now realized. In the debug spec 5.3 "Priority" it says about multiple triggers firing simultaneously:

> If one of these triggers has the “enter Debug Mode” action (1) and another trigger has the “raise a breakpoint exception” action (0), the preferred behavior is to have both actions take place. [...] If this is not implemented, then the hart must enter Debug Mode and ignore the breakpoint exception.

The above loop appears to return the action from the highest-index breakpoint that triggers, regardless of others that might have a different action.

It seems that `module_t::detect_memory_access_match` is similarly broken. I'll try to fix that in a separate PR.

scottj97 (changes requested):

YenHaoChen:
Rebase master for legalize_action().

riscv/triggers.cc
YenHaoChen: I want to make all triggers have consistent behavior. Thus, I choose the simplest solution and also apply it to other triggers first. We can change to the other more preferred solution in another PR.

Is this OK with you?

scottj97: Sounds good to me.

scottj97: I think I misunderstood. I see you added commit fb8a8d5cbc6665e9dd89961dab2cf24c33d1ed80 to address this in this PR. I also opened #1231 because I thought you were saying you'd fix this in a future PR.

I don't have good test coverage of this case, but AFAICT this will fix #1231.

scottj97 (approved):

scottj97:
@YenHaoChen do you think these spec changes will require any changes here?

* https://github.com/riscv/riscv-debug-spec/pull/797
* https://github.com/riscv/riscv-debug-spec/pull/798

scottj97:
(also please rebase onto latest master to resolve merge conflicts)

YenHaoChen:
This PR implements the behavior of the https://github.com/riscv/riscv-debug-spec/pull/797 and https://github.com/riscv/riscv-debug-spec/pull/798.
(rebase master to resolve the merge conflict)

scottj97 (approved):

PR #1224, opened 2023-01-11
For NS16550 UART, poll stdin less often

aswaterman:
On my Mac Mini, calling `poll()` on stdin takes around 10 us, and we are invoking it every 20 us or so.  Reduce the frequency of polling by 16x when not actively receiving data, thereby reducing the fraction of time spent in `poll()` to a trivial amount.


scottj97 (approved):
I didn't try to understand any of this, but it doesn't seem to break anything.

PR #1225, opened 2023-01-12
Run Spike and HTIF in a single thread, rather than two

aswaterman:
The two-thread approach was originally motivated by making Spike look as similar as possible to other HTIF targets.  But we can get the same semantics without threading by running the simulator inside of the HTIF host's idle loop instead of performing a context switch.

This was motivated by speeding up the simulator on Mac OS (it's worth around 20% because using pthread condition variables to force strict alternation is very slow).  But I think it also simplifies the control flow enough to justify it on that basis, too.

### DIFF
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -220,33 +220,15 @@ sim_t::~sim_t()
   delete debug_mmu;
 }
 
-void sim_thread_main(void* arg)
-{
-  ((sim_t*)arg)->main();
-}
-
-void sim_t::main()
+int sim_t::run()
 {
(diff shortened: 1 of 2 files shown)


scottj97 (approved):
I didn't try to understand it, but it doesn't seem to break anything.

i2h2:
I think this change breaks HTIF in interactive mode because the thread is stuck in sim_t::interactive() with no one serving the HTIF command buffer, so the target program (e.g. pk) gets stuck in an infinite loop.

aswaterman:
I’ll look into that.

PR #1229, opened 2023-01-14
Add more hint  when people only type a simple name

gr816ox:
when people type something like "spike pk" rather than "spike usr/local/bin/pk" and the path load_payload generated can't access, people now will get a new hit "**could not open usr/local/riscv64-unknown-elf/bin/pk  which is the absolute path spike generated by /$prefix/$with-target/bin/<target program>  ( did you correctly configure --prefix and --with-target when ../configure? did you misspell it? If VCS, did you forget +permissive/+permissive-off? )**"

Tonight was my first day with spike, and I spend 4 hours to read source code and finally figure out that the reason why "spike pk" can't work correctly is because default TARGET_ARCH is riscv64-unknown-elf. So the path spike looking is always ../riscv64-unknown-elf/bin/pk which is not existed in my linux file system. And I have no idea what's happening without reading source code. lt should be more easier for people to figure out what's wrong with the command they typed if we tell them the absoult path spike generated.

Signed-off-by: gr816ox <50945677+gr816ox@users.noreply.github.com>


fesvr/htif.cc
```diff
@@ -105,19 +105,27 @@ static void bad_address(const std::string& situation, reg_t addr)
...
   else if (payload.find('/') == std::string::npos)
   {
-    std::string test_path = PREFIX TARGET_DIR + payload;
+    test_path = PREFIX TARGET_DIR + payload;
     if (access(test_path.c_str(), F_OK) == 0)
       path = test_path;
   }
 
-  if (path.empty())
+  if (path.empty()){
```
aswaterman: Add a space before {

fesvr/htif.cc
```diff
@@ -105,19 +105,27 @@ static void bad_address(const std::string& situation, reg_t addr)
...
+    if (payload.find('/') == std::string::npos)
+      throw std::runtime_error(
+        "could not open " + test_path +
+        " which is the absolute path spike generated by /$prefix/$with-target/bin/<target program> " +
```
aswaterman: I think it would be easier to understand if you wrote “.  The search path is the current directory, followed by the ‘bin’ directory within the directory specified by the  configure —prefix.”

fesvr/htif.cc
```diff
@@ -117,6 +117,8 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
   if (path.empty())
     throw std::runtime_error(
         "could not open " + payload +
+        " The search path is the current directory, followed by the 'bin' directory within the directory specified by the configure —with-target"+
```
jerryz123:
```suggestion
        " The search path is the current directory, followed by the 'bin' directory within the directory specified by the configure --with-target"+
```

fesvr/htif.cc
```diff
@@ -117,6 +117,8 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
...
+        " within the directory specified by the configure —prefix(-prefix/-with-tartget/bin/<target program>)."+
```
jerryz123: This line is confusing. Is the first line sufficient here?

gr816ox: Yes, it's indeed confusing and there must be a better way to express, but I think it should be here. The search path is generated by both **--prefix** and **--with-target**. 

For example, one type `../configure --prefix=/usr/local --with-target=/riscv-linux-gnu` when he built spike. He type `spike pk`, spike will try to run `access("pk", F_OK)`. If it can't access, spike will generate a absolute path /usr/local/riscv-linux-gnu/bin/pk(--prefix/--with-target/bin/\<target program\>) and try to run `access("/usr/local/riscv-linux-gnu/bin/pk", F_OK)`. If it false again, the runtime_error will be thrown.

This is why `spike /usr/local/bin/pk` always work, while `spike pk` don't. The absolute path spike generated may be wrong with default --prefix(/usr/local) and default --with-target(riscv64-unknown-elf). If we mention how search path is generated, it will be easier to figure out what's wrong.

Maybe we could mention it in README?

gr816ox: > The search path is the current directory, followed by the 'bin' directory within the directory specified by the configure —prefix and —with-target (--prefix/--with-tartget/bin/\<target program\>).

Is it better?

scottj97: I agree the error message given in this case could be improved, though the new error message here is quite verbose and complex. How about something like this:
```
Error: could not open pk; searched paths:
  . (current directory)
  /usr/local/riscv-linux-gnu/bin/ (based on configured --prefix and --with-target)
```

gr816ox: I pull a new commit with @scottj97 's suggestion and add a `else` below `if(access(test_path,F_OK) == 0)` which will throw the error with searched path.

e.g.
`spike ls` will get:

```
  what():  could not open ls; searched paths:
        . (current directory)
        /usr/local/riscv64-linux-gnu/bin/ (based on configured --prefix and --with-target)
```
things like `spike /tmp/ls` will still get:
```
  what():  could not open /tmp/ls (did you misspell it? If VCS, did you forget +permissive/+permissive-off?)
```

fesvr/htif.cc
```diff
@@ -112,6 +112,12 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
     std::string test_path = PREFIX TARGET_DIR + payload;
     if (access(test_path.c_str(), F_OK) == 0)
       path = test_path;
+    else
+      throw std::runtime_error(
+        "could not open "+ payload + "; searched paths:\n" +
```
scottj97: Use consistent spacing around the `+` (you're missing a space before the first `+`)

scottj97 (changes requested):
I'm good with this, but @gr816ox please squash all these commits into one and force-push.

PR #1230, opened 2023-01-16
Add more hint when searched path is wrong

gr816ox:
Delete the old branch and pull a new one, because of a wrong force push. Git is not as easy as I think.

When people type `spike ls`, they will get:
```
  what():  could not open ls; searched paths:
        . (current directory)
        /usr/local/riscv64-unknown-elf/bin/ (based on configured --prefix and --with-target)
```

things like `spike /tmp/ls` will still get:
```
  what():  could not open /tmp/ls (did you misspell it? If VCS, did you forget +permissive/+permissive-off?)
```

Thanks for help from @scottj97 @jerryz123 @aswaterman .
Signed-off-by: gr816ox <50945677+gr816ox@users.noreply.github.com>

### DIFF
--- a/fesvr/htif.cc
+++ b/fesvr/htif.cc
@@ -112,6 +112,12 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
     std::string test_path = PREFIX TARGET_DIR + payload;
     if (access(test_path.c_str(), F_OK) == 0)
       path = test_path;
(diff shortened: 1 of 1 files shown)


scottj97 (approved):

aswaterman (approved):

aswaterman:
Thank you!

PR #1234, opened 2023-01-20
Zicond: implement Zicond (conditional integer operations)

ptomsich:
This implements the Zicond (conditional integer operations) extension, as of version 1.0-draft-20230120.

The Zicond extension acts as a building block for branchless sequences including conditional-arithmetic, conditional-logic and conditional-select/move.
The following instructions constitute Zicond:
 - czero.eqz rd, rs1, rs2  =>  rd = (rs2 == 0) ? 0 : rs1
 - czero.nez rd, rs1, rs2  =>  rd = (rs2 != 0) ? 0 : rs1

See
  https://github.com/riscv/riscv-zicond/releases/download/v1.0-draft-20230120/riscv-zicond_1.0-draft-20230120.pdf
for the proposed specification and usage details.


aswaterman (approved):
LGTM, thanks for contributing this.

riscv/encoding.h
```diff
@@ -709,6 +709,10 @@
 #define MASK_CTZ 0xfff0707f
 #define MATCH_CTZW 0x6010101b
 #define MASK_CTZW 0xfff0707f
+#define MATCH_CZERO_EQZ 0xe005033
+#define MASK_CZERO_EQZ 0xfe00707f
+#define MATCH_CZERO_NEZ 0xe007033
+#define MASK_CZERO_NEZ 0xfe00707f
```
scottj97, line 715: This `encoding.h` file is generated from the riscv-opcodes project. It should not be hand-edited. We should see the riscv-opcodes commit hash at the top of this file change to match the updated source, and it should be done as a separate commit in this PR.

scottj97 (changes requested):

riscv/encoding.h
aswaterman: Thanks for catching this.  However, we're in a pickle: see https://github.com/riscv/riscv-opcodes/pull/147

scottj97: Ugly, but you could make a PR on riscv-opcodes that forks from before riscv/riscv-opcodes#147, cherry-picks riscv/riscv-opcodes#152, then merges back to master with ultimately no changes on master. Then we'd have a git hash to build from here.

aswaterman: It's preferable to blocking this PR, which is unrelated to the breakage.  I'll do that.

aswaterman:
Superseded by #1241

PR #1242, opened 2023-02-02
Make htif->get_to/fromhost_addr methods public

jerryz123:
In #1232 I accidentally made the methods protected, not public, as intended


aswaterman (approved):

jerryz123:
Thanks

PR #1244, opened 2023-02-05
Only generate a dtb and bus devices if dtb_enabled

jerryz123:
Previously, spike would generate a dts/dtb even if `!dtb_enabled`, and add devices based on what was generated. This breaks the use case where a spike-as-library user wants to both:
 - provide a custom bootrom+dtb 
 - provide a custom set of bus devices
 
 `!dtb_enabled` now skips both dts/dtb generation and adding the default PLIC/CLINT/UART devices.

### DIFF
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -45,7 +45,6 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
     mems(mems),
     plugin_devices(plugin_devices),
     procs(std::max(cfg->nprocs(), size_t(1))),
-    dtb_file(dtb_file ? dtb_file : ""),
     dtb_enabled(dtb_enabled),
     log_file(log_path),
     cmd_file(cmd_file),
@@ -102,7 +101,7 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
                                log_file.get(), sout_);
   }
 
-  make_dtb();
+  make_dtb(dtb_file);
 
   void *fdt = (void *)dtb.c_str();
 
@@ -293,10 +292,10 @@ bool sim_t::mmio_store(reg_t paddr, size_t len, const uint8_t* bytes)
   return bus.store(paddr, len, bytes);
 }
 
-void sim_t::make_dtb()
+void sim_t::make_dtb(const char* dtb_file)
 {
-  if (!dtb_file.empty()) {
-    std::ifstream fin(dtb_file.c_str(), std::ios::binary);
+  if (dtb_file) {
+    std::ifstream fin(dtb_file, std::ios::binary);
     if (!fin.good()) {
       std::cerr << "can't find dtb file: " << dtb_file << std::endl;
       exit(-1);
@@ -317,7 +316,7 @@ void sim_t::make_dtb()
   int fdt_code = fdt_check_header(dtb.c_str());
   if (fdt_code) {
     std::cerr << "Failed to read DTB from ";
-    if (dtb_file.empty()) {
+    if (!dtb_file) {
       std::cerr << "auto-generated DTS string";
     } else {
       std::cerr << "`" << dtb_file << "'";
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -101,6 +101,10 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
                                log_file.get(), sout_);
   }
 
+  // When running without using a dtb, skip the fdt-based configuration steps
+  if (!dtb_enabled) return;
+
+  // Load dtb_file if provided, otherwise self-generate a dts/dtb
   make_dtb(dtb_file);
 
   void *fdt = (void *)dtb.c_str();
--- a/riscv/sim.h
+++ b/riscv/sim.h
@@ -49,7 +49,7 @@ public:
   void set_remote_bitbang(remote_bitbang_t* remote_bitbang) {
     this->remote_bitbang = remote_bitbang;
   }
-  const char* get_dts() { if (dts.empty()) reset(); return dts.c_str(); }
+  const char* get_dts() { return dts.c_str(); }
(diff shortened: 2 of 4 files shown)


scottj97 (changes requested):
I don't have time to investigate right now, but [b19eb27](https://github.com/riscv-software-src/riscv-isa-sim/pull/1244/commits/b19eb27903457d47f52f92d63cdd4450f33881e4) changes the way my custom tests (proprietary, sorry) run the reset sequence. The log diff:
```
 core   0: 3 0x0000000000001008 (0xf1402573) x10 0x0000000000000000
-core   0: 3 0x000000000000100c (0x0182b283) x5  0x0000000080002000 mem 0x0000000000001018
+core   0: 3 0x000000000000100c (0x0182b283) x5  0x0000000080000000 mem 0x0000000000001018
 core   0: 3 0x0000000000001010 (0x00028067)
```

So the memory value at 0x1018 is different, which changes the destination of the `jr`. I believe this code is coming from Spike's built-in reset vector.

jerryz123:
I think I see the issue. I'll update this in a moment.

jerryz123:
I've removed the offending commit from this PR.

riscv/sim.h
```diff
@@ -49,7 +49,7 @@ class sim_t : public htif_t, public simif_t
   void set_remote_bitbang(remote_bitbang_t* remote_bitbang) {
     this->remote_bitbang = remote_bitbang;
   }
-  const char* get_dts() { if (dts.empty()) reset(); return dts.c_str(); }
+  const char* get_dts() { return dts.c_str(); }
```
scottj97, line 52: Do you understand why the `reset()` was there before? I don't get it.

Also, it seems nobody calls `reset()` anymore, so that method can be removed entirely.

riscv/sim.cc
```diff
@@ -101,7 +101,10 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
                                log_file.get(), sout_);
   }
 
-  make_dtb(dtb_file);
+  // create dts, dtb, and bootrom device
+  if (dtb_enabled) {
+    make_dtb(dtb_file);
+  }
 
   void *fdt = (void *)dtb.c_str();
```
scottj97, line 110: I'm confused how this and the following code still works if `!dtb_enabled`. It's using `fdt` which I guess points to a `'\0'` now, so `fdt_parse_clint()` et al will find nothing, so devices won't be created. In fact the entire rest of this routine will do nothing.

If I'm correct about that, then let's do something like:
```
if (!dtb_enabled) return;
make_dtb(dtb_file);
...
```

Also the commit message for f1a43675be25166ddabf7299b94cfd89bd49d45c should say something about removing devices in the case of `!dtb_enabled`.

scottj97 (changes requested):

riscv/sim.h
jerryz123: I think its because spike-main calls `sim_t::get_dts` before it runs the htif main thread, which is responsible for calling `sim_t::reset`. 

Before https://github.com/riscv-software-src/riscv-isa-sim/commit/3b26740205c401498fe77c2c9a16c8907a139cd8, the reset here was necessary, as the DTS wouldn't be generated until `make_dtb` in `reset` was called. 

The sequence before https://github.com/riscv-software-src/riscv-isa-sim/commit/3b26740205c401498fe77c2c9a16c8907a139cd8:
 1. `sim_t::sim_t` - does not construct dts/dtb
 2. if `sim_t::get_dts` called, run 4
 3. `sim_t::run` - calls `htif::run`, which calls `sim_t::reset`
 4. `sim_t::reset`  - calls `sim_t::make_dtb` - constructs dts/dtb, queries htif for the reset vec, constructs the rom with reset vec + dtb


The sequence after https://github.com/riscv-software-src/riscv-isa-sim/commit/3b26740205c401498fe77c2c9a16c8907a139cd8:
 1. `sim_t::sim_t` - calls `sim_t::make_dtb` to construct dts + dtb
 2. if `sim_t::get_dts` called, run 4
 3. `sim_t::run` - calls `htif::run`, which calls `sim_t::reset`
 4. `sim_t::reset`  - calls `sim_t::set_rom` - queries htif for the reset vec, constructs the rom with the reset vec embedded

The sequence after this PR becomes
 1. `sim_t::sim_t` - if dtb_enabled, calls `sim_t::make_dtb` to construct dts + dtb,
 2. if `sim_t::get_dts` called, return the dts
 3. `sim_t::run` - calls `htif::run`, which calls `sim_t::reset`
 4. `sim_t::reset` - calls `sim_t::set_rom` - queries htif for the reset vec, constructs the rom with the reset vec embedded
 
 The awkwardness comes from constructing sim_t before htif has determined what the reset vector should be.

riscv/sim.cc
jerryz123: > I'm confused how this and the following code still works if !dtb_enabled. It's using fdt which I guess points to a '\0' now, so fdt_parse_clint() et al will find nothing, so devices won't be created. In fact the entire rest of this routine will do nothing.

Yeah that's the intent, it's more intuitive that dtb-parsed devices should not be added if `!dtb_enabled`.

I'll restructure this as you suggest.

jerryz123: I also had to change some of the processor_t constructor such that cfg-passed settings would be respected even if `!dtb_enabled`. Specifically for pmpregions, and matching the paging mode for both cases.

scottj97 (changes requested):
I think the commit messages for 6e1169b3a80f946a8837bf5c196657271c970acb  and 5c69bf9760d93cd877311a81b44822bb67de9951 need to explain the rationale. Imagine someone bisecting a regression and they land on one of those commits. The message needs to help them understand the _why_.

scottj97 (approved):

aswaterman (approved):

PR #1245, opened 2023-02-05
Make misa.C writable again, even in presence of Zc* extensions

### LINKED ISSUE #1172: Alignment  when 'C'  in misa CSR is cleared by software

aswaterman:
Builds on and supersedes #1176. See extensive discussion on that PR.  Fixes #1172.

`misa.C` is now writable whenever `C` is included in the ISA string.  When `misa.C` is writable, all Zc* extensions are also under its direct control.  (There is still no mechanism to individually disable Zc* extensions, nor is there a mechanism to disable Zc* extensions when `misa.C` is read-only).

cc @liweiwei90, who authored the first version of this

### DIFF
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -223,7 +223,7 @@ public:
     if (ext >= 'A' && ext <= 'Z')
       return state.misa->extension_enabled(ext);
     else
-      return isa->extension_enabled(ext);
+      return extension_enable_table[ext];
   }
   // Is this extension enabled? and abort if this extension can
   // possibly be disabled dynamically. Useful for documenting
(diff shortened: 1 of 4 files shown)


riscv/processor.h
```diff
@@ -232,12 +232,17 @@ class processor_t : public abstract_device_t
...
   bool extension_enabled_const(isa_extension_t ext) const {
-    if (ext >= 'A' && ext <= 'Z')
+    if (ext >= 'A' && ext <= 'Z') {
       return state.misa->extension_enabled_const(ext);
-    else
-      return isa->extension_enabled(ext);  // assume this can't change
+    } else {
+      assert(!extension_dynamic[ext]);
+      extension_assumed_const[ext] = true;
+      return extension_enabled(ext);
```
liweiwei90, line 240: This way seems not necessary now since extension_enabled_const(EXT_ZCMT) has been changed to  extension_enabled(EXT_ZCMT) in the previous commit and we can use dynamic extensions directly here (it can be initialized just from isa string).

aswaterman: Right.  But I think it was @scottj97's intention to future-proof the implementation--which, given he was able to do without duplicating the complicated enable logic, makes sense to me.

riscv/processor.h
```diff
@@ -304,6 +312,10 @@ class processor_t : public abstract_device_t
   bool check_triggers_icount;
   std::vector<bool> impl_table;
 
+  std::bitset<NUM_ISA_EXTENSIONS> extension_enable_table;
```
scottj97, line 316: I would add a comment here: "Note: does not include single-letter extensions in misa"

scottj97 (changes requested):
I would squash 54a4089f20390c86c1a356fefda1ebf31b24d93f back into ca6ef8e3eed702e928b12d4a6c9791a9e4d1f13c since that will keep everything working as intended at every step of the way. Also the commit message for 54a4089f20390c86c1a356fefda1ebf31b24d93f doesn't all make sense anymore so it can just be discarded (but maybe keep the comment about `mutable`).

aswaterman:
OK.  Attempt number 6, or whatever we're on now :)

scottj97 (approved):

PR #1246, opened 2023-02-06
Support discontiguous hart IDs in debug module

aswaterman:
Supersedes #1166

@scottj97 I'm backtracking on what I wrote earlier.  I think it's worth the effort to audit all of the places where we assume hart IDs are contiguous, because some of them have bad code smell.  Then we can support discontiguous hart IDs directly, rather than constructing all of them and skipping over some.  Are you OK with this direction?

This PR is sufficient to fix the segfault mentioned in #1166 and to make the debug module tolerate discontiguous hart IDs, but there are still other places where the behavior is surprising, if not incorrect: at minimum, the CLINT and the DeviceTree.  I've already done some of the work to fix them, but I'd like to reach agreement on this PR before continuing down that path.

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -32,18 +32,16 @@ static unsigned field_width(unsigned n)
 ///////////////////////// debug_module_t
 
 debug_module_t::debug_module_t(sim_t *sim, const debug_module_config_t &config) :
-  nprocs(sim->nprocs()),
   config(config),
   program_buffer_bytes((config.support_impebreak ? 4 : 0) + 4*config.progbufsize),
   debug_progbuf_start(debug_data_start - program_buffer_bytes),
   debug_abstract_start(debug_progbuf_start - debug_abstract_size*4),
   custom_base(0),
-  hartsellen(field_width(sim->nprocs())),
   sim(sim),
   // The spec lets a debugger select nonexistent harts. Create hart_state for
   // them because I'm too lazy to add the code to just ignore accesses.
-  hart_state(1 << field_width(sim->nprocs())),
-  hart_array_mask(sim->nprocs()),
+  hart_state(1 << field_width(sim->get_cfg().max_hartid() + 1)),
+  hart_array_mask(sim->get_cfg().max_hartid() + 1),
   rti_remaining(0)
 {
   D(fprintf(stderr, "debug_data_start=0x%x\n", debug_data_start));
@@ -51,9 +49,9 @@ debug_module_t::debug_module_t(sim_t *sim, const debug_module_config_t &config) 
   D(fprintf(stderr, "debug_abstract_start=0x%x\n", debug_abstract_start));
 
   const unsigned max_procs = 1024;
-  if (nprocs > max_procs) {
-    fprintf(stderr, "At most %u processors are supported (%u requested)\n",
-            max_procs, nprocs);
+  if (sim->get_cfg().max_hartid() >= max_procs) {
+    fprintf(stderr, "Hart IDs must not exceed %u (%zu harts with max hart ID %zu requested)\n",
+            max_procs - 1, sim->get_cfg().nprocs(), sim->get_cfg().max_hartid());
     exit(1);
   }
 
@@ -85,11 +83,8 @@ debug_module_t::~debug_module_t()
 
 void debug_module_t::reset()
 {
-  assert(sim->nprocs() > 0);
-  for (unsigned i = 0; i < sim->nprocs(); i++) {
-    processor_t *proc = sim->get_core(i);
-    if (proc)
-      proc->halt_request = proc->HR_NONE;
+  for (const auto& [hart_id, hart] : sim->get_harts()) {
+    hart->halt_request = hart->HR_NONE;
   }
 
   memset(&dmcontrol, 0, sizeof(dmcontrol));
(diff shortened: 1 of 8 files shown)


spike_main/spike.cc
```diff
@@ -299,10 +299,28 @@ static std::vector<int> parse_hartids(const char *s)
 
   int n;
   while (stream >> n) {
+    if (n < 0) {
```
scottj97, line 302: Why not just make `n` et al into `uint32_t`? That's the type passed into `processor_t::processor_t()`.

riscv/sim.h
```diff
@@ -51,8 +52,11 @@ class sim_t : public htif_t, public simif_t
   }
   const char* get_dts() { if (dts.empty()) reset(); return dts.c_str(); }
   processor_t* get_core(size_t i) { return procs.at(i); }
+  const cfg_t &get_cfg() { return *cfg; }
   unsigned nprocs() const { return procs.size(); }
 
+  const std::map<size_t, processor_t*>& get_harts() { return harts; }
```
scottj97, line 58: 2f5112f9dd94e0870f37711f3d99df4bfa6461d0 does far more than its commit message claims

scottj97 (commented):
I'm good with the approach.

timsifive:
FYI, OpenOCD currently assumes hart IDs and debug indices map 1-to-1, which won't work with discontiguous hart IDs. That's not a reason for spike to be held back.

timsifive (commented):
It looks like with this change, there's a 1-to-1 mapping of hart ID and hart index (as written to e.g. dmcontrol.hartsel). That is not right.

The debug module selects harts from 0 to N, continuously. These numbers usually correspond exactly to hart IDs, but the spec does not require that. So with sparse harts, you would have something like:
```
hartsel=0, hartid 0
hartsel=1, hartid 16
```

riscv/sim.h
aswaterman: Rebase screwup.  Most of that stuff should be part of the final commit.

aswaterman:
@timsifive Thanks--that's the kind of review I need, because it's the kind of assumption that's not at all obvious when the 1:1 assumption is baked into the code.

spike_main/spike.cc
aswaterman: Using `uint32_t` would ignore a negative number rather than rejecting it.  But we can do what you suggested and improve robustness further by checking the stream is empty after the loop.

scottj97: There's also a mix of `size_t` and `uint32_t` elsewhere used for hartids. I don't know if that's anything to worry about.

timsifive:
> Thanks--that's the kind of review I need, because it's the kind of assumption that's not at all obvious when the 1:1 assumption is baked into the code.

It's a short-cut I took way back when, and I guess I didn't even comment it. :(

spike_main/spike.cc
aswaterman: Yeah, I'm looking into that, too.

aswaterman: Turns out `stream >> unsigned` has unintuitive behavior (it happily accepts negative integers takes them mod 2^width) and so the existing code is actually reasonable.  I'm going to mark this as resolved.  (I'm still going to handle the `size_t` vs. `uint32_t` vs. `int` matter elsewhere, though.)

aswaterman:
@timsifive please review in detail again.  I think I got it right, but there are additional places where it's not obvious whether the contiguous `hartsel` space or discontiguous hart ID space should be used.  (I added `selected_hart_id()`, which returns the hart ID corresponding to `hartsel`, and `selected_hart_state()`, for DRY purposes.)

@scottj97 I think this is ready for your re-review, as well.

riscv/debug_module.cc
```diff
@@ -32,28 +32,26 @@ static unsigned field_width(unsigned n)
...
   debug_abstract_start(debug_progbuf_start - debug_abstract_size*4),
   custom_base(0),
-  hartsellen(field_width(sim->nprocs())),
   sim(sim),
   // The spec lets a debugger select nonexistent harts. Create hart_state for
   // them because I'm too lazy to add the code to just ignore accesses.
-  hart_state(1 << field_width(sim->nprocs())),
-  hart_array_mask(sim->nprocs()),
+  hart_state(1 << field_width(sim->get_cfg().max_hartid() + 1)),
+  hart_array_mask(sim->get_cfg().max_hartid() + 1),
```
timsifive, line 44: If these continue to be indexed by hart ID, they should probably become std::map in case people start using large hart IDs.

scottj97 (approved):
I did not review the debug module changes closely, nor do I have good test coverage of that code, but this doesn't seem to break anything else, at least.

aswaterman:
Thanks. I’ll wait for Tim’s scrutiny before merging.

riscv/debug_module.cc
```diff
@@ -487,7 +463,7 @@ bool debug_module_t::dmi_read(unsigned address, uint32_t *value)
           unsigned base = hawindowsel * 32;
           for (unsigned i = 0; i < 32; i++) {
             unsigned n = base + i;
-            if (n < nprocs && hart_array_mask[n]) {
+            if (n < hart_array_mask.size() && hart_array_mask[n]) {
```
timsifive: n here is a debug hart index, not a hart ID. So can't directly index into hart_array_mask.

riscv/debug_module.cc
```diff
@@ -846,15 +817,15 @@ bool debug_module_t::dmi_write(unsigned address, uint32_t value)
...
         return true;
 
       case DM_HAWINDOW:
         {
           unsigned base = hawindowsel * 32;
           for (unsigned i = 0; i < 32; i++) {
             unsigned n = base + i;
-            if (n < nprocs) {
+            if (sim->get_harts().find(n) != sim->get_harts().end()) {
               hart_array_mask[n] = (value >> i) & 1;
```
timsifive: Likewise, n is a hart index, not a hart ID.

timsifive (changes requested):
Mostly looks good. Just a couple of small comments.

riscv/debug_module.cc
aswaterman: We're still capping hart IDs as relatively small numbers, and there remain other (much larger) structures that allocate space proportional to the max hart ID.  We can revisit this if we really need to support hart IDs in the hundreds of millions.

aswaterman:
@timsifive can you re-review?  All that has changed are the two places where I used a debug index as a hart ID.

timsifive (approved):

PR #1250, opened 2023-02-15
Fix writable misa.F/V

liweiwei90:
- Disable Zfh{min} when misa.F is cleared
- Disable Zvfh{min} when misa.V is cleared
- Disable Zvfh when Zfhmin is disabled

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -642,6 +642,8 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   proc->set_extension_enable(EXT_ZCMT, proc->extension_enabled(EXT_ZCA));
   proc->set_extension_enable(EXT_ZFH, new_misa & (1L << ('F' - 'A')));
   proc->set_extension_enable(EXT_ZFHMIN, new_misa & (1L << ('F' - 'A')));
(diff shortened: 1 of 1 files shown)


scottj97 (approved):
I am increasingly concerned about the feature dependencies that must be expressed in two different ways (in [isa_parser.cc](https://github.com/riscv-software-src/riscv-isa-sim/blob/0e4f928b3cf53b323ccb5064f3f7d58d351a3060/riscv/isa_parser.cc#L100-L116) and also here in [csrs.cc](https://github.com/riscv-software-src/riscv-isa-sim/blob/0e4f928b3cf53b323ccb5064f3f7d58d351a3060/riscv/csrs.cc#L629-L642)). Also see [my comment here](https://github.com/riscv-software-src/riscv-isa-sim/commit/a883a5839dc1f09e5fc8aa80e563f4c841c4a4bc#commitcomment-99253767).

I'm okay with this change today but I'd like us all to think about how we can simplify the complex and duplicated descriptions of ISA dependencies.

aswaterman (approved):
Yes, I agree.  This is only going to get more complex going forward.  If the dependences in `isa_parser_t` were expressed as rules in a tabular form, we could avoid duplicating this logic.

Issue #1253, opened 2023-02-17
why update context when change plic threshold? 


WadeGao:
I have some questions about the function `bool plic_t::context_write(plic_context_t *c, reg_t offset, uint32_t val)` in riscv/plic.cc

when `offset == CONTEXT_THRESHOLD` and  `val > max_prio`, we set `update` and call `context_update` later. bu nothing of the current context changed. 

why update the context is necessary when meeting this case? 

![image](https://user-images.githubusercontent.com/49817941/219608985-fa18e288-4641-4047-99a7-cdabff841dfb.png)

aswaterman:
Generally, we prefer references to specific lines of code, rather than screenshots, which destroy much of the information.

@avpatel FYI this seems like your block

avpatel:
@WadeGao This looks like a bug which was introduced unintentionally. The "update" should be set whenever the context threshold is written so "update = true" should be under "if ()" and not under "else".

WadeGao:
> Generally, we prefer references to specific lines of code, rather than screenshots, which destroy much of the information.
> 
> @avpatel FYI this seems like your block

Thank you, I am afraid of that there may be some change of this file during our discussion. I will pay more attention to the open source community rules next time.

PR #1254, opened 2023-02-17
fix unintentionally plic context update

WadeGao:
fix a bug, the issue id is #1253 

unintentionally plic context update when:
- `offset == CONTEXT_THRESHOLD`
- `val > max_prio`

nothing of the current context changed, but still set `update` and call `context_update` later, this is a bug

### DIFF
--- a/riscv/plic.cc
+++ b/riscv/plic.cc
@@ -242,10 +242,12 @@ bool plic_t::context_write(plic_context_t *c,
   switch (offset) {
     case CONTEXT_THRESHOLD:
       val &= ((1 << PLIC_PRIO_BITS) - 1);
-      if (val <= max_prio)
+      if (val <= max_prio) {
         c->priority_threshold = val;
-      else
         update = true;
+      } else {
+        ret = false;
+      }
       break;
     case CONTEXT_CLAIM: {
       uint32_t id_word = val / 32;
@@ -259,7 +261,6 @@ bool plic_t::context_write(plic_context_t *c,
     }
     default:
       ret = false;
-      update = true;
       break;
   };
 


riscv/plic.cc
```diff
@@ -259,7 +262,7 @@ bool plic_t::context_write(plic_context_t *c,
     }
     default:
       ret = false;
-      update = true;
+      update = false;
```
aswaterman: Delete this line.  `update` is initialized to false, so setting it to false here is redundant.

riscv/plic.cc
```diff
@@ -242,10 +242,13 @@ bool plic_t::context_write(plic_context_t *c,
   switch (offset) {
     case CONTEXT_THRESHOLD:
       val &= ((1 << PLIC_PRIO_BITS) - 1);
-      if (val <= max_prio)
+      if (val <= max_prio) {
         c->priority_threshold = val;
-      else
         update = true;
+      }
+      else {
```
aswaterman: brace on same line as else, i.e. `} else {`

WadeGao: ok, fixed

riscv/plic.cc
WadeGao: I think it's more readable by assign true explicitly rather than deduce its value by source code context, so I add this line.  

I will make some change.👌🏻

aswaterman:
Thanks.  Finally, can you squash these into a single commit, so that the history stays clean?

WadeGao:
> Thanks.  Finally, can you squash these into a single commit, so that the history stays clean?

No problem👌🏻The recent conversations were operated by my Android Phone, so I will use my pc to squash it later.

Thank you for your detailed code review

aswaterman:
Cool.

WadeGao:
> Thanks. Finally, can you squash these into a single commit, so that the history stays clean?

done

aswaterman:
Seems like it didn't work; there are still three commits in the history.

WadeGao:
> Seems like it didn't work; there are still three commits in the history.

sorry for my disturb..... github has its cache, I have to refresh it even though I squash my commits and push it to the remote forked repo.

aswaterman (approved):
thanks again!

PR #1257, opened 2023-02-21
Implement mcontrol6.hit

YenHaoChen:
The debug spec revised the mcontrol6.hit in https://github.com/riscv/riscv-debug-spec/pull/795. This PR implements the revised behavior.


YenHaoChen:
The hit is separated into hit0 and hit1, i.e., hit={hit1,hit0}. I need help finding a more elegant implementation for the LSB and MSB operation on the hit field.
> tdata1 = set_field(tdata1, CSR_MCONTROL6_HIT1, hit & 2); // MSB of 2-bit field
> ...
> tdata1 = set_field(tdata1, CSR_MCONTROL6_HIT0, hit & 1); // LSB of 2-bit field
> ...
> hit = hit_t(2 * get_field(val, CSR_MCONTROL6_HIT1) + get_field(val, CSR_MCONTROL6_HIT0)); // 2-bit field {hit1,hit0}

YenHaoChen:
The CI fails. I passed the compilation on my PC and need help analyzing the failure.

aswaterman:
CI fails with:

```
../install/include/riscv/triggers.h:8:10: fatal error: debug_defines.h: No such file or directory
    8 | #include "debug_defines.h"
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
Error: Process completed with exit code 1.
```

aswaterman:
This failure mode indicates that a public header is including a private header: `triggers.h` should not include `debug_defines.h`.  (The alternative is to install `debug_defines.h`, but we really should not make that part of the API, so I suggest changing `triggers.h` so it does not need to rely on `debug_defines.h`.)

YenHaoChen:
> This failure mode indicates that a public header is including a private header: `triggers.h` should not include `debug_defines.h`. (The alternative is to install `debug_defines.h`, but we really should not make that part of the API, so I suggest changing `triggers.h` so it does not need to rely on `debug_defines.h`.)

There are two methods to avoid reliance on ```debug_defines.h```.
1. Define the macros in ```encoding.h```.
2. Hard-code the values in ```triggers.h```.

I choose the second method for simplicity.

timsifive:
Please hold off on merging this. The debug spec is tweaking this change a bit more. It should not have been merged into the spec yet. (The actual change relative to this PR will be minor, so this code will probably all apply.)

YenHaoChen:
It seems the debug spec has finalized the mcontrol6.hit. I believe this code of the PR provides the desired behavior.

riscv/triggers.cc
```diff
@@ -250,11 +250,11 @@ reg_t mcontrol6_t::tdata1_read(const processor_t * const proc) const noexcept {
   reg_t tdata1 = 0;
   tdata1 = set_field(tdata1, CSR_MCONTROL6_TYPE(xlen), CSR_TDATA1_TYPE_MCONTROL6);
   tdata1 = set_field(tdata1, CSR_MCONTROL6_DMODE(xlen), dmode);
+  tdata1 = set_field(tdata1, CSR_MCONTROL6_HIT1, hit & 2); // MSB of 2-bit field
```
timsifive: Shouldn't that be `hit >> 1`? I'm not exactly sure what `set_field` does, but I don't think it turns its argument into a bool for single-bit fields.

riscv/triggers.h
```diff
@@ -228,12 +235,22 @@ class mcontrol_t : public mcontrol_common_t {
 public:
   virtual reg_t tdata1_read(const processor_t * const proc) const noexcept override;
   virtual void tdata1_write(processor_t * const proc, const reg_t val, const bool allow_chain) noexcept override;
+
+  virtual void set_hit(hit_t UNUSED val) override { hit = true; }
```
timsifive: Seems safer to do: `hit = val != HIT_FALSE;`

riscv/triggers.cc
```diff
@@ -215,7 +215,7 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
   if (simple_match(xlen, value)) {
     /* This is OK because this function is only called if the trigger was not
      * inhibited by the previous trigger in the chain. */
-    hit = true;
+    set_hit(timing ? HIT_IMMEDIATELY_AFTER : HIT_BEFORE);
```
timsifive, line 236: Does this mean that for mcontrol6 triggers, timing is always immediately_after? That's a compliant implementation, but it would be better to implement the suggested timings:
image

riscv/triggers.h
YenHaoChen: The hit variable has the type hit_t, which cannot be assigned with a bool value.

riscv/triggers.h
```diff
@@ -51,6 +51,13 @@ struct match_result_t {
   action_t action;
 };
 
+typedef enum {
+  HIT_FALSE = 0,
+  HIT_BEFORE = 1,
+  HIT_AFTER = 2,
+  HIT_IMMEDIATELY_AFTER = 3
```
timsifive, line 58:
```suggestion
  HIT_FALSE = CSR_MCONTROL6_HIT0_FALSE,
  HIT_BEFORE = CSR_MCONTROL6_HIT0_BEFORE,
  HIT_AFTER = CSR_MCONTROL6_HIT0_AFTER,
  HIT_IMMEDIATELY_AFTER = CSR_MCONTROL6_HIT0_IMMEDIATELY_AFTER
```

riscv/triggers.cc
```diff
@@ -273,9 +273,9 @@ void mcontrol6_t::tdata1_write(processor_t * const proc, const reg_t val, const
   dmode = get_field(val, CSR_MCONTROL6_DMODE(xlen));
   vs = get_field(val, CSR_MCONTROL6_VS);
   vu = get_field(val, CSR_MCONTROL6_VU);
-  hit = get_field(val, CSR_MCONTROL6_HIT);
+  hit = hit_t(2 * get_field(val, CSR_MCONTROL6_HIT1) + get_field(val, CSR_MCONTROL6_HIT0)); // 2-bit field {hit1,hit0}
   select = get_field(val, CSR_MCONTROL6_SELECT);
-  timing = legalize_timing(val, CSR_MCONTROL6_TIMING, CSR_MCONTROL6_SELECT, CSR_MCONTROL6_EXECUTE, CSR_MCONTROL6_LOAD);
+  timing = legalize_timing(val, 0, CSR_MCONTROL6_SELECT, CSR_MCONTROL6_EXECUTE, CSR_MCONTROL6_LOAD);
```
timsifive: I think this means:
* data load triggers trigger right after the event
* all other triggers trigger right before the event

That means you can't combine a data load trigger with an address trigger to trigger on a load of a value at a given address. gdb doesn't support setting triggers in that way so possibly we don't care. I think it's worth clarifying the behavior with a comment, since it's not super obvious just from reading this code.

YenHaoChen: Add a comment in https://github.com/riscv-software-src/riscv-isa-sim/pull/1257/commits/33e992d5e4bd23cac673330efcf833e6bd3918f3 to clarify the behavior.

riscv/triggers.h
YenHaoChen: This requires the debug_defines.h to the triggers.h. The https://github.com/riscv-software-src/riscv-isa-sim/pull/1257#issuecomment-1438066439 wants the triggers.h need not to rely on the debug_defines.h. 

@timsifive I personally believe the debug_defines.h is a public header. Should we consider installing the debug_defines.h to the CI platform?

timsifive (approved):

aswaterman:
@YenHaoChen Is this ready to merge?  You might want to rebase on top of master since almost a year has elapsed.

YenHaoChen:
@aswaterman 
I do a review again today. 

+ The mcontrol6 trigger (hit0/hit1) is finalized (frozen) before this commit.
+ The Spike repository has not updated the header file "debug_defines.h" since this PR. Thus, the header file in this PR would work, although it is not the newest one from the debug repository.
+ After automatic rebasing (without a conflict), I could successfully compile the Spike binary on my PC.

I believe this PR is ready to merge.

rtwfroody:
I don't see the update to tinfo.version to indicate that mcontrol6 has the new hit0/hit1 fields.

YenHaoChen:
@rtwfroody Thanks for the note. I overlooked the tinfo.version. 

Bump the header "debug_defines.h" for the tinfo.version.

aswaterman (approved):

PR #1261, opened 2023-02-23
fesvr: elfloader: replace asserts after open and mmap by exceptions

viktor-prutyanov:
Asserts (especially without a message) aren't human readable way of error reporting. So, replace them by exceptions with messages with errno string.

Before:
```
spike: ../fesvr/elfloader.cc:30: std::map<std::__cxx11::basic_string<char>, long unsigned int> load_elf(const char*, memif_t*, reg_t*, unsigned int): Assertion `buf != MAP_FAILED' failed.
```
After:
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Specified ELF can't be mapped: No such device
```


aswaterman (approved):
LGTM, thanks for the improvement.

PR #1262, opened 2023-02-26
Fix --bootargs default value in spike usage help() function

jolivain:
Commit 191634d285 changed the default kernel boot args from "console=hvc0 earlycon=sbi" to "console=ttyS0 earlycon", but didn't updated the spike usage help() function.

This commit realign the usage help string with the actual value.

### DIFF
--- a/riscv/dts.cc
+++ b/riscv/dts.cc
@@ -34,10 +34,10 @@ std::string make_dts(size_t insns_per_rtc_tick, size_t cpu_hz,
     s << "    linux,initrd-start = <" << (size_t)initrd_start << ">;\n"
          "    linux,initrd-end = <" << (size_t)initrd_end << ">;\n";
     if (!bootargs)
-      bootargs = "root=/dev/ram console=ttyS0 earlycon";
+      bootargs = "root=/dev/ram " DEFAULT_KERNEL_BOOTARGS;
   } else {
     if (!bootargs)
-      bootargs = "console=ttyS0 earlycon";
+      bootargs = DEFAULT_KERNEL_BOOTARGS;
   }
     s << "    bootargs = \"";
   for (size_t i = 0; i < strlen(bootargs); i++) {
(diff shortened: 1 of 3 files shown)


jerryz123 (changes requested):
The default should be defined in a single place in a macro, like how DEFAULT_ISA is defined.

jolivain:
Hi,

I had few autoconf issues while implementing the proposed changes. I added few commits to address those issues, and regenerated the autoconf file with version 2.71 (the current latest version). Please tell me if you prefer I regenerate with another version (e.g. 2.69) or if other changes are needed.

Thanks! Best regards.

scottj97:
Thank you for updating the autoconf stuff, that is great.

In the case of `--bootargs` can you give some justification why it should be a configure-time option? What is the use case there, versus just a Spike runtime option?

jolivain:
Hi @scottj97

I understood the recommendation from @jerryz123  to do like the DEFAULT_ISA, including the AC_ARG_WITH() part.

Having this configure-time parameter just allow to change the defaults, for convenience.
For example, to allow the "old" SBI console behavior with a quiet kernel, configure spike with:

    ./configure --with-bootargs="console=hvc0 earlycon=sbi quiet"

Then, one could deviate from this new default with the runtime `--bootargs`.

If you think this is superfluous, I can rework the patch to remove the:

    AC_ARG_WITH(bootargs, ...)

and only keep the cpp macro:

    AC_DEFINE_UNQUOTED([DEFAULT_BOOTARGS], ["console=ttyS0 earlycon"], [Default value kernel bootargs])

What do you think?

jerryz123:
Apologies, I only meant that the default string should be deduplicated, I don't think its necessary to make it a compile-time flag.

platform.h might be a good place to store this, it looks like DEFAULT_RSTVEC is already there. @scottj97 what do you think?

jolivain:
Hi Jerry,

Thanks for your feedback. I updated the patch following your recommendation. I also removed autoconf patches from this pull request (since they are no longer needed here).  I'll propose them in another pull request.

Best regards,

Julien.

scottj97 (approved):
Excellent; thank you for this improvement.

jerryz123 (approved):

aswaterman (approved):

PR #1263, opened 2023-02-27
Improve icount trigger implementation

scottj97:
* Minor but easy performance improvement
* Initialization of icount fields
* Decrement counters on external interrupts

The current lack of initialization doesn't hurt anything today, because we don't instantiate any `icount_t` except right before calling `tdata1_write()` which then initializes all these fields. But in my design (in my fork of Spike) I have a fixed set of triggers which I initialize at reset, and the fields were coming up with garbage in them.

cc @YenHaoChen

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -263,7 +263,7 @@ void processor_t::step(size_t n)
             state.single_step = state.STEP_STEPPED;
           }
 
-          if (!state.serialized) {
+          if (!state.serialized && check_triggers_icount) {
             auto match = TM.detect_icount_match();
             if (match.has_value()) {
               assert(match->timing == triggers::TIMING_BEFORE);
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -696,6 +696,7 @@ void processor_t::take_interrupt(reg_t pending_interrupts)
     else
       abort();
 
+    if (check_triggers_icount) TM.detect_icount_match();
     throw trap_t(((reg_t)1 << (isa->get_max_xlen() - 1)) | ctz(enabled_interrupts));
   }
 }
--- a/riscv/triggers.h
+++ b/riscv/triggers.h
@@ -249,11 +249,11 @@ public:
   virtual std::optional<match_result_t> detect_icount_match(processor_t * const proc) noexcept override;
 
 private:
-  bool dmode;
-  bool hit;
-  unsigned count, count_read_value;
-  bool pending, pending_read_value;
-  action_t action;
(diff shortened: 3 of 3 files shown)


aswaterman (approved):

scottj97:
I added another commit to fix a bug where an external interrupt was not decrementing the count like it should.

aswaterman (approved):

scottj97:
I don't understand the CI build failure, but it doesn't seem to have anything to do with this change.

aswaterman:
@scottj97 It looks like a spurious failure of some kind, since https://github.com/riscv-software-src/riscv-isa-sim/releases/download/dummy-tag-for-ci-storage/spike-ci.tar is clearly accessible.  I saw a similar one, at a different phase of the build, last week.  Feel free to merge anyway (or wait til the reexecuted build passes).

scottj97:
I've added d69a788be7feea8f529cb2dab0e9854f665055f5 which I'm not 100% sure about. Would like to get @YenHaoChen and/or @timsifive feedback on this.

Edit: fixed an obvious bug and updated the git hash. I have no test coverage for this. :(

aswaterman:
d69a788be7feea8f529cb2dab0e9854f665055f5 passes my smell test; feel free to merge when you feel that you've gotten sufficient feedback from the others.

timsifive (approved):
Nothing is obviously wrong here, but I don't know much about the core loops being modified here.

There is an icount trigger smoke test in riscv-tests that could be run in CI if you edit debug-smoke.yml. OpenOCD needs 87f9e590b9d93d1be8f315ef1ab008b94a4e2098 or later, and riscv-tests needs 41eeb4aaee6a510cf471085bf7f897b8114ef452 or later.

scottj97:
> Nothing is obviously wrong here, but I don't know much about the core loops being modified here.

Thanks Tim. My question was more about the spec: should a pending icount breakpoint exception be taken even when textra doesn't match? See commit message in d69a788be7feea8f529cb2dab0e9854f665055f5

timsifive:
I think you interpreted the spec correctly, and I'm not sure the spec says what we want. I opened https://github.com/riscv/riscv-debug-spec/issues/812 to discuss that.

scottj97:
After [debug spec clarification](https://github.com/riscv/riscv-debug-spec/pull/826) I've removed the last commit, and rebased onto the latest master.

Would appreciate a re-review from @aswaterman @YenHaoChen

aswaterman (approved):
LGTM.

PR #1264, opened 2023-02-27
Allow HTIF to make forward progress while running inside interactive mode

aswaterman:
https://github.com/riscv-software-src/riscv-isa-sim/pull/1225#issuecomment-1442190137 pointed out that the HTIF doesn't get to run once the interactive mode takes control.  Fix this by periodically returning control to the HTIF.

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -252,6 +252,12 @@ static std::string readline(int fd)
 
 void sim_t::interactive()
 {
+  if (next_interactive_action.has_value()) {
+    auto f = next_interactive_action.value();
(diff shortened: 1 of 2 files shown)


scottj97 (approved):
I didn't study the code but I confirm this fixes a hang that I saw after #1225.

aswaterman:
Thanks.

It’s a situation where I’d love to have coroutines.

scottj97:
@aswaterman this fails compilation with `-O0`:
```
../riscv-isa-sim/riscv/interactive.cc:406: undefined reference to `sim_t::INTERLEAVE'
```

PR #1265, opened 2023-02-28
PLIC and CLINT fixes for heterogeneous harts/discontiguous hart IDs

aswaterman:
I went through the PLIC and CLINT to support discontiguous hart IDs.  It turns out the PLIC already supports them, but it wasn't handling the case of some harts having S-mode and some not.

Also, both the PLIC and CLINT were broken on big-endian hosts.  Especially for the CLINT, the fix for that dovetailed with the discontiguous hart ID fix.

### DIFF
--- a/riscv/clint.cc
+++ b/riscv/clint.cc
@@ -1,9 +1,10 @@
 #include <sys/time.h>
 #include "devices.h"
 #include "processor.h"
+#include "sim.h"
 
-clint_t::clint_t(std::vector<processor_t*>& procs, uint64_t freq_hz, bool real_time)
-  : procs(procs), freq_hz(freq_hz), real_time(real_time), mtime(0), mtimecmp(procs.size())
+clint_t::clint_t(sim_t* sim, uint64_t freq_hz, bool real_time)
+  : sim(sim), freq_hz(freq_hz), real_time(real_time), mtime(0)
 {
   struct timeval base;
 
@@ -29,16 +30,27 @@ clint_t::clint_t(std::vector<processor_t*>& procs, uint64_t freq_hz, bool real_t
 
 bool clint_t::load(reg_t addr, size_t len, uint8_t* bytes)
 {
+  if (len > 8)
+    return false;
+
   increment(0);
-  if (addr >= MSIP_BASE && addr + len <= MSIP_BASE + procs.size()*sizeof(msip_t)) {
-    std::vector<msip_t> msip(procs.size());
-    for (size_t i = 0; i < procs.size(); ++i)
-      msip[i] = !!(procs[i]->state.mip->read() & MIP_MSIP);
-    memcpy(bytes, (uint8_t*)&msip[0] + addr - MSIP_BASE, len);
-  } else if (addr >= MTIMECMP_BASE && addr + len <= MTIMECMP_BASE + procs.size()*sizeof(mtimecmp_t)) {
-    memcpy(bytes, (uint8_t*)&mtimecmp[0] + addr - MTIMECMP_BASE, len);
-  } else if (addr >= MTIME_BASE && addr + len <= MTIME_BASE + sizeof(mtime_t)) {
-    memcpy(bytes, (uint8_t*)&mtime + addr - MTIME_BASE, len);
+
+  if (addr >= MSIP_BASE && addr < MTIMECMP_BASE) {
+    if (len == 8) {
+      // Implement double-word loads as a pair of word loads
+      return load(addr, 4, bytes) && load(addr + 4, 4, bytes + 4);
+    }
+
+    const auto hart_id = (addr - MSIP_BASE) / sizeof(msip_t);
+    const msip_t res = sim->get_harts().count(hart_id) && (sim->get_harts().at(hart_id)->state.mip->read() & MIP_MSIP);
+    read_little_endian_reg(res, addr, len, bytes);
+    return true;
+  } else if (addr >= MTIMECMP_BASE && addr < MTIME_BASE) {
+    const auto hart_id = (addr - MTIMECMP_BASE) / sizeof(mtimecmp_t);
+    const mtime_t res = sim->get_harts().count(hart_id) ? mtimecmp[hart_id] : 0;
+    read_little_endian_reg(res, addr, len, bytes);
+  } else if (addr >= MTIME_BASE && addr < MTIME_BASE + sizeof(mtime_t)) {
+    read_little_endian_reg(mtime, addr, len, bytes);
   } else if (addr + len <= CLINT_SIZE) {
     memset(bytes, 0, len);
   } else {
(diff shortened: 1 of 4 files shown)


riscv/plic.cc
```diff
@@ -66,21 +67,28 @@
...
-  for (size_t i = 0; i < contexts.size(); i++) {
-    plic_context_t* c = &contexts[i];
-    memset(c, 0, sizeof(*c));
-    c->proc = procs[i / contexts_per_hart];
-    c->mmode = (i % contexts_per_hart == 0);
+  // PLIC contexts are contiguous in memory even if harts are discontiguous.
+  for (const auto& [hart_id, hart] : sim->get_harts()) {
+    contexts.push_back(plic_context_t());
+    plic_context_t* m_context = &contexts.back();
+    memset(m_context, 0, sizeof(*m_context));
```
scottj97: Yuck. Why not let `plic_context_t`'s constructor initialize all the fields? Including `proc` and `mmode`

riscv/plic.cc
```diff
@@ -340,17 +340,15 @@ bool plic_t::load(reg_t addr, size_t len, uint8_t* bytes)
     }
   }
 
-  if (ret) {
```
scottj97, line 343: The removal of the conditional here should be its own commit, so in case it happens to cause some problem, `git bisect` can pinpoint it.

riscv/plic.cc
```diff
@@ -340,17 +340,15 @@ bool plic_t::load(reg_t addr, size_t len, uint8_t* bytes)
...
-    memcpy(bytes, (uint8_t *)&val, len);
-  }
+  read_little_endian_reg(val, addr, len, bytes);
```
scottj97, line 332: 09db15baf422b0af769effe3ad5f09b722f619cf contains only dead code, which impairs `git bisect`; I'd squash it with this commit.

scottj97 (changes requested):

riscv/clint.cc
```diff
@@ -29,16 +30,27 @@ clint_t::clint_t(std::vector<processor_t*>& procs, uint64_t freq_hz, bool real_t
...
-    memcpy(bytes, (uint8_t*)&mtime + addr - MTIME_BASE, len);
+
+  if (addr >= MSIP_BASE && addr < MTIMECMP_BASE) {
+    if (len == 8) {
+      // Implement double-word loads as a pair of word loads
+      return load(addr, 4, bytes) && load(addr + 4, 4, bytes + 4);
+    }
+
+    const auto hart_id = (addr - MSIP_BASE) / sizeof(msip_t);
+    const msip_t res = sim->get_harts().count(hart_id) ? sim->get_harts().at(hart_id)->state.mip->read() & MIP_MSIP : 0;
```
scottj97: This causes MSIP to read out as 8 instead of 1

riscv/plic.cc
aswaterman: Ack.

riscv/clint.cc
aswaterman: Oops. Thanks for catching that.

riscv/plic.cc
aswaterman: Yes, that’s better still.

riscv/plic.cc
aswaterman: Ack.

riscv/plic.cc
aswaterman: For my edification, why do commits like this impair bisect?  Is it merely that a longer history makes it take longer to bisect, or is there something more subtle that I'm missing?

scottj97: If there was a bug in the dead code, `git bisect` would point to the following commit (where the dead code was first executed) instead of the commit where the bug was actually introduced.

Edit: it's not just for git bisect, though; I want every commit to be perfect, and dead code is not perfect. But perhaps that's overly pedantic.

aswaterman: Thanks for the explanation.  I think we should run with your preferred approach as a guideline, not a hard-and-fast rule.  Sometimes, the dead code is substantial enough that separating it into its own commit is justifiable on the basis of reducing cognitive load.  (This isn't such a case.)

riscv/devices.h
```diff
@@ -79,6 +79,11 @@ class clint_t : public abstract_device_t {
 #define PLIC_MAX_DEVICES 1024
 
 struct plic_context_t {
+  plic_context_t(processor_t* proc, bool mmode)
+    : proc(proc), mmode(mmode), priority_threshold(0), enable{}, pending{},
```
scottj97, line 85: Is there any significance to the mix of parentheses and curly braces?

scottj97: I think it would be preferable to use in-class member initializers for everything but `proc` and `mmode`. E.g:
```
uint8_t priority_threshold{};
uint32_t enable[PLIC_MAX_DEVICES/32]{};
```

scottj97 (approved):

riscv/clint.cc
```diff
@@ -29,16 +30,27 @@ clint_t::clint_t(std::vector<processor_t*>& procs, uint64_t freq_hz, bool real_t
...
+
+  if (addr >= MSIP_BASE && addr < MTIMECMP_BASE) {
+    if (len == 8) {
+      // Implement double-word loads as a pair of word loads
+      return load(addr, 4, bytes) && load(addr + 4, 4, bytes + 4);
+    }
+
+    const auto hart_id = (addr - MSIP_BASE) / sizeof(msip_t);
+    const msip_t res = sim->get_harts().count(hart_id) && (sim->get_harts().at(hart_id)->state.mip->read() & MIP_MSIP);
+    read_little_endian_reg(res, addr, len, bytes);
```
scottj97, line 46: If there's another device (or memory) just before MSIP_BASE, a misaligned `LD` to MSIP_BASE-2 could come here with len==6, resulting in abort:
```
spike: riscv-isa-sim/riscv/devices.h:176: void read_little_endian_reg(T, reg_t, size_t, uint8_t*) [with T = unsigned int; reg_t = long unsigned int; size_t = long unsigned int; uint8_t = unsigned char]: Assertion `len <= sizeof(T)' failed.
```

riscv/devices.h
aswaterman: Agreed; I'm mentally stuck in C++03.

riscv/clint.cc
aswaterman: It sure does.  I had been thinking that misaligned accesses cannot make it here, and I would be surprised if this is the only I/O device with a bug around that assumption.

riscv/devices.h
aswaterman: > Is there any significance to the mix of parentheses and curly braces?

I was misremembering that `{}` was syntax for array initialization only.

PR #1267, opened 2023-03-02
Initialize MTIP at reset

scottj97:
Since mtime and mtimecmp both initialize to 0, and "a machine timer interrupt becomes pending whenever mtime contains a value greater than or equal to mtimecmp," `mip.MTIP` should be 1.

Before this change, MTIP would be 0 until the first increment, or until reading/writing from CLINT regs (since `clint_t::load()` and `clint_t::store()` call `increment(0)`).


aswaterman (approved):
Funny, I had made the same change last night but didn't get around to PRing it.

PR #1268, opened 2023-03-03
Don't issue misaligned or non-power-of-2 MMIO accesses

aswaterman:
Rather than requiring each MMIO device to support arbitrary sizes and alignments, decompose misaligned loads and stores in such a way as to guarantee their constituent parts are always aligned.  (Specifically, they now always decompose to a sequence of one-byte accesses.)

This is not a semantic change for main-memory accesses, but it is a semantic change for I/O devices.  It makes them more realistic, in that most bus standards don't support non-power-of-2-sized accesses.

Main-memory accesses are not decomposed unless they span page boundaries (in which case they are decomposed into two accesses, possibly of non-power-of-two size).

This fixes misaligned accesses that span a device boundary into the bottom of the CLINT address space, mentioned in https://github.com/riscv-software-src/riscv-isa-sim/pull/1265#discussion_r1123301943

Also make misaligned stores to the CLINT MSIP registers behave more sanely (even though the behavior of misaligned MMIO accesses is unspecified).

Also improve a stylistic comment about initializing `plic_context_t`, https://github.com/riscv-software-src/riscv-isa-sim/pull/1265#discussion_r1123142619

### DIFF
--- a/riscv/devices.h
+++ b/riscv/devices.h
@@ -82,18 +82,17 @@ class clint_t : public abstract_device_t {
 
 struct plic_context_t {
   plic_context_t(processor_t* proc, bool mmode)
-    : proc(proc), mmode(mmode), priority_threshold(0), enable{}, pending{},
-      pending_priority{}, claimed{}
+    : proc(proc), mmode(mmode)
   {}
 
-	processor_t *proc;
-	bool mmode;
+  processor_t *proc;
+  bool mmode;
 
-	uint8_t priority_threshold;
-	uint32_t enable[PLIC_MAX_DEVICES/32];
-	uint32_t pending[PLIC_MAX_DEVICES/32];
-	uint8_t pending_priority[PLIC_MAX_DEVICES];
-	uint32_t claimed[PLIC_MAX_DEVICES/32];
+  uint8_t priority_threshold {};
+  uint32_t enable[PLIC_MAX_DEVICES/32] {};
+  uint32_t pending[PLIC_MAX_DEVICES/32] {};
+  uint8_t pending_priority[PLIC_MAX_DEVICES] {};
+  uint32_t claimed[PLIC_MAX_DEVICES/32] {};
 };
 
 class plic_t : public abstract_device_t, public abstract_interrupt_controller_t {
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -147,18 +147,35 @@ bool mmu_t::mmio_fetch(reg_t paddr, size_t len, uint8_t* bytes)
 
 bool mmu_t::mmio_load(reg_t paddr, size_t len, uint8_t* bytes)
 {
-  if (!mmio_ok(paddr, LOAD))
-    return false;
-
-  return sim->mmio_load(paddr, len, bytes);
+  return mmio(paddr, len, bytes, LOAD);
 }
 
 bool mmu_t::mmio_store(reg_t paddr, size_t len, const uint8_t* bytes)
 {
-  if (!mmio_ok(paddr, STORE))
-    return false;
+  return mmio(paddr, len, const_cast<uint8_t*>(bytes), STORE);
+}
 
-  return sim->mmio_store(paddr, len, bytes);
+bool mmu_t::mmio(reg_t paddr, size_t len, uint8_t* bytes, access_type type)
+{
+  bool power_of_2 = (len & (len - 1)) == 0;
+  bool naturally_aligned = (paddr & (len - 1)) == 0;
+
+  if (power_of_2 && naturally_aligned) {
+    if (!mmio_ok(paddr, type))
+      return false;
+
+    if (type == STORE)
+      return sim->mmio_store(paddr, len, bytes);
+    else
+      return sim->mmio_load(paddr, len, bytes);
+  }
+
+  for (size_t i = 0; i < len; i++) {
+    if (!mmio(paddr + i, 1, bytes + i, type))
+      return false;
+  }
+
+  return true;
 }
 
 void mmu_t::check_triggers(triggers::operation_t operation, reg_t address, std::optional<reg_t> data)
--- a/riscv/clint.cc
+++ b/riscv/clint.cc
@@ -71,12 +71,14 @@ bool clint_t::store(reg_t addr, size_t len, const uint8_t* bytes)
       return store(addr, 4, bytes) && store(addr + 4, 4, bytes + 4);
     }
 
-    msip_t msip = 0;
-    write_little_endian_reg(&msip, addr, len, bytes);
+    if (addr % sizeof(msip_t) == 0) {  // ignore in-between bytes
+      msip_t msip = 0;
+      write_little_endian_reg(&msip, addr, len, bytes);
 
-    const auto hart_id = (addr - MSIP_BASE) / sizeof(msip_t);
-    if (sim->get_harts().count(hart_id))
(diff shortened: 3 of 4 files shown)


scottj97:
I thought it was an improvement a few months ago when the misaligned behavior was changed away from doing a series of one-byte accesses. But this system doesn't have the drawbacks that the old one had, like funny mcontrol breakpoint behavior. So I like this and it nicely solves the problem.

However, wouldn't it be more efficient to break it into bytes only for MMIO and not for every misaligned memory access?

The only flaw I see here is that the way CLINT handles one-byte loads looks like misaligned loads can have some really weird outcomes, such as one MSIP value repeating in multiple bytes of the result. But I guess we don't care what happens to misaligned MMIO.

aswaterman:
> I thought it was an improvement a few months ago when the misaligned behavior was changed away from doing a series of one-byte accesses. But this system doesn't have the drawbacks that the old one had, like funny mcontrol breakpoint behavior. So I like this and it nicely solves the problem.

Agreed.

> However, wouldn't it be more efficient to break it into bytes only for MMIO and not for every misaligned memory access?

If you'd like me to look into improving performance for the main-memory case, I'm willing to do so.  I had been assuming it doesn't matter enough to bother.

Note that, for page-spanning accesses, it would require checking both pages first to determine whether either is MMIO.  This makes it seem pretty unattractive to me.

> The only flaw I see here is that the way CLINT handles one-byte loads looks like misaligned loads can have some really weird outcomes, such as one MSIP value repeating in multiple bytes of the result. But I guess we don't care what happens to misaligned MMIO.

I don't believe that's the case; the access should do the expected thing, with the MSIP bit showing up once.  `read_little_endian_reg` will extract the right byte from the aligned word, so you'll get mostly zeros in the misaligned load result.  I wrote a one-off test case and observed this behavior.

(With that said, I agree that the behavior for misaligned MMIO accesses can be unspecified.)

scottj97:
> Note that, for page-spanning accesses, it would require checking both pages first to determine whether either is MMIO. This makes it seem pretty unattractive to me.

I think it would be as simple as moving the check for unaligned to the point where we call `mmio_load()` or `mmio_store()` -- or even inside those methods of `mmu_t`.

While experimenting with this, I accidentally broke up every MMIO into bytes (even if aligned), and it broke clint's msip regs. I'm investigating, but if breaking up aligned accesses doesn't work, I'm not so sure that breaking up misaligned accesses will work correctly either. (Though we agree that misaligned MMIO is unspecified. But I want to understand why this didn't work as expected.)

riscv/devices.h
```diff
@@ -82,18 +82,17 @@ class clint_t : public abstract_device_t {
...
-	uint8_t priority_threshold;
-	uint32_t enable[PLIC_MAX_DEVICES/32];
-	uint32_t pending[PLIC_MAX_DEVICES/32];
-	uint8_t pending_priority[PLIC_MAX_DEVICES];
-	uint32_t claimed[PLIC_MAX_DEVICES/32];
+	uint8_t priority_threshold {};
+	uint32_t enable[PLIC_MAX_DEVICES/32] {};
+	uint32_t pending[PLIC_MAX_DEVICES/32] {};
+	uint8_t pending_priority[PLIC_MAX_DEVICES] {};
+	uint32_t claimed[PLIC_MAX_DEVICES/32] {};
```
scottj97: Replace tabs with spaces to be consistent with surrounding code

aswaterman:
> I think it would be as simple as moving the check for unaligned to the point where we call `mmio_load()` or `mmio_store()` -- or even inside those methods of `mmu_t`.

Right, if we first break up page-spanning accesses into two intra-page accesses, as we did before, then this would suffice.

I favor the current scheme because it only decomposes accesses in one place (into bytes) rather than two places (first into intrapage accesses, then into bytes).  But I can adopt your scheme if you prefer.

> While experimenting with this, I accidentally broke up every MMIO into bytes (even if aligned), and it broke clint's msip regs.

It might be the case that my recent tinkering with the CLINT fixed this.  A simple test case I wrote did reflect that an access to CLINT bytes [1, 4], i.e. spanning the MSIP fields for hart 0 and hart 1, did make hart 1's MSIP bit show up in bit 24, with other bits set to 0.

scottj97:
> Right, if we first break up page-spanning accesses into two intra-page accesses, as we did before, then this would suffice.
> 
> I favor the current scheme because it only decomposes accesses in one place (into bytes) rather than two places (first into intrapage accesses, then into bytes). But I can adopt your scheme if you prefer.

For your consideration: https://github.com/scottj97/riscv-isa-sim/commit/7f05c4611552de290d4c08b238444def761b0a92

This reduces unnecessary repeated tablewalks when accessing memory, if nothing else.

(Though I'd like to find a way to reduce the copypasta there.)


> It might be the case that my recent tinkering with the CLINT fixed this. A simple test case I wrote did reflect that an access to CLINT bytes [1, 4], i.e. spanning the MSIP fields for hart 0 and hart 1, did make hart 1's MSIP bit show up in bit 24, with other bits set to 0.

It was writes to MSIP that were broken, with addresses 1/2/3 clobbering over what had been written into byte 0: https://github.com/scottj97/riscv-isa-sim/commit/c2b94fa1f36e0942b5bdc254a3976ae435de4959 (Note that the diff displayed by GitHub sucks; it's mostly indenting that's changed, with one new `if()` condition)

aswaterman:
> For your consideration: [scottj97@7f05c46](https://github.com/scottj97/riscv-isa-sim/commit/7f05c4611552de290d4c08b238444def761b0a92)

Yeah, I'm happy with this approach.

> It was writes to MSIP that were broken, with addresses 1/2/3 clobbering over what had been written into byte 0: [scottj97@c2b94fa](https://github.com/scottj97/riscv-isa-sim/commit/c2b94fa1f36e0942b5bdc254a3976ae435de4959) (Note that the diff displayed by GitHub sucks; it's mostly indenting that's changed, with one new `if()` condition)

Oh, I see.  Your proposed fix looks good to me.

aswaterman:
@scottj97 can you re-review?  (Even though it's mostly your code now :))

riscv/mmu.cc
```diff
@@ -149,16 +149,31 @@ bool mmu_t::mmio_load(reg_t paddr, size_t len, uint8_t* bytes)
...
-  return sim->mmio_load(paddr, len, bytes);
+  bool power_of_2 = (len & (len - 1)) == 0;
+  bool naturally_aligned = (paddr & (len - 1)) == 0;
+  if (power_of_2 && naturally_aligned)
+    return sim->mmio_load(paddr, len, bytes);
+  for (size_t i = 0; i < len; i++) {
+    if (!sim->mmio_load(paddr + i, 1, bytes + i))
+      return false;
+  }
+  return true;
```
scottj97: Can you think of any way to remove all the copypasta? These two methods are the same except for the `sim->mmio_*` method called.

aswaterman: Sure, I’ll find a way to refactor it this weekend.

aswaterman: I also realized that we want to call `mmio_ok` on each byte, since these accesses might span boundaries of MMIO-ok-ness.

aswaterman: Done.  I'm not wild about the `const_cast` I ended up with, but it's reasonably contained.

scottj97 (approved):
Not too excited about the `const_cast`, but this is an improvement. Thanks.

PR #1272, opened 2023-03-07
make target_args() protected instead of private to allow subclasses to inject args

sagark:
see title

### DIFF
--- a/fesvr/htif.h
+++ b/fesvr/htif.h
@@ -63,6 +63,7 @@ class htif_t : public chunked_memif_t
   virtual void idle() {}
 
   const std::vector<std::string>& host_args() { return hargs; }
+  std::vector<std::string>& target_args() { return targs; }
 
   reg_t get_entry_point() { return entry; }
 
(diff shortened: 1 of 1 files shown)


jerryz123 (approved):
LGTM, pending CI.

scottj97:
I don't understand this. `target_args()` is not virtual, so a subclass couldn't override it. Even if it were virtual, subclasses in C++ can override private virtual methods.

What am I missing here?

jerryz123:
We have a subclass which needs to append to targs in its own constructor.

scottj97:
Sorry, I still don't get it...doesn't that mean that `targs` should be protected, not `target_args()`?

And why does this remove the `const` from the return type? Now the caller will be able to modify the private `targs`?

jerryz123:
Yeah, we are modifying the private targs from the returned reference.

scottj97:
I think it would be a lot clearer and safer to make `targs` protected instead. Leaving it private but modifying it from the child class is misleading.

jerryz123:
Sure. I'll fix this now.

jerryz123:
Actually I think I have another approach that bypasses the need to modify the targs member entirely... its a bit more invasive of a change

PR #1273, opened 2023-03-07
Support passing in htif target-arguments through plusargs

jerryz123:
This partially reverts #1272, and provides an alternative approach towards the problem that PR fixes.

HTIF now supports the `+target-argument=` plusarg as a mechanism for setting target arguments.

Ex: `spike +target-argument=pk +target-argument=hello` is the same as `spike pk hello`

This is only intended to be used in a specific scenario where we are unable to pass non-plusargs to the simulator that drives FESVR. As such, no documentation for this flag is provided in the help message for htif.

### DIFF
--- a/fesvr/htif.h
+++ b/fesvr/htif.h
@@ -145,6 +145,7 @@ TARGET (RISC-V BINARY) OPTIONS\n\
 {"chroot",    required_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 3 },     \
 {"payload",   required_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 4 },     \
 {"signature-granularity",    optional_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 5 },     \
+{"target-argument",          required_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 6 },     \
 {0, 0, 0, 0}
 
 #endif // __HTIF_H
--- a/fesvr/htif.h
+++ b/fesvr/htif.h
@@ -63,7 +63,7 @@ class htif_t : public chunked_memif_t
   virtual void idle() {}
 
   const std::vector<std::string>& host_args() { return hargs; }
-  std::vector<std::string>& target_args() { return targs; }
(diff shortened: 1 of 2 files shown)


fesvr/htif.h
```diff
@@ -63,7 +63,7 @@ class htif_t : public chunked_memif_t
   virtual void idle() {}
 
   const std::vector<std::string>& host_args() { return hargs; }
-  std::vector<std::string>& target_args() { return targs; }
+  const std::vector<std::string>& target_args() { return targs; }
```
scottj97, line 66: Does this still need to be `protected` instead of `private`?

fesvr/htif.h
```diff
@@ -144,7 +144,7 @@ TARGET (RISC-V BINARY) OPTIONS\n\
 {"signature", required_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 2 },     \
 {"chroot",    required_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 3 },     \
 {"payload",   required_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 4 },     \
-{"signature-granularity",    optional_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 5 },     \
+{"signature-granularity",    required_argument, 0, HTIF_LONG_OPTIONS_OPTIND + 5 },     \
```
scottj97, line 147: Why?

scottj97 (changes requested):

fesvr/htif.h
jerryz123: Based on the help text, this flag needs an argument. This fixes the getopt_long parsing to respect that, and provide the usage error text if no argument is provided.
```
      --signature-granularity=VAL           Size of each line in signature.\n\
       +signature-granularity=VAL\n\
```

fesvr/htif.h
jerryz123: Yes, we'd still like a way for subclasses to read targs after htif populates it.

scottj97 (approved):

aswaterman (approved):

PR #1274, opened 2023-03-07
Add +symbol_elf flag to fesvr to support loading in symbols from other elf files

jerryz123:
Currently, the `addr2symbol` table in `htif` is populated only with the symbols from the target binary. 

This PR adds the `+symbol-elf=` flag to htif, which allows easier debugging of control flow in payload binaries. For example, the Linux payload in OpenSBI.

### DIFF
--- a/fesvr/htif.cc
+++ b/fesvr/htif.cc
@@ -7,6 +7,7 @@
 #include "platform.h"
 #include "byteorder.h"
 #include "trap.h"
+#include "../riscv/common.h"
 #include <algorithm>
 #include <assert.h>
 #include <vector>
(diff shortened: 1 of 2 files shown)


fesvr/htif.cc
```diff
@@ -142,7 +142,8 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
   } preload_aware_memif(this);
 
   try {
-    return load_elf(path.c_str(), &preload_aware_memif, entry, expected_xlen);
+    return load_elf(path.c_str(), &preload_aware_memif, entry,
```
aswaterman: The implication is that the memory interface might disagree about the target endianness--is that really true, and should it be?

aswaterman: Ah, I see it's NULL in a later commit.  Can you instead construct a dummy memif that takes no action on read/write?  That would keep the ELF loader cleaner.

fesvr/htif.cc
```diff
@@ -34,6 +34,12 @@
 # define TARGET_DIR "/" TARGET_ARCH "/bin/"
 #endif
 
+#ifdef __GNUC__
+# define UNUSED __attribute__ ((unused))
+#else
+# define UNUSED
+#endif
+
```
aswaterman: Given this is source code and not a header, is there actually anything wrong with using the macro definitions in `../riscv/common.h` instead?

aswaterman (approved):
OK with me after you've considered my comment about the `UNUSED` definition.

fesvr/htif.cc
jerryz123: Good call, that makes this simpler

jerryz123:
Thanks

PR #1275, opened 2023-03-08
Fix ctrl-c responsiveness when running in interactive mode

jerryz123:
Fixes a bug introduced in #1264 which caused ctrl-c to be ignored in interactive runs.

Break out of the interactive_run->interactive_run loop when ctrlc_pressed


aswaterman (approved):

aswaterman:
Oops, thanks.

PR #1276, opened 2023-03-08
Use simif_t for devices

scottj97:
#1265 started using `sim_t` in both CLINT and PLIC. But there is also `simif_t` which is intended to be a generic interface for such devices to use.

This PR switches CLINT, PLIC, and the debug_module to use `simif_t` instead of `sim_t`.

@aswaterman what do you think? Is this in keeping with the spirit of `simif_t`?

### DIFF
--- a/riscv/simif.h
+++ b/riscv/simif.h
@@ -8,6 +8,7 @@
 #include "cfg.h"
 
 class processor_t;
+class mmu_t;
 
 // this is the interface to the simulator used by the processors and memory
 class simif_t
@@ -32,6 +33,7 @@ public:
 
   unsigned nprocs() const { return get_cfg().nprocs(); }
 
+  mmu_t* debug_mmu;  // debug port into main memory, for use by debug_module
 };
 
 #endif
--- a/riscv/simif.h
+++ b/riscv/simif.h
@@ -3,7 +3,11 @@
 #ifndef _RISCV_SIMIF_H
 #define _RISCV_SIMIF_H
 
+#include <map>
 #include "decode.h"
+#include "cfg.h"
+
+class processor_t;
 
 // this is the interface to the simulator used by the processors and memory
 class simif_t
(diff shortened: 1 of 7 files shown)


scottj97:
I will address the CI issues with a new PR (since I don't have `-Winconsistent-missing-override` in my compiler, so I need to rely on the CI here).

aswaterman (approved):
Oops.

aswaterman:
Yes, I think these new methods are generic enough that they should apply to any `simif_t` implementation.  However, longer term, they need some refactoring (I want to get rid of `get_core` and rename `nprocs` to match `get_harts`), which will be slightly annoying for other implementers of this interface.  Do you want to do that refactoring first?

riscv/simif.h
```diff
@@ -19,10 +24,17 @@ class simif_t
   // Callback for processors to let the simulation know they were reset.
   virtual void proc_reset(unsigned id) = 0;
 
+  virtual processor_t* get_core(size_t i) = 0;
+  virtual const cfg_t &get_cfg() = 0;
+  virtual unsigned nprocs() const = 0;
+  virtual const std::map<size_t, processor_t*>& get_harts() = 0;
+
```
jerryz123, line 29: I think these can be deduplicated so we can have a narrower interface for `simif_t`. For example, it looks like debug_module only uses `get_cfg` to get `nprocs` and `hartids`.

aswaterman: Longer term, I want to get rid of `get_core`, but AFAICS its only usage is inside of `sim_t` itself.  So we can avoid adding it to the `simif_t` interface.

I have a feeling `get_cfg` will keep rearing its head, even though you appear to be right about its current usage.

`nprocs` is just a convenience.  Rather than making it virtual, we could define it here as a non-virtual method that returns `get_cfg().nprocs()`.

scottj97: I took out `get_core` from `simif_t` (not sure why I put it there in the first place) and implemented `nprocs()` inside `simif_t`.

jerryz123 (approved):

aswaterman (approved):
👍

PR #1279, opened 2023-03-08
Improve ctrlc responsiveness in interactive mode

jerryz123:
5f4cabc was only a partial fix.
This avoids some more corner cases where the ctrlc is registered after the last point in the loop where the condition is checked.

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -252,7 +252,13 @@ static std::string readline(int fd)
 
 void sim_t::interactive()
 {
+  if (ctrlc_pressed) {
+    next_interactive_action = std::nullopt;
+    ctrlc_pressed = false;
+  }
+
   if (next_interactive_action.has_value()) {
+    ctrlc_pressed = false;
     auto f = next_interactive_action.value();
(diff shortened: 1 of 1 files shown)


aswaterman:
I think that means we can undo this change, right? https://github.com/riscv-software-src/riscv-isa-sim/blob/a411096f9c318d0dd58de25ae8be290cbc82ce9f/riscv/interactive.cc#L414

To keep the logic in fewer places, can we delete https://github.com/riscv-software-src/riscv-isa-sim/blob/a411096f9c318d0dd58de25ae8be290cbc82ce9f/riscv/interactive.cc#L407 and https://github.com/riscv-software-src/riscv-isa-sim/blob/a411096f9c318d0dd58de25ae8be290cbc82ce9f/riscv/interactive.cc#L748, and instead perform that action here? https://github.com/riscv-software-src/riscv-isa-sim/blob/a411096f9c318d0dd58de25ae8be290cbc82ce9f/riscv/interactive.cc#L256

aswaterman (approved):

Issue #1280, opened 2023-03-13
PLIC Pending bits not implemented


jiegec:
The pending bits are not implemented in plic:

```
 * base + 0x000000: Reserved (interrupt source 0 does not exist)
 * base + 0x000004: Interrupt source 1 priority
 * base + 0x000008: Interrupt source 2 priority
 * ...
 * base + 0x000FFC: Interrupt source 1023 priority
 * base + 0x001000: Pending 0
 * base + 0x001FFF: Pending
 * base + 0x002000: Enable bits for sources 0-31 on context 0
 * base + 0x002004: Enable bits for sources 32-63 on context 0
```

Code: https://github.com/riscv-software-src/riscv-isa-sim/blob/a35865f0f5559d8af81920b2d832e15af6caa123/riscv/plic.cc#L316-L320

It only handled the interrupt source priority and enable fields.

aswaterman:
cc @avpatel

jiegec:
Done in #1281.

PR #1281, opened 2023-03-13
Implement pending bits for plic

jiegec:
Fix issue #1280

### DIFF
--- a/riscv/plic.cc
+++ b/riscv/plic.cc
@@ -48,6 +48,9 @@
 #define PRIORITY_BASE           0
 #define PRIORITY_PER_ID         4
 
+/* Each interrupt source has a pending bit associated with it. */
+#define PENDING_BASE            0x1000
+
 /*
  * Each hart context has a vector of interupt enable bits associated with it.
  * There's one bit for each interrupt source.
@@ -152,6 +155,21 @@ bool plic_t::priority_write(reg_t offset, uint32_t val)
     val &= ((1 << PLIC_PRIO_BITS) - 1);
     priority[id] = val;
   }
+
+  return true;
+}
+
+bool plic_t::pending_read(reg_t offset, uint32_t *val)
+{
+  uint32_t id_word = (offset >> 2);
+
+  if (id_word < num_ids_word) {
+    *val = 0;
+    for (auto context: contexts) {
+        *val |= context.pending[id_word];
+    }
+  } else
+    *val = 0;
 
   return true;
 }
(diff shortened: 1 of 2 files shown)


aswaterman:
cc @avpatel

scottj97 (approved):
I'm not that familiar with PLIC; I assume these are supposed to be read-only? Since you didn't touch `plic_t::store`.

aswaterman:
@avpatel can you please review?

riscv/plic.cc
```diff
@@ -313,8 +316,16 @@ bool plic_t::load(reg_t addr, size_t len, uint8_t* bytes)
       return false;
   }
 
-  if (PRIORITY_BASE <= addr && addr < ENABLE_BASE) {
+  if (PRIORITY_BASE <= addr && addr < PENDING_BASE) {
     ret = priority_read(addr, &val);
+  } else if (PENDING_BASE <= addr && addr < ENABLE_BASE) {
+    uint32_t id_word = (addr - PENDING_BASE) >> 2;
```
avpatel: To be consistent with other parts of PLIC emulation, add `bool plic_t::pending_read(reg_t offset, uint32_t *val)` and use it here.

riscv/plic.cc
```diff
@@ -313,8 +316,16 @@ bool plic_t::load(reg_t addr, size_t len, uint8_t* bytes)
...
+    if (id_word < num_ids_word) {
+      val = level[id_word];
```
avpatel: This is not correct. The pending register should be an ORing of pending state in each context

```
val = 0;
for (i = 0; i < contexts.size(); i++) {
    val |= contexts.pending[id_word];
}
```

riscv/plic.cc
scottj97: I assume you meant `pending_read()`?

avpatel: Ahh, yes. It was a typo. I have fixed it.

jiegec: Done.

riscv/plic.cc
jiegec: Done.

riscv/plic.cc
```diff
@@ -156,6 +159,21 @@ bool plic_t::priority_write(reg_t offset, uint32_t val)
   return true;
 }
 
+bool plic_t::pending_read(reg_t offset, uint32_t *val)
+{
+  uint32_t id_word = (offset >> 2);
+
+  if (id_word > 0 && id_word < num_ids_word) {
```
avpatel: Only `id_word < num_ids_word` check is required here.

avpatel:
@jiegec Only minor comment, otherwise this looks good.

riscv/plic.cc
jiegec: Thanks!

avpatel:
@aswaterman This looks good to me now. Please merge it.

riscv/plic.cc
```diff
@@ -156,6 +159,21 @@ bool plic_t::priority_write(reg_t offset, uint32_t val)
   return true;
 }
 
+bool plic_t::pending_read(reg_t offset, uint32_t *val)
+{
+  uint32_t id_word = (offset >> 2);
+
+  if (id_word < num_ids_word) {
+    *val = 0;
+    for (size_t i = 0; i < contexts.size(); i++) {
```
scottj97: Please use a modern C++ loop: `for (auto context: contexts)`

scottj97 (changes requested):

riscv/plic.cc
jiegec: Done

scottj97 (approved):

aswaterman (approved):

PR #1285, opened 2023-03-16
Fix the relationship between Zce and Zcf

liweiwei90:
Zcf is RV32 only.

### DIFF
--- a/riscv/isa_parser.cc
+++ b/riscv/isa_parser.cc
@@ -157,7 +157,7 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
       extension_table[EXT_ZCB] = true;
       extension_table[EXT_ZCMT] = true;
       extension_table[EXT_ZCMP] = true;
-      if (extension_table['F'])
+      if (extension_table['F'] && max_xlen == 32)
         extension_table[EXT_ZCF] = true;
     } else if (ext_str == "zca") {
       extension_table[EXT_ZCA] = true;


aswaterman (approved):

scottj97:
Does [this line](https://github.com/riscv-software-src/riscv-isa-sim/blob/f29dcd0d34bfc8c7d8982c9d03dc2e40bbc2f212/riscv/csrs.cc#L638) need a corresponding change too?
```
  proc->set_extension_enable(EXT_ZCF, (new_misa & (1L << ('F' - 'A'))) && proc->extension_enabled(EXT_ZCA));
```

aswaterman:
@scottj97 Turns out it's benign, because all code paths that check whether Zcf is enabled are gated by XLEN being 32.  But it might be better to keep this code in sync with the ISA parser, in case we later attempt to refactor and share the logic.

scottj97:
> @scottj97 Turns out it's benign, because all code paths that check whether Zcf is enabled are gated by XLEN being 32. But it might be better to keep this code in sync with the ISA parser, in case we later attempt to refactor and share the logic.

Does that mean that [00066cb](https://github.com/riscv-software-src/riscv-isa-sim/pull/1285/commits/00066cb3655314f07f7b13f225d98e66d32bfd87) was also benign?

aswaterman:
Yes, I hadn't thought that through when I approved it.  In any case, I think we should be consistent and add the check to `csrs.cc`, too.

PR #1286, opened 2023-03-17
Implement Smrnmi extension

aswaterman:
We don't model any sources of RNMI, so this is mostly vestigial.

cc @ptomsich


aswaterman:
The spec for this feature is a chapter in the privileged ISA manual, BTW.  https://github.com/riscv/riscv-isa-manual/blob/faa77da9701329bc212241d5f9c04684358bb28f/src/rnmi.tex

riscv/csrs.h
```diff
@@ -256,6 +256,20 @@ class mstatus_csr_t final: public base_status_csr_t {
 
 typedef std::shared_ptr<mstatus_csr_t> mstatus_csr_t_p;
 
+class mnstatus_csr_t final: public csr_t {
```
scottj97: This could be `basic_csr_t` and only override `unlogged_write()`. That would be more consistent with the other simple CSRs.

riscv/insns/mnret.h
```diff
@@ -0,0 +1,15 @@
+require_extension(EXT_SMRNMI);
+require_privilege(PRV_M);
+set_pc_and_serialize(p->get_state()->mnepc->read());
+reg_t s = STATE.mnstatus->read();
+reg_t prev_prv = get_field(s, MNSTATUS_MNPP);
+reg_t prev_virt = get_field(s, MNSTATUS_MNPV);
+if (prev_prv != PRV_M) {
+  reg_t mstatus = STATE.mstatus->read();
+  mstatus = set_field(mstatus, MSTATUS_MPRV, 0);
+  p->put_csr(CSR_MSTATUS, mstatus);
```
scottj97: Why not `STATE.mstatus->write(mstatus)`? That would avoid an unnecessary map lookup in `csrmap`. (Same comment for the other `put_csr()` just below.)

Could even add a simple method to `mstatus_csr_t` so this could do `STATE.mstatus->clear_mprv()`.

riscv/mmu.h
```diff
@@ -406,6 +406,7 @@ class mmu_t
   inline bool in_mprv()
   {
     return proc != nullptr
+           && !(proc->state.mnstatus && !get_field(proc->state.mnstatus->read(), MNSTATUS_NMIE))
```
scottj97, line 409: Would it make sense/improve anything if we always instantiated `state.mnstatus` and then only put it into `csrmap` if Smrnmi is enabled. Then we wouldn't need to test for null in at least three places I see.

scottj97 (changes requested):

riscv/mmu.h
aswaterman: The only reason I didn't do that to begin with is that the reset value of the NMIE bit is 0 if the extension is implemented, but to suppress the behavior we need to set it to 1.  I'll try it and see how I feel about it.

riscv/csrs.h
aswaterman: Thanks for the suggestion.  I had forgotten about `basic_csr_t`.

riscv/insns/mnret.h
aswaterman: Copy-pasta from the implementation of the MRET instruction.

I decided against defining `clear_mprv`, since inlining the code isn't so bad and I don't expect the method to get much reuse.

riscv/mmu.h
aswaterman: I don't think it turns out to be an improvement.  Since we need NMIE to be 1 to suppress the new behavior, we'll have a conundrum if we ever add devices to Spike that can generate NMIs: NMIs would appear to be enabled even though the extension isn't present, so we'd need a special-case check over there instead.

riscv/csrs.cc
```diff
@@ -516,6 +516,24 @@ reg_t mstatus_csr_t::compute_mstatus_initial_value() const noexcept {
...
 
+// implement class mnstatus_csr_t
+mnstatus_csr_t::mnstatus_csr_t(processor_t* const proc, const reg_t addr):
+  basic_csr_t(proc, addr, 0) {
+}
+
+bool mnstatus_csr_t::unlogged_write(const reg_t val) noexcept {
+  // NMIE can be set but not cleared
+  const reg_t mask = (~read() & MNSTATUS_NMIE)
+                   | (proc->extension_enabled('H') ? MNSTATUS_MNPV : 0)
```
scottj97, line 527: I haven't read the spec but I'm guessing this MNPV should be 0 if no hypervisor? What if we write it to 1, and then disable H?

Most bits that are H-only are masked on read, not write, for this reason. Or written to 0 when H is cleared in `misa_csr_t::unlogged_write()`.

riscv/csrs.h
```diff
@@ -256,6 +256,15 @@ class mstatus_csr_t final: public base_status_csr_t {
 
 typedef std::shared_ptr<mstatus_csr_t> mstatus_csr_t_p;
 
+class mnstatus_csr_t final: public basic_csr_t {
+ public:
+  mnstatus_csr_t(processor_t* const proc, const reg_t addr);
+ protected:
+  virtual bool unlogged_write(const reg_t val) noexcept override;
+ private:
+  reg_t val;
```
scottj97: Hides the parent class's `val`. Instead call `basic_csr_t::unlogged_write()` from within `mnstatus_csr_t::unlogged_write()`.

scottj97 (changes requested):

riscv/csrs.h
aswaterman: Oops, I should've deleted this when I changed the choice of base class.

riscv/csrs.cc
aswaterman: I'm following our implementation of `mstatus.MPV`.  Is the implication that needs to change, too?

aswaterman: Nevermind, I'm not accurately following our implementation of `mstatus.MPV`; I missed that writing `misa` also alters that bit.  I've updated the patch to mimic that behavior.

scottj97 (approved):

aswaterman:
@scottj97 Thanks for helping me improve this PR.

PR #1297, opened 2023-03-27
Stop unconditionally adding zicntr/zihpm to supported extensions

jerryz123:
See https://github.com/riscv-software-src/riscv-isa-sim/pull/926 for prior discussion.

This makes it possible to model systems w.o zicntr and zihpm. I'm not sure what the backwards compatibility being prioritized is, but I've changed the default ISA str to include zicntr/zihpm

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -323,7 +323,7 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
 
   csrmap[CSR_MEDELEG] = medeleg = std::make_shared<medeleg_csr_t>(proc, CSR_MEDELEG);
   csrmap[CSR_MIDELEG] = mideleg = std::make_shared<mideleg_csr_t>(proc, CSR_MIDELEG);
-  const reg_t counteren_mask = 0xffffffffULL;
+  const reg_t counteren_mask = (proc->extension_enabled_const(EXT_ZICNTR) ? 0x7UL : 0x0) | (proc->extension_enabled_const(EXT_ZIHPM) ? 0xfffffff8ULL : 0x0);
   mcounteren = std::make_shared<masked_csr_t>(proc, CSR_MCOUNTEREN, counteren_mask, 0);
   if (proc->extension_enabled_const('U')) csrmap[CSR_MCOUNTEREN] = mcounteren;
   csrmap[CSR_SCOUNTEREN] = scounteren = std::make_shared<masked_csr_t>(proc, CSR_SCOUNTEREN, counteren_mask, 0);
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -247,7 +247,7 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
     auto mcounter = std::make_shared<const_csr_t>(proc, which_mcounter, 0);
     csrmap[which_mcounter] = mcounter;
 
-    if (proc->extension_enabled_const(EXT_ZICNTR) && proc->extension_enabled_const(EXT_ZIHPM)) {
+    if (proc->extension_enabled_const(EXT_ZIHPM)) {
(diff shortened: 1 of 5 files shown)


aswaterman (approved):
I think you also need to manipulate the declaration of `mcounteren` and `scounteren` so that the corresponding mask bits cannot be set.

I made a patch to bbl/pk so that this will be less disruptive than it otherwise would be.  It relies on mcycle/minstret continuing to exist (which is fine, because their existence isn't predicated on Zicntr).

https://github.com/riscv-software-src/riscv-pk/commit/3ed18cfbc707c75af83941e8fdf7f68f5a3b4803

aswaterman (changes requested):
I think you also need to manipulate the declaration of mcounteren and scounteren so that the corresponding mask bits cannot be set.

I made a patch to bbl/pk so that this will be less disruptive than it otherwise would be. It relies on mcycle/minstret continuing to exist (which is fine, because their existence isn't predicated on Zicntr).

https://github.com/riscv-software-src/riscv-pk/commit/3ed18cfbc707c75af83941e8fdf7f68f5a3b4803

scottj97 (changes requested):
Can you explain in 36bb5cd8a0f36f294bc56e2653064774d93fa2ae what you did to update these files? Was it just a matter of running `autoconf`? What version specifically?

Otherwise this looks great. Backward compatibility is a concern but we need some way to allow disabling optional extensions, and this is the best approach I can see.

jerryz123:
I ran `autoreconf` in the top directory, version 2.69

scottj97 (approved):

riscv/processor.cc
```diff
@@ -323,7 +323,8 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
 
   csrmap[CSR_MEDELEG] = medeleg = std::make_shared<medeleg_csr_t>(proc, CSR_MEDELEG);
   csrmap[CSR_MIDELEG] = mideleg = std::make_shared<mideleg_csr_t>(proc, CSR_MIDELEG);
-  const reg_t counteren_mask = 0xffffffffULL;
+  const reg_t counteren_mask = (proc->extension_enabled_const(EXT_ZICNTR) && proc->extension_enabled_const(EXT_ZIHPM)) ?
+    0xffffffffULL : 0x7ULL;
```
aswaterman: What's the logic here?  I would have expected this to be `(zihpm ? fffffff8 : 0) | (zicntr ? 7 : 0)`, but I might be misremembering something.

jerryz123: The hpms seem to only be added if `(zihpm && zicntr)` currently. https://github.com/riscv-software-src/riscv-isa-sim/blob/a6eb6c2a99ec7a23640d2aee5f3c05dc35c223d9/riscv/processor.cc#L250-L252

Is that a bug? This seems to not be correct for zihpm && !zicntr

aswaterman: I suppose it's fine if Spike requires Zicntr if Zihpm is present, but we should enforce that in the ISA parser if so.

But that's not the question I was actually asking.  I was wondering why we are unconditionally enabling mask bits 2:0, even if Zicntr is not enabled.

jerryz123: I prefer to support zihpm && !zicntr. I added a commit to this series to support that case.

Fixed the low bits for zicntr

scottj97 (approved):
I haven't studied the behavior when these two options are not both configured, but at least it doesn't break things when they are both requested via the `--isa` string.

aswaterman (approved):

aswaterman:
@scottj97 @jerryz123 This reminds me that we should add ISA-parser support for the recently ratified profile names (e.g. `--isa=rva22`), which is the right way of keeping the ISA strings compact while maintaining maximal optionality.  Furthermore, we should probably adopt the convention that the default ISA string (unless overridden by `configure` argument) is the latest RVA profile.

PR #1300, opened 2023-03-31
Adjust cache stats printout

Jakub-Palider:
With this change the statistics can be printed on demand and there is no longer need to wait for destructor to trigger this action.
It is also possible to finetune naming for various caches keeping backward API compatibility.
Finally, stats are printed regardless whether any accesses have been requested or not. This is for better insight and potential scripting input.

### DIFF
--- a/riscv/cachesim.h
+++ b/riscv/cachesim.h
@@ -100,6 +100,10 @@ class cache_memtracer_t : public memtracer_t
   {
     cache->set_log(log);
   }
+  void print_stats(void)
+  {
+    cache->print_stats();
+  }
 
  protected:
   cache_sim_t* cache;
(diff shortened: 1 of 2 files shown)


riscv/cachesim.h
```diff
@@ -100,6 +100,10 @@ class cache_memtracer_t : public memtracer_t
   {
     cache->set_log(log);
   }
+  void print_stats(void)
```
scottj97, line 103: No need for the `(void)`; empty parentheses suffice

scottj97 (changes requested):
This appears to be three related changes:
1. Print stats even when no accesses
2. Allow client to override default names of caches ("I$" or "D$")
3. Add `print_stats()` helper to enable printing any time during run

Those are each independent and therefore should be three separate commits, each with a commit message that explains what & why.

riscv/cachesim.h
Jakub-Palider: Thank you for your feedback. I will submit patches with corresponding changes shortly.

scottj97:
When I asked for three separate commits, I did not mean three separate Pull Requests.

But, don't spend any more time redoing this now.

Closing as superseded by #1305 #1306 and #1307.

Jakub-Palider:
Ok, thank you for merging. Next time I will try to better address the pull-request flow.

PR #1309, opened 2023-04-06
Use different tlbs respectively to match riscv spec

liguang-hxd:
Hi all 

There has only one tlb for load/store.
but h-extension has three stage: HS_stage VS_stage and G_stage
hfence.gvma/hfence.vvma sfence.vma instructions has its tlb scope for different stage.
I just modify the load/store tlb to adapt to this behavior above.

@aswaterman @scottj97

Tests:
hypervisor hfence.gvma hfence.vvma sfence tlb pass

1. hfence.gvma only flush G-stage tlb
2. hfence.vvma only flush VS-stage tlb
3. when V=0 sfence.vma only flush HS-stage tlb; when V=1 sfence.vma only flush VS-stage tlb

as just in riscv-privilege SPEC

image-20230406143850942

image-20230406144103436

image-20230406144200107

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -434,8 +434,10 @@ void base_status_csr_t::maybe_flush_tlb(const reg_t newval) noexcept {
   if ((newval ^ read()) &
       (MSTATUS_MPP | MSTATUS_MPRV
        | (has_page ? (MSTATUS_MXR | MSTATUS_SUM) : 0)
-      ))
-    proc->get_mmu()->flush_tlb();
+      )) {
+    if(state->v) proc->get_mmu()->flush_tlb(VS_STAGE);
+    else proc->get_mmu()->flush_tlb();
+  }
 }
 
 namespace {
@@ -911,8 +913,12 @@ base_atp_csr_t::base_atp_csr_t(processor_t* const proc, const reg_t addr):
 
 bool base_atp_csr_t::unlogged_write(const reg_t val) noexcept {
   const reg_t newval = proc->supports_impl(IMPL_MMU) ? compute_new_satp(val) : 0;
-  if (newval != read())
-    proc->get_mmu()->flush_tlb();
+  if (newval != read()) {
+    if(address == CSR_VSATP || state->v)
+      proc->get_mmu()->flush_tlb(VS_STAGE);
+    else
+      proc->get_mmu()->flush_tlb(HS_STAGE);
+  }
   return basic_csr_t::unlogged_write(newval);
 }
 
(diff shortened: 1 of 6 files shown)


aswaterman (changes requested):
I disagree with the premise of this PR.  Microarchitectures are free to implement address translation however they wish, so long as they comply with the architecture.  Spike is already compliant.

scottj97 (changes requested):
There are so many problems with this PR:
1. Using one giant commit instead of a series of small steps (see [Understanding the Git Workflow](https://sandofsky.com/workflow/git-workflow/))
2. Inconsistent code formatting (especially, space after `if`)
3. Far too much copypasta
4. Behavioral changes: in some of my (proprietary, sorry) tests, I can see `hlvx` succeed where it previously took a page fault. This particular test should not have any platform-dependent behavior; it flushes TLBs everywhere necessary, so this looks like an implementation bug to me.
5. Why does this apply only to data and not to ifetch?
6. Confusing/misleading variable names: `phys` and `paddr`: what's the difference?
7. Performance: `load()` now always calls `load_slow_path()`; the optimized "fast path" is gone
8. Magic numbers: `3` appears in many places

But ultimately it boils down to what Andrew said: Spike is already architecturally compliant, so you'll need a better justification for adding all this extra complexity.

liguang-hxd:
@scottj97  @aswaterman 
I don't understand "Spike is already architecturally compliant", How to explain the behavior of spike's hfence/sfence does not comply with riscv privilege SPEC of h-extension. 

And I have another question of tlb entry just save the offset of va and pa, but not the pte. When hit the tlb, there has no permission check in fast load/store/fetch. This will leads to lacking permission instruction always pass.

Pls help to check. Thanks

scottj97:
> @scottj97 @aswaterman I don't understand "Spike is already architecturally compliant", How to explain the behavior of spike's hfence/sfence does not comply with riscv privilege SPEC of h-extension.

How exactly do you think it does not comply with that spec?

Page translation caching (i.e. "TLBs") is strictly a performance thing. A spec-compliant implementation could have no TLBs at all, tablewalking on every memory access. The TLB fence instructions would be implemented as NOPs.

The spec requires certain page translation caches to be invalidated by these fence instructions. Spike invalidates those caches, and more. The spec does not forbid any page translations from being invalidated at any time.


> And I have another question of tlb entry just save the offset of va and pa, but not the pte. When hit the tlb, there has no permission check in fast load/store/fetch. This will leads to lacking permission instruction always pass.

Do you have a failing test case?

I have not studied the code recently, but I suspect it works correctly because Spike has separate fetch/load/store TLBs. If a translation would cause an access fault for the given access type, it never gets put into that TLB. Changing privilege flushes the entire TLB.

scottj97:
Closing since OP has disappeared.

PR #1311, opened 2023-04-11
explicitly show D(-mode) instead of M(-mode) when in debug mode

YenHaoChen:
Debug mode (D-mode) can be considered an additional privilege mode. This commit proposes printing D(-mode) instead of M(-mode) for D-mode.


aswaterman (changes requested):
Seems reasonable to me.

aswaterman (approved):
Seems reasonable to me.

scottj97 (approved):

PR #1313, opened 2023-04-11
riscv: sim.cc: Parse for "sifive,clint0" if "riscv,clint0" is absent

endeneer:
"riscv,clint0" and "sifive,clint0" in device tree's "compatible" string point to the same driver, as can be seen from drivers/clocksource/timer-clint.c in Linux kernel. https://github.com/torvalds/linux/commit/2ac6795fcc085e8d03649f1bbd0d70aaff612cad


jerryz123 (approved):
I've rebased against master, this should be GTG, pending CI

aswaterman (approved):

PR #1314, opened 2023-04-11
riscv: sim.cc: Parse for other compatible strings if "riscv,plic0" is absent

endeneer:
There are three more compatible strings in device tree that point to the same driver,
- "sifive,plic-1.0.0"
- "andestech,nceplic100"
- "thead,c900-plic" 

as can be seen from drivers/irqchip/irq-sifive-plic.c in Linux kernel. https://github.com/torvalds/linux/commit/5873ba559101fa37ad9764e79856f71bf54021aa

### DIFF
--- a/riscv/plic.cc
+++ b/riscv/plic.cc
@@ -418,7 +418,8 @@ std::string plic_generate_dts(const sim_t* sim)
 plic_t* plic_parse_from_fdt(const void* fdt, const sim_t* sim, reg_t* base, const std::vector<std::string>& sargs)
 {
   uint32_t plic_ndev;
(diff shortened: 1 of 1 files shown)


scottj97:
The kernel file and commit you pointed to say that thead and andes PLICs have different behavior than the first two. (plic_init vs plic_edge_init)

endeneer:
Thanks @scottj97 for pointing that out, I have amended my commit accordingly. :-)

jerryz123 (approved):
I've rebased this on master. This should be good to merge, pending CI

aswaterman (approved):

PR #1315, opened 2023-04-11
riscv: sim.cc: Consider cpu-map node in cpus node

endeneer:
cpu-map node may be present in cpus node.
https://github.com/torvalds/linux/commit/124e46a86580c71e0eee8459c5da7649318118db
So this commit adds logic to cater that.

### DIFF
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -149,14 +149,18 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
   }
 
   //per core attribute
-  int cpu_offset = 0, rc;
+  int cpu_offset = 0, cpu_map_offset, rc;
   size_t cpu_idx = 0;
   cpu_offset = fdt_get_offset(fdt, "/cpus");
+  cpu_map_offset = fdt_get_offset(fdt, "/cpus/cpu-map");
   if (cpu_offset < 0)
     return;
 
   for (cpu_offset = fdt_get_first_subnode(fdt, cpu_offset); cpu_offset >= 0;
        cpu_offset = fdt_get_next_subnode(fdt, cpu_offset)) {
+
+    if (!(cpu_map_offset < 0) && cpu_offset == cpu_map_offset)
+      continue;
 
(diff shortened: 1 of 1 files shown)


riscv/sim.cc
```diff
@@ -149,15 +149,19 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
   }
 
   //per core attribute
-  int cpu_offset = 0, rc;
+  int cpu_offset = 0, cpu_map_offset=0, rc;
   size_t cpu_idx = 0;
   cpu_offset = fdt_get_offset(fdt, "/cpus");
-  if (cpu_offset < 0)
+  cpu_map_offset = fdt_get_offset(fdt, "/cpus/cpu-map");
+  if (cpu_offset < 0 || cpu_map_offset < 0)
```
jerryz123: Doesn't this change the behavior for DTS without the `cpu-map` node? As in, a DTS with `cpu` nodes, but no `cpu-map` node will skip the code below?

endeneer: Sorry for the silly mistake, I have corrected that below.

scottj97 (approved):
I don't understand what this does or why we need it, but it doesn't seem to break anything.

endeneer:
@scottj97 In the presence of cpu-map in cpus node, without this commit, the cpu-map will be wrongly considered as a cpu node.
```
cpus {
		#address-cells = <1>;
		#size-cells = <0>;
		cpu-map {      <------- ***This is not a cpu node***
			socket0 {
				cluster0 {
					core0 {
						cpu = <&CPU1>;
					};
					core1 {
						cpu = <&CPU2>;
					};
				};
			};
		};

		CPU1: cpu@1 {
			device_type = "cpu";
			compatible = "sifive,rocket0", "riscv";
			reg = <0x1>;
		}

		CPU2: cpu@2 {
			device_type = "cpu";
			compatible = "sifive,rocket0", "riscv";
			reg = <0x2>;
		}
	}
};
```

PR #1316, opened 2023-04-12
implement Debug Specification Section 5.4 Native Triggers

YenHaoChen:
This commit implements Debug Specification Section 5.4 Native Triggers, which resolves the reentrancy problem.

**Reentrancy problem**
Native triggers, i.e., tdata1.action=0, cause a breakpoint exception when matching. Firing a breakpoint exception in a trap handler may result in reentrancy to the trap handler, which overwrites trap-related CSRs (such as mcause and mepc) and leaves the hart unable to resume normal execution.

To avoid the reentrancy problem, native triggers should prevent causing a breakpoint exception to a trap handler, where the hart is already in. The specification allows two solutions to avoid the reentrancy problem, as shown below. This commit chooses the first solution because the second one targets implementations without S-mode.
image

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -106,6 +106,21 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
   return true;
 }
 
+bool trigger_t::allow_action(const state_t * const state) const
+{
+  if (get_action() == ACTION_DEBUG_EXCEPTION) {
+    const bool mstatus_mie = state->mstatus->read() & MSTATUS_MIE;
+    const bool sstatus_sie = state->sstatus->read() & MSTATUS_SIE;
+    const bool vsstatus_sie = state->vsstatus->read() & MSTATUS_SIE;
+    const bool medeleg_breakpoint = (state->medeleg->read() >> CAUSE_BREAKPOINT) & 1;
+    const bool hedeleg_breakpoint = (state->hedeleg->read() >> CAUSE_BREAKPOINT) & 1;
+    return (state->prv != PRV_M || mstatus_mie) &&
+           (state->prv != PRV_S || state->v || !medeleg_breakpoint || sstatus_sie) &&
+           (state->prv != PRV_S || !state->v || !medeleg_breakpoint || !hedeleg_breakpoint || vsstatus_sie);
+  }
+  return true;
+}
+
 reg_t disabled_trigger_t::tdata1_read(const processor_t * const proc) const noexcept
 {
   auto xlen = proc->get_xlen();
@@ -212,7 +227,7 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
     value &= 0xffffffff;
   }
 
-  if (simple_match(xlen, value)) {
+  if (simple_match(xlen, value) && allow_action(proc->get_state())) {
     /* This is OK because this function is only called if the trigger was not
      * inhibited by the previous trigger in the chain. */
     hit = true;
@@ -289,7 +304,7 @@ void mcontrol6_t::tdata1_write(processor_t * const proc, const reg_t val, const 
 
 std::optional<match_result_t> icount_t::detect_icount_match(processor_t * const proc) noexcept
 {
-  if (!common_match(proc))
+  if (!common_match(proc) || !allow_action(proc->get_state()))
     return std::nullopt;
 
   std::optional<match_result_t> ret = std::nullopt;
@@ -389,7 +404,7 @@ std::optional<match_result_t> trap_common_t::detect_trap_match(processor_t * con
   bool interrupt = (t.cause() & ((reg_t)1 << (xlen - 1))) != 0;
   reg_t bit = t.cause() & ~((reg_t)1 << (xlen - 1));
   assert(bit < xlen);
-  if (simple_match(interrupt, bit)) {
+  if (simple_match(interrupt, bit) && allow_action(proc->get_state())) {
     hit = true;
     return match_result_t(TIMING_AFTER, action);
   }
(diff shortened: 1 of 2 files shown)


YenHaoChen:
The spec says the native trigger prevents **_matching_** to resolve the reentrancy problem. Additionally, the spec lets icount trigger countdowns when **_matches_**.
image
image
I am slightly confused about whether the native icount trigger will countdown when mstatus.MIE=0 in M-mode. I assume mstatus.MIE only prevents the native icount trigger from firing instead of countdowns.
@timsifive Do I understand the spec correctly?

riscv/triggers.cc
```diff
@@ -106,6 +106,18 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
...
 
+bool suppress_breakpoint_triggers(state_t * const state)
+{
+  const bool mstatus_mie = state->mstatus->read() & MSTATUS_MIE;
+  const bool sstatus_sie = state->sstatus->read() & MSTATUS_SIE;
+  const bool vsstatus_sie = state->vsstatus->read() & MSTATUS_SIE;
+  const bool medeleg_breakpoint = (state->medeleg->read() >> CAUSE_BREAKPOINT) & 1;
+  const bool hedeleg_breakpoint = (state->hedeleg->read() >> CAUSE_BREAKPOINT) & 1;
+  return (state->prv == PRV_M && !mstatus_mie) ||
+         (state->prv == PRV_S && medeleg_breakpoint && !sstatus_sie) ||
```
scottj97: Should this qualify with `!state->v`?

riscv/triggers.cc
```diff
@@ -212,7 +224,7 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
     value &= 0xffffffff;
   }
 
-  if (simple_match(xlen, value)) {
+  if (simple_match(xlen, value) && (action != ACTION_DEBUG_EXCEPTION || !suppress_breakpoint_triggers(proc->get_state()))) {
```
scottj97: This long expression appears 3 times. How about a common method `allow_action(action)`? It might make sense to fold this into `suppress_breakpoint_triggers()`.

scottj97 (changes requested):

scottj97:
> I am slightly confused about whether the native icount trigger will countdown when mstatus.MIE=0 in M-mode. I assume mstatus.MIE only prevents the native icount trigger from firing instead of countdowns.

The icount breakpoint should not decrement its count if it cannot trigger.

This came up recently in the context of `tcontrol`: https://github.com/riscv/riscv-debug-spec/pull/829

YenHaoChen:
> > I am slightly confused about whether the native icount trigger will countdown when mstatus.MIE=0 in M-mode. I assume mstatus.MIE only prevents the native icount trigger from firing instead of countdowns.
> 
> The icount breakpoint should not decrement its count if it cannot trigger.
> 
> This came up recently in the context of `tcontrol`: [riscv/riscv-debug-spec#829](https://github.com/riscv/riscv-debug-spec/pull/829)

That means the mstatus.MIE may prevent icount native trigger from decreasing (matching) and firing if pending. 
@scottj97 Is this correct?
(Commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1316/commits/f14fcd35eea489123b3d835860fc4ac0be13354f implements this behavior.)

timsifive:
> The icount breakpoint should not decrement its count if it cannot trigger.

That is right. icount triggers decrement when they match. Section 5.4 explicitly says matching is prevented. My gut says that firing should also be prevented, although I can't quite come up with the case for that. I've reopened https://github.com/riscv/riscv-debug-spec/issues/812 to take a look at that.

YenHaoChen:
@timsifive 
Consider a writable icount.pending. One can set the icount.pending in D-mode and return to M-mode with mstatus.MIE=0. In this case, merely preventing icount from decreasing (matching) does not implicitly prevents icount from firing.

riscv/triggers.cc
```diff
@@ -106,6 +106,21 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
   return true;
 }
 
+bool allow_action(state_t * const state, const action_t action)
```
scottj97: Any reason this isn't a method on `trigger_t`? Then you wouldn't have to pass in `action`

scottj97 (approved):

riscv/triggers.cc
YenHaoChen: Move the allow_action() function to trigger_t.

riscv/triggers.cc
```diff
@@ -389,7 +404,7 @@ std::optional<match_result_t> trap_common_t::detect_trap_match(processor_t * con
   bool interrupt = (t.cause() & ((reg_t)1 << (xlen - 1))) != 0;
   reg_t bit = t.cause() & ~((reg_t)1 << (xlen - 1));
   assert(bit < xlen);
-  if (simple_match(interrupt, bit)) {
+  if (simple_match(interrupt, bit) && (action != ACTION_DEBUG_EXCEPTION || allow_action(proc->get_state()))) {
```
scottj97: `allow_action` already checks for `!= ACTION_DEBUG_EXCEPTION`

riscv/triggers.h
```diff
@@ -90,6 +90,7 @@ class trigger_t {
 protected:
   static action_t legalize_action(reg_t val, reg_t action_mask, reg_t dmode_mask) noexcept;
   bool common_match(processor_t * const proc) const noexcept;
+  bool allow_action(state_t * const state);
```
scottj97: Param could be `const state_t * const state`

Method could be `const`

scottj97 (changes requested):

scottj97 (approved):

aswaterman (approved):

PR #1318, opened 2023-04-12
Fix syscall gettimeofday in 32-bit emulator

qwe661234:
When running the 32-bit emulator, we've discovered that the proxy kernel generates inaccurate time results when the `gettimeofday` system call is made. However, the 64-bit emulator doesn't seem to be impacted by this issue. To address this problem, we've made a modification to the system by forwarding the 'gettimeofday' system call to the Spike front-end server. This allows us to receive the correct time results and resolve the problem. The modification includes both the proxy kernel and Spike.

## Issue reproduction
* [dhrystone source code](https://github.com/qwe661234/rv32emu/blob/master/tests/dhrystone.c)
### 1. Compile
```
riscv32-unknown-elf-gcc -march=rv32g -O2 ./dhrystone.c -o dhrystone.elf
```
### 2. Execute
```
/path/to/spike --isa=rv32g /path/to/32-bit_proxy_kernel dhrystone.elf
```

Related: https://github.com/riscv-software-src/riscv-pk/pull/294

### DIFF
--- a/fesvr/syscall.cc
+++ b/fesvr/syscall.cc
@@ -7,6 +7,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/stat.h>
+#include <sys/time.h>
 #include <limits.h>
 #include <errno.h>
 #include <stdlib.h>
@@ -164,6 +165,7 @@ syscall_t::syscall_t(htif_t* htif)
   table[79] = &syscall_t::sys_fstatat;
   table[80] = &syscall_t::sys_fstat;
   table[93] = &syscall_t::sys_exit;
+  table[169] = &syscall_t::sys_gettimeofday;
   table[291] = &syscall_t::sys_statx;
   table[1039] = &syscall_t::sys_lstat;
   table[2011] = &syscall_t::sys_getmainvars;
@@ -214,6 +216,14 @@ void syscall_t::handle_syscall(command_t cmd)
 reg_t syscall_t::sys_exit(reg_t code, reg_t a1, reg_t a2, reg_t a3, reg_t a4, reg_t a5, reg_t a6)
 {
   htif->exitcode = code << 1 | 1;
+  return 0;
+}
+
+reg_t syscall_t::sys_gettimeofday(reg_t buf, reg_t a1, reg_t a2, reg_t a3, reg_t a4, reg_t a5, reg_t a6)
+{
+  struct timeval tv;
+  gettimeofday(&tv, NULL);
+  memif->write(buf, sizeof(timeval), &tv);
   return 0;
 }
 
--- a/fesvr/syscall.h
+++ b/fesvr/syscall.h
@@ -47,6 +47,7 @@ class syscall_t : public device_t
   std::string undo_chroot(const char* fn);
 
   reg_t sys_exit(reg_t, reg_t, reg_t, reg_t, reg_t, reg_t, reg_t);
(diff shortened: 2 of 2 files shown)


scottj97:
Please fix the PR title and the first line of the commit message to spell the function name correctly

aswaterman (changes requested):
Spike currently has two modes of operation: the default, which is deterministic, and `--real-time-clint`, which is not.  I would think we would not want `sys_gettimeofday` to return the host time if `--real-time-clint` is specified.

In any case, I agree with @scottj97's comment on https://github.com/riscv-software-src/riscv-pk/pull/294: it's not clear why this PR is really needed.  I'd rather fix the fundamental RV32 problem you identified, rather than changing how the syscall works on RV32.

Finally, note that frontend system calls are extremely slow on some real implementations.  `gettimeofday` is expected to be fast.  So I think this approach is problematic for that reason, too.

qwe661234:
> Spike currently has two modes of operation: the default, which is deterministic, and `--real-time-clint`, which is not. I would think we would not want `sys_gettimeofday` to return the host time if `--real-time-clint` is specified.
> 
> In any case, I agree with @scottj97's comment on [riscv-software-src/riscv-pk#294](https://github.com/riscv-software-src/riscv-pk/pull/294): it's not clear why this PR is really needed. I'd rather fix the fundamental RV32 problem you identified, rather than changing how the syscall works on RV32.
> 
> Finally, note that frontend system calls are extremely slow on some real implementations. `gettimeofday` is expected to be fast. So I think this approach is problematic for that reason, too.

The reason I want to change the 'gettimeofday' system call is to obtain accurate time results. Otherwise, the results of any benchmark using 'gettimeofday' would be incorrect. However, I acknowledge the statement that frontend system calls can be significantly slow on certain real-world implementations. Perhaps, only the code within the proxy kernel needs to be modified in order to achieve this objective.

qwe661234:
May I ask another question? If the objective is to evaluate the performance of Spike running on the host, would the time results obtained from the proxy kernel be considered accurate?

scottj97:
> The reason I want to change the 'gettimeofday' system call is to obtain accurate time results. Otherwise, the results of any benchmark using 'gettimeofday' would be incorrect. However, I acknowledge the statement that frontend system calls can be significantly slow on certain real-world implementations. Perhaps, only the code within the proxy kernel needs to be modified in order to achieve this objective.

Spike is not a performance model and using it for benchmarking is not going to give you any useful information no matter what `gettimeofday` returns.

If you're trying to benchmark Spike itself, you can do that using gprof or any other host-based profiling tool. You don't need anything special in Spike for that.

You haven't explained why you think RV64 should work differently than RV32 here, as your PR implements.

qwe661234:
>Spike is not a performance model and using it for benchmarking is not going to give you any useful information no matter what gettimeofday returns.
>If you're trying to benchmark Spike itself, you can do that using gprof or any other host-based profiling tool. You don't need anything special in Spike for that.
>You haven't explained why you think RV64 should work differently than RV32 here, as your PR implements.

Thank you, I understand your point. The result of the `gettimeofday` system call in Spike is emulated time. However, since my objective is to evaluate the performance of the 32-bit Spike emulator running on the host machine, precise execution time is necessary. That is the reason why I opted to add the `gettimeofday` system call to the front-end server. As your suggestion, I will use a performance examination tool instead of running benchmarks on Spike. Furthermore, upon double-checking the 'dhrystone' results of both the 32-bit and 64-bit Spike emulators, it was determined that the time results of both emulators are emulated time rather than execution time on the host machine. Therefore, my previous statement was incorrect.

PR #1320, opened 2023-04-14
Add clear in_wfi interface to proc_t

jerryz123:
This enables more flexible modelling of wfi behavior

### DIFF
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -293,7 +293,8 @@ public:
   void set_mmu_capability(int cap);
 
   const char* get_symbol(uint64_t addr);
-  
+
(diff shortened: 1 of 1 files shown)


aswaterman (commented):
I skimmed the relevant code and believe this to be safe.  OK with you, too, @scottj97?

scottj97 (approved):
Not sure I understand the use case here, but it doesn't hurt anything.

jerryz123:
Some implementations may exit WFI arbitrarily. This lets me match implementation behavior with spike by clearing spike's wfi state before calling step.

aswaterman (approved):

PR #1321, opened 2023-04-14
Add support for BF16 extensions

liweiwei90:
This PR tries to add support for BF16 extensions v0.2.1. 
The spec can be found in: https://github.com/riscv/riscv-bfloat16.


aswaterman (changes requested):
I haven't reviewed this yet, but we should wait until the opcode conflict mentioned on https://github.com/riscv/riscv-opcodes/pull/174 is resolved.

softfloat/bf16_to_f32.c
```diff
@@ -0,0 +1,88 @@
...
+        goto uiZ;
+    }
+    /*------------------------------------------------------------------------
+    *------------------------------------------------------------------------*/
+    if ( ! exp ) {
+        if ( ! frac ) {
+            uiZ = packToF32UI( sign, 0, 0 );
+            goto uiZ;
+        }
+    }
```
aswaterman: This can be deleted.  Zero is not a special case for this conversion; the code on the line below does the exact same thing.

softfloat/f32_to_bf16.c
```diff
@@ -0,0 +1,87 @@
...
+=============================================================================*/
+
+#include <stdbool.h>
+#include <stdint.h>
+#include "platform.h"
+#include "internals.h"
+#include "specialize.h"
+#include "softfloat.h"
+
+bfloat16_t f32_to_bf16( float32_t a )
```
aswaterman, line 44: This routine doesn't work correctly for subnormals.  For the input 0x00000001, it produces 0x0040.  Depending on the rounding mode, the correct result is, of course, 0x0000 or 0x0001.

The issue is that FP32 to BF16 is a special case: the mantissa narrows but the exponent does not.  A simple workaround is to implement this operation by first converting to FP64 then converting to BF16.  It isn't high performance, but it's simple, and that's fine for our purposes.  Here is a prototype that seems to work: https://github.com/riscv-software-src/riscv-isa-sim/commit/84e2dca8c41fe436dae82dd0413a89a335f9a7f8

riscv/insns/flh.h
```diff
@@ -1,3 +1,3 @@
-require_extension(EXT_ZFHMIN);
+require_either_extensions(EXT_ZFHMIN, EXT_ZFBFMIN, EXT_ZVFBFMIN);
```
aswaterman: An alternative approach here would be to define a new Spike-internal extension, say, `EXT_INTERNAL_ZFH_MOVE`, which includes only these four instructions.  Then we would enable this extension when any of those three extensions is included.  (We would not add any command-line support.)

I could go either way on this.  What do you think, @scottj97?

aswaterman: The alternate approach would also avoid the copy-pasta in the disassembler: https://github.com/riscv-software-src/riscv-isa-sim/pull/1321/commits/107ee46f29f0a531a7542bc48537cfd0d565c067#diff-6ec72161b4427824dd85867de3c0998cd73d7ebfaad6d7af568240a38767b58dR1243

riscv/overlap_list.h
```diff
@@ -9,3 +9,5 @@ DECLARE_OVERLAP_INSN(cm_mva01s, EXT_ZCMP)
 DECLARE_OVERLAP_INSN(cm_mvsa01, EXT_ZCMP)
 DECLARE_OVERLAP_INSN(cm_jalt, EXT_ZCMT)
 DECLARE_OVERLAP_INSN(c_fsd, EXT_ZCD)
+DECLARE_OVERLAP_INSN(fround_h, EXT_ZFA)
+DECLARE_OVERLAP_INSN(fcvt_bf16_s, EXT_ZFBFMIN)
```
aswaterman: I assume we can delete this now.

softfloat/bf16_to_f32.c
liweiwei90: OK. I'll delete this.

riscv/overlap_list.h
liweiwei90: OK. I'll update the encode for fcvt_bf16_s and delete this.

riscv/insns/flh.h
liweiwei90: It's acceptable to me.

softfloat/f32_to_bf16.c
liweiwei90: OK. I'll try to fix this.

scottj97 (approved):
I haven't reviewed the code, or tested the new BF16 instructions, but I've confirmed that it doesn't break anything else.

softfloat/f32_to_bf16.c
liweiwei90: I think we can covert it to normal value format instead of FP64. And it will produce the same result as FP64 in my test.

aswaterman (approved):

softfloat/f32_to_bf16.c
aswaterman: Thanks!

PR #1322, opened 2023-04-14
Remove unused parameter for macro in v_ext_macro.h

liweiwei90:
BODY8/16 and CHECK8/16 are unused in these cases.

### DIFF
--- a/riscv/insns/vfncvt_f_f_w.h
+++ b/riscv/insns/vfncvt_f_f_w.h
@@ -1,4 +1,4 @@
-// vfncvt.f.f.v vd, vs2, vm
+// vfncvt.f.f.w vd, vs2, vm
 VI_VFP_NCVT_FP_TO_FP(
   { vd = f32_to_f16(vs2); },           // BODY32
   { vd = f64_to_f32(vs2); },           // BODY64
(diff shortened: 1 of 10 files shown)


riscv/insns/vfncvt_f_f_w.h
```diff
@@ -1,9 +1,7 @@
-// vfncvt.f.f.v vd, vs2, vm
+// vfncvt.f.f.w vd, vs2, vm
```
aswaterman, line 1: The comment changes can be a separate commit.

riscv/insns/vfncvt_f_f_w.h
```diff
@@ -1,9 +1,7 @@
...
 VI_VFP_NCVT_FP_TO_FP(
-  {;},                                 // BODY16
```
aswaterman, line 3: Deleting BODY16 suggests that the PR description is incomplete.  I guess BODY16 is removed for narrowing ops and BODY8 is removed for single-width ops?

aswaterman (changes requested):
Seems like a good improvement, modulo my comments.

riscv/insns/vfncvt_f_f_w.h
liweiwei90: OK. I have updated it.

riscv/insns/vfncvt_f_f_w.h
liweiwei90: OK. I have udpated the description and add a new commit to update the parameters for VI_VFP_NCVT_* macros to be consistent with these comments.

aswaterman (approved):
Thanks!

PR #1326, opened 2023-04-18
Avoid declaring potentially out-of-bounds references in vmfeq etc.

### LINKED ISSUE #1325: Vector Floating-Point Compare Instructions vd index overflow issue
for some Vector Floating-Point Compare Instructions like vmfeq.vf instructions

there were 2 &vd defined
image

image

in VFP_VF_PARAMS macro where defined &vd as vector elements while vd shoule be single mask register
for example, if vd is v29, lmum=4, the second define macro may extract an illegal vector register index like v32 which may cause problem.

aswaterman:
I don't think this was actually broken, since the references weren't used, but it was certainly unsanitary.

Fixes #1325

### DIFF
--- a/riscv/v_ext_macros.h
+++ b/riscv/v_ext_macros.h
@@ -341,12 +341,12 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
   type_usew_t<x>::type vs2 = P.VU.elt<type_usew_t<x>::type>(rs2_num, i);
 
 #define VV_PARAMS(x) \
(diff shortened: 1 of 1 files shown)


jerryz123 (changes requested):
Should the same change be applied to the integer macros as well?
https://github.com/riscv-software-src/riscv-isa-sim/blob/87e9eab73ba0eac9e83511212af9eac76711effc/riscv/v_ext_macros.h#L349

aswaterman:
@jerryz123 The integer versions appear to be structured a bit differently and don't have this problem.  However, I was able to remove the UNUSED qualifier without any warnings (which makes me wonder why it was there in the first place).

jerryz123 (approved):

PR #1333, opened 2023-04-20
Close file descriptors in destructor of syscall_t

poemonsense:
In the constructor of `syscall_t`, `dup `is called to duplicate the pipes. However, they are not correctly closed when `syscall_t` is destroyed. If we instantiate and delete it multiple times (it's used by `sim_t`, which are normally used to drive the simulation and may be called multiple times), it will cause too many files to be opened and lead to errors.

This PR adds the destructor and calls `close` for the descriptors.


aswaterman (commented):
Seems right to me, but can you take a second look, @scottj97?

scottj97 (approved):
I know little about fesvr but this seems good to me.

PR #1337, opened 2023-04-21
triggers: icount: not to decrease on firing icount trigger with Debug Mode action

YenHaoChen:
Firing a icount trigger does not execute current instruction and should not decrease.

Update (2023.06.19) https://github.com/riscv/riscv-debug-spec/issues/842#issuecomment-1533703993: 
The icount decreases on firing trigger action=breakpoint but not on firing trigger action=debug_mode when icount.count=1.

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -317,7 +317,7 @@ std::optional<match_result_t> icount_t::detect_icount_match(processor_t * const 
     ret = match_result_t(TIMING_BEFORE, action);
   }
 
-  if (count >= 1) {
+  if (count >= 1 && (ret == std::nullopt || action != MCONTROL_ACTION_DEBUG_MODE)) {
     if (count == 1)
       pending = 1;
     count = count - 1;


scottj97 (changes requested):
Can you explain the case here? It seems this would apply only in the screwy case that someone wrote (via csrw) `pending=1 && count>0`.

Even in that case, icount breakpoints decrement on traps, too, so why shouldn't it decrement on this trap?

YenHaoChen:
@scottj97 is right. The icount should decrease the counter on firing with action=breakpoint. Sorry for the misunderstanding of the specification.

From the current implementation, the icount does not decrease the counter on entering debug mode due to debug interrupt but does on entering debug mode due to trigger firing. Although both conditions enter debug mode, the icount.count behaves differently under those conditions.
@timsifive Can you share some words about this? Thank you.

YenHaoChen:
Clarifying desired behavior in https://github.com/riscv/riscv-debug-spec/issues/842.

YenHaoChen:
The issue (https://github.com/riscv/riscv-debug-spec/issues/842) clarifies that the icount decreases on firing a breakpoint action but not on entering Debug Mode action. For example, consider the case by @scottj97, where  ```pending=1 && count>0```. Whether to decrease the count depends on the action.

The commit (https://github.com/riscv-software-src/riscv-isa-sim/pull/1337/commits/4fce5a42defab4756fa810943c05189b56ab3577) updates the behavior correspondingly.

riscv/triggers.cc
```diff
@@ -296,7 +296,10 @@ std::optional<match_result_t> icount_t::detect_icount_match(processor_t * const
   if (pending) {
     pending = 0;
     hit = true;
-    ret = match_result_t(TIMING_BEFORE, action);
+    if (action == ACTION_DEBUG_EXCEPTION) // decrease count only on firing breakpoint exception action
+      ret = match_result_t(TIMING_BEFORE, ACTION_DEBUG_EXCEPTION);
+    else
+      return match_result_t(TIMING_BEFORE, action);
```
scottj97: I don't like the duplicate expressions here in the if and else clauses. How about set `ret` first (the existing code), then test the action and `return ret` if needed?

scottj97 (changes requested):

riscv/triggers.cc
YenHaoChen: Thanks for the review and comment. The if-else clauses here could be more transparent. It messes up the execution flow and exits the function early, reducing the code's readability. 

How about simplifying the function's structure by keeping the function always goes through three steps: 1) setting ret=action, 2) decreasing counter, and 3) return ret. Although this implementation has a more complicated checking on whether to decrease the counter, it provides a better overall readability for future maintainers.

YenHaoChen:
The CI fails with ```mkdir: build: File exists```. I need help dealing with this issue.

scottj97 (approved):

aswaterman (approved):

PR #1338, opened 2023-04-24
fixup sb_write/sb_read to handle exceptions

aap-sc:
system bus read/write operations could lead to a variety of memory-related exceptions. Before this patch not every memory exception was handled. This could lead to simulator crashes: an example is when debugger (like OpenOCD) issues non-aligned memory read.

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -314,7 +314,7 @@ void debug_module_t::sb_read()
     } else {
       sbcs.error = 3;
     }
-  } catch (trap_load_access_fault& t) {
+  } catch (const mem_trap_t& ) {
     sbcs.error = 2;
   }
 }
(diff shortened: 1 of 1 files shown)


aswaterman:
@timsifive Does this change result in the desired behavior if the debugger performs a misaligned access?  (Obviously, the old behavior of crashing Spike was not the desired behavior :))

aap-sc:
@timsifive gentle ping.. Would you kindly take a look at this one?

aswaterman (approved):
@aap-sc I'm going to merge this because it's obviously better than crashing.  If @timsifive has any feedback post-merge, please work with him to make any necessary changes.

aap-sc:
@aswaterman thanks!

@timsifive - just for the reference. I've run the whole riscv-tests/gdb testsuite for OpenOCD on spike with this patch applied and all tests passed. I've also had some local testing to make sure that we won't crash on unaligned access initiated by the debugger.

timsifive (commented):
This is clearly an improvement. The sbcs.error field does support encodings for different types of failures. Specifically:
0 (none): There was no bus error.
1 (timeout): There was a timeout.
2 (address): A bad address was accessed.
3 (alignment): There was an alignment error.
4 (size): An access of unsupported size was re-
quested.
7 (other): Other.

Probably 2, 3, and 4 are relevant for us, and we could probably do a bit better mapping the actual exceptions to those error codes. In practice OpenOCD only puts the number in an error message, and I doubt anyone really cares. (On real hardware which is more opaque and complex, it probably matters more.)

While touching this code, you should probably use the constants from debug_defines.h:
```
#define DM_SBCS_SBERROR_NONE                0
#define DM_SBCS_SBERROR_TIMEOUT             1
#define DM_SBCS_SBERROR_ADDRESS             2
#define DM_SBCS_SBERROR_ALIGNMENT           3
#define DM_SBCS_SBERROR_SIZE                4
#define DM_SBCS_SBERROR_OTHER               7
```

PR #1339, opened 2023-04-25
csrrw/csrrwi should NOT read CSR when rd is zero

zhongchengyong:
According to the RISC-V spec, csrrw/csrrwi shouldn't read CSR when rd is zero, otherwise may cause side effects on user-customized registers

### DIFF
--- a/riscv/insns/csrrwi.h
+++ b/riscv/insns/csrrwi.h
@@ -1,5 +1,9 @@
 int csr = validate_csr(insn.csr(), true);
-reg_t old = p->get_csr(csr, insn, true);
+bool read = insn.rd() != 0;
+reg_t old = 0;
+if (read) {
+  old = p->get_csr(csr, insn, true);
+}
 p->put_csr(csr, insn.rs1());
 WRITE_RD(sext_xlen(old));
 serialize();
--- a/riscv/insns/csrrw.h
+++ b/riscv/insns/csrrw.h
@@ -1,5 +1,9 @@
 int csr = validate_csr(insn.csr(), true);
-reg_t old = p->get_csr(csr, insn, true);
+bool read = insn.rd() != 0;
+reg_t old = 0;
+if (read) {
+  old = p->get_csr(csr, insn, true);
+}
 p->put_csr(csr, RS1);
 WRITE_RD(sext_xlen(old));
 serialize();


riscv/insns/csrrwi.h
```diff
@@ -1,5 +1,9 @@
 int csr = validate_csr(insn.csr(), true);
-reg_t old = p->get_csr(csr, insn, true);
+bool read = insn.rd() != 0;
+reg_t old = 0;
+if (read) {
+  old = p->get_csr(csr, insn, true);
+}
```
aswaterman, line 6: A more compact version of this would be something like

    // Avoid actioning read side effects when rd is 0
    reg_t old = insn.rd() != 0 ? p->get_csr(csr, insn, true) : 0;

aswaterman (commented):
See this comment: https://github.com/riscv-software-src/riscv-isa-sim/blob/19a7f343af6fa197f3234e0630d10bd1e8db9849/riscv/processor.cc#L968-L970

This is probably an incomplete fix; for example, printing a register in interactive mode will also inadvertently action read side effects.

Since we don't have standard CSRs that have side effects on reads (and since I expect RVIA will maintain that discipline), this isn't a problem.  It's not clear to me that we want to merge an incomplete fix to something that isn't a problem for standard implementations.  What do you think @scottj97?

scottj97:
See these comments on `csr_t`:
```
  // read() returns the architectural value of this CSR. No permission                                                                           
  // checking needed or allowed. Side effects not allowed.                                                                                       
  virtual reg_t read() const noexcept = 0;
```

Side effects not allowed -- which is why the method is `const`. If side effects are needed, `class csr_t` needs a new method that's called from the `csrr*` family of instructions (or `get_csr` somehow) to trigger the side effect.

riscv/insns/csrrwi.h
zhongchengyong: Got it.

zhongchengyong:
> See these comments on `csr_t`:
> 
> ```
>   // read() returns the architectural value of this CSR. No permission                                                                           
>   // checking needed or allowed. Side effects not allowed.                                                                                       
>   virtual reg_t read() const noexcept = 0;
> ```
> 
> Side effects not allowed -- which is why the method is `const`. If side effects are needed, `class csr_t` needs a new method that's called from the `csrr*` family of instructions (or `get_csr` somehow) to trigger the side effect.

If a new function add for csr_t, such as "get_csr_se", we need call different function by checking rd==0. How about adding a parameter "peek" for the read function and make it non-const,  set the "peek" argument to true when rd==0 in "csrr*" family of instructions? In short,  the csr_t should support csr read which have side effects.

zhongchengyong:
> See this comment:
> 
> https://github.com/riscv-software-src/riscv-isa-sim/blob/19a7f343af6fa197f3234e0630d10bd1e8db9849/riscv/processor.cc#L968-L970
> 
> This is probably an incomplete fix; for example, printing a register in interactive mode will also inadvertently action read side effects.

Thanks for the kind reply. AFAIK, printing register value doesn't rely on the csrr instruction?

scottj97 (changes requested):
> > Side effects not allowed -- which is why the method is `const`. If side effects are needed, `class csr_t` needs a new method that's called from the `csrr*` family of instructions (or `get_csr` somehow) to trigger the side effect.
> 
> If a new function add for csr_t, such as "get_csr_se", we need call different function by checking rd==0. How about adding a parameter "peek" for the read function and make it non-const, set the "peek" argument to true when rd==0 in "csrr*" family of instructions? In short, the csr_t should support csr read which have side effects.

Skipping the call to `get_csr()` is a broken solution, because the permissions check is done there, and not inside `put_csr()`. This PR will allow unprivileged code to write to privileged CSRs.

I believe what Andrew was saying is that because Spike does not currently implement any CSRs with read side-effects, we don't want to add infrastructure here to support that. Unused code adds maintenance costs with no benefit to the project.

You are, of course, free to implement read side-effects however you want in your own fork of Spike with your custom CSRs. My advice as to the best way to do so would be to add a new method to `csr_t` to action the side effects, and call that from `get_csr()` based on a new boolean parameter, much like how `verify_permissions()` is called based on the `peek` param.

Then from the `csrr*` instructions, you set that param accordingly, based on `rd != 0`. Other places that call `get_csr()` (e.g. interactive printing) will set the new param to false so they don't trigger the read side-effects.

aswaterman:
I agree with everything @scottj97 said.

zhongchengyong:
> 

I got it. Thank you.

PR #1342, opened 2023-04-25
feat: add a dense memory backend implementation

tianrui-wei:
Currently, the mem_t only supports a sparse memory implementation. This
commit adds in dense memory support. It behaves as follows:

1. it will try to take advantage of HugeTLB and Transparent Huge Pages
in modern Linux kernels
2. it will allocate once/free once for every memory region, instead of
on demand
3. paves the way for a simple memory backed block device implementation

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>

### DIFF
--- a/riscv/devices.h
+++ b/riscv/devices.h
@@ -41,19 +41,33 @@ class mem_t : public abstract_device_t {
  public:
   mem_t(reg_t size);
   mem_t(const mem_t& that) = delete;
-  ~mem_t();
+  virtual ~mem_t() = default;
 
   bool load(reg_t addr, size_t len, uint8_t* bytes) { return load_store(addr, len, bytes, false); }
   bool store(reg_t addr, size_t len, const uint8_t* bytes) { return load_store(addr, len, const_cast<uint8_t*>(bytes), true); }
-  char* contents(reg_t addr);
   reg_t size() { return sz; }
-  void dump(std::ostream& o);
+  virtual char* contents(reg_t addr) = 0;
+  virtual void dump(std::ostream& o) = 0;
+
+ protected:
+  reg_t sz;
 
  private:
-  bool load_store(reg_t addr, size_t len, uint8_t* bytes, bool store);
+  virtual bool load_store(reg_t addr, size_t len, uint8_t* bytes, bool store) = 0;
+};
+
+class sparse_mem_t : public mem_t {
+ public:
+  sparse_mem_t(reg_t size);
+  ~sparse_mem_t();
+
+  virtual char* contents(reg_t addr) override;
+  virtual void dump(std::ostream& o) override;
+
+ private:
+  bool load_store(reg_t addr, size_t len, uint8_t* bytes, bool store) override;
 
   std::map<reg_t, char*> sparse_memory_map;
-  reg_t sz;
 };
 
 class clint_t : public abstract_device_t {
--- a/spike_main/spike.cc
+++ b/spike_main/spike.cc
@@ -270,7 +270,7 @@ static std::vector<std::pair<reg_t, mem_t*>> make_mems(const std::vector<mem_cfg
   std::vector<std::pair<reg_t, mem_t*>> mems;
   mems.reserve(layout.size());
   for (const auto &cfg : layout) {
-    mems.push_back(std::make_pair(cfg.get_base(), new mem_t(cfg.get_size())));
+    mems.push_back(std::make_pair(cfg.get_base(), new sparse_mem_t(cfg.get_size())));
   }
   return mems;
 }
--- a/ci-tests/testlib.c
+++ b/ci-tests/testlib.c
@@ -7,7 +7,7 @@ static std::vector<std::pair<reg_t, mem_t*>> make_mems(const std::vector<mem_cfg
   std::vector<std::pair<reg_t, mem_t*>> mems;
   mems.reserve(layout.size());
   for (const auto &cfg : layout) {
-    mems.push_back(std::make_pair(cfg.get_base(), new mem_t(cfg.get_size())));
+    mems.push_back(std::make_pair(cfg.get_base(), new sparse_mem_t(cfg.get_size())));
   }
   return mems;
 }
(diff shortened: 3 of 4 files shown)


spike_main/spike.cc
```diff
@@ -214,6 +216,12 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
     if (!*p || *p != ':')
       help();
     auto size = strtoull(p + 1, &p, 0);
+    if (*p && *p == ':') {
+      if (!*(p + 1))
+        help();
+      auto dense = *(p + 1) == 'd';
```
scottj97: This variable does not seem to be used anywhere?

spike_main/spike.cc
```diff
@@ -214,6 +216,12 @@ static std::vector<mem_cfg_t> parse_mem_layout(const char* arg)
...
+      p  = p + 2;
```
scottj97: Remove extraneous space; and why not `p += 2;`?

riscv/devices.h
```diff
@@ -41,21 +41,50 @@ class mem_t : public abstract_device_t {
...
+ private:
+  virtual bool load_store(reg_t addr, size_t len, uint8_t* bytes, bool store) = 0;
+};
+
+class sparse_mem_t : public mem_t {
+ public:
+  sparse_mem_t(reg_t size);
+  ~sparse_mem_t();
+
+  char* contents(reg_t addr);
```
scottj97: Use `virtual` and `override` keywords on overridden virtual methods

scottj97 (changes requested):
This does far too much in one commit:
1. Adding command-line option `:d`
2. Creating class hierarchy in `mem_t`
3. Adding `dense` to `mem_cfg_t`
4. Implementing `dense_mem_t`

That's at least 4 separate commits.

The bugfix just pushed should be squashed back in so there is no bug in any commit.

But more importantly, I think this needs more justification. What is the benefit here? Does this improve performance when simulating programs that actually use a large block of memory? By how much?

tianrui-wei:
Hi Scott,

Thanks for reviewing my PR. I'll address your comments in subsequent commits. I'll split this single commit as you suggested.

The benefit is that the current lazy allocation scheme is inefficient compared to allocating the entire backing memory at once. For example, in https://github.com/ucb-bar/chipyard/pull/1438, we unify the backing memory in both spike cosim and the actual RTL simulation to detect divergence. At the start of the cosimulation, the entire RTL memory is copied into spike, which in turn calls calloc many times.

The other reason is so that this could serve as the backing memory for a block device for using with drivers like https://github.com/u-boot/u-boot/blob/master/drivers/mmc/piton_mmc.c. If you'd like I could run some benchmark comparison on memcpy benchmarks.

Thanks,
Tianrui

aswaterman:
I suspect we can improve the current scheme to be more efficient, e.g. by using memalign and allocating larger chunks (say, 2 MiB to match the superpage size).  This would require some experimentation, but it seems preferable to having two different schemes.

I suspect the block device will end up being a separate `device_t` anyway, so I question whether this PR really serves as a building block towards that.

tianrui-wei:
I would argue that multiple backends are beneficial, and is also practical in other projects as qemu: https://github.com/qemu/qemu/blob/master/backends/hostmem.c. For one, the dense allocation will crash if allocated physical memory is larger than host memory, but the sparse backend could effectively test memory address > 39 bits for example. The dense allocation will also make sharing memory easier and more flexible instead of performing a lookup.

jerryz123:
@tianrui-wei for our use case, we can just provide our own implementation of mem_t that uses a dense memory in our code that links with spike, and pass it to the sim_t constructor.

tianrui-wei:
Could we perhaps only cherry pick https://github.com/riscv-software-src/riscv-isa-sim/pull/1342/commits/e71df645b199be1f3f2c85aecbadccbd93cc25d7 that implements an abstract class for memory interface, and expose a flexible interface in libriscv?

scottj97 (changes requested):
Commit e71df645b199be1f3f2c85aecbadccbd93cc25d7 fails to compile:
```
/local_home/sjohnson/spike-regress/riscv-isa-sim/spike_main/spike.cc: In function ‘std::vector<std::pair<long unsigned int, mem_t*> > make_mems(const std::vector<mem_cfg_t>&)’:
/local_home/sjohnson/spike-regress/riscv-isa-sim/spike_main/spike.cc:273:75: error: invalid new-expression of abstract class type ‘mem_t’
             mems.push_back(std::make_pair(cfg.get_base(), new mem_t(cfg.get_size())));
                                                                                   ^
```

riscv/devices.h
```diff
@@ -41,21 +41,50 @@ class mem_t : public abstract_device_t {
...
+  virtual bool load_store(reg_t addr, size_t len, uint8_t* bytes, bool store) = 0;
+};
+
+class sparse_mem_t : public mem_t {
+ public:
+  sparse_mem_t(reg_t size);
+  ~sparse_mem_t();
+
+  char* contents(reg_t addr) override;
+  void dump(std::ostream& o) override;
```
scottj97: You added `override` but not `virtual`. Please add `virtual` to be consistent with other code in Spike

tianrui-wei: Thank you, addressed in [`69e70ad` (#1342)](https://github.com/riscv-software-src/riscv-isa-sim/pull/1342/commits/69e70ad0bdbbe68d9f3fc2868f46174da7e49478)

tianrui-wei:
Hi Scott,

Thank you for reviewing and shepherding the PR. I've updated the commit to address your reviews and only cherry-picked the first commit.

Thanks,
Tianrui

riscv/devices.h
```diff
@@ -41,19 +41,33 @@ class mem_t : public abstract_device_t {
...
+class sparse_mem_t : public mem_t {
+ public:
+  sparse_mem_t(reg_t size);
+  ~sparse_mem_t();
+
+  virtual char* contents(reg_t addr) override;
+  virtual void dump(std::ostream& o) override;
+
+ private:
+  bool load_store(reg_t addr, size_t len, uint8_t* bytes, bool store) override;
```
scottj97, line 68: Needs `virtual`

scottj97 (changes requested):
I'll leave it up to @aswaterman but this seems like a pointless change now.

jerryz123:
Resolved by #1408

Issue #549, opened 2020-09-22
Make target `install-pc` fails


huksys:
Hello, the latest commit on master, 59d450e, unfortunately introduced a bug into the build system.

The make target `install-pc`, which is a prerequisite of the `install` target, fails due to a missing rule for `riscv-disasm.pc`:
> make: *** No rule to make target 'riscv-disasm.pc', needed by 'install-pc'.  Stop.

It seems like the `install_pcs` variable is appended to by the `subproject_define` macro ([Makefile.in#L342](https://github.com/riscv/riscv-isa-sim/blob/59d450e58646bdc0af5e9250df7d6267c3c791fc/Makefile.in#L342)) which is used for each enabled subproject ([Makefile#L348](https://github.com/riscv/riscv-isa-sim/blob/59d450e58646bdc0af5e9250df7d6267c3c791fc/Makefile.in#L348)). Unfortunately, I cannot provide a patch due to my inexperience with autotools.

aswaterman:
Thanks for the bug report.  Not clear why I didn't see that during local testing.  Will be fixed presently.

PR #1346, opened 2023-05-02
Install meta pkg-config file for Spike.

Wojciech-Bartczak:
Using a Spike a stand-alone library is interesting option to develop own solutions based on Spike without the need to directly intake it as source code.
It also creates interface that prevents unrelated changes to Spike itself.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -337,7 +337,7 @@ install_hdrs += $$(addprefix $(src_dir)/$(1)/, $$($(2)_install_hdrs))
 install_libs += $$(if $$($(2)_install_lib),lib$(1).a,)
 install_libs += $$(if $$($(2)_install_shared_lib),lib$(1).so,)
 install_exes += $$($(2)_install_prog_exes)
-install_pcs  += $$(if $$($(2)_install_lib),riscv-$(1).pc,)
+install_pcs  += $$(if $$($(2)_install_pcs),riscv-$(1).pc,)
 
 endef
 
--- a/configure
+++ b/configure
@@ -6362,6 +6362,8 @@ ac_config_files="$ac_config_files riscv-fesvr.pc"
 
 ac_config_files="$ac_config_files riscv-disasm.pc"
 
+ac_config_files="$ac_config_files riscv-riscv.pc"
+
 cat >confcache <<\_ACEOF
 # This file is a shell script that caches the results of configure
 # tests run on this system so they can be shared between configure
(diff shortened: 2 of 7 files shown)


Wojciech-Bartczak:
Patch addresses problem in fix #549. However, it doesn't remove option to use Spike as a library.
This definitely simplifies creating tools and projects based on Spike without temptation to change Spike itself.

Wojciech-Bartczak:
The example usage is available here:
https://github.com/Wojciech-Bartczak/spike-as-lib-example

scottj97 (dismissed):
I don't understand the Make flow well enough to understand what this does, who it benefits, how to use it, etc. So I can't really form an opinion on whether we should merge this or not.

In any case, PRs like this should not have embedded merges in them. If you want to update this to work on the latest master, please use rebase and force-push.

aswaterman:
This is OK with me, once this is rebased and the commit history is cleaned up.

Wojciech-Bartczak:
Hi Andrew, Hi Scott,

Thank you for reviewing the change. Although I'm not very familiar with Github, I will ensure that the patch is fixed to meet the standards.

Regarding the functionality, prior to #549, each subproject (`riscv`, `disasm`, etc.) had its own `*.pc` file (a `pkg-config` configuration file for the library), and all libraries were installed. This allowed Spike to be integrated as a library using `pkg-config`, benefiting tooling and internal programs that utilized Spike.

However, one of the subprojects caused a compilation error, leading to #549. As a result, only `disasm` and `fesrv` remained as installed libraries accessible to users through the `pkg-config` mechanism. Although `riscv.so` was still available, it lacked an attached `*.pc` file. Consequently, users had to manually specify the include list and paths for the compiler to build binaries based on `riscv.so`.

This issue arose because the `install-pcs` target was only evaluated for static libraries (specifically, the `install-libs` target).

The patch I created restores half of the libraries to their original state, giving you the option to choose which libraries are delivered/exposed to users through `*.pc` files.

To achieve this, I added an intermediate make variable that is set to `yes` for libraries intended to be externally visible through `pkg-config`. Subsequently, this variable is used to generate the `*.pc` files during the make install step.

I hope this clarifies the rationale behind the changes.

Kind regards,
Wojciech Bartczak.

aswaterman (approved):

aswaterman:
Thanks for writing up the explanation.  I believe Scott is out of pocket, so I'll override his review and merge.

PR #1347, opened 2023-05-03
Fix for #872 mstatus.GVA

rbuchner-aril:
Replaces xlate_flags with a data structure that is passed around and know the privilege of a load or store.

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -391,7 +391,7 @@ private:
   bool mmio_store(reg_t paddr, size_t len, const uint8_t* bytes);
   bool mmio(reg_t paddr, size_t len, uint8_t* bytes, access_type type);
   bool mmio_ok(reg_t paddr, access_type type);
-  void check_triggers(triggers::operation_t operation, reg_t address, std::optional<reg_t> data = std::nullopt);
+  void check_triggers(triggers::operation_t operation, reg_t address, bool virt, std::optional<reg_t> data = std::nullopt);
   reg_t translate(mem_access_info_t access_info, reg_t len);
 
   reg_t pte_load(reg_t pte_paddr, reg_t addr, bool virt, access_type trap_type, size_t ptesize) {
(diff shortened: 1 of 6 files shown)


scottj97 (approved):
I've worked with @rbuchner-aril on this internally. Once the CI issues are fixed (do I need to update my GCC to find this? Or is this only a clang thing?) then I think this is good to go.

aswaterman:
Thank you, @rbuchner-aril.  I'll review tomorrow.

@scottj97 It appears to fail only on the Ubuntu target and not the MacOS target, so it's likely that upgrading GCC is necessary to replicate the bug locally.  But it looks trivial:

```
/home/runner/work/riscv-isa-sim/riscv-isa-sim/ci-tests/../riscv/mmu.cc:63:8: error: unused variable ‘hlvx’ [-Werror=unused-variable]
   63 |   bool hlvx = access_info.flags.hlvx;
```

riscv/mmu.h
```diff
@@ -38,6 +38,12 @@ struct tlb_entry_t {
   reg_t target_offset;
 };
 
+struct xlate_flags_t {
```
aswaterman, line 41: For reasons I didn't have time to thoroughly investigate, passing this struct by value reduces performance substantially on my Mac M1.

If you make it a bitfield struct (i.e. the below), the perf problem goes away while preserving the semantics.

```
struct xlate_flags_t {
  const bool forced_virt : 1;
  const bool hlvx : 1;
  const bool lr : 1;
};
```

aswaterman (changes requested):
Looks great; once my minor performance comment is addressed, we should merge this.

riscv/mmu.h
rbuchner-aril: Updated.

aswaterman: Thanks!

aswaterman (approved):
LGTM now.

PR #1348, opened 2023-05-04
triggers: implement {mcontrol,mcontrol6}.match = {not equal,not napot,not mask low,not mask high}

YenHaoChen:
This PR implements {mcontrol,mcontrol6}.match = {not equal,not napot,not mask low,not mask high}.

### DIFF
--- a/riscv/triggers.h
+++ b/riscv/triggers.h
@@ -194,7 +194,11 @@ public:
     MATCH_GE = MCONTROL_MATCH_GE,
     MATCH_LT = MCONTROL_MATCH_LT,
     MATCH_MASK_LOW = MCONTROL_MATCH_MASK_LOW,
-    MATCH_MASK_HIGH = MCONTROL_MATCH_MASK_HIGH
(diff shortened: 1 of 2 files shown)


riscv/triggers.h
YenHaoChen, line 1: The encoding.h does not have MCONTROL_MATCH_NOT_EQUAL, MCONTROL_MATCH_NOT_NAPOT, MCONTROL_MATCH_NOT_MASK_LOW, MCONTROL_MATCH_NOT_MASK_HIGH. Without modifying encoding.h (or without including debug_defines.h), we must hardcode the values. However, that makes this part inconsistent. Should we also hardcode the values of MATCH_EQUAL, MATCH_NAPOT, MATCH_GE, MATCH_LT, MATCH_MASK_LOW, and MATCH_MASK_HIGH?

YenHaoChen:
@timsifive Would you share some words on this?

riscv/triggers.h
timsifive: What's the problem with including debug_defines.h? That seems easier to maintain, etc. Are there macro name conflicts?

PR #1350, opened 2023-05-08
triggers: Fix etrigger match on exceptions

atulkharerivos:
This change is composed of four patches.

1.  The first two patches add new fields to processor state
3. The third patch adds additional parameters to mode_match() in anticipation of forthcoming changes.
4. The final patch adds a new parameter to common_match() and uses it to match trigger edge cases.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -197,7 +197,7 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   auto xlen = proc->get_isa().get_max_xlen();
 
   prv = PRV_M;
-  v = false;
+  v = prev_v = false;
   csrmap[CSR_MISA] = misa = std::make_shared<misa_csr_t>(proc, CSR_MISA, max_isa);
   mstatus = std::make_shared<mstatus_csr_t>(proc, CSR_MSTATUS);
 
@@ -747,17 +747,16 @@ void processor_t::set_virt(bool virt)
   if (state.prv == PRV_M)
     return;
 
-  if (state.v != virt) {
-    /*
-     * Ideally, we should flush TLB here but we don't need it because
-     * set_virt() is always used in conjucter with set_privilege() and
-     * set_privilege() will flush TLB unconditionally.
-     *
-     * The virtualized sstatus register also relies on this TLB flush,
-     * since changing V might change sstatus.MXR and sstatus.SUM.
-     */
-    state.v = virt;
-  }
+  /*
+    * Ideally, we should flush TLB here but we don't need it because
+    * set_virt() is always used in conjucter with set_privilege() and
+    * set_privilege() will flush TLB unconditionally.
+    *
+    * The virtualized sstatus register also relies on this TLB flush,
+    * since changing V might change sstatus.MXR and sstatus.SUM.
+    */
+  state.prev_v = state.v;
+  state.v = virt;
 }
 
 void processor_t::enter_debug_mode(uint8_t cause)
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -196,7 +196,7 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   // mstatus_csr_t::unlogged_write()):
   auto xlen = proc->get_isa().get_max_xlen();
 
-  prv = PRV_M;
+  prv = prev_prv = PRV_M;
   v = prev_v = false;
   csrmap[CSR_MISA] = misa = std::make_shared<misa_csr_t>(proc, CSR_MISA, max_isa);
   mstatus = std::make_shared<mstatus_csr_t>(proc, CSR_MSTATUS);
@@ -717,6 +717,7 @@ reg_t processor_t::legalize_privilege(reg_t prv)
 void processor_t::set_privilege(reg_t prv)
 {
   mmu->flush_tlb();
+  state.prev_prv = state.prv;
   state.prv = legalize_privilege(prv);
 }
 
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -84,6 +84,7 @@ struct state_t
   std::unordered_map<reg_t, csr_t_p> csrmap;
   reg_t prv;    // TODO: Can this be an enum instead?
   bool v;
+  bool prev_v;
   misa_csr_t_p misa;
   mstatus_csr_t_p mstatus;
   csr_t_p mstatush;
--- a/riscv/processor.h
+++ b/riscv/processor.h
@@ -83,6 +83,7 @@ struct state_t
   // control and status registers
   std::unordered_map<reg_t, csr_t_p> csrmap;
   reg_t prv;    // TODO: Can this be an enum instead?
+  reg_t prev_prv;
   bool v;
   bool prev_v;
   misa_csr_t_p misa;
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -56,15 +56,16 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
 }
 
 bool trigger_t::common_match(processor_t * const proc) const noexcept {
-  return mode_match(proc->get_state()) && textra_match(proc);
+  auto state = proc->get_state();
+  return mode_match(state->prv, state->v) && textra_match(proc);
 }
 
-bool trigger_t::mode_match(state_t * const state) const noexcept
+bool trigger_t::mode_match(reg_t prv, bool v) const noexcept
 {
-  switch (state->prv) {
+  switch (prv) {
     case PRV_M: return m;
-    case PRV_S: return state->v ? vs : s;
-    case PRV_U: return state->v ? vu : u;
+    case PRV_S: return v ? vs : s;
+    case PRV_U: return v ? vu : u;
     default: assert(false);
   }
 }
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -55,9 +55,11 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
   sselect = (sselect_t)((proc->extension_enabled_const('S') && get_field(val, CSR_TEXTRA_SSELECT(xlen)) <= SSELECT_MAXVAL) ? get_field(val, CSR_TEXTRA_SSELECT(xlen)) : SSELECT_IGNORE);
 }
 
-bool trigger_t::common_match(processor_t * const proc) const noexcept {
+bool trigger_t::common_match(processor_t * const proc, bool use_prev_prv) const noexcept {
   auto state = proc->get_state();
-  return mode_match(state->prv, state->v) && textra_match(proc);
+  auto prv = use_prev_prv ? state->prev_prv : state->prv;
+  auto v = use_prev_prv ? state->prev_v : state->v;
+  return mode_match(prv, v) && textra_match(proc);
 }
 
 bool trigger_t::mode_match(reg_t prv, bool v) const noexcept
@@ -398,7 +400,8 @@ void itrigger_t::tdata1_write(processor_t * const proc, const reg_t val, const b
 
 std::optional<match_result_t> trap_common_t::detect_trap_match(processor_t * const proc, const trap_t& t) noexcept
 {
-  if (!common_match(proc))
+  // Use the previous privilege for matching
+  if (!common_match(proc, true))
     return std::nullopt;
 
   auto xlen = proc->get_xlen();
--- a/riscv/triggers.h
+++ b/riscv/triggers.h
@@ -102,7 +102,7 @@ protected:
 
 private:
   unsigned legalize_mhselect(bool h_enabled) const noexcept;
-  bool mode_match(state_t * const state) const noexcept;
+  bool mode_match(reg_t prv, bool v) const noexcept;
   bool textra_match(processor_t * const proc) const noexcept;
 
   struct mhselect_interpretation {
--- a/riscv/triggers.h
+++ b/riscv/triggers.h
@@ -90,7 +90,7 @@ public:
 
 protected:
   static action_t legalize_action(reg_t val, reg_t action_mask, reg_t dmode_mask) noexcept;
-  bool common_match(processor_t * const proc) const noexcept;
+  bool common_match(processor_t * const proc, bool use_prev_prv = false) const noexcept;
   bool allow_action(const state_t * const state) const;
   reg_t tdata2;
 


riscv/execute.cc
```diff
@@ -304,13 +304,15 @@ void processor_t::step(size_t n)
...
+      // save exception PC before it gets changed by take_trap
+      const reg_t exc_pc = pc;
       take_trap(t, pc);
       n = instret;
 
       // Trigger action takes priority over single step
       auto match = TM.detect_trap_match(t);
       if (match.has_value())
-        take_trigger_action(match->action, 0, state.pc);
+        take_trigger_action(match->action, 0, exc_pc);
```
YenHaoChen: This part of the modification seems to address another issue. Perhaps, we can have another commit for this modification.

riscv/processor.h
```diff
@@ -83,6 +83,7 @@ struct state_t
   // control and status registers
   std::unordered_map<reg_t, csr_t_p> csrmap;
   reg_t prv;    // TODO: Can this be an enum instead?
+  reg_t prev_prv;
```
YenHaoChen, line 86: We can use the existing variable "last_inst_priv" of the same struct "state_t".

YenHaoChen:
The execution flow calls the trap-taking function, i.e., take_trap(t, pc), before the itrigger/etrigger-checking function, i.e., TM.detect_trap_match(t). The trap-taking function may change the current state, including state.pc and state.prv. However, the trigger-checking function should take the unchanged (stashed) state as input. 

The commit message points out a solid case of the issue. Without this PR, an etrigger monitoring only U-mode ecall fails to provide the functionality since the state.prv can never be U-mode on the etrigger checking.

This PR proposes to use the stashed state of both PC and privilege mode in the trigger-checking function to resolve the issue.

riscv/processor.h
atulkharerivos: Updated per the suggestion. Note that since `last_inst_priv` was being updated in the logging path, the code was changed to make the update unconditional in both the fast and logged execution paths.

scottj97: > Updated per the suggestion. Note that since `last_inst_priv` was being updated in the logging path, the code was changed to make the update unconditional in both the fast and logged execution paths.

This sounds like a bug fix if the `last_inst_priv` wasn't being updated in the fast path before?

If I understand correctly, the bug fix should be a separate commit explaining that change in isolation.

riscv/execute.cc
```diff
@@ -14,7 +14,7 @@ static void commit_log_reset(processor_t* p)
   p->get_state()->log_mem_write.clear();
 }
 
-static void commit_log_stash_privilege(processor_t* p)
+static void stash_privilege(processor_t* p)
```
scottj97: Why change the function name? It clutters up this commit with other things, obscuring the substantive changes. If it really makes sense to change the function name, that should be a separate commit.

riscv/processor.h
atulkharerivos: That is correct. Please see if the update (and new commit) reflects the intent of the change.

riscv/execute.cc
atulkharerivos: Please see new commit to see if it better explains the rationale.

YenHaoChen: The stashed state information saves the previous state before the execution of the instruction. The previous implementation uses the information only for logging after the execution and, thus, maintains the variables only when logging is enabled.

This PR points out the stashed state information is also required for the desired functionality, i.e., state.last_inst_priv for trigger checking. Thus, the PR proposes maintaining the information unconditionally regardless of whether in fast/slow path or logging is enabled or not. (Otherwise, we may need to introduce another variable with the same value if valid. That was the previous proposal of this PR. https://github.com/riscv-software-src/riscv-isa-sim/pull/1350#discussion_r1191791795)

Also, changing variable validity by modifying the calling condition and renaming the function for better description could be two rational independent commits, which may provide more clear reasoning for this PR.

atulkharerivos: @YenHaoChen Good suggestion; updated per the above.

riscv/execute.cc
```diff
@@ -14,7 +14,7 @@ static void commit_log_reset(processor_t* p)
   p->get_state()->log_mem_write.clear();
 }
 
-static void commit_log_stash_privilege(processor_t* p)
+static void commit_stash_privilege(processor_t* p)
```
YenHaoChen: This commit can keep the function name "commit_log_stash_privilege()."

atulkharerivos: My bad...cut and paste error...fixed.

YenHaoChen (approved):
The trigger matching function, i.e., trigger_t::mode_match(), requires the stashed last_inst_priv for the desired functionality regardless of fast/slow path or logging enabled. Thus, this commit proposes maintaining the stashed variables unconditionally.

I believe this is the easiest way to maintain the last_inst_priv. However, this commit also maintains additional unused stashed variables, i.e., last_inst_xlen and last_inst_flen. Additionally, the commit introduces another function call to the fast path. Others may give further comments on that.

YenHaoChen (approved):
The new name clarifies that the function is no longer only for logging but also generally for desired behavior.

YenHaoChen (approved):
The mode_match() should consider the state->last_inst_priv, i.e., the privilege mode before the execution of the current instruction, instead of the state->prv.

For instance, an etrigger monitoring only U-mode ecall fails to provide the functionality because the state->prv can never be U-mode on the etrigger checking without this commit.

YenHaoChen (approved):

scottj97 (changes requested):
I am seeing this PR cause differences on icount breakpoints and possibly others. I'm investigating.

riscv/triggers.cc
```diff
@@ -61,7 +61,7 @@ bool trigger_t::common_match(processor_t * const proc) const noexcept {
 
 bool trigger_t::mode_match(state_t * const state) const noexcept
 {
-  switch (state->prv) {
+  switch (state->last_inst_priv) {
```
scottj97: This change will apply to `icount` and `mcontrol` type breakpoints as well, and will break them, because it will be considering the privilege of the previous instruction instead of the current.

It seems that it's only for `detect_trap_match()` that we want this behavior.

Could you get that behavior by calling `detect_trap_match()` just before calling `take_trap()`? I suspect that would work, and then you wouldn't need to use `last_inst_priv` at all.

Edit: I don't think it's that simple...`common_match()` needs to consider the previous privilege, but `allow_action()` needs to consider the new privilege, IIUC. Both of these are elements of `detect_trap_match()`.

scottj97 (changes requested):

riscv/triggers.cc
atulkharerivos: @scottj97 A coworker suggested something along the lines of the following based on your comment. It seems reasonable, but maybe there be dragons?

```
diff --git a/riscv/execute.cc b/riscv/execute.cc
index 295879d4..b92fffdb 100644
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -304,16 +304,19 @@ void processor_t::step(size_t n)
     }
     catch(trap_t& t)
     {
-      take_trap(t, pc);
       n = instret;
 
       // Trigger action takes priority over single step
       auto match = TM.detect_trap_match(t);
-      if (match.has_value())
+      if (match.has_value()) {
+        take_trap(t, pc);
         take_trigger_action(match->action, 0, state.pc, 0);
-      else if (unlikely(state.single_step == state.STEP_STEPPED)) {
-        state.single_step = state.STEP_NONE;
-        enter_debug_mode(DCSR_CAUSE_STEP);
+      } else {
+        take_trap(t, pc);
+        if (unlikely(state.single_step == state.STEP_STEPPED)) {
+          state.single_step = state.STEP_NONE;
+          enter_debug_mode(DCSR_CAUSE_STEP);
+        }
       }
     }
     catch (triggers::matched_t& t)
```

scottj97: In `detect_trap_match()` you need to know both the pre-trap privilege (for `common_match()`) and post-trap privilege (for `allow_action()`), yes? I could be wrong about this but it seems so since the breakpoint trap is being taken from the post-trap privilege (mepc will point to the first instruction of the handler, yes?).

I don't see how this proposed change could possibly do that.

scottj97: Also if it turns out that I'm wrong and your proposed change above is correct, I think it could be done much cleaner by simply moving the `take_trap()` call to be between the `auto match` and `if match` lines.

atulkharerivos: @scottj97 Since it's a little difficult to talk about things in the abstract, let's consider the following scenario:

1. Trigger is set on an ECALL from U-mode
2. U-mode ECALL instruction is at pc = 0xAA
3. MTVEC handler at pc = 0xBB

It seems that with the suggested change, when pc = 0xAA:

- `detect_trap_match()` will succeed because we still haven't changed the privilege level. This is effectively the "pre-trap privilege level"
- Following the match, `take_trap()` is called. This will result in the "post-trap" privilege level (M) and  state.pc = 0xBB, which is the first instruction of the MTVEC handler as you noted. The mepc is set to 0xAA
-` take_trigger_action()` is now called with the new pc (0xBB)

In my mind, this is indeed the expected behavior. As you pointed out, we can implement it more succinctly by moving` trap_trap()` after the `auto match`, but it seems like the approach should work.

scottj97: Assume that neither ECALL nor breakpoint traps are delegated beyond M-mode. Assume an etrigger with action=0 (breakpoint exception).

The problem in your example is that the trigger action should have been suppressed by the mechanism in #1316, which is intended to prevent a reentrant M-mode exception from clobbering `mepc` and making recovery impossible. Inside `detect_trap_match()`, it calls `allow_action()` which should detect that we're already in M-mode and `mstatus.mie==0` therefore we should not allow this action.

It will fail to detect this because privilege is still set to U-mode. The result will be two consecutive M-mode exceptions, and runaway software, because the breakpoint trap will set mepc to 0xBB and the original 0xAA address will be lost.

atulkharerivos: @scottj97 Thanks for pointing out the issue with the alternative approach.

Based on your previous feedback, it seems that one of the main issues with matching on the previous privilege was that it was causing icount match tests to fail. Having looked at the code again, the likely cause was that all those code paths use  `common_match()`.  I have attempted to resolve that issue by special casing the trigger matches for traps...PTAL.

riscv/triggers.h
```diff
@@ -102,7 +102,7 @@ class trigger_t {
 
 private:
   unsigned legalize_mhselect(bool h_enabled) const noexcept;
-  bool mode_match(state_t * const state) const noexcept;
+  bool mode_match(state_t * const state, bool use_last_inst_priv = false) const noexcept;
```
YenHaoChen: Perhaps, passing the prv as a parameter directly will be cleaner. Additionally, should we pass the last_inst_v?
```bool mode_match(reg_t prv, bool v) const noexcept;```

File: riscv/triggers.h [hunk 7]
```diff
@@ -91,6 +91,8 @@ class trigger_t {
 protected:
   static action_t legalize_action(reg_t val, reg_t action_mask, reg_t dmode_mask) noexcept;
   bool common_match(processor_t * const proc) const noexcept;
+  // Same as common_match, but based on previous privilege
+  bool trap_match(processor_t * const proc) const noexcept;
```
YenHaoChen: We only call the common_match() twice and the trap_match() only once. We may remove (inline) the common_match() and trap_match() functions because those function names barely present additional messages.

scottj97:
I like @YenHaoChen 's suggestions.

I confirm that as of 093d9ee08629a10bbd2e09d7ed5b12e6025cbb0f I can no longer see an impact on icount & mcontrol breakpoint types.

riscv/triggers.cc
```diff
@@ -55,13 +55,15 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
   sselect = (sselect_t)((proc->extension_enabled_const('S') && get_field(val, CSR_TEXTRA_SSELECT(xlen)) <= SSELECT_MAXVAL) ? get_field(val, CSR_TEXTRA_SSELECT(xlen)) : SSELECT_IGNORE);
 }
 
-bool trigger_t::common_match(processor_t * const proc) const noexcept {
-  return mode_match(proc->get_state()) && textra_match(proc);
+bool trigger_t::common_match(processor_t * const proc, bool use_last_inst_priv) const noexcept {
+  auto state = proc->get_state();
+  auto prv = use_last_inst_priv ? state->last_inst_priv : state->prv;
+  return mode_match(state, prv) && textra_match(proc);
```
YenHaoChen: We can inline those code.
```!common_match(proc)```  ->  ```!mode_match(proc->get_state()->prv, proc->get_state()->v) || !textra_match(proc)```

File: riscv/triggers.h [hunk 9]
```diff
@@ -102,7 +102,7 @@ class trigger_t {
 
 private:
   unsigned legalize_mhselect(bool h_enabled) const noexcept;
-  bool mode_match(state_t * const state) const noexcept;
+  bool mode_match(state_t * const state, reg_t prv) const noexcept;
```
YenHaoChen: How about explicitly passing "prv" and "v" parameters?
```bool mode_match(reg_t prv, bool v) const noexcept;```

File: riscv/triggers.cc [hunk 10]
```diff
@@ -55,16 +55,18 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
   sselect = (sselect_t)((proc->extension_enabled_const('S') && get_field(val, CSR_TEXTRA_SSELECT(xlen)) <= SSELECT_MAXVAL) ? get_field(val, CSR_TEXTRA_SSELECT(xlen)) : SSELECT_IGNORE);
 }
 
-bool trigger_t::common_match(processor_t * const proc) const noexcept {
-  return mode_match(proc->get_state()) && textra_match(proc);
+bool trigger_t::common_match(processor_t * const proc, bool use_last_inst_priv) const noexcept {
+  auto state = proc->get_state();
+  auto prv = use_last_inst_priv ? state->last_inst_priv : state->prv;
+  return mode_match(prv, state->v) && textra_match(proc);
```
scottj97: You need to use `state->v` from the previous instruction, too!

scottj97 (changes requested):
70333658d8196ebae78ef937cef9fcc26b53f59b ought to be two commits: change the params to `mode_match()`, then separately add the `use_last_inst_priv`.

I would also combine the `use_last_inst_priv` change with the next commit, so that it gets used in the same commit where the code is added. (Otherwise there's one commit where the new code is unused, which means `git bisect` wouldn't be able to find problems in it.)

riscv/triggers.cc
atulkharerivos: Good catch...would have required another patch down the road. I have added an additional commit that introduces a last_v state field.

atulkharerivos:
> [7033365](https://github.com/riscv-software-src/riscv-isa-sim/commit/70333658d8196ebae78ef937cef9fcc26b53f59b) ought to be two commits: change the params to `mode_match()`, then separately add the `use_last_inst_priv`.
> 
> I would also combine the `use_last_inst_priv` change with the next commit, so that it gets used in the same commit where the code is added. (Otherwise there's one commit where the new code is unused, which means `git bisect` wouldn't be able to find problems in it.)

Updated.

scottj97 (approved):
Thank you for your patience with the review process.

aswaterman (changes requested):
Unconditionally stashing the privilege information has a large impact on performance (25% or so).  My intuition is that this can be accomplished without doing so, as it only changes on exceptions and interrupts, perhaps in the `processor_t::take_trap` routine.  @scottj97 might have some ideas here.

Note the difference in the main simulation loop (before and after):

```
  a1d030: 48 8b 34 24           mov    (%rsp),%rsi
  a1d034: 48 8d 56 01           lea    0x1(%rsi),%rdx                                                     
  a1d038: 48 3b 54 24 08        cmp    0x8(%rsp),%rdx
  a1d03d: 74 27                 je     a1d066 <_ZN11processor_t4stepEm+0x126>                             
  a1d03f: 48 89 14 24           mov    %rdx,(%rsp)                                                        
  a1d043: 49 89 c7              mov    %rax,%r15
  a1d046: 48 89 45 70           mov    %rax,0x70(%rbp)                                                    
  a1d04a: 4c 8b 63 18           mov    0x18(%rbx),%r12                                                    
  a1d04e: 48 8b 4b 10           mov    0x10(%rbx),%rcx                                                    
  a1d052: 4c 89 fa              mov    %r15,%rdx                                                          
  a1d055: 4c 89 e6              mov    %r12,%rsi
  a1d058: 48 89 ef              mov    %rbp,%rdi
  a1d05b: ff d1                 callq  *%rcx
  a1d05d: 48 8b 5b 08           mov    0x8(%rbx),%rbx                                                     
  a1d061: 48 39 03              cmp    %rax,(%rbx)                                                        
  a1d064: 74 ca                 je     a1d030 <_ZN11processor_t4stepEm+0xf0>                              
```

```
  a1d068: 48 8b 74 24 08        mov    0x8(%rsp),%rsi                                                     
  a1d06d: 48 8d 56 01           lea    0x1(%rsi),%rdx                                                     
  a1d071: 48 3b 54 24 10        cmp    0x10(%rsp),%rdx                                                    
  a1d076: 0f 84 8a 00 00 00     je     a1d106 <_ZN11processor_t4stepEm+0x196>                             
  a1d07c: 48 89 54 24 08        mov    %rdx,0x8(%rsp)                                                     
  a1d081: 48 89 04 24           mov    %rax,(%rsp)                                                        
  a1d085: 48 89 43 70           mov    %rax,0x70(%rbx)                                                    
  a1d089: 4c 8b 45 10           mov    0x10(%rbp),%r8                                                     
  a1d08d: 4c 8b 65 18           mov    0x18(%rbp),%r12                                                    
  a1d091: 48 8b 83 b0 03 00 00  mov    0x3b0(%rbx),%rax                                                   
  a1d098: ba 80 00 00 00        mov    $0x80,%edx                                                         
  a1d09d: 48 89 83 48 0e 00 00  mov    %rax,0xe48(%rbx)                                                   
  a1d0a4: 8b 83 64 0e 00 00     mov    0xe64(%rbx),%eax                                                   
  a1d0aa: 89 83 50 0e 00 00     mov    %eax,0xe50(%rbx)                                                   
  a1d0b0: 48 8b 83 c0 03 00 00  mov    0x3c0(%rbx),%rax                                                   
  a1d0b7: 48 8b 48 28           mov    0x28(%rax),%rcx                                                    
  a1d0bb: f7 c1 00 00 01 00     test   $0x10000,%ecx                                                      
  a1d0c1: 75 15                 jne    a1d0d8 <_ZN11processor_t4stepEm+0x168>                             
  a1d0c3: ba 40 00 00 00        mov    $0x40,%edx                                                         
  a1d0c8: f6 c1 08              test   $0x8,%cl                                                           
  a1d0cb: 75 0b                 jne    a1d0d8 <_ZN11processor_t4stepEm+0x168>                             
  a1d0cd: 31 d2                 xor    %edx,%edx                                                          
  a1d0cf: 83 e1 20              and    $0x20,%ecx                                                         
  a1d0d2: 0f 95 c2              setne  %dl                                                                
  a1d0d5: c1 e2 05              shl    $0x5,%edx                                                          
  a1d0d8: 0f b6 83 b8 03 00 00  movzbl 0x3b8(%rbx),%eax                                                   
  a1d0df: 89 93 54 0e 00 00     mov    %edx,0xe54(%rbx)                                                   
  a1d0e5: 4c 89 e6              mov    %r12,%rsi                                                          
  a1d0e8: 48 89 df              mov    %rbx,%rdi                                                          
  a1d0eb: 48 8b 14 24           mov    (%rsp),%rdx                                                        
  a1d0ef: 88 83 58 0e 00 00     mov    %al,0xe58(%rbx)                                                    
  a1d0f5: 41 ff d0              callq  *%r8                                                               
  a1d0f8: 48 8b 6d 08           mov    0x8(%rbp),%rbp                                                     
  a1d0fc: 48 39 45 00           cmp    %rax,0x0(%rbp)                                                     
  a1d100: 0f 84 62 ff ff ff     je     a1d068 <_ZN11processor_t4stepEm+0xf8>                              
```

scottj97:
> Unconditionally stashing the privilege information has a large impact on performance (25% or so). My intuition is that this can be accomplished without doing so, as it only changes on exceptions and interrupts, perhaps in the `processor_t::take_trap` routine. @scottj97 might have some ideas here.

It can also change with `mret`, `sret`, and `dret`.

All these use `processor_t::set_privilege()`. But the problem is that `last_inst_priv` needs to be changed not when the privilege changes, but one instruction later.

Perhaps set a flag `processor_t::privilege_changed`, force the slow path when that flag is set, and test that flag after every slow-path instruction executes?

In any case @atulkharerivos please don't go back and rework the commits here. We have fully functional code as of https://github.com/riscv-software-src/riscv-isa-sim/commit/7a2ff14bff461f2c7adfbf407e11527f55d920db; performance improvements should be done as additional commits beyond that, so we can easily demonstrate that the performance improvements have not impacted functionality. (Of course, that assumes we have good stimulus, which I do not.)

aswaterman:
> Perhaps set a flag `processor_t::privilege_changed`, force the slow path when that flag is set, and test that flag after every slow-path instruction executes?

`mret` etc. already force us onto the PC_SERIALIZE_AFTER code path, namely https://github.com/riscv-software-src/riscv-isa-sim/blob/76b0027c177113fcf083ba2c95b3c35feb642957/riscv/execute.cc#L234

Does that observation help in any way?  At minimum, it might be the place to check `processor_t::privilege_changed`, if that flag actually proves necessary.

atulkharerivos:
@aswaterman That's an intriguing finding, and it's a little odd that ~four load / store unconditional operations have such an outsize impact. Also, I would have expected most of those operations to hit the cache given that all of them are in proessor_t::state. 

Anyway, I tried a small tweak on the side to make stash_privilege a member function of processor_t. It appeared that this single change reduced the number of assembly instructions to 1226 (from 1386 bytes), so as @scottj97 suggested, perhaps we should check the performance with it before attempting further surgery? 

Note that I am gcc 11.3 so YMMV.

With processor_t::stash_privilege() - 1226 bytes:
```
disassemble processor_t::step
Dump of assembler code for function _ZN11processor_t4stepEm:
Address range 0x7cd750 to 0x7cdc1a:
```
With standalone stash_privilege() -- 1386 bytes:
```
disassemble processor_t::step
Dump of assembler code for function _ZN11processor_t4stepEm:
Address range 0x7cd6f0 to 0x7cdc5a:
   0x00000000007cd6f0 <+0>:	endbr64 
   0x00000000007cd6f4 <+4>:	push   %r1
   ```

**Comment by aswaterman** [comment]
Spike's hot code paths are pretty optimized (about two dozen host instructions to simulate an ADDI), so an extra handful of loads and stores will have a large impact, despite being cache hits.

I don't think it should be too hard to call `stash_privilege` selectively.  The following probably isn't quite right but serves as a sketch of what I was thinking: 0e83fe66fbf7b918602476bbaacfa6198a90d337

**Comment by atulkharerivos** [comment]
Thanks @aswaterman.

AFAICT, your suggestion will work just fine, and for purposes of this fix, I think it might be sufficient to call it only from `take_trap()`. Perhaps @YenHaoChen / @scottj97  have other suggestions.

**Comment by aswaterman** [comment]
OK, @scottj97 seemed concerned about MRET etc., so let's see what was on his mind.  Of course, it would be nice if it suffices to do the work in `take_trap` only.

**Comment by atulkharerivos** [comment]
@aswaterman @scottj97 What if we updated the `last_inst_priv`  in `set_privilege()` instead? It handles the xret cases and shouldn't have any performance overhead (while being simpler)? I don't know if  `last_v` needs to be updated, but can it ever change on a xret?

```
void processor_t::set_privilege(reg_t prv)
{
  mmu->flush_tlb();
+ state.last_inst_priv = state.prv;
  state.prv = legalize_privilege(prv);
}
```

**Comment by aswaterman** [comment]
@atulkharerivos Updating `last_inst_priv` in `set_privilege` makes sense to me.  Yes, xRET can change `v`, but you might be able to update `last_v` within `set_virt`, analogous to your proposal for `last_inst_priv` and `set_privilege`.

**Comment by scottj97** [comment]
> @aswaterman @scottj97 What if we updated the `last_inst_priv` in `set_privilege()` instead? It handles the xret cases and shouldn't have any performance overhead (while being simpler)? I don't know if `last_v` needs to be updated, but can it ever change on a xret?

So `last_inst_priv` only changes in `set_privilege()`? That makes no sense...it needs to be updated one instruction *after* the privilege changes, not simultaneously. Along with `last_v` (and yes xret can change `state.v`).

**Comment by atulkharerivos** [comment]
@scottj97 Basically, the suggestion was a more ergonomic approach to [0e83fe66](https://github.com/riscv-software-src/riscv-isa-sim/commit/0e83fe66fbf7b918602476bbaacfa6198a90d337l). So, would you recommend setting a flag to go down the slow(er) path in `set_privilege()` and `set_virt()`?

**Comment by scottj97** [comment]
> @scottj97 Basically, the suggestion was a more ergonomic approach to [0e83fe66](https://github.com/riscv-software-src/riscv-isa-sim/commit/0e83fe66fbf7b918602476bbaacfa6198a90d337l). So, would you recommend setting a flag to go down the slow(er) path in `set_privilege()` and `set_virt()`?

I'm not stuck on any particular approach, though that sounds reasonable.

My criticism of the previous idea was because I can't understand how that could possibly work properly. But I could be misunderstanding something.

**Comment by atulkharerivos** [comment]
@scottj97 Please see if the revised approach fits the bill. This separates commit log path from the previous privilege and virtual mode tracking that's used for trigger matching.

File: riscv/processor.cc [hunk 11]
```diff
@@ -756,6 +757,7 @@ void processor_t::set_virt(bool virt)
      * The virtualized sstatus register also relies on this TLB flush,
      * since changing V might change sstatus.MXR and sstatus.SUM.
      */
+    state.prev_v = state.v;
```
scottj97: I think I've talked myself into accepting this approach.

Because etrigger only fires on an exception, and because an exception always calls `set_privilege()` and `set_virt()`, the previous privilege used for the etrigger check will always be the privilege of the immediately preceding instruction.

However I suspect there might still be a bug here. Scenario:
1. SRET from HS to VU: `prev_v` is set to 0
2. Trap from VU to VS: because `state.v` is not changing, `prev_v` doesn't change and remains 0
3. etrigger that's set to break on a VU-mode trap won't match properly because `prev_v` is incorrect

I think this can be solved simply by removing the conditional `if (state.v != virt)` around these two lines.

scottj97 (changes requested):
Can you tell us how you've tested this? Do you have a failing test case that's now passing after this change? Does it include an etrigger breakpoint on a trap that goes from VU to VS mode?

riscv/processor.cc
atulkharerivos: Good catch; changed.

> Can you tell us how you've tested this? Do you have a failing test case that's now passing after this change? Does it include an etrigger breakpoint on a trap that goes from VU to VS mode?M

The testing didn't involve VU/VS mode (it was focused on U/S triggers for ECALLs / interrupts). Note that the original patch didn't have a prev_v, and it was added in response to feedback. However,  the scenario you described above seems perfectly viable.

scottj97 (approved):

aswaterman (approved):

PR #1353, opened 2023-05-09
Zfa: fix NX handling for the fround/froundnx family

### LINKED ISSUE #1351: Zfa: fround vs froundnx
In Ch.25.3 of unpriv, FROUNDNX is defined as the same as FROUND, except that it sets NX if the input differs from the rounded result (and is not NaN).

In the Spike code, both of these instructions call softfloat/f*_roundToInt, while the "exact" operand is true for fround but false for froundnx.  Setting NX is done under some conditions only if the "exact" operand is true.  This appears to conflict with the definition in unpriv.

Am I understanding this incorrectly?  If not, is unpriv correct, or is Spike?

ptomsich:
The initial implementation (together with the SAIL code and the tests) had gotten the NX variants backwards (as in 'an inexact result is ok' vs. 'set NX is inexact').  Update all 4 instructions.

Resolves #1351

### DIFF
--- a/riscv/insns/fround_d.h
+++ b/riscv/insns/fround_d.h
@@ -1,5 +1,5 @@
 require_extension('D');
 require_extension(EXT_ZFA);
 require_fp;
-WRITE_FRD_D(f64_roundToInt(FRS1_D, RM, true));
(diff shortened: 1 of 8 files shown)


aswaterman (approved):
Thanks!

PR #1356, opened 2023-05-09
Zfa: fix missing set_fp_exceptions for fleq/fltq

### LINKED ISSUE #1355: Zfa: fleq/fltq should set NV flag if source is SNaN
According to unpriv, the difference between fle/flt and fleq/fltq is that the "q" versions defined in Zfa do not set NV if the source is QNaN, but should set NV if the source is SNaN.  However, the Spike code fleq_*.h/fltq_*.h does not invoke "set_fp_exceptions", meaning SNaN won't set NV.

ptomsich:
Resolves #1355


aswaterman (approved):
Thanks, @ptomsich!

PR #1357, opened 2023-05-09
Zfa: fix exception behaviour for fcvtmod.w.d

### LINKED ISSUE #1312: 2 issue about fcvtmod_w_d.h instruction
hi , we found 2 issue in fcvtmod_w_d.h file 
1. in fcvtmod_w_d.h line 40, if shift <= -64 , whether do we need to rasie a inexact exception ? due to the frac bit got shifted out 
2. in  fcvtmod_w_d.h line 36 & line 46 , inexact exception and invalid exception can raise together meanwhile , but from IEEE754  description this two exception cannot raise together , same input under fcvt_w_d instruction also only raise  invalid exception

ptomsich:
Fixup the exception behavior of fcvtmod.w.d:
- if the fraction is shifted out entirely, we need to signal inexact as we will have shifted out the implicit bit
- when signaling invalid, we may not signal inexact

Resolves #1312

### DIFF
--- a/riscv/insns/fcvtmod_w_d.h
+++ b/riscv/insns/fcvtmod_w_d.h
@@ -38,12 +38,14 @@ if (exp == 0) {
   } else {
     /* The fraction is shifted out entirely.  */
     frac = 0;
+    inexact = true;
   }
 
-  /* Notice overflow or inexact exceptions.  */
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
LGTM.  Thanks.

PR #1362, opened 2023-05-18
Specify -march and -mabi in README.md example

MarekVCodasip:
Some toolchains (notably the The xPack GNU RISC-V Embedded GCC) build 32 bit binaries by default, spike runs 64 by default, causing an error. This pull request specifies needed parameters to avoid this problem in the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,13 +238,13 @@ SECTIONS
   .text : { *(.text) }
   .data : { *(.data) }
 }
-$ riscv64-unknown-elf-gcc -g -Og -o rot13-64.o -c rot13.c
-$ riscv64-unknown-elf-gcc -g -Og -T spike.lds -nostartfiles -o rot13-64 rot13-64.o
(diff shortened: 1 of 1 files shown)


aswaterman:
This one cuts both ways.  Some toolchains don't include multilibs for rv64imac/lp64, so this may break elsewhere.  I would think that using `riscv64-unknown-elf-gcc` with no args is more likely to work for more people, and really should be producing RV64 code (`riscv32-unknown-elf-gcc` would seem like the right name for a compiler that defaults to RV32 code).

MarekVCodasip:
So in that case, would adding a note about it suffice? Something along the lines of:

NOTE: Depending on your toolchain, it might build 32 bit binaries by default, add -march=rv64imac -mabi=lp64 to GCC to prevent this.

aswaterman:
Yeah, seems reasonable.

PR #1363, opened 2023-05-22
Let mstatus.MPP initially be M-mode if unsupporting U-mode

YenHaoChen:
This commit lets the mstatus.MPP be a valid value if unsupporting U-mode. Without this commit, the mret may result in a corrupted state without properly setting the MPP to M-mode (if unsupporting U-mode).

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -510,6 +510,7 @@ reg_t mstatus_csr_t::compute_mstatus_initial_value() const noexcept {
                               | (proc->extension_enabled_const('S') ? MSTATUS_SBE : 0)
                               | MSTATUS_MBE;
   return 0
+         | set_field((reg_t)0, MSTATUS_MPP, proc->extension_enabled_const('U') ? PRV_U : PRV_M)
          | (proc->extension_enabled_const('U') && (proc->get_const_xlen() != 32) ? set_field((reg_t)0, MSTATUS_UXL, xlen_to_uxl(proc->get_const_xlen())) : 0)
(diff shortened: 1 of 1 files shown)


riscv/csrs.cc
```diff
@@ -510,6 +510,7 @@ reg_t mstatus_csr_t::compute_mstatus_initial_value() const noexcept {
                               | (proc->extension_enabled_const('S') ? MSTATUS_SBE : 0)
                               | MSTATUS_MBE;
   return 0
+         | (proc->extension_enabled_const('U') ? PRV_U : PRV_M)
```
aswaterman: Don't you need to make this something like

    (proc->extension_enabled_const('U') ? 0 : set_field((reg_t)0, MSTATUS_MPP, PRV_M))

or

    set_field((reg_t)0, MSTATUS_MPP, proc->extension_enabled_const('U') ? PRV_U : PRV_M)

so the bits get shifted into the right position within `mstatus`?

YenHaoChen: @aswaterman Thank you for the review. Yes, we need to shift into the proper position. The commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1363/commits/5b6b9e20c9ba1f701ce3b8ed28a2ab6a2d42a019 updates the code correspondingly.

BTW, should we simplify the source code by letting the MPP always initialize to M-mode?
```| MSTATUS_MPP # Initialize MPP to M-mode```

aswaterman: Although I agree that would be valid, I would prefer to avoid changing the behavior for the case that U-mode is present.

aswaterman (approved):

aswaterman:
thanks!

PR #1364, opened 2023-05-24
Zacas extension

glg-rv:
This implements the Zacas extension (Atomic compare-and-swap instructions) as of Version 0.1, 4/2023.

Specs: https://github.com/riscv/riscv-zacas/tree/main


aswaterman (changes requested):
I think it would be cleaner to factor this a little differently.  Instead of defining the very generic primitive `mmu_t::amo_opt`, just define `mmu_t::compare_and_swap`.  This will reduce code duplication without removing any useful generality, because it seems unlikely we'll use `amo_opt` for anything besides CAS.

Second, rather than defining these `pair` versions, use `uint128_t` for AMOCAS.Q.  We already depend on the existence of this type for the V extension.  To avoid breaking legacy 32-bit hosts that don't have this type, conditionalize Zacas on the existence of this type, as we did with V:

https://github.com/riscv-software-src/riscv-isa-sim/blob/fa79c20a4eea06650c96986360a06403732d88a7/riscv/riscv.mk.in#L1377

and

https://github.com/riscv-software-src/riscv-isa-sim/blob/fa79c20a4eea06650c96986360a06403732d88a7/riscv/processor.cc#L46

glg-rv:
> I think it would be cleaner to factor this a little differently. Instead of defining the very generic primitive `mmu_t::amo_opt`, just define `mmu_t::compare_and_swap`. This will reduce code duplication without removing any useful generality, because it seems unlikely we'll use `amo_opt` for anything besides CAS.

Agreed. I had two ways to go towards this. One would be to change amo() in always returning an optional, and in the other AMO functions always returning a value. Or, as you suggested, having a specific cas function. I sent this RFC exactly to gather opinions on what would be the best way. Will do this, thanks!

> 
> Second, rather than defining these `pair` versions, use `uint128_t` for AMOCAS.Q. We already depend on the existence of this type for the V extension. To avoid breaking legacy 32-bit hosts that don't have this type, conditionalize Zacas on the existence of this type, as we did with V:
> 

Ack. As the spec defines this order in particular I wanted to be as specific as the pseudocode.

glg-rv:
Right. Updated with `amo_compare_and_swap`.

A bit of a concatenated events of bug fixing:

- In order to support load and store in mmu of `uint128_t`, I had to define its `swap` operation in `byteorder.h`
- Also, on 128-bit load/store, `reg_from_bytes` in `check_triggers` would fail because it would be too long to fit in a register, so I had to call triggers in chunks of eight bytes until the remaining bytes are short enough to fit in a reg. I am not sure if we have a requirement to only support individual loads/stores of 64-bit.
- Instead of using `__SIZEOF_INT128__` to check for support of `__int128`, I introduced a `config.h` variable HAVE_INT128. This way what we decide to compile on configure is consistent with the way we compile code.
-  When tested, the check in `processor.cc` for `V` enabled when 128 bits are not supported did segfault. It is too early to run `extension_enabled`, so I resorted to call `isa->extension_enabled`.

aswaterman:
Thanks, I'll review this when I have some more time and will offer feedback on the approach.  It might be that we do want to make `mmu_t::compare_and_swap<uint128_t>` a special case via template specialization, so as to avoid teaching the rest of the code base about memory accesses greater than 64 bits.  Or maybe not.  We'll see.

glg-rv:
Sure. In the meanwhile I will send the secondary patches -- bug fixes for processor.cc and HAVE_INT128 define -- on a separate PR to make this patch simpler and only focus on the zacas mechansim itself.

I will gain some insight on the inner working of load/store, and be sure that we can get away with not having a special `pair` case for amocas_d in RV32 and amocas_q in RV64.

glg-rv:
Updated branch with fixes for big endian. Now that we access the memory via load/store in 128-bit,we need to recreate the value to compare as that of `AMOCAS.Q` in RV64 and `AMOCAS.D` in RV32 are not strictly 128 and 64 bit comparison.

glg-rv:
Updated once again, as some of the side-changes have been merged in PRs #1367 and #1368.

Here's a summary to what's happening in this PR.

An interesting part of the AMOCAS instructions is the AMOCAS.D and AMOCAS.Q behaviour in respectively RV32 and RV64. Rather than being (respectively) 64-bit and 128-bit comparisons of values as loaded from memory, the spec specifies which address goes into what register, and then it's a register pair comparison.

My initial implementation had this register loading explicit.

It was then changed upon reviewing allowing 64-bit and 128-bit loads/stores and then compare the 64-bit/128-bit values. Doing this requires the instruction implementation to be endianness-dependent. This is what is currently on review.

I think the outstanding questions for this PR are:

Should we:
a. use 64-bit/128-bit load/store as implemented right now
b. having dual loads into different registers as originally implemented?

If A, should we:
1. Add support to 128-bit load/stores as implemented in this patch
2. Specialise load<T> and store<T> for `uint128_t`, avoiding changes to the load/store mechanism but requiring the specialisation to be endianness-dependent?

riscv/mmu.cc
```diff
@@ -242,6 +242,11 @@ void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, xlate_flags_t
       load_slow_path_intrapage(len - len_page0, bytes + len_page0, access_info.split_misaligned_access(len_page0));
   }
 
+  while (len > 8) {
+    check_triggers(triggers::OPERATION_LOAD, addr, access_info.effective_virt, reg_from_bytes(8, bytes));
+    len -= 8;
+    bytes += 8;
```
aswaterman: Rather than hardcoding `8` on each of these lines, please use `sizeof(reg_t)` to make the intent more obvious.

riscv/mmu.cc
```diff
@@ -275,8 +280,16 @@ void mmu_t::store_slow_path_intrapage(reg_t len, const uint8_t* bytes, mem_acces
...
   auto access_info = generate_access_info(addr, STORE, xlate_flags);
-  if (actually_store)
-    check_triggers(triggers::OPERATION_STORE, addr, access_info.effective_virt, reg_from_bytes(len, bytes));
+  if (actually_store) {
+    reg_t trig_len = len;
+    const uint8_t* trig_bytes = bytes;
+    while (trig_len > 8) {
+      check_triggers(triggers::OPERATION_STORE, addr, access_info.effective_virt, reg_from_bytes(8, trig_bytes));
+      trig_len -= 8;
+      trig_bytes += 8;
```
aswaterman: Same here.

aswaterman:
@jerryz123 Can I get a second opinion on the approach here?  IMO, it doesn't infect too much of the code with int128 gunk, so it seems reasonable to me.

@scottj97 ditto, if you can spare the time.

@glg-rv assuming we have agreement, can you rebase on HEAD of master and resolve conflicts?  You'll need to regenerate `encoding.h`.  Note you're racing with PR #1303, so whoever gets in first might cause another round of conflicts :)

jerryz123:
That looks fine to me, I don't think it causes any problems with the shared library use case.

glg-rv:
> @glg-rv assuming we have agreement, can you rebase on HEAD of master and resolve conflicts? You'll need to regenerate `encoding.h`. Note you're racing with PR #1303, so whoever gets in first might cause another round of conflicts :)

Sure, am more than okay with rebases. I'll re-update with current master of https://github.com/riscv/riscv-opcodes

glg-rv:
Ack, updated with change requested and rebased. :)

aswaterman (approved):
LGTM now; thanks for the contribution!

PR #1366, opened 2023-05-24
Implement dcsr.v and make DRET use it

### LINKED ISSUE #1365: dret doesn't handle VS/VU modes
The `dcsr.v` bit is described by the debug spec:

> Extends the prv field with the virtualization mode the hart was operating in when Debug Mode was entered. The encoding is described in Table 4.6. A debugger can change this value to change the hart’s virtualization mode when exiting Debug Mode.

The [`dret` implementation](https://github.com/riscv-software-src/riscv-isa-sim/blob/23bc4cd63a1a644566a4fc4e51d75377353df968/riscv/insns/dret.h) in Spike does not look at this bit, or call `set_virt()`, so won't properly return to VS  or VU mode.

I have not looked, but I suspect entry into Debug Mode has a similar problem (failing to set `dcsr.v` appropriately).

cc @timsifive

aswaterman:
Fixes #1365.

I also refactored the `set_privilege`/`set_virt` interface to prevent this kind of bug from cropping up again.  I'd appreciate a careful review, as I only have limited tests for this kind of thing.

cc @timsifive

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -13,6 +13,8 @@
 #include "trap.h"
 // For require():
 #include "insn_macros.h"
+// For CSR_DCSR_V:
+#include "debug_defines.h"
 
 // STATE macro used by require_privilege() macro:
 #undef STATE
@@ -1234,6 +1236,7 @@ dcsr_csr_t::dcsr_csr_t(processor_t* const proc, const reg_t addr):
   ebreaks(false),
   ebreaku(false),
   halt(false),
+  v(false),
   cause(0) {
 }
 
@@ -1255,6 +1258,7 @@ reg_t dcsr_csr_t::read() const noexcept {
   result = set_field(result, DCSR_CAUSE, cause);
   result = set_field(result, DCSR_STEP, step);
   result = set_field(result, DCSR_PRV, prv);
+  result = set_field(result, CSR_DCSR_V, v);
   return result;
 }
 
@@ -1267,12 +1271,14 @@ bool dcsr_csr_t::unlogged_write(const reg_t val) noexcept {
   ebreaks = get_field(val, DCSR_EBREAKS);
   ebreaku = get_field(val, DCSR_EBREAKU);
   halt = get_field(val, DCSR_HALT);
+  v = proc->extension_enabled('H') ? get_field(val, CSR_DCSR_V) : false;
   return true;
 }
 
-void dcsr_csr_t::write_cause_and_prv(uint8_t cause, reg_t prv) noexcept {
+void dcsr_csr_t::write_cause_and_prv(uint8_t cause, reg_t prv, bool v) noexcept {
   this->cause = cause;
   this->prv = prv;
+  this->v = v;
   log_write();
 }
 
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1244,18 +1244,18 @@ void dcsr_csr_t::verify_permissions(insn_t insn, bool write) const {
 }
 
 reg_t dcsr_csr_t::read() const noexcept {
-  uint32_t v = 0;
-  v = set_field(v, DCSR_XDEBUGVER, 1);
-  v = set_field(v, DCSR_EBREAKM, ebreakm);
-  v = set_field(v, DCSR_EBREAKH, ebreakh);
-  v = set_field(v, DCSR_EBREAKS, ebreaks);
-  v = set_field(v, DCSR_EBREAKU, ebreaku);
-  v = set_field(v, DCSR_STOPCYCLE, 0);
-  v = set_field(v, DCSR_STOPTIME, 0);
-  v = set_field(v, DCSR_CAUSE, cause);
-  v = set_field(v, DCSR_STEP, step);
-  v = set_field(v, DCSR_PRV, prv);
-  return v;
+  reg_t result = 0;
+  result = set_field(result, DCSR_XDEBUGVER, 1);
+  result = set_field(result, DCSR_EBREAKM, ebreakm);
+  result = set_field(result, DCSR_EBREAKH, ebreakh);
+  result = set_field(result, DCSR_EBREAKS, ebreaks);
+  result = set_field(result, DCSR_EBREAKU, ebreaku);
+  result = set_field(result, DCSR_STOPCYCLE, 0);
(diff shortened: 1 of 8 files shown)


riscv/processor.cc
```diff
@@ -831,10 +814,9 @@ void processor_t::take_trap(trap_t& t, reg_t epc)
     s = set_field(s, MSTATUS_SPP, state.prv);
     s = set_field(s, MSTATUS_SIE, 0);
     state.sstatus->write(s);
-    set_privilege(PRV_S);
+    set_privilege(PRV_S, true);
```
timsifive, line 817: This looks correct, and also like a behavior change. Does it fix a bug, or am I misreading it?

timsifive (commented):
Debug changes look good.

scottj97 (changes requested):
3732e8db1e772c180ece89a432660cd6664624e2 definitely broke something...I'm investigating.

riscv/processor.cc
scottj97: I don't think it's a behavior change. `vsdeleg` would be 0 unless `state.v` is already true, so we won't get here unless it's already true.

aswaterman:
Darn.  Thanks for testing.  In parallel, I'll stare at the code a bit more.

scottj97:
> Darn. Thanks for testing. In parallel, I'll stare at the code a bit more.

I see one trap going to the VS-mode handler when it should have gone to the HS-mode handler. Still investigating.

Edit: might have been in the wrong mode before the trap occurred. Or might be something about the trap-taking mechanism.

aswaterman:
Unfortunately, staring at the code isn't revealing anything to me... hopefully you don't end up wasting too much time tracking down the divergence.

scottj97:
> Unfortunately, staring at the code isn't revealing anything to me... hopefully you don't end up wasting too much time tracking down the divergence.

I pushed a [branch](https://github.com/scottj97/riscv-isa-sim/tree/intermediate-1366) breaking this down into smaller commits. The divergence happens here: https://github.com/scottj97/riscv-isa-sim/commit/4b9cbc1e547f0d944e99550863554435e690c939

I'm still investigating, but maybe that will give you a clue.

scottj97:
I get it. `state->stvec()` used to read architectural `stvec` but is now reading architectural `vstvec` because V=1 still.

We need to use `state->nonvirtual_stvec` etc for all the virtualizable CSRs here (`stvec`, `scause`, `sepc`, `stval`).

riscv/processor.cc
```diff
@@ -741,30 +743,11 @@ const char* processor_t::get_privilege_string()
...
-  state.prev_v = state.v;
-  state.v = virt;
-}
-
 void processor_t::enter_debug_mode(uint8_t cause)
 {
   state.debug_mode = true;
   state.dcsr->write_cause_and_prv(cause, state.prv, state.v);
-  set_privilege(PRV_M);
+  set_privilege(PRV_M, false);
```
scottj97, line 750: This looks like a functional change, since it wasn't forcing V=0 previously?

aswaterman: I guess so, but it was definitely a bug before, because prv=M and V=1 is paradoxical.  I'll separate it into another commit.

aswaterman:
> We need to use state->nonvirtual_stvec etc for all the virtualizable CSRs here (stvec, scause, sepc, stval).

Ah-ha!  By "here" you mean the HS case of `take_trap`, right?

scottj97:
> > We need to use state->nonvirtual_stvec etc for all the virtualizable CSRs here (stvec, scause, sepc, stval).
> 
> Ah-ha! By "here" you mean the HS case of `take_trap`, right?

Yes. I'm already underway on a rework of this since it's right up my alley with all these CSRs.

aswaterman:
OK, thanks.  Note I already pushed over my branch to separate out the commit mentioned in https://github.com/riscv-software-src/riscv-isa-sim/pull/1366#discussion_r1204814853

aswaterman:
@timsifive can you look into why the debug tests are failing?  Presumably it's just because of the addition of `dcsr.v`, but maybe it's something else.

riscv/processor.cc
```diff
@@ -765,6 +765,7 @@ void processor_t::enter_debug_mode(uint8_t cause)
   state.debug_mode = true;
   state.dcsr->write_cause_and_prv(cause, state.prv, state.v);
   set_privilege(PRV_M);
+  set_virt(false);
```
scottj97: Sadly, this commit will be a NOP, because privilege is already PRV_M so this change will [get blocked](https://github.com/riscv-software-src/riscv-isa-sim/blob/ebf9a94af86f7c56e24d12f104ef9e983e836f99/riscv/processor.cc#L748-L749).

I will fix this in my rework.

aswaterman: Ah, right... okay, thanks again.

scottj97:
@aswaterman see [my branch](https://github.com/scottj97/riscv-isa-sim/tree/intermediate-1366) for a reworked proposal.

I am testing it more thoroughly now.

aswaterman:
@scottj97 Yeah, that version looks good to me.

scottj97:
> @scottj97 Yeah, that version looks good to me.

I force-pushed some silly bug fixes to it.

Do you want to take that and replace the branch you have here? Or maybe I somehow have permission to do so?

aswaterman:
You should be able to force-push over this branch, but if not, just open a new PR.

scottj97 (approved):
I do not have any test coverage of debug mode, but everything else looks good to me now.

aswaterman:
Thanks for the help!

riscv/csrs.cc
```diff
@@ -1255,6 +1258,7 @@ reg_t dcsr_csr_t::read() const noexcept {
   v = set_field(v, DCSR_CAUSE, cause);
   v = set_field(v, DCSR_STEP, step);
   v = set_field(v, DCSR_PRV, prv);
+  v = set_field(v, CSR_DCSR_V, v);
```
timsifive: v here is an internal variable, and does not refer to the class member v as intended.
```suggestion
  v = set_field(v, CSR_DCSR_V, this->v);
```

(Great example of how even people don't even notice obvious bugs when reviewing code.)

aswaterman: Yikes.

PR #1367, opened 2023-05-26
decode_macros: move 'is_aligned' from 'v_ext_macros.h'

glg-rv:
Simple nit: `is_aligned()` is only used in `decode_macros.h` while being defined in `v_ext_macros.h`.
Move it to `decode_macros.h`, that looks more like its place.


aswaterman (approved):
Indeed.

PR #1368, opened 2023-05-26
Minor fixes (mostly related to __int128 support)

glg-rv:
These are minor fixes related somehow to spike support of uint128_t.

Patch 1&2 use a `configure` defined `HAVE_INT128`  rather than using `__SIZEOF_INT128__`. If anything, the advantage to this is to be able to test `!HAVE_INT128`. Possibly not a fix, this is something that at least IMHO makes things a little bit better.

Patch 3 fixes a crash in spike happening when we do not have `HAVE_INT128` support.

### DIFF
--- a/riscv/decode.h
+++ b/riscv/decode.h
@@ -19,11 +19,6 @@ typedef int64_t sreg_t;
 typedef uint64_t reg_t;
 typedef float128_t freg_t;
 
-#ifdef __SIZEOF_INT128__
-typedef __int128 int128_t;
-typedef unsigned __int128 uint128_t;
-#endif
-
 const int NXPR = 32;
 const int NFPR = 32;
 const int NVPR = 32;
--- a/config.h.in
+++ b/config.h.in
@@ -39,6 +39,9 @@
 /* Dynamic library loading is supported */
 #undef HAVE_DLOPEN
 
+/* __int128_t is supported */
+#undef HAVE_INT128
+
 /* Define to 1 if you have the <inttypes.h> header file. */
 #undef HAVE_INTTYPES_H
 
--- a/configure
+++ b/configure
@@ -4742,6 +4742,9 @@ ac_fn_cxx_check_type "$LINENO" "__int128_t" "ac_cv_type___int128_t" "$ac_include
 if test "x$ac_cv_type___int128_t" = xyes; then :
   HAVE_INT128=yes
 
+
+$as_echo "#define HAVE_INT128 1" >>confdefs.h
+
 fi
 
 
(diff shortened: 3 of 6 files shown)


riscv/decode.h
```diff
@@ -7,6 +7,7 @@
 # error spike requires a two''s-complement c++ implementation
 #endif
 
+#include "config.h"
```
jerryz123: Adding `#include "config.h"` here breaks libraries which want to include spike's installed headers.
`decode.h` is an installed header, while `decode_macros.h` is not.

Is it possible to move the int128 typedefs to `decode_macros.h`, or `v_ext_macros.h`?

jerryz123 (changes requested):

riscv/decode.h
glg-rv: Oh sorry, completely missed this. Let me have a look at this.

glg-rv:
Updated. Moved to `decode_macros.h`, as @jerryz123 suggested.

It is good for now, because uint128_t is only used by instruction implementations. This might become an odd place to have it (see #1364) if we ever used outside of that.

In general, possibly having our own `types.h` that is project wide would be nice. Although I am still not completely aware of the overall structure of the project.

riscv/decode.h
glg-rv: Updated, thanks!

aswaterman (changes requested):
The title of this PR suggests that multiple things are being fixed, but I don't see that anything is being fixed.  Could you rewrite the title and description to clearly indicate both the purpose and effect?

glg-rv:
> The title of this PR suggests that multiple things are being fixed, but I don't see that anything is being fixed. Could you rewrite the title and description to clearly indicate both the purpose and effect?

? Hm.

I think you possibly missed the commit messages.

There's a spike crash that is fixed. See description. I consider that a fix.
Also, defining HAVE_INT128 in configure.ac makes sure that the support of __int128 is consistent both in code and compilation. This is a very, very minor fix. In commit message.
And I think also that now we make the code _slightly_ better by removing the definition of uint128_t out of an installed header.

In my mind, these are minor fixes. Possibly _related_ to uin128_t support, rather than _to_.

glg-rv:
Updated the description. I can see why it could have been confusing. I usually leave commit message to explanations.

aswaterman:
Thanks!  Naturally, we also obey the convention that each commit's message should explain its rationale, and I appreciate that you did so.

Your updated PR message is quite helpful!  Going forward, the most important thing is the summary of what was done.  Explaining each commit is not usually necessary, except when the relationship between commits would be non-obvious to a casual reviewer.  But the PR description absolutely must contain a good summary.

aswaterman (approved):

aswaterman:
@jerryz123 please merge if you feel your concerns have been addressed

jerryz123 (approved):

PR #1373, opened 2023-06-02
dscr.ebreakh is now dcsr.ebreakv[su]

timsifive:
This change was made ages ago in the spec.

I did not actually test that the new privilege checks in ebreak and c.ebreak are correct, but all the existing debug tests still pass.


aswaterman (approved):
LGTM.

PR #1374, opened 2023-06-06
Use device factories to control dts-construction/dts-parsing/device-instantiation

jerryz123:
This unifies the dts-construction/dts-parsing/instantiation of devices behind `device_factory_t` singletons. 
The primary motivation is to allow plugin devices to both 
- add custom device nodes to the DTS 
- parse their existence from a FDT.

Now the flow to instantiate the set of devices is:
1. Register a list of plugin `device_factory_t`, specified by the `--extlib` flag
2. Specify a set of registered device factories to provide to the simulator, with the `--device=` flag
3. If no custom DTS is specified, combine the the specified device factories with the default clint/plic/ns16550 factories, and generate a set of DTS nodes to append to the generated DTS
4. For each factory in the combined list of device factories, attempt to parse the device from the FDT


riscv/sim.h
```diff
@@ -70,6 +70,7 @@ class sim_t : public htif_t, public simif_t
   std::string dts;
   std::string dtb;
   bool dtb_enabled;
+  std::vector<std::shared_ptr<abstract_device_t>> devices;
```
aswaterman, line 77: Since the bus already needs to maintain a list of all devices, can we avoid some redundancy by just calling `bus.tick()` and having that tick everything?

aswaterman:
Looks like a significant improvement to me.

riscv/sim.h
jerryz123: I think we can make the bus tick everything, but the list of devices should still be maintained by `sim_t`, to avoid dealing with shared ownership, and to avoid introducing `shared_ptrs` to more places.

aswaterman: The ticking was less my concern; it was more that we are appending every device to both the bus and the device list. If there’s some other way to be DRY about that, I’d be OK with it.

jerryz123: If you are ok with changing device pointers to  `shared_ptr` everywhere, then we can DRY it. 

The problem is that some devices can be solely owned by the bus (rom, plugins), while others need to be accessed through the sim interface as well.

scottj97 (dismissed):
fc2f48a370972f89ffc3250e3075989f35db17ae seems to have unnecessary debug messages printing:
```
Attempting to register clint
Attempting to register plic
Attempting to register ns16550
```

(I haven't looked any deeper)

riscv/sim.h
jerryz123: Nvm I see what you are saying

jerryz123: Actually I think the current set up is best.
`sim_t.devices` maintains ownership of the devices under sim_t, while `sim_t.bus.devices` tracks which devices are managers under the bus.

The code is already pretty DRY I think, only the boot_rom has some special handling since it is instantiated at `sim_t::reset`, not the constructor.

jerryz123:
> [fc2f48a](https://github.com/riscv-software-src/riscv-isa-sim/commit/fc2f48a370972f89ffc3250e3075989f35db17ae) seems to have unnecessary debug messages printing:
> 
> ```
> Attempting to register clint
> Attempting to register plic
> Attempting to register ns16550
> ```
> 
> (I haven't looked any deeper)

Oops, fixed.

riscv/sim.h
aswaterman: Yeah, I'm OK with only the boot ROM being a special case in this respect, since, as you point out, it is a special case.

aswaterman (approved):

PR #1375, opened 2023-06-08
Replace ternary operator with std:min

demin-han:


aswaterman (approved):
I agree, this is clearer.

PR #1376, opened 2023-06-09
Fix: PMP checking region on CBO instructions

YenHaoChen:
The starting address of the cbo instruction should be implicitly aligned to the cache block. The previous implementation checks the PMP region starting from the misaligned address and crossing the cache block boundary.

This PR corrects the issue by giving the starting address to the translate() function. Additionally, to preserve the desired tval, the clean_inval() function catches the exception from the translate() function and converts the tval.

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -214,7 +214,7 @@ public:
 
   void clean_inval(reg_t addr, bool clean, bool inval) {
     convert_load_traps_to_store_traps({
-        const reg_t paddr = translate(generate_access_info(addr, LOAD, {false, false, false}), blocksz) & ~(blocksz - 1);
+      const reg_t paddr = translate(generate_access_info(addr, LOAD, {false, false, false}), 1);
       if (sim->reservable(paddr)) {
         if (tracer.interested_in_range(paddr, paddr + PGSIZE, LOAD))
           tracer.clean_invalidate(paddr, blocksz, clean, inval);


riscv/mmu.h
```diff
@@ -206,14 +206,33 @@ class mmu_t
...
+    } catch (trap_store_page_fault& t) { \
+      throw trap_store_page_fault(t.has_gva(), addr, 0, t.get_tinst()); \
+    } catch (trap_load_access_fault& t) { \
+      throw trap_store_access_fault(t.has_gva(), addr, 0, t.get_tinst()); \
+    } catch (trap_store_access_fault& t) { \
+      throw trap_store_access_fault(t.has_gva(), addr, 0, t.get_tinst()); \
+    } catch (trap_load_guest_page_fault& t) { \
+      throw trap_store_guest_page_fault(addr, (addr & ((reg_t(2) << (proc->xlen-1))-1)) >> 2, t.get_tinst()); \
+    } catch (trap_store_guest_page_fault& t) { \
+      throw trap_store_guest_page_fault(addr, (addr & ((reg_t(2) << (proc->xlen-1))-1)) >> 2, t.get_tinst()); \
```
YenHaoChen: I am unfamiliar with the tval2 of the hypervisor. The equation is by source code tracing.

riscv/mmu.h
```diff
@@ -206,15 +206,34 @@ class mmu_t
     return (float128_t){load<uint64_t>(addr), load<uint64_t>(addr + 8)};
   }
 
+  // AMO/Zicbom faults should be reported as store faults
+  // Zicbom instruction has different tval and starting address
+  #define convert_cbo_traps(BODY) \
```
aswaterman: Having a second copy of these is quite unfortunate.  Can't we do this in a DRY way and reuse the existing `convert_load_traps_to_store_traps`, e.g. by aligning `addr` to cache-line boundary as soon as you enter the `clean_inval` function?

aswaterman: In other words, what's wrong with reporting `addr & -blocksz` in `tval`?  I realize it's a behavioral change, but is it actually wrong?

YenHaoChen: We need the original address for the tval when raising exceptions.

aswaterman: > We need the original address for the tval when raising exceptions.

Yeah, I'm questioning that. Is that really required?

aswaterman: Sounds like this won't fly.  But I have a better idea.

The spec says "The PMP access control bits shall be the same for all physical addresses in the cache block [... else] the behavior of a CBO instruction is UNSPECIFIED."

So, we don't need to check the entire cache block.  We only need to check the byte rs1 points to.  So, the following change suffices.  Can you update the PR?

```
diff --git a/riscv/mmu.h b/riscv/mmu.h
index 5a4835c3..efc6e9de 100644
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -214,7 +214,7 @@ public:
 
   void clean_inval(reg_t addr, bool clean, bool inval) {
     convert_load_traps_to_store_traps({
-        const reg_t paddr = translate(generate_access_info(addr, LOAD, {false, false, false}), blocksz) & ~(blocksz - 1);
+      const reg_t paddr = translate(generate_access_info(addr, LOAD, {false, false, false}), 1);
       if (sim->reservable(paddr)) {
         if (tracer.interested_in_range(paddr, paddr + PGSIZE, LOAD))
           tracer.clean_invalidate(paddr, blocksz, clean, inval);
```

YenHaoChen: Thank you for the suggestion. That cleans up the code elegantly.

riscv/mmu.h
YenHaoChen: Dropping the above modification.

aswaterman (approved):

PR #1377, opened 2023-06-12
Have CI run on each commit in a PR, instead of just the HEAD

jerryz123:


aswaterman (approved):
I didn't review the code, but I approve of the idea... feel free to merge once you've got it working.

jerryz123:
Obviously the penalty on CI time is high, but I hope this can reduce some manual labor. We'll see how this goes.

PR #1381, opened 2023-06-14
Add Smcntrpmf support

atulkharerivos:
This adds support for the "Cycle & Instret Privilege Mode Filtering Architecture" extension ( https://github.com/riscv/riscv-smcntrpmf/releases/download/v1.0_rc2/riscv-smcntrpmf-v1.0_rc2.pdf).


riscv/isa_parser.cc
```diff
@@ -240,6 +240,9 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
       extension_table[EXT_ZVFBFWMA] = true;
     } else if (ext_str == "sstc") {
         extension_table[EXT_SSTC] = true;
+    }
+    else if (ext_str == "smcntrpmf") {
```
liweiwei90: It's  better to add "else if" to last line.

riscv/csrs.cc
```diff
@@ -996,7 +996,19 @@ reg_t wide_counter_csr_t::read() const noexcept {
 }
 
 void wide_counter_csr_t::bump(const reg_t howmuch) noexcept {
-  val += howmuch;  // to keep log reasonable size, don't log every bump
+  auto config_val = 0;
+
+  if (proc->extension_enabled_const(EXT_SMCNTRPMF)) {
+    if (address == CSR_MCYCLE) {
+      config_val = state->csrmap[CSR_MCYCLECFG].get()->read();
```
liweiwei90: I think you can just add mcyclecfg and following minstretcfg to the state structure, and use them directly.

liweiwei90 (changes requested):

riscv/csrs.cc
```diff
@@ -1014,6 +1026,22 @@ reg_t wide_counter_csr_t::written_value() const noexcept {
...
+bool wide_counter_csr_t::is_counting_inhibited(const reg_t config_val) const noexcept {
+  bool count_inhibited = false;
+  auto prv = state->prv_changed ? state->prev_prv : state->prv;
+  if (prv == PRV_M) {
+    count_inhibited = (config_val & MHPMEVENT_MINH) != 0;
+  } else if (prv == PRV_S) {
+    count_inhibited = state->v ? (config_val & MHPMEVENT_VSINH) != 0 : (config_val & MHPMEVENT_SINH) != 0;
+  } else {
+    count_inhibited = state->v ? (config_val & MHPMEVENT_VUINH) != 0 : (config_val & MHPMEVENT_UINH) != 0;
+  }
```
scottj97: Too much repetition here when the only difference is one constant.

Instead, use one expression to calculate the mask based on prv/v, then use that in the `return` line. No need for a separate `count_inhibited` variable because the method name already explains the return value.

riscv/csrs.cc
```diff
@@ -1014,6 +1026,22 @@ reg_t wide_counter_csr_t::written_value() const noexcept {
   return this->val + 1;
 }
 
+
+// Returns whether counting is inhibited by Smcntrpmf.
+// Note that minstretcfg / mcyclecfg / mhpmevent* share the same inhibit bits.
+bool wide_counter_csr_t::is_counting_inhibited(const reg_t config_val) const noexcept {
+  bool count_inhibited = false;
+  auto prv = state->prv_changed ? state->prev_prv : state->prv;
```
scottj97, line 1041: It seems what you're looking for is the privilege of the immediately previous instruction that committed?

Shouldn't you also use `prev_v` in that case too?

riscv/csrs.cc
```diff
@@ -996,7 +996,19 @@ reg_t wide_counter_csr_t::read() const noexcept {
 }
 
 void wide_counter_csr_t::bump(const reg_t howmuch) noexcept {
-  val += howmuch;  // to keep log reasonable size, don't log every bump
+  auto config_val = 0;
+
+  if (proc->extension_enabled_const(EXT_SMCNTRPMF)) {
+    if (address == CSR_MCYCLE) {
```
scottj97: Instead of changing behavior based on the CSR's address, it would be better to add a constructor argument to define the behavior difference.

In this case it's easy, because the difference is only which CSR controls the config.

If you move the creation of these new config regs to just above the creation of `mcycle` and `minstret`, then you can pass them in to the `wide_counter_csr_t` constructor.

To handle the case where Smcntrpmf is not enabled, you can always instantiate the config regs, and always pass them into the two counter CSRs, but do not add them to `csrmap`. Then there's no further need to check for this extension being enabled. (I'm assuming the reset value of these config CSRs results in "normal" behavior of `mcycle`, as if this feature was not configured.)

However that might cause a performance problem for the no-Smcntrpmf case, since these counters get bumped every instruction. To solve that, you can check for `proc->extension_enabled(EXT_SMCNTRPMF)` here in `bump()`, but please add that performance boost as a separate commit so the commit history can explain that this is strictly for performance and not necessary for functionality.

riscv/csrs.cc
```diff
@@ -996,7 +996,19 @@ reg_t wide_counter_csr_t::read() const noexcept {
 }
 
 void wide_counter_csr_t::bump(const reg_t howmuch) noexcept {
-  val += howmuch;  // to keep log reasonable size, don't log every bump
+  auto config_val = 0;
+
+  if (proc->extension_enabled_const(EXT_SMCNTRPMF)) {
```
scottj97: The difference in `extension_enabled()` and `extension_enabled_const()` is that the latter is for code that assumes this feature won't be dynamically enabled or disabled by software. Because this code executes every counter bump, I don't see any such assumption here? So I think this could be just plain `extension_enabled()`.

riscv/csrs.cc
scottj97: Yes, that is preferable to using `csrmap` which requires a map lookup every time. Though in this case, adding these to the state structure is not necessary if the config register is passed in to the constructor, as I requested in another comment.

riscv/csrs.cc
```diff
@@ -996,7 +996,19 @@ reg_t wide_counter_csr_t::read() const noexcept {
...
+    if (address == CSR_MCYCLE) {
+      config_val = state->csrmap[CSR_MCYCLECFG].get()->read();
+    } else if (address == CSR_MINSTRET) {
+      config_val = state->csrmap[CSR_MINSTRETCFG].get()->read();
+    }
+  }
+
+  if (!is_counting_inhibited(config_val)) {
```
scottj97: Should the method be `is_counting_enabled()` instead, to avoid the double negative here?

scottj97 (changes requested):
I don't know anything about the spec for this feature, but I have some comments on the implementation.

riscv/isa_parser.cc
atulkharerivos: Done.

riscv/csrs.cc
atulkharerivos: Changed.

riscv/csrs.cc
atulkharerivos: Yes indeed, changed.

riscv/csrs.cc
atulkharerivos: Updated.

riscv/csrs.cc
atulkharerivos: Done.

riscv/csrs.cc
atulkharerivos: Please see updates.

atulkharerivos:
@scottj97 Sorry for taking so long to turnaround, but for some reason, I wasn't receiving any notifications on the comments. The specification can be found here: https://github.com/riscv/riscv-smcntrpmf/releases/download/v1.0_rc2/riscv-smcntrpmf-v1.0_rc2.pdf

scottj97 (approved):
I have reviewed only the implementation, and not its adherence to the spec, but this looks good to me.

riscv/processor.cc
```diff
@@ -724,8 +745,10 @@ void processor_t::set_privilege(reg_t prv, bool virt)
   mmu->flush_tlb();
   state.prev_prv = state.prv;
   state.prev_v = state.v;
+  state.prv_changed = true;
   state.prv = legalize_privilege(prv);
   state.v = virt && state.prv != PRV_M;
+  state.v_changed = state.v != state.prev_v;
```
aswaterman, line 791: This probably isn't a bug, but I thought it was worth asking about.  `prv_changed` is unconditionally set to `true` here, even though it might not have actually changed.  (Consider: MRET returning from M-mode to M-mode.)  However, we are being precise about `v_changed`.

I feel like we should be consistent here.  Is there any harm in setting `v_changed` more sloppily?  Alternatively, is there any harm in setting `prv_changed` more precisely?

atulkharerivos: @aswaterman --good suggestion -- will post an update. BTW, this PR has some `encoding.h` artifacts from the approved, but not yet merged https://github.com/riscv-software-src/riscv-isa-sim/pull/1383, so if you planning on committing that soon, I can post a revised version of that as well.

aswaterman: Merged #1383.

atulkharerivos: Thanks @aswaterman. Please see if the update fits the bill.

aswaterman (approved):

riscv/processor.cc
aswaterman: LGTM.

atulkharerivos: Great -- I will push an update  for the `encoding.h ` commit when https://github.com/riscv/riscv-opcodes/pull/181 gets merged.

aswaterman: Merged that, too.  Sorry for the long dependence chain... :)

atulkharerivos: Done -- think all of us know the joys of working with sub-modules :).

PR #1383, opened 2023-06-15
Add Smcsrind / Sscsrind support

atulkharerivos:
Adds preliminary support for Smcsrind / Sscsrind extension. Future extensions like sscdeleg are expected to layer the proxy CSR mapping for the ireg? CSRs.


riscv/encoding.h
```diff
@@ -172,6 +172,8 @@
 #define MSTATEEN0_FCSR     0x00000002
 #define MSTATEEN0_JVT      0x00000004
 #define MSTATEEN0_HCONTEXT 0x0200000000000000
+#define MSTATEEN0_SCSRIND  0x1000000000000000
+#define HSTATEEN0_SCSRIND  0x1000000000000000
```
scottj97: We should not be hand-editing `encoding.h`. It needs to be generated from the riscv-opcodes repo, and any change to it here should update the comment at the top that explains which git hash it was generated from.

riscv/processor.h
```diff
@@ -168,6 +168,28 @@ struct state_t
   csr_t_p stimecmp;
   csr_t_p vstimecmp;
 
+  // Smcsrind / Sscsrind
+  sscsrind_sel_csr_t::sscsrind_sel_csr_t_p miselect;
```
scottj97: Do these all need to be here? They don't seem to be used today except in `state_t::reset()` so they could be local to that method.

riscv/processor.cc
```diff
@@ -519,6 +519,58 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
+  if (proc->extension_enabled_const(EXT_SMCSRIND)) {
+    miselect = std::make_shared<sscsrind_sel_csr_t>(proc, CSR_MISELECT, 0);
+    mireg = std::make_shared<sscsrind_reg_csr_t>(proc, CSR_MIREG, 0, miselect);
+    mireg2 = std::make_shared<sscsrind_reg_csr_t>(proc, CSR_MIREG2, 0, miselect);
+    mireg3 = std::make_shared<sscsrind_reg_csr_t>(proc, CSR_MIREG3, 0, miselect);
+    mireg4 = std::make_shared<sscsrind_reg_csr_t>(proc, CSR_MIREG4, 0, miselect);
+    mireg5 = std::make_shared<sscsrind_reg_csr_t>(proc, CSR_MIREG5, 0, miselect);
+    mireg6 = std::make_shared<sscsrind_reg_csr_t>(proc, CSR_MIREG6, 0, miselect);
+
+    csrmap[CSR_MISELECT] = miselect;
```
scottj97: Why not keep the existing pattern of assigning to both `csrmap` and `miselect` on a single line, instead of having to list each register twice here? With two paragraphs, there's a much bigger risk of someone changing one but not the other.

scottj97 (changes requested):

riscv/csrs.h
```diff
@@ -787,4 +791,37 @@ class jvt_csr_t: public basic_csr_t {
...
+ public:
+  typedef std::shared_ptr<sscsrind_sel_csr_t> sscsrind_sel_csr_t_p;
+  sscsrind_sel_csr_t(processor_t* const proc, const reg_t addr, const reg_t init,
+                     sscsrind_sel_csr_t_p virt_csr = nullptr);
+  reg_t read() const noexcept override;
+  virtual void verify_permissions(insn_t insn, bool write) const override;
+ protected:
+  virtual bool unlogged_write(const reg_t val) noexcept override;
+ private:
+  sscsrind_sel_csr_t_p virt_csr;
```
scottj97: These classes appear to be reimplementing `virtualized_csr_t`. Is there any reason these classes can't be simple nonvirtualized CSRs, then instantiated using `virtualized_csr_t`? See `stval` for a typical example:

```
  nonvirtual_stval = std::make_shared<basic_csr_t>(proc, CSR_STVAL, 0);
  csrmap[CSR_VSTVAL] = vstval = std::make_shared<basic_csr_t>(proc, CSR_VSTVAL, 0);
  csrmap[CSR_STVAL] = stval = std::make_shared<virtualized_csr_t>(proc, nonvirtual_stval, vstval);
```

scottj97 (changes requested):

riscv/encoding.h
atulkharerivos: Agree -- this was expedient and intended to ensure that it compiles :).

riscv/processor.cc
atulkharerivos: This has undergone significant revision -- will post an update.

riscv/csrs.h
atulkharerivos: One of the limitations of `virtualized_csr_t` is that doesn't override `verify_permissions()`. That said, the code has undergone significant changes; will post an update.

riscv/processor.h
atulkharerivos: Right, will address it in the update.

riscv/encoding.h
atulkharerivos: Dropped -- see https://github.com/riscv/riscv-opcodes/pull/178 for the relevant changes.

riscv/processor.cc
atulkharerivos: Changed.

riscv/csrs.h
atulkharerivos: Another alternative is to add `verify_permissions()` to `virtualized_csr_t`, but didn't want to risk introducing unintended side-effects.

scottj97:
Compile fails on 87a9472296963ab30e82aac103edc309facc06e8

riscv/processor.cc
```diff
@@ -524,6 +524,35 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
+  if (proc->extension_enabled_const(EXT_SSCSRIND)) {
+    vsiselect = std::make_shared<basic_csr_t>(proc, CSR_VSISELECT, 0);
+    csrmap[CSR_VSISELECT] = vsiselect;
+    siselect = std::make_shared<basic_csr_t>(proc, CSR_SISELECT, 0);
+    csrmap[CSR_SISELECT] = std::make_shared<virtualized_csr_t>(proc, siselect, vsiselect);
+  }
+
+  if (proc->extension_enabled_const(EXT_SMCSRIND)) {
+    const reg_t mireg_csrs[] = { CSR_MIREG, CSR_MIREG2, CSR_MIREG3, CSR_MIREG4, CSR_MIREG5, CSR_MIREG6 };
+    for (auto i = 0; i < 6; i++) {
```
scottj97: This is insidious duplication. The length of the array should not be hardcoded on a different line of code, even the very next line. Use `std::vector` and a modern C++ iterator instead.

It's unfortunate that the CSR addresses are not consecutive, so we wouldn't need any array here.

riscv/csrs.cc
```diff
@@ -1574,3 +1574,66 @@ void jvt_csr_t::verify_permissions(insn_t insn, bool write) const {
     }
   }
 }
+
+virtualized_indirect_csr_t::virtualized_indirect_csr_t(processor_t* const proc, csr_t_p orig, csr_t_p virt):
+  virtualized_csr_t(proc, orig, virt) {
+}
+
+void virtualized_indirect_csr_t::verify_permissions(insn_t insn, bool write) const {
+  csr_t::verify_permissions(insn, write);
```
scottj97: The parent class is `virtualized_csr_t` so this should call that class's `verify_permissions`. Otherwise if `virtualized_csr_t` one day overrides this method there could be a subtle bug here.

riscv/csrs.cc
```diff
@@ -1574,3 +1574,66 @@ void jvt_csr_t::verify_permissions(insn_t insn, bool write) const {
...
+sscsrind_reg_csr_t::sscsrind_reg_csr_t(processor_t* const proc, const reg_t addr, const reg_t init, csr_t_p iselect) :
+  basic_csr_t(proc, addr, init),
+  iselect(iselect) {
+}
+
+void sscsrind_reg_csr_t::verify_permissions(insn_t insn, bool write) const {
+  // Don't call base verify_permission for VS registers because the addresses always fail
+  // the permission check when remapped to S-mode
+  if (!(address >= CSR_VSIREG && address <= CSR_VSIREG6))
+    basic_csr_t::verify_permissions(insn, write);
```
scottj97: This seems like it could be a problem if VS-level code tries to access `vsireg` (0x251). It should not be accessible to VS-level code at that address, but this will skip that permission check.

riscv/processor.cc
```diff
@@ -524,6 +524,35 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
...
+      mireg[i] = std::make_shared<sscsrind_reg_csr_t>(proc, mireg_csrs[i], 0, miselect);
+      csrmap[mireg_csrs[i]] = mireg[i];
+    }
+  }
+
+  if (proc->extension_enabled_const(EXT_SSCSRIND)) {
+    const reg_t vsireg_csrs[] = { CSR_VSIREG, CSR_VSIREG2, CSR_VSIREG3, CSR_VSIREG4, CSR_VSIREG5, CSR_VSIREG6 };
+    for (auto i = 0; i < 6; i++) {
+      vsireg[i] = std::make_shared<sscsrind_reg_csr_t>(proc, vsireg_csrs[i], 0, vsiselect);
+      csrmap[vsireg_csrs[i]] = (sireg[i]);
```
scottj97: Shouldn't the right side be `vsireg[i]`?

riscv/isa_parser.cc
```diff
@@ -287,9 +287,38 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
...
+    } else if (ext_str == "sscsrind") {
+      extension_table[EXT_SSCSRIND] = true;
     } else if (ext_str[0] == 'x') {
       extension_table['X'] = true;
-      if (ext_str.size() == 1) {
+      if (ext_str == "xbitmanip") {
+        extension_table[EXT_XZBP] = true;
+        extension_table[EXT_XZBS] = true;
+        extension_table[EXT_XZBE] = true;
+        extension_table[EXT_XZBF] = true;
```
scottj97: What's all these bitmanip-related changes doing here?

atulkharerivos:
@scottj97 My bad -- had accidentally dropped the commit for the regenerated encoding.h -- fixed.

scottj97 (changes requested):

riscv/csrs.h
```diff
@@ -6,8 +6,12 @@
 #include "encoding.h"
 // For reg_t:
 #include "decode.h"
+// For std::map
+#include <map>
```
scottj97: I guess this should be `unordered_map`? I can't see `std::map` used below, and this fails compile for me because `unordered_map` is undefined.

atulkharerivos: Yes indeed -- will push an update along with resolving other rebase artifacts (like the bitmap features in isaparser.cc).

riscv/csrs.cc
atulkharerivos: Great catch -- fixed now. If VS mode access is redirected, the CSR address won't match the instruction, but if it tries to access it directly, it will match, so we need to prevent that.

atulkharerivos:
@scottj97 Any further thoughts on the changes?

riscv/processor.cc
scottj97: Still one of these array lengths hardcoded here, as of da7389761c11be070d237e21aa93a272e98ece11

scottj97:
Other than the one hard-coded array length I commented on, this looks good to me.

riscv/processor.cc
atulkharerivos: Fixed.

atulkharerivos:
Thanks -- the array length has been fixed.

scottj97 (approved):
Thank you.

It looks like a rebase is needed before this can be merged.

atulkharerivos:
Rebased.

aswaterman (approved):

scottj97 (approved):

PR #1384, opened 2023-06-17
Remove Xbitmanip instructions

aswaterman:
This change is motivated by the fact that Zicond extension support is broken, as pointed out in https://github.com/riscv/riscv-zicond/issues/12, because the ratified Zicond opcodes overlap the unratified Zbt opcode for CMOV.

We gave the unratified bitmanip instructions a two-year grace period in Spike, which I think was generous.  And when some of these instructions are eventually revived, their implementation can be found in the git history.


ptomsich (approved):

ptomsich:
I'll follow this up with a disasm.cc change to correctly recognize the czero.{eqz.nez}, as that had been missed on the original submission.

aswaterman:
Thanks.

PR #1385, opened 2023-06-17
In CI, test that instruction opcodes don't overlap

aswaterman:
The intent is to catch accidents like the one described in https://github.com/riscv/riscv-zicond/issues/12.  We assume that users of DECLARE_OVERLAP_INSN know what they're doing, and so we simply filter those out.

The test uses a naive quadratic-time algorithm, but even with 10,000 instructions, it should complete in less than a second.

This should fail until #1384 is merged.


jerryz123 (approved):
Neat. I agree the naive algorithm is the cleaner thing to do here.

PR #1386, opened 2023-06-18
Makefile: fix type in check target

ptomsich:
The check target processes the output using grep; however, one of the patterns misspelled 'Segmenetation'.  Fixing the typo.


jerryz123 (approved):

aswaterman:
whoops, thanks for the fix.

PR #1387, opened 2023-06-18
Add Zicond to disassembler

ptomsich:
Disassembler-support had been missing from the original Zicond enablement. Adding it now.


aswaterman (approved):
Thanks!

PR #1388, opened 2023-06-19
Zvk disassembler support (on top of PR 1303)

egouriou-rivos:
Zvk: disassembler support

This is meant to be pulled *after* https://github.com/riscv-software-src/riscv-isa-sim/pull/1303 has gone in. Only one commit is relevant and all previous ones will disappear once rebased.

Add disassembler support for all instructions in Zvk extensions:
 - Zvbb (bitmanip)
 - Zvbc (carryless multiplication)
 - Zvkg (GMAC)
 - Zvkned (AES)
 - Zvknha / Zvknhb (SHA-256, SHA-512)
 - Zvksed (SM4)
 - Zvksh (SM3)

Macros are used to limit code duplication, following the example
of the base V extension. Because the V extension undefines some
of its macros after their use, there Zvk support does define some
similar macros.

Co-authored-by: Gianluca Guida <gianluca@rivosinc.com>
Signed-off-by: Eric Gouriou <ego@rivosinc.com>

### DIFF
--- a/disasm/disasm.cc
+++ b/disasm/disasm.cc
@@ -189,6 +189,12 @@ struct : public arg_t {
 
 struct : public arg_t {
   std::string to_string(insn_t insn) const {
+    return std::to_string(insn.v_zimm6());
+  }
+} v_zimm6;
+
+struct : public arg_t {
+  std::string to_string(insn_t insn) const {
     int32_t target = insn.sb_imm();
     std::string s = target >= 0 ? "pc + " : "pc - ";
     s += std::to_string(abs(target));
@@ -676,6 +682,11 @@ static void NOINLINE add_vector_vi_insn(disassembler_t* d, const char* name, uin
 static void NOINLINE add_vector_viu_insn(disassembler_t* d, const char* name, uint32_t match, uint32_t mask)
 {
   d->add_insn(new disasm_insn_t(name, match, mask, {&vd, &vs2, &zimm5, opt, &vm}));
+}
+
+static void NOINLINE add_vector_viu_z6_insn(disassembler_t* d, const char* name, uint32_t match, uint32_t mask)
+{
+  d->add_insn(new disasm_insn_t(name, match, mask, {&vd, &vs2, &v_zimm6, opt, &vm}));
 }
 
 static void NOINLINE add_vector_vvm_insn(disassembler_t* d, const char* name, uint32_t match, uint32_t mask)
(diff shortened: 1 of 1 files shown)


egouriou-rivos:
@aswaterman
@nibrunieAtSi5

I am not crazy about the macro duplications, neither between V and the Zvk extensions, nor between the Zvk extensions themselves. Because I followed the V example I undef the macros after each extension, leading to some duplication between extensions. Feedback is appreciated.

I wish we could use more functions, but I don't really see how to avoid macros when dealing with identifier generation (`vror` -> `match_vror_vv`, `match_vror_vx`, etc.).

riscv/encoding.h
```diff
@@ -421,6 +421,12 @@
 #define MASK_AMOAND_D 0xf800707f
 #define MATCH_AMOAND_W 0x6000202f
 #define MASK_AMOAND_W 0xf800707f
+#define MATCH_AMOCAS_D 0x2800302f
+#define MASK_AMOCAS_D 0xf800707f
+#define MATCH_AMOCAS_Q 0x2800402f
+#define MASK_AMOCAS_Q 0xf800707f
+#define MATCH_AMOCAS_W 0x2800202f
+#define MASK_AMOCAS_W 0xf800707f
```
nibrunieAtSi5: I assume those are side effects of the `riscv-opcodes` update (and not directly related to vector crypto) ?

nibrunieAtSi5:
I think the PR should have as base branch the branch of https://github.com/riscv-software-src/riscv-isa-sim/pull/1303, namely `rivosinc:zvk-vector-crypto` at least until the PR merge (this would make only the delta change appear for review in this PR).

riscv/encoding.h
egouriou-rivos: Indeed. The encoding update is the one from https://github.com/riscv-software-src/riscv-isa-sim/pull/1303. It is made from riscv-opcode's master branch at HEAD because the previous encoding update contained commits very close to (at?) HEAD but seemingly not from the master branch or any commit that appears in riscv-opcodes's repo.

egouriou-rivos: With the merge of https://github.com/riscv-software-src/riscv-isa-sim/pull/1303, those changes disappeared from this PR.

aswaterman (approved):
Thanks. Auto-merge should do its thing once the PR clears CI.

aswaterman:
Oddly, auto-merge took effect before CI passed. Please babysit the CI task and create a fixup PR if there are any failures.

egouriou-rivos:
For the record the new version that got accepted had those changes:
 - rebase, making all commits disappear but the disassembler one;
 - fixed the vaes{ed}{mf}.{vv,vs} and vsm4r.{vv,vs} instructions to use the "_V" decoding, as they all are of the form "<op> vd, vs2" instead of the "vs, vs2, vs1" that the _VV macros encode.

Thanks for the merge!

PR #1394, opened 2023-06-27
Let debugger control hart availability

timsifive:
This change lets me test OpenOCD's behavior when harts become available. It only affects how things look to the debugger. Harts that are "unavailable" still execute code as usual.

Control is implemented through the 2 LSBs of the DMCUSTOM register in the Debug Module.

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -73,6 +73,9 @@ debug_module_t::debug_module_t(simif_t *sim, const debug_module_config_t &config
           jal(ZERO, debug_abstract_start - DEBUG_ROM_WHERETO));
 
   memset(debug_abstract, 0, sizeof(debug_abstract));
+  for (unsigned i = 0; i < sizeof(hart_available_state) / sizeof(*hart_available_state); i++) {
+    hart_available_state[i] = true;
+  }
 
   reset();
 }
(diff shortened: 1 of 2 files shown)


riscv/debug_module.cc
```diff
@@ -43,7 +43,7 @@ debug_module_t::debug_module_t(simif_t *sim, const debug_module_config_t &config
   // them because I'm too lazy to add the code to just ignore accesses.
   hart_state(1 << field_width(sim->get_cfg().max_hartid() + 1)),
   hart_array_mask(sim->get_cfg().max_hartid() + 1),
-  rti_remaining(0)
+  rti_remaining(0), hart_available_state{true, true}
```
aswaterman: Seems clunky that the array's size of 2 is repeated multiple times in this code.  If you invert the polarity and make this `hart_unavailable_state` instead, you could use the initializer `{false}` to do the right thing regardless of size.

riscv/debug_module.cc
```diff
@@ -522,6 +536,10 @@ bool debug_module_t::dmi_read(unsigned address, uint32_t *value)
       case DM_DMCS2:
         result = set_field(result, DM_DMCS2_GROUP, selected_hart_state().haltgroup);
         break;
+      case DM_CUSTOM:
+        result = set_field(result, 1, hart_available_state[0]);
+        result = set_field(result, 2, hart_available_state[1]);
```
aswaterman: Similarly, it would be cleaner if the hardcoded array size of 2 weren't duplicated here; could use a loop that iterates until `sizeof(hart_available_state) / sizeof(*hart_available_state)` instead.

riscv/debug_module.cc
timsifive: I've turned everything into loops. I prefer to preserve the name as `hart_available_state` to avoid a negative in the variable name.

aswaterman (approved):

PR #1405, opened 2023-07-10
Link spike main with --whole-archive

jerryz123:
The `--extlib` and `--extension` features require runtime dynamic loading of externally compiled libraries. Spike main must make all symbols of `libriscv` available for those libraries. Without `--whole-archive`, some symbols of `libriscv` would not be linked into spike, causing dynamic loader errors when using `--extension/--extlib` with plugins that reference those symbols.

Jamming `libriscv.a` into `LDFLAGS` is a bit of a hack... I don't see a cleaner way to do this with the existing Makefile.
EDIT: It turns out `--whole-archive` isn't available in clang anyways :grimacing: 

An alternative solution would be to dynamically link the library dependencies into the spike main binaries. Not sure which is preferable

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -264,7 +264,7 @@ $$($(2)_test_objs) : %.o : %.cc
 	$(COMPILE) -c $$<
 
 $$($(2)_test_exes) : %-utst : %.t.o $$($(2)_test_libnames)
-	$(LINK) -o $$@ $$< $$($(2)_test_libnames) $(LIBS)
+	$(LINK) $$($(2)_LDFLAGS) -o $$@ $$< $$($(2)_test_libnames) $(LIBS)
 
 $(2)_deps += $$($(2)_test_deps)
 $(2)_junk += \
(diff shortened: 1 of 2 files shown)


scottj97 (approved):

aswaterman:
I see no problem with the `libriscv.a`-in-LDFLAGS hack.  I suspect there's some way to do the equivalent in Clang.

Note also that dynamic linking on Mac OS works a little differently, though I don't recall the details offhand.  Whatever workaround you end up settling on, you'll want to test it out on Mac OS, in addition to making sure CI passes.

jerryz123:
I'm going to close this, https://github.com/riscv-software-src/riscv-isa-sim/pull/1409 does this without using a GCC-only flag.

> Note also that dynamic linking on Mac OS works a little differently, though I don't recall the details offhand. Whatever workaround you end up settling on, you'll want to test it out on Mac OS, in addition to making sure CI passes.

It seems like the --extension feature never worked properly in MacOS to begin with, from testing with the "cflush" extension. #1409 should compile under both GCC and Clang at least.

PR #1406, opened 2023-07-10
build: Dynamically link installed binaries

jerryz123:
An alternative solution to the same problem as #1405 

With static linking, some symbols are duplicated when a dynamically loaded extlib/extension plugin loads libriscv.so again. Notably, the default clint/plic/ns16550 extensions will re-register themselves.

Dynamic linking resolves this, since then the dynamic load of the plugin library could freely load symbols from `libriscv.so`.


scottj97 (approved):
I have only a basic understanding of static/dynamic linking issues, but this seems cleaner to me than #1405.

aswaterman:
Within SiFive, dynamically linking against `libriscv.so` etc. has been problematic for certain use cases.  I'd like to push on the #1405 approach instead.

PR #1409, opened 2023-07-11
Add all symbols from extension.o to spike main

jerryz123:
A more refined approach than https://github.com/riscv-software-src/riscv-isa-sim/pull/1405

Prepending extension.o to the linker command will force inclusion of the critical missing symbols.
Unlike `--whole-archive`, this approach is supported for both GCC and Clang.

Note that the `--extension` flag support appears to have never worked properly for MacOS. Even the example "cflush" extension fails to load.


aswaterman (approved):
This approach is fine with me—though CI seems not to agree.

jerryz123:
As I debug this, I become confused over how this worked in the first place.
`disasm.o` in `libdisasm` references `isa_parser_t::isa_parser_t` in `libriscv`, yet with the current linking, the order is `libriscv.a libdisasm.a`.

It seems there is a circular dependency across these two libraries. I haven't figured out why the change in this PR is exposing this, and only for certain binaries (not spike-main, for example). 

I think there's two ways to resolve this.
1) Use `--start-group` and `--end-group` to force the static linker to iterate over the archives until the symbol table is resolved. This seems like bad practice
2) Attempt to remove the dependence of `libdisasm` on `libriscv`, by moving `isa_parser_t` into `libdisasm`. AFAICT, the interface between `isa_parser_t` and the reset of `libriscv` is narrow, only minor adjustments would be necessary. 

Option 2) is what I'll go with.

aswaterman (approved):
Interesting.

I agree the source-code change is more hygienic than the archive-group approach.

PR #1410, opened 2023-07-12
mcontrol/mcontrol6 triggers on cbo.flush/clean

YenHaoChen:
The mcontrol/mcontrol6 store address before has a higher priority over page faults and access faults. Thus, trigger checking should before the translate().

Reference: Debug spec 1.0, 5.5.3 Cache Operations
Reference: CMO spec 1.0.1, 2.5.4 Breakpoint Exceptions and Debug Mode Entry

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -224,6 +224,7 @@ public:
   }
 
   void clean_inval(reg_t addr, bool clean, bool inval) {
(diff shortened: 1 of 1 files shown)


YenHaoChen:
Merge this PR into https://github.com/riscv-software-src/riscv-isa-sim/pull/1413.

PR #1413, opened 2023-07-13
mcontrol/mcontrol6 on CBO

YenHaoChen:
This PR fixes two mcontrol/mcontrol6 trigger issues on cbo instructions.

1. The tval should capture the effective address when a (trigger) exception on cbo.zero. (https://github.com/riscv/riscv-isa-manual/issues/1317). The previous implementation does not pass the effective address on checking_triggers(). This PR proposes extending the checking_triggers() with an additional parameter "tval" and passing the effective address.
2. The mcontrol/mcontrol6 trigger must match cbo.clean/inval as stores. This PR checks all addresses in the cache block appointed by the cbo instruction.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -169,7 +169,7 @@ bool mmu_t::mmio(reg_t paddr, size_t len, uint8_t* bytes, access_type type)
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


aswaterman:
@scottj97 may we rely on you to run this one through your internal debug tests?

scottj97 (approved):
I don't have any coverage of CBO instructions, but this doesn't seem to break anything else.

PR #1415, opened 2023-07-14
Make methods of `mem_t` virtual to allow overriding

### LINKED ISSUE #1408: Allow overriding `mem_t`
Hi! We'd like to be able to inject custom implementation of `mem_t` (that contains just a single `char*` for its memory) when building `sim_t`. Currently this is not possible since the methods of `mem_t` are not `virtual`, and the implementation of `mem_t` is sparse (splits things into page-sized allocations and stores them in `std::map`). 

There are a few options how this could be supported:

1. Change `mem_t` to support both the current sparse version of memory as well as just a `char*` (through something like `std::variant` or a manual tag).
2. Change `mem_t` to be an abstract class, where its current implementation would move to its own class `sparse_mem_t` (similarly to what #1342 does).
3. Make the methods of `mem_t` virtual. Currently only `contents()` would really need to be virtual.

Let me know what you think about it and if you have any better ideas! Thanks in advance! 😃

michalt:
This allows the users to provide their own custom implementations of `mem_t` when constructing `sim_t`.

Fixes #1408.

### DIFF
--- a/ci-tests/testlib.c
+++ b/ci-tests/testlib.c
@@ -2,9 +2,9 @@
 
 // Copied from spike main.
 // TODO: This should really be provided in libriscv
-static std::vector<std::pair<reg_t, mem_t*>> make_mems(const std::vector<mem_cfg_t> &layout)
+static std::vector<std::pair<reg_t, abstract_mem_t*>> make_mems(const std::vector<mem_cfg_t> &layout)
 {
-  std::vector<std::pair<reg_t, mem_t*>> mems;
+  std::vector<std::pair<reg_t, abstract_mem_t*>> mems;
   mems.reserve(layout.size());
   for (const auto &cfg : layout) {
     mems.push_back(std::make_pair(cfg.get_base(), new mem_t(cfg.get_size())));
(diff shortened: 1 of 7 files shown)


michalt:
@aswaterman @jerryz123 I know this wasn't your preferred option when discussing this on #1408, but looking at the code now, I think it'd be a bit nicer if we separated the interface from the implementation: We could have an`abstract_mem_t` just for the interface, and a `mem_t` (or a `sparse_mem_t`) that implements the interface; `sim_t` would take `abstract_mem_t` to allow injecting custom implementations. That being said, I don't feel strongly, and the current version of the PR does solve the problem for my team. 😃

aswaterman:
I'm OK with that alternative if @jerryz123 is.

jerryz123:
I have no problems with that.

michalt:
Great, I updated the PR with the changes. Let me know what you think! Thanks! 😄

scottj97 (approved):

aswaterman:
Seems like there are some CI failures.

jerryz123:
Can you squash the two commits together? Each commit in PRs to this project is checked for functionality.

michalt:
@jerryz123 Done! 😃  (BTW is there a way for me to trigger the tests?)

michalt:
FYI: I missed one `mem_t` and had to push again.

jerryz123 (approved):

PR #1416, opened 2023-07-17
Legalize xenvcfg.CBIE

YenHaoChen:
The value 2 of xenvcfg.CBIE is reserved. This PR legalizes it to 0.

The values 2 and 0 of xenvcfg.CBIE do not have corresponding macros. Thus, this PR hard-codes those values.

Reference: https://github.com/riscv/riscv-CMOs/issues/65

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1476,7 +1476,7 @@ void sstateen_csr_t::verify_permissions(insn_t insn, bool write) const {
 // implement class senvcfg_csr_t
 senvcfg_csr_t::senvcfg_csr_t(processor_t* const proc, const reg_t addr, const reg_t mask,
                              const reg_t init):
-  masked_csr_t(proc, addr, mask, init) {
+  envcfg_csr_t(proc, addr, mask, init) {
 }
 
 void senvcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -916,7 +916,7 @@ bool envcfg_csr_t::unlogged_write(const reg_t val) noexcept {
 
 // implement class henvcfg_csr_t
 henvcfg_csr_t::henvcfg_csr_t(processor_t* const proc, const reg_t addr, const reg_t mask, const reg_t init, csr_t_p menvcfg):
-  masked_csr_t(proc, addr, mask, init),
+  envcfg_csr_t(proc, addr, mask, init),
   menvcfg(menvcfg) {
 }
 
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -901,6 +901,19 @@ bool masked_csr_t::unlogged_write(const reg_t val) noexcept {
   return basic_csr_t::unlogged_write((read() & ~mask) | (val & mask));
 }
 
+envcfg_csr_t::envcfg_csr_t(processor_t* const proc, const reg_t addr, const reg_t mask,
+                             const reg_t init):
+  masked_csr_t(proc, addr, mask, init) {
+  // In unlogged_write() we WARLize this field for all three of [msh]envcfg
+  assert(MENVCFG_CBIE == SENVCFG_CBIE && MENVCFG_CBIE == HENVCFG_CBIE);
+}
+
+bool envcfg_csr_t::unlogged_write(const reg_t val) noexcept {
+  const reg_t cbie_reserved = 2; // Reserved value of xenvcfg.CBIE
+  const reg_t adjusted_val = get_field(val, MENVCFG_CBIE) != cbie_reserved ? val : set_field(val, MENVCFG_CBIE, 0);
+  return masked_csr_t::unlogged_write(adjusted_val);
+}
+
 // implement class henvcfg_csr_t
 henvcfg_csr_t::henvcfg_csr_t(processor_t* const proc, const reg_t addr, const reg_t mask, const reg_t init, csr_t_p menvcfg):
   masked_csr_t(proc, addr, mask, init),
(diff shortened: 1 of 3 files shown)


aswaterman (commented):
Rather than repeating the constant `2` in multiple places, could we at least `#define CBIE_RESERVED 2` in `csrs.cc`?

YenHaoChen:
Add ```#define CBIE_RESERVED 2``` in ```csrs.cc``` as suggested by @aswaterman .

riscv/csrs.cc
```diff
@@ -1460,6 +1462,17 @@ void sstateen_csr_t::verify_permissions(insn_t insn, bool write) const {
...
+// implement class menvcfg_csr_t
+menvcfg_csr_t::menvcfg_csr_t(processor_t* const proc, const reg_t addr, const reg_t mask,
+                             const reg_t init):
+  masked_csr_t(proc, addr, mask, init) {
+}
+
+bool menvcfg_csr_t::unlogged_write(const reg_t val) noexcept {
+  reg_t adjusted_val = get_field(val, MENVCFG_CBIE) != CBIE_RESERVED ? val : set_field(val, MENVCFG_CBIE, 0);
+  return masked_csr_t::unlogged_write(adjusted_val);
+}
```
scottj97: It's pretty ugly to have the same WARLization in three separate places here.

Since CBIE is the same for all three of these `*envcfg` CSRs, how about having `senvcfg` and `henvcfg` inheriting from a parent class that does the WARLization? In that parent class's constructor you can `assert(MENVCFG_CBIE == SENVCFG_CBIE && MENVCFG_CBIE == HENVCFG_CBIE);` to make sure the one implementation will work properly for all three CSRs.

scottj97 (changes requested):

riscv/csrs.cc
YenHaoChen: Add ```envcfg_csr_t``` as the parent class that does the WARLization and assertion.

YenHaoChen:
Replace the global macro ```CBIE_RESERVED``` with a local constant ```cbie_reserved```.

~2023.07.21 update: use the global macro instead of the local constant.~

riscv/csrs.cc
```diff
@@ -901,6 +901,18 @@ bool masked_csr_t::unlogged_write(const reg_t val) noexcept {
...
 
+envcfg_csr_t::envcfg_csr_t(processor_t* const proc, const reg_t addr, const reg_t mask,
+                             const reg_t init):
+  masked_csr_t(proc, addr, mask, init) {
+  assert(MENVCFG_CBIE == SENVCFG_CBIE && MENVCFG_CBIE == HENVCFG_CBIE);
+}
+
+bool envcfg_csr_t::unlogged_write(const reg_t val) noexcept {
+  const reg_t cbie_reserved = 2; // Reserved value of xenvcfg.CBIE
+  reg_t adjusted_val = get_field(val, MENVCFG_CBIE) != cbie_reserved ? val : set_field(val, MENVCFG_CBIE, 0);
```
scottj97: Could be `const`

riscv/csrs.cc
```diff
@@ -901,6 +901,18 @@ bool masked_csr_t::unlogged_write(const reg_t val) noexcept {
   return basic_csr_t::unlogged_write((read() & ~mask) | (val & mask));
 }
 
+envcfg_csr_t::envcfg_csr_t(processor_t* const proc, const reg_t addr, const reg_t mask,
+                             const reg_t init):
+  masked_csr_t(proc, addr, mask, init) {
+  assert(MENVCFG_CBIE == SENVCFG_CBIE && MENVCFG_CBIE == HENVCFG_CBIE);
```
scottj97, line 908: I would suggest a comment here explaining why we assert this. Something like: "In unlogged_write() we WARLize this field for all three of [msh]envcfg"

riscv/csrs.cc
```diff
@@ -901,6 +901,18 @@ bool masked_csr_t::unlogged_write(const reg_t val) noexcept {
...
+}
+
+bool envcfg_csr_t::unlogged_write(const reg_t val) noexcept {
+  const reg_t cbie_reserved = 2; // Reserved value of xenvcfg.CBIE
```
scottj97, line 912: I sorta liked this better as a macro so that it's next to all the other definitions of CBIE. When it's isolated here, if someone changes the interpretation of this field, they are liable to change only the macros and not realize this code exists.

scottj97 (changes requested):

riscv/csrs.cc
```diff
@@ -20,6 +20,8 @@
 #undef STATE
 #define STATE (*state)
 
+#define CBIE_RESERVED 2 // Reserved value of xenvcfg.CBIE
```
scottj97: When I [asked](https://github.com/riscv-software-src/riscv-isa-sim/pull/1416#discussion_r1269366332) to make this a macro, I explained my reasoning: so it would be next to the other CBIE definitions.

But now I see that there are no other CBIE definitions. (Is that right? Are the interpretations of 0/1/3 defined anywhere? I can't find any.)

In that case, this is better the previous way (as a local `const`).

scottj97 (changes requested):
Looks like this needs a rebase.

riscv/csrs.cc
scottj97: Sorry to go back and forth like this. I was assuming last week that somewhere else we had something like:
```
#define CBIE_FOO 0
#define CBIE_BAR 1
#define CBIE_FOOBAR 3
```

YenHaoChen: To my knowledge, there is no public header file that includes all definitions of CSR values. Although the ```debug_defines.h``` includes the value definitions of debug CSRs, it does not include other CSRs. The other possible public header file for global definitions of CSR values is ```encoding.h```, which also includes some value definitions of debug CSRs. However, I believe the ```encoding.h``` should only contain the definition of CSR fields instead of values. 
Nevertheless, I agree there should be a public header file that includes all definitions of CSR values. By the naming rule in ```debug_defines.h``` and ```encoding.h```, we should have something like:
```
#define MENVCFG_CBIE_EXCEPTION 0
#define MENVCFG_CBIE_FLUSH 1
#define MENVCFG_CBIE_RESERVED 2
#define MENVCFG_CBIE_INVALIDATE 3
```
image

scottj97 (approved):
Thanks for your patience.

Issue #1418, opened 2023-07-18
xtval update values on cross-boundary instruction exceptions


poemonsense:
Spike just makes me confused when co-simulating it with rocket. I assume rocket should be correct while Spike not.

Here is the logs:

```
// The last instruction is a ret, and the CPU jumps to 0x1fffe with virtual memory enabled.
// These commit logs are from Rocket
[19] commit pc 000000000001d8fe inst 00008082 wen 1 dst 00 data 0000000000000000 ret
[20] exception pc 000000000001fffe inst 0000ffff cause 0000000000000001 unknown <--

// Here comes the Spike logs.
// This is the last ret instruction
[Spike] pc = 0x1d8fe inst 0x8082

// Starting fetching the next instruction in refill_cache
[Spike] refill_icache addr: 1fffe 

// Fetch the first two bytes
[Spike] mem_ptw   addr: 0x8280b000, data: 0x0000000020a02401, len: 8
[Spike] ptw: level 2, vaddr 0x1fffe, pg_base 0x8280b000, p_pte 0x8280b000, pte.val 0x20a02401
[Spike] mem_ptw   addr: 0x82809000, data: 0x0000000020a03401, len: 8
[Spike] ptw: level 1, vaddr 0x1fffe, pg_base 0x82809000, p_pte 0x82809000, pte.val 0x20a03401
[Spike] mem_ptw   addr: 0x8280d0f8, data: 0x0000000020eacc5b, len: 8
[Spike] ptw: level 0, vaddr 0x1fffe, pg_base 0x8280d000, p_pte 0x8280d0f8, pte.val 0x20eacc5b

// after fetching the first two bytes, it returns 0xffff
// insn_length gives 8 so it tries to fetch the following 6 bytes in the orders of 6, 4, 2
[Spike] refill_icache addr: 1fffe length: 8, insn ffff  

// Now it starts fetching the following bytes (in the next-page) at base+6
[Spike] mem_ptw   addr: 0x8280b000, data: 0x0000000020a02401, len: 8
[Spike] ptw: level 2, vaddr 0x20004, pg_base 0x8280b000, p_pte 0x8280b000, pte.val 0x20a02401
[Spike] mem_ptw   addr: 0x82809000, data: 0x0000000020a03401, len: 8
[Spike] ptw: level 1, vaddr 0x20004, pg_base 0x82809000, p_pte 0x82809000, pte.val 0x20a03401
[...]

PR #1419, opened 2023-07-18
mmu: fetch instruction bytes in ascending order

poemonsense:
Fetching instruction bytes in descending order would result in wrong xtval update values.

See https://github.com/riscv-software-src/riscv-isa-sim/issues/1418


aswaterman (approved):
@scottj97 and/or @jerryz123 please provide a sanity-check on this one.  I think it's the right solution to #1418.

jerryz123 (approved):
The analysis in #1418 makes sense to me

PR #1427, opened 2023-07-26
triggers: fix textra.sbytemask

YenHaoChen:
Ignore corresponding bytes to the scontext and textra.svalue.

Update: Cast 0xff to reg_t for the 34-bit textra64.svalue.

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -84,7 +84,7 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
     assert(CSR_TEXTRA32_SBYTEMASK_LENGTH < CSR_TEXTRA64_SBYTEMASK_LENGTH);
     for (int i = 0; i < CSR_TEXTRA64_SBYTEMASK_LENGTH; i++)
       if (sbytemask & (1 << i))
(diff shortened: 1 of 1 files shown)


riscv/triggers.cc
```diff
@@ -84,7 +84,7 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
     assert(CSR_TEXTRA32_SBYTEMASK_LENGTH < CSR_TEXTRA64_SBYTEMASK_LENGTH);
     for (int i = 0; i < CSR_TEXTRA64_SBYTEMASK_LENGTH; i++)
       if (sbytemask & (1 << i))
-        mask &= 0xff << (i * 8);
+        mask &= ~(0xff << (i * 8));
```
aswaterman: Don't you need to cast `0xff` to `reg_t` for this to do the right thing for mask values >= 2^32?

YenHaoChen: Cast 0xff to reg_t for the 34-bit textra64.svalue.

aswaterman (approved):
OK with you, @timsifive?

timsifive (approved):
Looks correct to me.

PR #1434, opened 2023-08-03
 update model_test.h: set_msw/clear_msw/set_mtimer/clear_mtimer

dansmathers:
Added ifndef to clint addresses instead of hard-coding
Added clear_msw and clear mtimer
Tested against Sail/isa-sim with new proposed Smclint/Ssclint arch-tests
https://github.com/riscv-non-isa/riscv-arch-test/pull/372
Building a baseline of interrupt tests that changes to SAIL/isa-sim can be tested against when other interrupt extensions are added.

### DIFF
--- a/arch_test_target/spike/model_test.h
+++ b/arch_test_target/spike/model_test.h
@@ -49,16 +49,50 @@
 //RVTEST_IO_ASSERT_DFPR_EQ
 #define RVMODEL_IO_ASSERT_DFPR_EQ(_D, _R, _I)
 
-#define RVMODEL_SET_MSW_INT       \
- li t1, 1;                         \
- li t2, 0x2000000;                 \
- sw t1, 0(t2);
+#ifndef RVMODEL_MCLINTBASE    
+        #define RVMODEL_MCLINTBASE 0x02000000
+#endif
 
-#define RVMODEL_CLEAR_MSW_INT     \
- li t2, 0x2000000;                 \
- sw x0, 0(t2);
+#ifndef RVMODEL_MSIP_OFFSET    
+        #define RVMODEL_MSIP_OFFSET 0x0
+#endif
 
-#define RVMODEL_CLEAR_MTIMER_INT
+#ifndef RVMODEL_MTIMECMP_OFFSET    
+        #define RVMODEL_MTIMECMP_OFFSET 0x4000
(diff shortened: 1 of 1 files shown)


aswaterman:
@neelgala Can you review this PR since you made the initial arch-test framework contribution?

dansmathers:
@neelgala, did you get a chance to look at this?   I'm not aware of any other tests currently using these macros so I think it is a pretty simple pull request.

dansmathers:
@aswaterman, can anyone else review this?  My understanding is @neelgala is quite busy with a new job.  Note that set and clear mtimer int macros are currently broken in model_test.h so no one is clearly usings these macros yet.
Thanks and regards,
Dan

aswaterman:
Maybe @allenjbaum?  I can't really review this PR since I'm unfamiliar with the codebase.  I can simply merge it as-is, but that should be a last resort.

allenjbaum:
The codebase here is the Spike codebase, so I'm a bit surprised.
The issue here is that to properly test interrupts, we need to be able to
set and clear the IP bits, and the way to do it is inherently model
specific.
That may be as simple as an MMIO write to a specific address with a
specific value, or something else.
So, if Spike has a way to clear (and/or set) the 3 architecturally
defined interrupts
(which might be as simple as reading the claim register for external
interrupts, or writing mtimecmp for timer interrupts),
then that's what these macros should do.
But it may not be possible for an Smode handler to do some of those things,
in which case it should be wrapped inside an ACT specific GOTO_MMODE,
GOTO_LOWER_MODE macro,
and that's part of the codebase you may not be familiar with.

I'll review this PR when I get a chance, but I'm not sure that I'll be able
to verify that they actually work for SPIKE since I'm not familiar with
that code base.

IF you can think of another way that this can be included or reviewed, let
me lknow.
I may not get to this for another week or more.

neelgala (approved):

neelgala:
@dansmathers @aswaterman apologies for the delay on this - this looks good. External interrupts will require a lot more thought (on ACT side as well) before we can do something for it here. We can merge this.

aswaterman:
@neelgala Thank you for taking the time to review it!

aswaterman (approved):
Approving based on @neelgala's review

PR #1435, opened 2023-08-09
term: init termios only when it's used

f0rget-the-sad:
Currently termios is initialized as global static objects, this could
cause several problems:

a) When two or more process are launched from the same terminal it's
possible to have an incorrect sequence of saving and restoring flags.

For example:
process 1: get flags (with enabled ICANON | ECHO)
process 1: store original flags
process 1: set flags (with disabled ICANON | ECHO)
process 2: get flags (with already disabled ICANON | ECHO)
process 2: store 'original' flags <- This flags were modified by process 1
process 2: set flags (with disabled ICANON | ECHO)
process 1: restores original flags
process 2: restores 'original' flags <- Terminal in wrong state

To resolve this issue, store original flags only when ICANON | ECHO flags
were set in the configuration, this would allow only the first process
to modify and restore the flags.

b) Flag restoring relies on static object destructor, which will be
called on exit(), but not in the case of abort().
There's handle registered for this case in htif_t() constructor:

  ...
  signal(SIGABRT, &handle_signal); // we still want to call static destructors

But in case when htif_t object is not created, it could lead to problems
with restoring a terminal state in case of abort(), since the terminal
object always exists as global static.

To resolve this issue, introduce canonical_terminal_t::init() function
that will initialize the terminal only when it's actually used by
other components (i. e. when proper signal handler is registered).


f0rget-the-sad:
Kind reminder to all maintainers, please review and ack/comment.
Let me know if some details are required from my side, or it's just pending in review list.

fesvr/term.cc
```diff
@@ -30,7 +31,10 @@ class canonical_termios_t
   bool restore_tios;
 };
 
-static canonical_termios_t tios; // exit() will clean up for us
+void canonical_terminal_t::init()
+{
+	static canonical_termios_t tios; // exit() will clean up for us
```
jerryz123:
```suggestion
  static canonical_termios_t tios; // exit() will clean up for us
```

jerryz123 (commented):
Fix spacing

jerryz123 (commented):
Would it not be sufficient to set `restore_tios` to true only if this process is the one which constructs the static tios?

f0rget-the-sad:
I'm not sure that I properly understand the suggestion.

`tios` is global static variable, which it is constructed in every process that links with term.cc
Each process constructs it's own `tios` at the begging of execution, so it's not clear how to distinguish them here.

This commit attempts to limit `tios`'s scope to function static (which mean's it will be initialized on first use) and
set `restore_tios` to true only if the required flags is not set.

PR #1436, opened 2023-08-14
Rename *envcfg.HADE to *envcfg.ADUE

ved-rivos:


aswaterman (approved):
Thanks for following up with these changes.

PR #1438, opened 2023-08-16
fdt: Install header files `fdt.h` and `libfdt_env.h` as needed by `libfdt.h`

liuyu81:
All three header files `libfdt.h`, `fdt.h` and `libfdt_env.h` need to be installed, for correctly compiling periphery devices, as exemplified in [https://github.com/ucb-bar/spike-devices](https://github.com/ucb-bar/spike-devices).


aswaterman (approved):
I do see that `libfdt.h` includes the other two headers, so it's pointless to install `libfdt.h` without installing the others.

PR #1445, opened 2023-08-25
use correct sstatten csr number

chihminchao:


chihminchao:
@liweiwei90  You may be interested at that.

aswaterman (commented):
I think these should be separate PRs, but individually both LGTM.

chihminchao:
Separate the bf16 part into  https://github.com/riscv-software-src/riscv-isa-sim/pull/1446

aswaterman (approved):

aswaterman:
Waiting for @liweiwei90's review before I merge this.

PR #1448, opened 2023-08-27
A/D updates in G-stage PTE

ved-rivos:
This PR includes two commits:
1. Invoke G-stage to check write permission to VS-stage page table before setting A and/or D bit in VS-stage PTE. This will also cause A and/or D bit to be set in G-stage PTE mapping the VS-stage PT appropriately.
2. Report the pseudo-instructions specified in "Transformed Instruction or Pseudoinstruction for mtinst or htinst" in Priv specification for guest page fault that may occur when doing an address translation for VS-stage PT address. This PR assumes that HSXLEN == VSXLEN - as spike does not support dynamic XLEN change.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -536,6 +536,9 @@ reg_t mmu_t::walk(mem_access_info_t access_info)
 
       if ((pte & ad) != ad) {
         if (hade) {
+          // Check for write permission to the first-stage PT in second-stage
+          // PTE and set the D bit in the second-stage PTE if needed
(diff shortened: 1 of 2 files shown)


defermelowie:
The spec. explicitly states about the pseudo-instruction that 
> ... the value written to `mtinst` or `htinst` must be taken from Table 9.12; zero is not allowed. ~ Section 9.6.3

Therefore, the current behavior of writing zero is incorrect? What's the current status on this PR which seems to fix the issue?

aswaterman (approved):
@defermelowie I think it just got lost in the shuffle; I agree we should merge it.

PR #1452, opened 2023-09-01
fix condition of executing cbo.inval as a flush operation

viktoryou:
As CMO spec described in Chapter 3, page 13:

> if (((priv_mode != M) && (menvcfg.CBIE == 01)) ||
> ((priv_mode == U) && (senvcfg.CBIE == 01)) ||
> ((priv_mode == VS) && (henvcfg.CBIE == 01)) ||
> ((priv_mode == VU) && ((henvcfg.CBIE == 01) || (senvcfg.CBIE == 01))))
> {
> <execute CBO.INVAL and perform flush operation>

the condition of executing cbo.inval as a flush operation has a clerical error. Please help to check that.


aswaterman:
Worse still, the previous line has an assignment `sCBIE = 1` instead of a comparison `sCBIE == 1`.  I'll fix both.

aswaterman:
Manually merged via 3c8320ecd76b1d98b5d0955dd521bf5ee1616136

cc @liweiwei90

PR #1458, opened 2023-09-13
Hit multiple icount triggers with different actions

YenHaoChen:
Entering debug mode does not decrease icount triggers (https://github.com/riscv/riscv-debug-spec/issues/842).
Thus, if any pending icount enters debug mode, all icount triggers shall not decrease.

This PR provides the behavior by decoupling firing and decreasing routines. The revised code checks whether any pending icount enters debug mode before decreasing all icount triggers.


timsifive (approved):
LGTM

aswaterman (approved):
LGTM

PR #1459, opened 2023-09-13
fix: mcontrol/mcontrol6 on fetch

YenHaoChen:
The fetch address and data are in refill_icache() instead of fetch_slow_path(). This commit moves the mcontrol/mcontrol6 checkings from fetch_slow_path() to refill_icache() for correct fetch values (https://github.com/riscv-software-src/riscv-isa-sim/commit/2128daa0e700c0109c1931f997d78ea70d60dbd3).

The last commit (https://github.com/riscv-software-src/riscv-isa-sim/pull/1459/commits/1779c358130b2cd92adef4e9ebe4befcbcba009b) is for performance improvement. I need help verifying the effectiveness.

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -290,6 +290,7 @@ public:
     if (matched_trigger)
       throw *matched_trigger;
 
+    auto access_info = generate_access_info(addr, FETCH, {false, false, false});
     auto tlb_entry = translate_insn_addr(addr);
     insn_bits_t insn = from_le(*(uint16_t*)(tlb_entry.host_offset + addr));
     int length = insn_length(insn);
@@ -307,6 +308,7 @@ public:
       insn |= (insn_bits_t)from_le(*(const uint16_t*)translate_insn_addr_to_host(addr + 4)) << 32;
       insn |= (insn_bits_t)from_le(*(const uint16_t*)translate_insn_addr_to_host(addr + 6)) << 48;
     }
+    check_triggers(triggers::OPERATION_EXECUTE, addr, access_info.effective_virt, insn);
 
     insn_fetch_t fetch = {proc->decode_insn(insn), insn};
     entry->tag = addr;
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -291,6 +291,7 @@ public:
       throw *matched_trigger;
 
     auto access_info = generate_access_info(addr, FETCH, {false, false, false});
+    check_triggers(triggers::OPERATION_EXECUTE, addr, access_info.effective_virt);
     auto tlb_entry = translate_insn_addr(addr);
     insn_bits_t insn = from_le(*(uint16_t*)(tlb_entry.host_offset + addr));
     int length = insn_length(insn);
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -290,8 +290,10 @@ public:
     if (matched_trigger)
       throw *matched_trigger;
 
+    reg_t vpn = addr >> PGSHIFT;
     auto access_info = generate_access_info(addr, FETCH, {false, false, false});
-    check_triggers(triggers::OPERATION_EXECUTE, addr, access_info.effective_virt);
+    if (unlikely(tlb_insn_tag[vpn % TLB_ENTRIES] != vpn))
+      check_triggers(triggers::OPERATION_EXECUTE, addr, access_info.effective_virt);
     auto tlb_entry = translate_insn_addr(addr);
     insn_bits_t insn = from_le(*(uint16_t*)(tlb_entry.host_offset + addr));
     int length = insn_length(insn);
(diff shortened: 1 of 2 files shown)


timsifive (commented):
LGTM, but I'm not really familiar with this part of the code.

scottj97:
@YenHaoChen can you explain the failing scenario that 2128daa0e700c0109c1931f997d78ea70d60dbd3 fixes?

I can see one of my testcases was previously taking a fetch breakpoint, and no longer is after that change. I will study it to see which behavior is correct.

scottj97:
The case where I see behavior change:

tdata1[2]=0x200000000010010c (mcontrol match=2(ge) u=1 execute=1)
tdata1[2]=0x0000000080002795
PC that triggered is 80002794, which contains a 32-bit opcode

After this PR, this PC executes without taking the breakpoint trap.

I don't think the spec says one way or the other about this case, but it is a functional change that needs to be documented in the PR. I guess it's only comparing the first byte of the fetch, instead of each byte?

YenHaoChen:
@scottj97 
Split https://github.com/riscv-software-src/riscv-isa-sim/commit/2128daa0e700c0109c1931f997d78ea70d60dbd3 into two commits with additional commit messages for better readability.

The first commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1459/commits/0578364d13a0094ca62c5b36ffd2b750883fcbbb fixes the instruction data comparison. For instance, an instruction data 0xe9830313 cannot match a data trigger with tdata2=0xe9830313 but instead ones with  tdata2={0xe983,0x0313} without this commit.
The second commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1459/commits/146e41e27fa50434d8c08d9dac2268a0349599e2 fixes the instruction address (PC) comparison. For instance, a 4-byte instruction at PC 0x80002794 also matches an address trigger with tdata2=0x80002796 without this commit. (The spec requires to match tdata2=0x80002794 and recommends to match tdata2={0x80002794,80002795,80002796,80002797}. Matching tdata2={0x80002794,80002796} only does not make sense from a user perspective.)

image

scottj97 (approved):
I don't have good test coverage of data triggers, so I'm not surprised that I didn't see that difference.

The last commit I am unsure about and @aswaterman should review.

riscv/mmu.h
```diff
@@ -290,8 +290,10 @@ class mmu_t
     if (matched_trigger)
       throw *matched_trigger;
 
+    reg_t vpn = addr >> PGSHIFT;
     auto access_info = generate_access_info(addr, FETCH, {false, false, false});
-    check_triggers(triggers::OPERATION_EXECUTE, addr, access_info.effective_virt);
+    if (unlikely(tlb_insn_tag[vpn % TLB_ENTRIES] & TLB_CHECK_TRIGGERS))
```
aswaterman: This doesn't look right to me.  What if this is a TLB conflict miss and this is an unrelated TLB entry (with the same VPN LSBs)?  This check is only valid if we've already refilled the TLB.  And in fact we never refill the TLB for the MMIO case, so moving this check after the call to `translate_insn_addr` wouldn't suffice, either.

The old design where the call to `check_triggers` lived in `fetch_slow_path` made more sense to me.  We knew we were guaranteed to end up in `fetch_slow_path` if `TLB_CHECK_TRIGGERS` was true, because the comparison in `translate_insn_addr` would never match.

aswaterman (changes requested):

riscv/mmu.h
YenHaoChen: Update the commit according to @aswaterman's suggestion.

riscv/mmu.h
```diff
@@ -290,8 +290,10 @@ class mmu_t
     if (matched_trigger)
       throw *matched_trigger;
 
+    reg_t vpn = addr >> PGSHIFT;
     auto access_info = generate_access_info(addr, FETCH, {false, false, false});
-    check_triggers(triggers::OPERATION_EXECUTE, addr, access_info.effective_virt);
+    if (unlikely(tlb_insn_tag[vpn % TLB_ENTRIES] != vpn))
+      check_triggers(triggers::OPERATION_EXECUTE, addr, access_info.effective_virt);
```
aswaterman, line 296: I believe this will spuriously invoke `check_triggers` on ITLB misses.  Note that fetches from MMIO devices accesses always result in ITLB misses, so this means `check_triggers` will be invoked on every MMIO fetch.  Maybe this is functionally correct and is only a perf issue, but it seems undesirable.

What's wrong with my original suggestion of moving this check back into `fetch_slow_path`?  By construction, we should not need to check the triggers on the fetch fast path...

YenHaoChen: The current implementation splits instruction fetching into multiple 2-byte operations, which call ```fetch_slow_path``` separately. This results in that a 4-byte instruction 0xe9830313 matches data triggers with tdata2={0xe983, 0x0313} instead of one with tdata2=0xe9830313 incorrectly. Additionally, a 4-byte instruction with PC at 0x4000 matches address triggers with tdata2={0x4000, 0x4002}, which is unexpected from a user perspective.

The behavior of the data trigger in the current implementation does not conform to the specification. It is weird to me that the address trigger matches PCs of 2-byte granularity. Clarify whether the specification allows the address trigger to match an arbitrary subset of the accessing addresses: https://github.com/riscv/riscv-debug-spec/issues/877.

Update 2023.09.20:
The https://github.com/riscv/riscv-debug-spec/issues/877 clarifies that  the specification allows the address trigger to match an arbitrary subset of the accessing addresses. Thus, the specification allows the original behavior of the address trigger. Possibly, we can improve performance by keeping address trigger checking in ```fetch_slow_path```. However, the data trigger truly does not work in ```fetch_slow_path```. Moving address triggers out from ```fetch_slow_path``` also provides a more reasonable behavior from a user perspective.
@aswaterman How do you feel about this?

PR #1476, opened 2023-10-10
Bump configure to autoconf 2.71

Phantom1003:
Hi, I found the configure generated by Ubuntu22's upstream autoconf is broken.

 I suspect that the latest version has some changes that cause erroneous expansion of commas.
As a result, the generated macro definition is truncated.

image

Since the default value of the DEFAULT_VARCH doesn't require shell expansions, I think we can use AC_DEFINE to bypass this problem.

I keep the buggy configure to help us identify any changes in the working version.
Once approved, I will remove it.


Phantom1003:
The CI was failed because the buggy version [a809228](https://github.com/riscv-software-src/riscv-isa-sim/pull/1476/commits/a8092281256f492a74875bb0f3fa2e1dac425115), the fixed version should work

jerryz123:
> The CI was failed because the buggy version [a809228](https://github.com/riscv-software-src/riscv-isa-sim/pull/1476/commits/a8092281256f492a74875bb0f3fa2e1dac425115), the fixed version should work

To maintain a bisectable commit history, the CI checks every commit in a PR, not just the HEAD of the PR branch. 

I believe if you reorder these commits, the tests should pass?

Phantom1003:
> To maintain a bisectable commit history, the CI checks every commit in a PR, not just the HEAD of the PR branch.

Understand.

> I believe if you reorder these commits, the tests should pass?

Sure, the purpose of putting the buggy version first is to highlight the changed lines in the diff.

Phantom1003:
@jerryz123 can you review this patch?

jerryz123:
If you remove the latest commit b587af0 (I'm not sure what this commit is doing), then this looks fine to me

Phantom1003:
@jerryz123 Sorry, that commit should have been generated by accidentally clicking on the sync button.

jerryz123 (approved):
This looks good for me, let's wait for @aswaterman + @scottj97

aswaterman (approved):
OK with me, too.

riscv/riscv.ac
```diff
@@ -24,7 +24,7 @@ AC_ARG_WITH(varch,
 	[AS_HELP_STRING([--with-varch=vlen:128,elen:64],
 		[Sets the default vector config])],
   AC_DEFINE_UNQUOTED([DEFAULT_VARCH], "$withval", [Default value for --varch switch]),
-  AC_DEFINE_UNQUOTED([DEFAULT_VARCH], ["vlen:128,elen:64"], [Default value for --varch switch]))
+  AC_DEFINE([DEFAULT_VARCH], ["vlen:128,elen:64"], [Default value for --varch switch]))
```
scottj97, line 27: The commit message needs to explain what was broken here, and how this fixes it.

scottj97 (changes requested):
This seems to be two different changes: fixing something that was broken, and updating autoconf. ~~Those should be two different PRs.~~

Edit: I understand it was only the newer autoconf that exposed the problem. So I'm okay with it being one PR. But the commit message still needs to explain more, so people don't have to come view the PR to understand it.

riscv/riscv.ac
Phantom1003: The commit message has been updated

scottj97 (approved):

PR #1477, opened 2023-10-10
Update dtm.h with switch_to_* functions

abejgonzalez:
This PR adds `switch_to_{host,target}` functions matching the same functionality as `tsi.cc` (https://github.com/riscv-software-src/riscv-isa-sim/blob/d1efcdffffee57bab0fdbd2b377c6132b37556fd/fesvr/tsi.h#L27, https://github.com/riscv-software-src/riscv-isa-sim/blob/d1efcdffffee57bab0fdbd2b377c6132b37556fd/fesvr/tsi.h#L38).

### DIFF
--- a/fesvr/dtm.h
+++ b/fesvr/dtm.h
@@ -55,6 +55,12 @@ class dtm_t : public htif_t
 
   void producer_thread();
 
+  void switch_to_host() {
+    // update the target with the current context
+    target = context_t::current();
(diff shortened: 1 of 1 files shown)


jerryz123 (approved):
This looks wholly innocuous, so SFTM.

aswaterman (approved):

PR #1482, opened 2023-10-17
Remove vector AMO

a4lg:
Not only this is not a part of the vector extension version 1.0, 8-bit vector AMO instructions will conflict to byte-granularity AMO instructions of the `Zabha` extension (containing subword AMO instructions) in the fast track.

The current vector spec (draft) also states that vector AMO encodings are likely to change, making keeping vector AMO as-is less usable.

This commit removes all vector AMO instructions but not `VI_AMO` macro.


a4lg:
Seems it conflicts with #1481.

chihminchao:
What a coincident!!!

aswaterman:
Funny.  I'll close this one since I've merged its twin.

PR #1468, opened 2023-09-23
Add unratified Zcmop extension

ved-rivos:
Specification: https://github.com/riscv/riscv-isa-manual/blob/main/src/zimop.adoc
Sail ref: https://github.com/riscv/sail-riscv/pull/309
Arch tests: https://github.com/riscv-non-isa/riscv-arch-test/pull/388

PR #1486, opened 2023-10-18
Add unratified Zcmop extension

aswaterman:
Attempt 3.  See #1468


aswaterman:
Merging, as this is a lightly modified version of approved #1468

PR #1490, opened 2023-10-28
Add support for setting PMP Granularity over CLI

### LINKED ISSUE #1478: Changing PMP Grain through CLI
Hi

I'm aware of the option to change the number of PMP entries using the CLI option ``--pmpregions=<n>``. I think it would be helpful if we could also set the PMP grain in a similar fashion. I want to use spike with a pmp grain other than the default value. I went through the code base a little bit and found [this](https://github.com/riscv-software-src/riscv-isa-sim/blob/b5d13f36053d08e310da01e92e8ebc2d29623483/riscv/processor.cc#L669) function. Can we expose it to be called through the CLI ? Perhaps an option ``--pmpgrain=<n>`` ?

Writing a grain that is less than the minimum would simply be ignored based on the current framework, and all of that seems right. I would appreciate it if someone can throw some more light on this. As of this moment, in order to change the PMP grain, i'd have to do [this](https://github.com/riscv-software-src/riscv-isa-sim/issues/814) and then rebuild spike. Please correct me if I'm wrong. Thanks in advance!

iamkarthikbk:
Resolves #1478

### DIFF
--- a/spike_main/spike.cc
+++ b/spike_main/spike.cc
@@ -41,6 +41,7 @@ static void help(int exit_code = 1)
   fprintf(stderr, "  --debug-cmd=<name>    Read commands from file (use with -d)\n");
   fprintf(stderr, "  --isa=<name>          RISC-V ISA string [default %s]\n", DEFAULT_ISA);
   fprintf(stderr, "  --pmpregions=<n>      Number of PMP regions [default 16]\n");
+  fprintf(stderr, "  --pmpgranularity=<n>  PMP Granularity in bytes [default 4]\n");
   fprintf(stderr, "  --priv=<m|mu|msu>     RISC-V privilege modes supported [default %s]\n", DEFAULT_PRIV);
   fprintf(stderr, "  --varch=<name>        RISC-V Vector uArch string [default %s]\n", DEFAULT_VARCH);
   fprintf(stderr, "  --pc=<address>        Override ELF entry point\n");
@@ -366,6 +367,7 @@ int main(int argc, char** argv)
             /*default_misaligned=*/false,
             /*default_endianness*/endianness_little,
             /*default_pmpregions=*/16,
+            /*default_pmpgranularity=*/(1 << PMP_SHIFT),
             /*default_mem_layout=*/parse_mem_layout("2048"),
             /*default_hartids=*/std::vector<size_t>(),
             /*default_real_time_clint=*/false,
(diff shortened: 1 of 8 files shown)


jerryz123:
To support bisectable commit history, this project requires that each commit in a PR pass the tests.
This project also requires that PRs be up-to-date with the master branch before merge, so you should rebase your changes on top of master, rather than merge master into your feature branch.

iamkarthikbk:
Hi @jerryz123 . Now there's a single commit that should work. How should I trigger the CI workflows ?

iamkarthikbk:
Can you help me understand why the test failed? I don't think I understand it well enough to find what's broken and fix it.

spike_main/spike.cc
```diff
@@ -41,6 +41,7 @@ static void help(int exit_code = 1)
   fprintf(stderr, "  --debug-cmd=<name>    Read commands from file (use with -d)\n");
   fprintf(stderr, "  --isa=<name>          RISC-V ISA string [default %s]\n", DEFAULT_ISA);
   fprintf(stderr, "  --pmpregions=<n>      Number of PMP regions [default 16]\n");
+  fprintf(stderr, "  --pmpgranularity=<n>  PMP Granularity [default 0]\n");
```
scottj97: What are the units here? Bytes?

0 does not make sense as a granularity. In RISC-V, 4 bytes is the minimum PMP granularity.

scottj97: Upon closer reading of the code, it seems this is the **G** value described in the PMP spec. But that's not the PMP granularity! Note this line from a footnote in the spec:

> PMP granularity is 2**(G+2) bytes

iamkarthikbk: Spike already implements PMPs as per the riscv priv spec. [This](https://github.com/riscv-software-src/riscv-isa-sim/blob/5a1145742e701597eb45825855311dfad21232a6/riscv/processor.cc#L66) line sets the sets the default granularity. 4 is the minimum PMP granularity, and that is taken care of, by the macro `PMP_SHIFT`. In order to set a coarser grain, we would have to increment the value of that shift on the LSB side of pmpaddr registers. This CLI arg should contain the value to be added to `PMP_SHIFT` while setting the grain.

this line makes that change:
```
set_pmp_granularity(1 << (PMP_SHIFT+cfg->pmpgranularity));
```

If the granularity is to remain `4B`, then `0` should be present in the value `cfg->pmpgranularity`.
Please correct me if im wrong.

iamkarthikbk: Maybe the help message should have been:
```
fprintf(stderr, "  --pmpgranularity=<n>  PMP Granularity [default 0 (4B)]\n");
```
instead of:
```
fprintf(stderr, "  --pmpgranularity=<n>  PMP Granularity [default 0]\n");
```
Please let me know if this is correct, or if the whole method of setting the granularity is wrong in this PR.

scottj97: You are wrong because you are conflating **G** with PMP granularity. They are not synonymous. The PMP granularity is 2**(G+2) bytes which has a minimum of 4, not 0.

Several options here:
* Rename the command line argument to "PMP G" (which is not very self-explanatory except to people who have the PMP spec fresh in their minds)
* Calculate G from the byte value given on the command line
* Directly pass the byte value given on the command line to `set_pmp_granularity()`, which already has legality checks

The last option seems cleanest and easiest.

iamkarthikbk: You're right. Directly accepting the granularity in bytes seems like a much more straightforward approach. Thanks for pointing it out. I've made the changes like you suggested.

iamkarthikbk:
Hi @scottj97 . Could you please review the changes now ?

scottj97 (approved):

aswaterman (approved):

Issue #1495, opened 2023-11-07
SC Instruction Returns Incorrect Value on Trap.


youzi27:
### Description
According to the RISC-V ISA specification for the A extension, the `SC` (Store Conditional ) instruction requires that the address in `rs1` be naturally aligned to the size of the operand. This means that for a 64-bit `SC.D` instruction, the address must be 8-byte aligned. If the address is not naturally aligned, an address misaligned exception or an access-fault exception should be raised.

However, during testing with the Spike simulator, it has been observed that when the SC.D instruction is executed with a misaligned address in rs1, an exception is raised. Additionally, the RD register does not accurately reflect the success or failure of the SC execution. For instance, if the RD register is initially set to 0, it remains 0 even when the SC instruction encounters a trap, suggesting successful execution, which contradicts the expected behavior.

### Steps to Reproduce
1. Execute an `SC` instruction with a misaligned address(or any value that causes the SC to trap) in `rs1` in the Spike simulator.
2. Observe the value written to the destination register (should be non-zero to indicate failure, according to the specification).

According to section 8.2 of the RISC-V Unprivileged ISA Specification: If the SC.W fails, the instruction does not write to memory, and it **writes a nonzero value to rd**. 
This behavior is mandated to indicate the failure of the store conditional operation.

### Expected Behavior
The `SC` instruction should not succeed when given a misaligned address and should return a non-zero value in the destination register to indicate failure, in accordance with the specification.

### Actual Behavior
When executing sc, due to traps occurring internally, such as misaligned addresses, a non-zero value will not be written to the rd register, implying that the sc execution has failed. Especially when the rd register is originally 0, it is easy to mislead that the execution has been successful.

### Additional Information
- This behavior has been consistently observed on the latest versions of the Spike simulator[v1.1.1].
- The expected behavior is outlined in the RISC-V ISA specification, section8.2.

It appears that there may be an issue with how the misaligned address exception is being handled, or how the return value of the `SC` instruction is being set under these exceptional circumstances.

### Proposed Solution
When executing SC, the RD register should be set to '!false' before performing MMU operations.

I am looking forward to your feedback on this matter.

Thank you for your attention to this issue.

youzi27:
I have submitted a Pull Request that addresses this issue: #1496. Please take a moment to review it. Thank you!

PR #1496, opened 2023-11-07
SC Instruction Returns Incorrect Value on Trap.

youzi27:
_See issue #1495._
### Description
According to the RISC-V ISA specification for the A extension, the SC (Store Conditional ) instruction requires that the address in rs1 be naturally aligned to the size of the operand. This means that for a 64-bit SC.D instruction, the address must be 8-byte aligned. If the address is not naturally aligned, an address misaligned exception or an access-fault exception should be raised.

However, during testing with the Spike simulator, it has been observed that when the SC.D instruction is executed with a misaligned address in rs1, an exception is raised. Additionally, the RD register does not accurately reflect the success or failure of the SC execution. For instance, if the RD register is initially set to 0, it remains 0 even when the SC instruction encounters a trap, suggesting successful execution, which contradicts the expected behavior.

### Expected Behavior
The SC instruction should not succeed when given a misaligned address and should return a non-zero value in the destination register to indicate failure, in accordance with the specification.

### Actual Behavior
When executing sc, due to traps occurring internally, such as misaligned addresses, a non-zero value will not be written to the rd register, implying that the sc execution has failed. Especially when the rd register is originally 0, it is easy to mislead that the execution has been successful.

### Proposed Solution
When executing SC, the RD register should be set to '!false' before performing MMU operations.

### DIFF
--- a/riscv/insns/sc_d.h
+++ b/riscv/insns/sc_d.h
@@ -1,6 +1,8 @@
 require_extension('A');
 require_rv64;
 
+WRITE_RD(!false);
+
 bool have_reservation = MMU.store_conditional<uint64_t>(RS1, RS2);
 
 WRITE_RD(!have_reservation);
--- a/riscv/insns/sc_w.h
+++ b/riscv/insns/sc_w.h
@@ -1,5 +1,7 @@
 require_extension('A');
 
+WRITE_RD(!false);
+
 bool have_reservation = MMU.store_conditional<uint32_t>(RS1, RS2);
 
 WRITE_RD(!have_reservation);


youzi27:
According to section 8.2 of the RISC-V Unprivileged ISA Specification: If the SC.W fails, the instruction does not write to memory, and it writes a nonzero value to rd.
This behavior is mandated to indicate the failure of the store conditional operation.

scottj97 (changes requested):
An instruction that takes an exception never writes to its destination register or memory. In effect, the instruction never executed, so the SC never had a chance to succeed or fail. The fact of the exception is sufficient to show that the SC never completed.

scottj97:
Also, for next time, please learn about [writing good commit messages](https://cbea.ms/git-commit/).

youzi27:
> Also, for next time, please learn about [writing good commit messages](https://cbea.ms/git-commit/).

I see, thank you very much for your explanation.

PR #1498, opened 2023-11-09
fesvr/htif: allow exit on SIGINT.

f0rget-the-sad:
Currently signal handler would call exit() only on second received signal, this prevents proper program cleanup.
Instead use signal flag to exit loop.


aswaterman (approved):
Thanks, this looks like an oversight, given we are checking `signal_exit` in the loop condition at https://github.com/riscv-software-src/riscv-isa-sim/blob/f04be9849356cf013f4071429f6fd152a8e5e518/fesvr/htif.cc#L260

PR #1500, opened 2023-11-10
Update debug smoketest action.

timsifive:
To get https://github.com/riscv-software-src/riscv-tests/pull/522, which fixes an intermittent failure.


aswaterman (approved):
Preemptive approval.

PR #1501, opened 2023-11-13
Add Ssdtso support

cmuellner:
The second draft of the Ssdtso extension was published recently:
  https://lists.riscv.org/g/tech-arch-review/message/183
This PR adds support for this new extension.

Note, that the Ssdtso specification is in development state (i.e., not frozen or even ratified).


aswaterman (commented):
Let's wait til the riscv-opcodes PR is merged, but LGTM otherwise.

Issue #1502, opened 2023-11-14
[Feature Request] Remove the limitation of VLEN <= 4K


sequencer:
We are designing [t1](https://github.com/chipsalliance/t1), which VLEN can be configured to be larger than 4K, in RISC-V Vector specification, The VLEN can reach to 64K:
https://github.com/riscv/riscv-v-spec/blob/fc76ec73fc1dd4531360ee8f3138f79f02e8b1b0/v-spec.adoc?plain=1#L67-L72
However riscv-isa-sim has this special limitation: 
https://github.com/riscv-software-src/riscv-isa-sim/blob/be5dee0bafb413c9ac8845ca144db9b7641941b2/riscv/processor.cc#L175-L177
Thus spike cannot emulate the RISC-V Vector behavior which VLEN is larger than 4K.

We wanna request spike to remove this limitation.

luojia65:
I'd like to see this requirement relaxed to at least 65,536 bytes for that we need to compare [rustsbi/zemu](https://gitee.com/rustsbi/zemu) to an existing RISC-V software emulation model to ensure security critical firmware implementations work as expected.

aswaterman:
Of course, I have no objection with lifting this restriction, but you will need to contribute it (after testing it thoroughly).

PR #1503, opened 2023-11-14
Relax VLEN requirement to 65,536 bytes in processor module

luojia65:
Related issue: https://github.com/riscv-software-src/riscv-isa-sim/issues/1502

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -172,9 +172,9 @@ void processor_t::parse_varch_string(const char* s)
   if (vlen < elen)
     bad_varch_string(s, "vlen must be >= elen");
 
-  /* spike requirements. */
-  if (vlen > 4096)
-    bad_varch_string(s, "vlen must be <= 4096");
+  /* Vector spec requirements. */
+  if (vlen > 65536)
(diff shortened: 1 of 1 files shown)


sequencer:
> (after testing it thoroughly)

How did you test for it?

luojia65:
I'm sorry that I don't have a test environment locally, it would be somehow hard to test. If there are more test cases other than CI, please note me and I'll try to run them locally. By now I'd wait before CI result.

sequencer:
Basically the community has
https://github.com/ksco/riscv-vector-tests
For test case generation
You can run it to check there are no internal exceptions from spike.

Since for riscv community, spike is the golden model that doesn’t have a
diamond model which can be used for testing, neither do I have any idea to
verify the correctness.

However testing spike against t1 for cross reference(we already have a
corresponding test framework) is a possible solution, if you are willing
to, you can send PR to t1, and our engineers will help you debugging for
any mismatches on both spike and t1.

aswaterman:
Yeah, I can’t accept this until I hear a more solid testing story.

PR #1511, opened 2023-11-23
stimecmp: perform menvcfg.STCE permission check when accessing vstimecmp in HS-mode

YenHaoChen:
The spec requires menvcfg.STCE=1 on accessing stimecmp or vstimecmp in a mode other than M-mode. The previous implementation does not check the permission on accessing vstimecmp in HS-mode. This commit fixes the issue by moveing the permission check from virtualized_stimecmp_csr_t to stimecmp_csr_t, which implements the vstimecmp.
image


aswaterman:
@i2h2 please review this PR.

i2h2:
I think it looks good. Thanks for fixing!

aswaterman (approved):
Approving on the basis of @i2h2's review.

PR #1512, opened 2023-11-27
fix: vector crypto: VWSLL.VI and VWSLL.VX do not use vs1

YenHaoChen:
The vector crypto instructions VWSLL.VI and VWSLL.VX do not use vs1. Prvious implementation checks as using vs1.


YenHaoChen:
A previous PR addresses this issue.
https://github.com/riscv-software-src/riscv-isa-sim/pull/1506

PR #1514, opened 2023-11-28
remote_bitbang: make send_buf class member

f0rget-the-sad:
Currently send buffer is static variable in function, which makes it's impossible to have multiple concurrent instances of remote bitbang class, since all of them would share this static buffer. Thus, make send_buf a class member.


timsifive (approved):
Looks reasonable. It seems likely there are other parts of the debug code that don't work right when you have multiple remote bitbang connections. At least, it's not a use case I ever considered when doing the spike implementation.

aswaterman (approved):
Yeah, this is definitely an improvement.

PR #1515, opened 2023-11-29
Move isa_parser_t to libdisasm out of libriscv (cont.)

YenHaoChen:
A previous commit [1] moves isa_parser.cc from riscv/ to disasm/. This commit proposes putting the header along with the source and moving isa_parser.h from riscv/ to disasm/.

This commit is a refactoring attempt. Although I’m not sure if it’s desirable, I believe that combining header and source files in the same directory could make the file structure more apparent.

[1] https://github.com/riscv-software-src/riscv-isa-sim/commit/47ab8926b04a592128b924da0bdbd622213568f8


YenHaoChen:
CI seems unhappy about this. I guess that is because the isa_parser.h isn't part of the global header.

Issue #1518, opened 2023-12-02
The BASE field of mtvec appears to lack alignment checking for 4-byte alignment.


youzi27:
When I attempted to write a value to _mtvec_, the BASE field of _mtvec_ does not check for **4-byte alignment.** In other words, the lowest bit of the BASE field can be binary 1.

Privilege Specification 3.1.7 states: 

> The value in the BASE field must always be aligned on a 4-byte boundary.

How to reproduce:
```
...
addi    s11, s11, -64
ori x11, x11, 0xff
csrw mtvec, x11 
...
```
The log in Spike displays:
```
core   0: 0x0000000080000044 (0xfc0d8d93) addi    s11, s11, -64
(spike) 
core   0: 0x0000000080000048 (0x0ffded93) ori     s11, s11, 255
(spike) 
core   0: 0x000000008000004c (0x305d9073) csrw    mtvec, s11
(spike) reg 0 mtvec
0x000000008005f0fd
```
If I am mistaken in any way, I appreciate any corrections. Thank you.

scottj97:
See explanation in #1519.

youzi27:
The BASE area is MXLEN-2 (from bit 63 to bit 2, where bits 1 to 0 are the MODE field, unrelated to BASE, in 64-bit), not BASE+MODE. The four-byte alignment you mentioned includes the MODE field, right? But BASE and MODE are calculated separately. Besides, MODE can be set to 1 (it's just not supported by default in spike), but according to your method, MODE can only be 0, which is obviously incorrect.
If I have any mistakes, I hope you can point them out.

youzi27:
> See explanation in #1519.

In other words, the mtvec register consists of two completely different functional segments, which are not directly related to each other. When calculating BASE, the MODE bits should not be considered. That is, BASE is the higher part of mtvec (bits 2-63, in a 64-bit system).

PR #1519, opened 2023-12-03
fix:Align the BASE field of the tvec register to 4-byte boundaries.

youzi27:
See issue #1518 

image
image

The specification requires the BASE field of the tvec register to be aligned to four-byte boundaries, but in practice, it is possible to write a non-four-byte aligned address.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -374,7 +374,8 @@ reg_t tvec_csr_t::read() const noexcept {
 }
 
 bool tvec_csr_t::unlogged_write(const reg_t val) noexcept {
-  this->val = val & ~(reg_t)2;
+  //Preserve the MODE field and adhere to a 4-byte aligned BASE field.
+  this->val = val & ~(reg_t)0xe;
   return true;
 }
 


scottj97:
Look at the diagram. Only bits `[MXLEN-1:2]` are stored. The address is naturally 4-byte aligned because there is no storage for bits 1 or 0.

This PR is proposing to zero out bits 3 and 2, forcing 16-byte alignment.

youzi27:
> Look at the diagram. Only bits `[MXLEN-1:2]` are stored. The address is naturally 4-byte aligned because there is no storage for bits 1 or 0.
> 
> 
> 
> This PR is proposing to zero out bits 3 and 2, forcing 16-byte alignment.
> 
> 

The BASE area is MXLEN-2 (from bit 63 to bit 2, where bits 1 to 0 are the MODE field, unrelated to BASE, in 64-bit), not BASE+MODE. The four-byte alignment you mentioned includes the MODE field.
The specification states that the pc is equal to the value of BASE, and I believe it seems to refer to bits 63-2 of the mtvec.
If I have any mistakes, I hope you can point them out.

scottj97:
> The BASE area is MXLEN-2 (from bit 63 to bit 2, where bits 1 to 0 are the MODE field, unrelated to BASE, in 64-bit), not BASE+MODE. The four-byte alignment you mentioned includes the MODE field. The specification states that the pc is equal to the value of BASE, and I believe it seems to refer to bits 63-2 of the mtvec. If I have any mistakes, I hope you can point them out.

Look at the diagram. The bits of `mtvec` are numbered above the rectangle. The bits from `MXLEN-1` down to bit `2` contain the value `BASE[MXLEN-1:2]` (shown inside the rectangle). If this were the entire BASE value, it would simply read `BASE`, or perhaps `BASE[MXLEN-3:0]`. But it's not the entire BASE. It's a subset of the entire BASE value.

On an exception, the PC is set to `BASE[MXLEN-1:0]`. Bits 1 and 0 are not here in this CSR. Where are they?

We know, since it is specified that BASE must be aligned to 4 bytes, that bits 1 and 0 must be 0. Hence, for 64 bit (in Verilog syntax):

```
bit [63:0] BASE = {mtvec[63:2], 2'b00};
```

youzi27:
Hi @scottj97
Thank you for the information, but after reviewing the source code, I've found that Spike's implementation of this still has issues. 

```
-     // Handle the trap in M-mode
-     const reg_t vector = (state.mtvec->read() & 1) && interrupt ? 4 * bit : 0;
-     const reg_t trap_handler_address = (state.mtvec->read() & ~(reg_t)1) + vector;
```

Should I resubmit a PR (Pull Request) or open a new issue for this?

youzi27:
> > The BASE area is MXLEN-2 (from bit 63 to bit 2, where bits 1 to 0 are the MODE field, unrelated to BASE, in 64-bit), not BASE+MODE. The four-byte alignment you mentioned includes the MODE field. The specification states that the pc is equal to the value of BASE, and I believe it seems to refer to bits 63-2 of the mtvec. If I have any mistakes, I hope you can point them out.
> 
> Look at the diagram. The bits of `mtvec` are numbered above the rectangle. The bits from `MXLEN-1` down to bit `2` contain the value `BASE[MXLEN-1:2]` (shown inside the rectangle). If this were the entire BASE value, it would simply read `BASE`, or perhaps `BASE[MXLEN-3:0]`. But it's not the entire BASE. It's a subset of the entire BASE value.
> 
> On an exception, the PC is set to `BASE[MXLEN-1:0]`. Bits 1 and 0 are not here in this CSR. Where are they?
> 
> We know, since it is specified that BASE must be aligned to 4 bytes, that bits 1 and 0 must be 0. Hence, for 64 bit (in Verilog syntax):
> 
> ```
> bit [63:0] BASE = {mtvec[63:2], 2'b00};
> ```

Hi @scottj97
Thank you for the information, but after reviewing the source code, I've found that Spike's implementation of this still has issues.

```
-     // Handle the trap in M-mode
-     const reg_t vector = (state.mtvec->read() & 1) && interrupt ? 4 * bit : 0;
-     const reg_t trap_handler_address = (state.mtvec->read() & ~(reg_t)1) + vector; ***
-     // RNMI exception vector is implementation-defined.  Since we don't model
-     // RNMI sources, the feature isn't very useful, so pick an invalid address.
-     const reg_t rnmi_trap_handler_address = 0;
-     const bool nmie = !(state.mnstatus && !get_field(state.mnstatus->read(), MNSTATUS_NMIE));
-     state.pc = !nmie ? rnmi_trap_handler_address : trap_handler_address;

```
From the marked *** section, it is evident that it only ensures 2-byte alignment, which does not conform to the specification.

Should I resubmit a PR (Pull Request) or open a new issue for this?

aswaterman:
> From the marked *** section, it is evident that it only ensures 2-byte alignment, which does not conform to the specification.

Not exactly.  The `unlogged_write` implementation clears bit 1, and the line you quoted clears bit 0, so collectively, 4-byte alignment is enforced.

youzi27:
> > From the marked *** section, it is evident that it only ensures 2-byte alignment, which does not conform to the specification.
> 
> Not exactly. The `unlogged_write` implementation clears bit 1, and the line you quoted clears bit 0, so collectively, 4-byte alignment is enforced.

Great, thank you for your patient answers, much appreciated!

aswaterman:
You're welcome!

PR #1520, opened 2023-12-03
Add blockdevice 

joonho3020:
Adds blockdevice support for spike.
Can boot Linux with `*.img` files now.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+
+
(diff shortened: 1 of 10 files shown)


jerryz123:
I believe the block device was originally removed from Spike to minimize the scope of this project. 

With the new device library APIs, it should be possible to implement the block device as a separate library, that can be dynamically loaded into spike with the right command line flags.

PR #1522, opened 2023-12-04
Fix Spike --device option to pass on args to downstream plugins

joonho3020:
Relate to this PR https://github.com/ucb-bar/spike-devices/pull/2.
Basically, as the MMIO devices are suppose to be implemented as plugins, we need a way to pass down arguments to the plugins from the top.
Hence, we change the `--device` option to specify the name of the device along with the arguments.
E.g., `--device="mydevice,arg1=hello"`.

### DIFF
--- a/riscv/abstract_device.h
+++ b/riscv/abstract_device.h
@@ -8,6 +8,7 @@
 #include <string>
 #include <map>
 #include <stdexcept>
+#include <vector>
 
 class sim_t;
 
@@ -26,10 +27,15 @@ public:
   virtual abstract_device_t* parse_from_fdt(const void* fdt, const sim_t* sim, reg_t* base) const = 0;
   virtual std::string generate_dts(const sim_t* sim) const = 0;
   virtual ~device_factory_t() {}
+  void set_sargs(std::vector<std::string> sargs) { this->sargs = sargs; }
+  std::vector<std::string> get_sargs() { return sargs; }
+
+protected:
+  std::vector<std::string> sargs;
 };
 
 // Type for holding all registered MMIO plugins by name.
-using mmio_device_map_t = std::map<std::string, const device_factory_t*>;
+using mmio_device_map_t = std::map<std::string, device_factory_t*>;
 
 mmio_device_map_t& mmio_device_map();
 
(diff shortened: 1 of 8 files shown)


riscv/ns16550.cc
```diff
@@ -341,7 +341,7 @@ std::string ns16550_generate_dts(const sim_t* sim)
   return s.str();
 }
 
-ns16550_t* ns16550_parse_from_fdt(const void* fdt, const sim_t* sim, reg_t* base)
+ns16550_t* ns16550_parse_from_fdt(const void* fdt, const sim_t* sim, reg_t* base, std::vector<std::string> sargs)
```
scottj97: Pass by const ref, instead of passing by value, which will make an unnecessary copy of all the strings.

riscv/sim.cc
```diff
@@ -38,7 +38,7 @@ extern device_factory_t* ns16550_factory;
 
 sim_t::sim_t(const cfg_t *cfg, bool halted,
              std::vector<std::pair<reg_t, abstract_mem_t*>> mems,
-             std::vector<const device_factory_t*> plugin_device_factories,
+             std::vector<device_factory_t*> plugin_device_factories,
```
scottj97, line 41: It's unfortunate to have to remove `const` here. Isn't there a way you could pass the args into the factories' constructors? So we can keep `const` here?

scottj97 (changes requested):
The rationale for this PR needs to be explained here. Pointing to a PR in another project -- especially one that also has no documented rationale! -- is insufficient.

scottj97:
Can you provide a realistic example of an MMIO device that needs arguments passed in via cmdline?

riscv/abstract_device.h
```diff
@@ -26,10 +27,12 @@ class device_factory_t {
   virtual abstract_device_t* parse_from_fdt(const void* fdt, const sim_t* sim, reg_t* base) const = 0;
   virtual std::string generate_dts(const sim_t* sim) const = 0;
   virtual ~device_factory_t() {}
+  void set_sargs(std::vector<std::string> sargs) { _sargs = sargs; }
+  std::vector<std::string> _sargs;
```
jerryz123: The `_private_var` is a Pythonism. You should just declare this as private.

riscv/sim.cc
jerryz123: I don't think so, for dynamically loaded extension libraries, the factories are constructed by the loading mechanism

joonho3020:
> Can you provide a realistic example of an MMIO device that needs arguments passed in via cmdline?

The ucb-bar block device is implemented as an MMIO device.
We would like to pass an argument to the block device such that it can read a file and initialize its contents at startup.
However, with the current spike's top level arguments this is impossible.

scottj97:
Please squash the commits together. We don't want to have commits that fix issues from an earlier commit in the same PR.

scottj97 (approved):

aswaterman (approved):

aswaterman:
Because this changes the API, the CI tests that are programmed against that API need to change, too.  The failing one is https://github.com/riscv-software-src/riscv-isa-sim/blob/d94fe56b4fbc000d24f68597eb32e13cb73802ef/ci-tests/testlib.c

jerryz123 (approved):

PR #1523, opened 2023-12-06
miselect: support miselect when enabling smcsrind

YenHaoChen:
This commit adds the missing miselect CSR.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -543,6 +543,9 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   sscsrind_reg_csr_t::sscsrind_reg_csr_t_p vsireg[6];
 
   if (proc->extension_enabled_const(EXT_SMCSRIND)) {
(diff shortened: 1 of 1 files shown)


riscv/processor.cc
```diff
@@ -543,6 +543,9 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   sscsrind_reg_csr_t::sscsrind_reg_csr_t_p vsireg[6];
 
   if (proc->extension_enabled_const(EXT_SMCSRIND)) {
+    miselect = std::make_shared<basic_csr_t>(proc, CSR_MISELECT, 0);
```
aswaterman: Can you move the declaration of the `miselect` variable into this scope, making the declration and initialization a single statement?  (And also make the analogous improvement for `siselect` and `vsiselect`.)

YenHaoChen: Add a new commit that combines the declaration and initialization of miselect into a single statement  (as well as siselect and vsiselect).

aswaterman (approved):
Thanks! LGTM.

PR #1526, opened 2023-12-07
Add default cfg_t and debug_module_config_t constructor to libriscv

jerryz123:
Adding additional config options to cfg_t has, up to now, required propagating the flag to all code which instantiates cfg_t, including projects which use Spike as a library.

Now, the default constructor for `cfg_t` will construct the default spike parameters.

This should be merged after #1522 to avoid merge conflicts.

### DIFF
--- a/ci-tests/testlib.c
+++ b/ci-tests/testlib.c
@@ -14,21 +14,7 @@ static std::vector<std::pair<reg_t, abstract_mem_t*>> make_mems(const std::vecto
 
 int main()
 {
-  std::vector<mem_cfg_t> mem_cfg { mem_cfg_t(0x80000000, 0x10000000) };
-  std::vector<size_t> hartids = {0};
-  cfg_t cfg(std::make_pair(0, 0),
-            nullptr,
-            "rv64gcv",
-            "MSU",
-            "vlen:128,elen:64",
-            false,
-            endianness_little,
-            16,
-            (1 << PMP_SHIFT),
-            mem_cfg,
-            hartids,
-            false,
-            4);
+  cfg_t cfg;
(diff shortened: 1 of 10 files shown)


aswaterman:
I didn't look at the code, but the rationale checks out.  Probably worth pinging this thread after #1522 is merged.

scottj97:
I'm not sure I understand the removal of `cfg_arg_t`. (There's only one place left that uses it after this change.)

@rswarbrick introduced this in #946 to keep track of which config vars had been overridden from their initial values (e.g. via command-line). Do we not need that capability anymore?

jerryz123:
> @rswarbrick introduced this in https://github.com/riscv-software-src/riscv-isa-sim/pull/946 to keep track of which config vars had been overridden from their initial values (e.g. via command-line). Do we not need that capability anymore?

That capability was never used once the `cfg_t` was passed into `sim_t`. I believe some other code in spike main uses the `cfg_arg_t` ... the `explicit_hartids` stuff.

My understanding is that the `overridden` API in `cfg_t` was added to eventually support a use case where the user provides cmdline args, and a DTS/DTB, and spike would implement some code to check that the DTS/DTB is consistent with the command line args (https://github.com/riscv-software-src/riscv-isa-sim/pull/938). However, none of the consistency checking has been implemented. I'm not actually sure this "consistency check" approach makes sense... do we need to support the use case of providing both DTB and cmdline args which affect the same system configurations?

jerryz123:
@aswaterman @scottj97 This is ready now.

aswaterman:
Still OK with me; let's wait for @scottj97

scottj97:
I don’t care

aswaterman (approved):

PR #1529, opened 2023-12-08
Add unratified Zimop extension

ved-rivos:
Specification: https://github.com/riscv/riscv-isa-manual/blob/main/src/zimop.adoc
Sail ref: https://github.com/riscv/sail-riscv/pull/321
Arch tests: https://github.com/riscv-non-isa/riscv-arch-test/pull/387

PR #1531, opened 2023-12-08
Add unratified Zimop extension

aswaterman:
This PR just reorganizes the commits in #1529.


aswaterman:
Merging on the basis of the approval of #1529

PR #1546, opened 2023-12-21
UB in negate in mulh/mulhsu

nibrunieAtSi5:
follow-up to https://github.com/riscv-software-src/riscv-isa-sim/pull/1539.

Ensuring No negation on -2^63(int64_t) is performed during mulh/mulhsu

### DIFF
--- a/riscv/arith.h
+++ b/riscv/arith.h
@@ -32,15 +32,15 @@ inline uint64_t mulhu(uint64_t a, uint64_t b)
 inline int64_t mulh(int64_t a, int64_t b)
 {
   int negate = (a < 0) != (b < 0);
-  uint64_t res = mulhu(a < 0 ? -a : a, b < 0 ? -b : b);
+  uint64_t res = mulhu(a < 0 ? -(uint64_t)a : a, b < 0 ? -(uint64_t)b : b);
   return negate ? ~res + ((uint64_t)a * (uint64_t)b == 0) : res;
 }
 
 inline int64_t mulhsu(int64_t a, uint64_t b)
 {
   int negate = a < 0;
-  uint64_t res = mulhu(a < 0 ? -a : a, b);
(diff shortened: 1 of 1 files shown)


riscv/arith.h
```diff
@@ -32,14 +32,14 @@ inline uint64_t mulhu(uint64_t a, uint64_t b)
...
+  uint64_t res = mulhu(a < 0 ? -(uint64_t)a : a, b < 0 ? -(uint64_t)b : b);
   return negate ? ~res + ((uint64_t)a * (uint64_t)b == 0) : res;
 }
 
 inline int64_t mulhsu(int64_t a, uint64_t b)
 {
   int negate = a < 0;
-  uint64_t res = mulhu(a < 0 ? -a : a, b);
+  uint64_t res = mulhu(a < 0 ? -(uint64_t)a : a, b);
   return negate ? ~res + (a * b == 0) : res;
```
nibrunieAtSi5: Maybe a cast is required here too
```suggestion
  return negate ? ~res + ((uint64_t)a * b == 0) : res;
```

aswaterman: I think that’s right.

aswaterman (approved):
Ping me again when it’s ready to merge.

aswaterman:
@nibrunieAtSi5 Returning to this eight months later, can you rebase onto master and then squash the commits together?

nibrunie:
@jerryz123 could you review this PR since you reviewed https://github.com/riscv-software-src/riscv-isa-sim/pull/1539 ?

compiler link to reproduce the possible UB before this PR: https://godbolt.org/z/bn9xh3G1s (using `-fsanitize=undefined`)
issue disappear after the change: https://godbolt.org/z/svWMn9bn5

jerryz123 (approved):
I think you need to change it from Draft to Ready (I can't do this from my phone). 

Edit: nvm

PR #1547, opened 2023-12-22
typo: correct sstateen CSR address

YenHaoChen:
This commit corrects the CSR addresses of sstateen.


aswaterman (approved):

scottj97:
Let me point out that the CSR address passed to the constructor affects only permission checks on the CSR, not its address as seen by `csrr` etc. instructions. (That comes from the index into `csrmap`, which was already correct.)

PR #1552, opened 2023-12-30
Add Ssqosid extension

ved-rivos:
1. Refresh encoding.h to add `mstateen0/mstateen0h.PRIV114` field, `srmcfg` CSR fields (RCID/RMID) and address
2. Add `srmcfg` CSR introduced by Ssqosid extension

Sail: https://github.com/riscv/sail-riscv/pull/379
Arch-tests: https://github.com/riscv-non-isa/riscv-arch-test/pull/420
Spec: https://github.com/riscv/riscv-ssqosid


aswaterman (approved):
LGTM.

PR #1559, opened 2024-01-07
Add B extension

ved-rivos:


aswaterman (changes requested):
IMO, it would be cleaner and more consistent to mirror the pattern of C/Zca for B/{Zba, Zbb, Zbs}.  See `misa_csr_t::unlogged_write`, where we toggle the availability of Zca when misa.C is toggled. (And if C is unavailable in the ISA string but Zca is available in the ISA string, Zca remains available unconditionally.)

Then, we shouldn't need any changes to the individual instructions.

ved-rivos:
> IMO, it would be cleaner and more consistent to mirror the pattern of C/Zca for B/{Zba, Zbb, Zbs}. 

Agree. Updating to match that.

aswaterman (approved):
LGTM.  Thanks!

PR #1561, opened 2024-01-09
Fix order of H extension in ISA string

demin-han:
According to Subset Naming Convention section of ISA Manual Volume I, H extension should be after V extension in ISA string.


aswaterman (approved):
Although I expect those requirements to be deprecated or otherwise superseded, this change is fine for the time being.

PR #1563, opened 2024-01-11
Refactor put_csr to direct write

demin-han:
1. put_csr needs search
2. MSTATUS_MPV not written back for RV32

### DIFF
--- a/riscv/insns/mret.h
+++ b/riscv/insns/mret.h
@@ -9,5 +9,5 @@ s = set_field(s, MSTATUS_MIE, get_field(s, MSTATUS_MPIE));
 s = set_field(s, MSTATUS_MPIE, 1);
 s = set_field(s, MSTATUS_MPP, p->extension_enabled('U') ? PRV_U : PRV_M);
 s = set_field(s, MSTATUS_MPV, 0);
-p->put_csr(CSR_MSTATUS, s);
+STATE.mstatus->write(s);
 p->set_privilege(prev_prv, prev_virt);


scottj97 (changes requested):
I agree this is an improvement -- no reason to require a lookup into `csrmap` -- but I don't understand the part about "MSTATUS_MPV not written back for RV32".

Is there a bug that this fixes? If so we need a lot more explanation of the case.

I can't see how this change would affect MPV on RV32, but maybe I'm missing something.

demin-han:
@scottj97 
MPV located in mstatush when RV32. 

if (xlen == 32) {
    csrmap[CSR_MSTATUS] = std::make_shared<rv32_low_csr_t>(proc, CSR_MSTATUS, mstatus);
    csrmap[CSR_MSTATUSH] = mstatush = std::make_shared<rv32_high_csr_t>(proc, CSR_MSTATUSH, mstatus);
  } else {
    csrmap[CSR_MSTATUS] = mstatus;
  }

CSR_MSTUTUS is associated with lower part of mstatus when RV32.
so put_csr(CSR_MSTATUS, s) only update lower part of mstatus, MPV in mstatush not updated.

scottj97 (approved):
Yes. I see the problem now.

I think RV32 Hypervisor has seen very limited usage (or none at all) because I suspect this has been broken all along, since MPV was first added.

I've confirmed that this PR doesn't change any RV64 Hypervisor behavior, but I don't have any test coverage of RV32 Hypervisor.

aswaterman (approved):
Ah, that's subtle.

PR #1564, opened 2024-01-12
Add missing log commit of mstatush

demin-han:


scottj97 (approved):
This matches what we do in [take_trap](https://github.com/riscv-software-src/riscv-isa-sim/blob/f80fc52e8842d3a5270b787a0fc57028edca5dab/riscv/processor.cc#L938).

It kinda sucks, but...?

aswaterman:
It's a little unfortunate, but if this is the extent of the use of this pattern, I guess it's OK.

Here is an alternative pattern we could consider in the future: make `mstatus` a 64-bit register for logging purposes, even for RV32.  Writes to `mstatush` would never show up in the log (i.e. a write to `mstatush` would produce a log entry for `mstatus`).  This would also match how the privileged architecture spec describes the relationship between the two registers.

PR #1566, opened 2024-01-13
Prototype of Keccak multi-round instruction

nibrunieAtSi5:
Implementing prototype of multi-round Keccak instruction designed by https://github.com/mjosaarinen 

Can be tested/used


nibrunieAtSi5:
Not the correct base repo ...

PR #1570, opened 2024-01-18
vcpop.m, vfirst.m: delete unnecessary VSTART write.

h-cheng:
For vector mask intructions like vcpop.m, vfirst.m, vmsbf.m, vmsif.m, vmsof.m, viota.m, VSTART should be zero and it is checked in Spike. So it is optional to reset VSTART at the end of these instructions.
However, there is kind of inconsistency in Spike that VSTART is reset for vcpop.m and vfirst.m, but not for vmsbf.m, vmsif.m, vmsof.m, viota.m. I would suggest doing the same thing for them as well.
One additional benefit for this change is making the commit logs of all vector instructions contain a "c8_vstart" segment, which is easy to parse during verification.

### DIFF
--- a/riscv/insns/shfli.h
+++ b/riscv/insns/shfli.h
@@ -1,4 +1,5 @@
 // Zbkb contains zip but not general shfli
+require_rv32;
 require(((insn.rs2() == (xlen / 2 - 1)) && p->extension_enabled(EXT_ZBKB)));
 require(SHAMT < (xlen/2));
 reg_t x = RS1;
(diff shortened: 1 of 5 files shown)


aswaterman (commented):
> So it is optional to reset VSTART at the end of these instructions.

For this reason, the extra code to reset vstart to 0 strikes me as unnecessary clutter.

> However, there is kind of inconsistency in Spike that VSTART is reset for vcpop.m and vfirst.m

I agree this is inconsistent, but it would seem cleaner to resolve it in the opposite manner, by deleting the explicit `vstart` write from `vcpop.m` and `vfirst.m`.

h-cheng:
Sounds good. I submitted a new commit here that deletes `vstart` from `vcpop.m` and `vfirst.m`.

scottj97:
Please update the title of the PR to match the new implementation, and squash the two commits together. No need to add code in one commit and then immediately remove it in the next commit.

h-cheng:
Sorry. I mistakenly touched previous two existing commits when I squash mine. I will close this PR and submit a clean one later.

PR #1573, opened 2024-01-19
vcpop.m, vfirst.m: delete unnecessary VSTART write.

h-cheng:
An updated PR of previous closed [#1570](https://github.com/riscv-software-src/riscv-isa-sim/pull/1570)


aswaterman (approved):
Thanks.

PR #1582, opened 2024-01-24
Support Zicfilp

mylai-mtk:
This PR adds support of the mechanism defined in the [Zicfilp spec](https://github.com/riscv/riscv-cfi/blob/main/cfi_forward.adoc) (commit 079216b).
It is part of the effort to support Zicfilp in spike/pk.


mylai-mtk:
Related PRs:
+ riscv-software-src/riscv-pk#311
+ riscv/riscv-opcodes#225

riscv/encoding.h
```diff
@@ -4,7 +4,7 @@
 
 /*
  * This file is auto-generated by running 'make' in
- * https://github.com/riscv/riscv-opcodes (a014979)
+ * https://github.com/riscv/riscv-opcodes (5238af5)
```
scottj97: We should wait to merge this PR until the riscv-opcodes PR is merged, and make sure this git hash is still relevant. (It could get rebased away before merge.)

riscv/insns/lpad.h
```diff
@@ -0,0 +1,9 @@
+if (p->extension_enabled(EXT_ZICFILP)) {
+  if (ZICFILP_READ_XLPE(STATE.v, STATE.prv) && STATE.elp == elp_t::LP_EXPECTED) {
+    software_check(pc % 4 == 0 &&
+                   ((READ_REG(7) >> 12) == insn.u_imm() || insn.u_imm() == 0),
```
scottj97: This is giving a warning from my (old) compiler:
```
.../riscv-isa-sim/riscv/insns/lpad.h: In function ‘reg_t logged_rv32e_lpad(processor_t*, insn_t, reg_t)’:
.../riscv-isa-sim/riscv/insns/lpad.h:4:41: warning: comparison of integer expressions of different signedness: ‘long unsigned int’ and ‘int64_t’ {aka ‘long int’} [-Wsign-compare]
    ((READ_REG(7) >> 12) == insn.u_imm() || insn.u_imm() == 0),
     ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
```

riscv/csrs.h
```diff
@@ -675,7 +680,7 @@ class dcsr_csr_t: public csr_t {
   dcsr_csr_t(processor_t* const proc, const reg_t addr);
   virtual void verify_permissions(insn_t insn, bool write) const override;
   virtual reg_t read() const noexcept override;
-  void write_cause_and_prv(uint8_t cause, reg_t prv, bool v) noexcept;
+  void update_fields(uint8_t cause, reg_t prv, bool v, const elp_t pelp) noexcept;
```
scottj97: Be consistent about adding `const` here. In the definition, you added `const` on all 4 params. (It's not necessary to add `const` here at all, but if you want to add it, at least add it to all 4.)

riscv/insns/sret.h
```diff
@@ -13,7 +13,6 @@ reg_t prev_prv = get_field(s, MSTATUS_SPP);
 s = set_field(s, MSTATUS_SIE, get_field(s, MSTATUS_SPIE));
 s = set_field(s, MSTATUS_SPIE, 1);
 s = set_field(s, MSTATUS_SPP, PRV_U);
-STATE.sstatus->write(s);
```
scottj97, line 16: Is there any reason to move the `sstatus` write further down in the code?

I can't see any problem that doing so might cause, but I've seen subtle bugs result from reordering things like this. Absent any reason, I'd say the `sstatus` write should stay here.

riscv/decode_macros.h
```diff
@@ -321,3 +321,35 @@ inline long double to_f(float128_t f) { long double r; memcpy(&r, &f, sizeof(r))
...
+      switch (prv) { \
+      case PRV_U: csr_val = STATE.menvcfg->read(); mask = MENVCFG_LPE; break; \
+      case PRV_M: csr_val = STATE.mseccfg->read(); mask = MSECCFG_MLPE; break; \
+      default: abort(); \
+      } \
+    } \
+    get_field(csr_val, mask); \
+  })
+#define ZICFILP_EXPECT_LP(reg_num) \
+  (((reg_num) != 1 && (reg_num) != 5 && (reg_num) != 7) ? \
```
scottj97, line 342: The magic numbers here need some explanation...I assume this is from the spec?

riscv/insns/c_jalr.h
```diff
@@ -3,3 +3,10 @@ require(insn.rvc_rs1() != 0);
 reg_t tmp = npc;
 set_pc(RVC_RS1 & ~reg_t(1));
 WRITE_REG(X_RA, tmp);
+
+if (p->extension_enabled(EXT_ZICFILP)) {
+  if (ZICFILP_READ_XLPE(STATE.v, STATE.prv)) {
+    STATE.elp = ZICFILP_EXPECT_LP(insn.rvc_rs1());
+  }
+}
+
```
scottj97: Delete extraneous newline

riscv/insns/c_jr.h
```diff
@@ -1,3 +1,10 @@
 require_extension(EXT_ZCA);
 require(insn.rvc_rs1() != 0);
 set_pc(RVC_RS1 & ~reg_t(1));
+
+if (p->extension_enabled(EXT_ZICFILP)) {
+  if (ZICFILP_READ_XLPE(STATE.v, STATE.prv)) {
+    STATE.elp = ZICFILP_EXPECT_LP(insn.rvc_rs1());
+  }
+}
+
```
scottj97: Delete extraneous newline

riscv/insns/jalr.h
```diff
@@ -1,3 +1,10 @@
 reg_t tmp = npc;
 set_pc((RS1 + insn.i_imm()) & ~reg_t(1));
 WRITE_RD(tmp);
+
+if (p->extension_enabled(EXT_ZICFILP)) {
+  if (ZICFILP_READ_XLPE(STATE.v, STATE.prv)) {
+    STATE.elp = ZICFILP_EXPECT_LP(insn.rs1());
+  }
+}
+
```
scottj97: Delete extraneous newline

riscv/insns/lpad.h
```diff
@@ -0,0 +1,9 @@
...
+                   SCE_XTVAL_LANDING_PAD_FAULT);
+    STATE.elp = elp_t::NO_LP_EXPECTED;
+  }
+}
+
```
scottj97: Delete extraneous newline

riscv/decode_macros.h
```diff
@@ -321,3 +321,35 @@ inline long double to_f(float128_t f) { long double r; memcpy(&r, &f, sizeof(r))
...
 
 #endif
+
+#define software_check(x, tval) (unlikely(!(x)) ? throw trap_software_check_fault(tval) : (void) 0)
+#define ZICFILP_READ_XLPE(v, prv) \
+  ({ \
+    reg_t csr_val, mask; \
+    if (p->extension_enabled('S')) { \
+      if (v) { \
+        switch (prv) { \
```
scottj97: See `DECLARE_XENVCFG_VARS` in this file. I suspect it could simplify a lot of this.

scottj97 (changes requested):
I very much appreciate the way this is broken into small, atomic commits. It makes review much easier.

I'm not familiar with the Zicfilp spec so my review is only around the implementation in Spike and not its semantics.

This PR is based off of a commit from #1573 that was never the HEAD of `master`. Please rebase onto the latest `master`.

riscv/csrs.cc
```diff
@@ -893,6 +896,7 @@ bool medeleg_csr_t::unlogged_write(const reg_t val) noexcept {
     | (1 << CAUSE_LOAD_PAGE_FAULT)
     | (1 << CAUSE_STORE_PAGE_FAULT)
     | (proc->extension_enabled('H') ? hypervisor_exceptions : 0)
+    | (proc->extension_enabled(EXT_ZICFILP) ? (1 << CAUSE_SOFTWARE_CHECK_FAULT) : 0)
```
ved-rivos, line 899: Please rename "SOFTWARE_CHECK_FAULT" to "SOFTWARE_CHECK"

riscv/insns/mret.h
```diff
@@ -9,6 +9,16 @@ s = set_field(s, MSTATUS_MIE, get_field(s, MSTATUS_MPIE));
 s = set_field(s, MSTATUS_MPIE, 1);
 s = set_field(s, MSTATUS_MPP, p->extension_enabled('U') ? PRV_U : PRV_M);
 s = set_field(s, MSTATUS_MPV, 0);
+
+if (p->extension_enabled(EXT_ZICFILP)) {
+  if (ZICFILP_READ_XLPE(prev_virt, prev_prv)) {
+    STATE.elp = static_cast<elp_t>(get_field(s, MSTATUS_MPELP));
+  } else {
+    STATE.elp = elp_t::NO_LP_EXPECTED;
+  }
```
ved-rivos: else is not required as the state is already NO_LP_EXPECTED when mret executes.

riscv/processor.cc
```diff
@@ -383,7 +391,8 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
     (1 << CAUSE_USER_ECALL) |
     (1 << CAUSE_FETCH_PAGE_FAULT) |
     (1 << CAUSE_LOAD_PAGE_FAULT) |
-    (1 << CAUSE_STORE_PAGE_FAULT);
+    (1 << CAUSE_STORE_PAGE_FAULT) |
+    (proc->extension_enabled(EXT_ZICFILP) ? (1 << CAUSE_SOFTWARE_CHECK_FAULT) : 0);
   csrmap[CSR_HEDELEG] = hedeleg = std::make_shared<masked_csr_t>(proc, CSR_HEDELEG, hedeleg_mask, 0);
```
ved-rivos, line 387: Software Check is not depend on ZICFILP. It is part of Priv 1.13. Zicfilp defines 2 specific trap values for Software Check.

riscv/processor.cc
```diff
@@ -819,8 +833,13 @@ const char* processor_t::get_privilege_string()
 
 void processor_t::enter_debug_mode(uint8_t cause)
 {
+  const bool has_zicfilp = extension_enabled(EXT_ZICFILP);
   state.debug_mode = true;
-  state.dcsr->write_cause_and_prv(cause, state.prv, state.v);
+  state.dcsr->update_fields(cause, state.prv, state.v,
+                            (has_zicfilp ? state.elp : elp_t::NO_LP_EXPECTED));
+  if (has_zicfilp) {
+    state.elp = elp_t::NO_LP_EXPECTED;
```
ved-rivos: Why not just is if ( extension_enabled[EXT_ZICFILP] )?

riscv/decode_macros.h
ved-rivos: could use the spec name for the macro - ZICFILP_IS_LP_EXPECTED. Could add the extension_enabled qualifier and xLPE qualifier here.

riscv/insns/c_jalr.h
ved-rivos: could name ZICFILP_READ_XLPE to ZICFILP_xLPE to match spec.

riscv/trap.h
```diff
@@ -119,5 +119,6 @@ DECLARE_MEM_GVA_TRAP(CAUSE_FETCH_GUEST_PAGE_FAULT, instruction_guest_page_fault)
 DECLARE_MEM_GVA_TRAP(CAUSE_LOAD_GUEST_PAGE_FAULT, load_guest_page_fault)
 DECLARE_INST_TRAP(CAUSE_VIRTUAL_INSTRUCTION, virtual_instruction)
 DECLARE_MEM_GVA_TRAP(CAUSE_STORE_GUEST_PAGE_FAULT, store_guest_page_fault)
+DECLARE_INST_TRAP(CAUSE_SOFTWARE_CHECK_FAULT, software_check_fault)
```
ved-rivos, line 123: Just software_check.

riscv/processor.cc
```diff
@@ -94,6 +94,14 @@ processor_t::~processor_t()
   delete disassembler;
 }
 
+static void zicfilp_maybe_expect_lpad(const elp_t elp, insn_t insn)
+{
+  if (unlikely(elp == elp_t::LP_EXPECTED)) {
+    // also see riscv/lpad.h for more checks performed
+    software_check((insn.bits() & MASK_LPAD) == MATCH_LPAD, SCE_XTVAL_LANDING_PAD_FAULT);
+  }
```
ved-rivos, line 102: Missing check for pc alignment

riscv/insns/lpad.h
```diff
@@ -0,0 +1,9 @@
+if (p->extension_enabled(EXT_ZICFILP)) {
+  if (ZICFILP_READ_XLPE(STATE.v, STATE.prv) && STATE.elp == elp_t::LP_EXPECTED) {
```
ved-rivos: Is elp_t:: required?

riscv/insns/dret.h
```diff
@@ -1,5 +1,12 @@
 require(STATE.debug_mode);
 set_pc_and_serialize(STATE.dpc->read());
+if (p->extension_enabled(EXT_ZICFILP)) {
+  if (ZICFILP_READ_XLPE(STATE.dcsr->v, STATE.dcsr->prv)) {
+    STATE.elp = STATE.dcsr->pelp;
+  } else {
+    STATE.elp = elp_t::NO_LP_EXPECTED;
+  }
```
ved-rivos: else not required

riscv/processor.cc
```diff
@@ -832,8 +832,13 @@ const char* processor_t::get_privilege_string()
 
 void processor_t::enter_debug_mode(uint8_t cause)
 {
+  const bool has_zicfilp = extension_enabled(EXT_ZICFILP);
   state.debug_mode = true;
-  state.dcsr->write_cause_and_prv(cause, state.prv, state.v);
+  state.dcsr->update_fields(cause, state.prv, state.v,
+                            (has_zicfilp ? state.elp : elp_t::NO_LP_EXPECTED));
```
ved-rivos: If extension is not enabled then state.elp should be just NO_LP_EXPECTED - so can remove the has_zicfilp based select.

riscv/processor.cc
ved-rivos: should not need has_zicfilp qualification here.

riscv/processor.cc
```diff
@@ -902,6 +908,10 @@ void processor_t::take_trap(trap_t& t, reg_t epc)
     s = set_field(s, MSTATUS_SPIE, get_field(s, MSTATUS_SIE));
     s = set_field(s, MSTATUS_SPP, state.prv);
     s = set_field(s, MSTATUS_SIE, 0);
+    if (has_zicfilp) {
```
ved-rivos: can just remove has_zicfilp qualification

mylai-mtk:
Update: Address comments

riscv/processor.cc
mylai-mtk: Fixed.

riscv/processor.cc
mylai-mtk: Good point. Fixed.

riscv/insns/dret.h
mylai-mtk: Fixed.

riscv/insns/lpad.h
mylai-mtk: `elp_t` is an `enum struct`, so I need this `elp_t::` to access its enumerators.

riscv/trap.h
mylai-mtk: If the `CAUSE_SOFTWARE_CHECK_FAULT` cannot be renamed, do you still want here to be just `software_check`?

riscv/processor.cc
mylai-mtk: > Why not just is if ( extension_enabled[EXT_ZICFILP] )?

Just felt that `extension_enabled(EXT_ZICFILP)` looks a bit expensive from the call site.

Fixed.

riscv/processor.cc
mylai-mtk: Fixed.

riscv/insns/mret.h
mylai-mtk: Good point. Fixed.

riscv/csrs.cc
mylai-mtk, line 909: This name is generated by `riscv-opcodes`, and that particular line giving this name was contributed long ago (by you) and is not touched in this PR series. Given that `riscv-opcodes` generations are included in other projects, I think it's a bit too late to rename what's already in it.

riscv/decode_macros.h
mylai-mtk: Sorry, I can't see how the approach taken by `DECLARE_XENVCFG_VARS` could help in an obvious way. In this `xLPE` case, it is essentially a runtime table lookup from an integer pair (virtual, privilege) to a (CSR id, CSR bit mask) pair, which is quite different from the paradigm of `DECLARE_XENVCFG_VARS`.

If you have a good idea about this part, please do let me know =)

riscv/insns/lpad.h
mylai-mtk: Fixed

riscv/insns/jalr.h
mylai-mtk: Fixed

riscv/insns/c_jr.h
mylai-mtk: Fixed

riscv/insns/c_jalr.h
mylai-mtk: Fixed

riscv/decode_macros.h
mylai-mtk: @scottj97 Yeah, the numbers come from the spec, which looks like this:
>   is_lp_expected = ( (JALR || C.JR || C.JALR) &&
>                     (rs1 != x1) && (rs1 != x5) && (rs1 != x7) ) ? 1 : 0;

Since the plain numbers just lay themselves out like that in the spec, I suppose for someone looking into this part of Zicfilp-specific code, it should not be hard to get the connection to spec. I can add a comment referencing the spec to clarify where the numbers came from, if this could address your concern?

-----

@ved-rivos I agree `ZICFILP_IS_LP_EXPECTED` is a naturally better name. But I think putting those test conditions you mentioned in this macro would lead to some drawbacks. Consider the macro containing the conditions:
```c++
#define ZICFILP_IS_LP_EXPECTED(reg_num) (has_zicfilp && xLPE && reg_num == 1/5/7 ? LP_EXPECTED : NO_LP_EXPECT)
```
This would cause the use site to become:
```c++
STATE.elp = ZICFILP_IS_LP_EXPECTED(rs1);
```
without any mention of testing whether Zicfilp is enabled, which
1. By my own standard, is a bit confusing to readers who would expect an `extension_enabled()` test to mark out optional behaviors of instructions, and
2. Forces `elp` update unnecessarily.

Due to these concerns, I think keeping the current content of `ZICFILP_EXPECT_LP` is better for readability and performance. What do you say?

riscv/insns/sret.h
mylai-mtk: There's a data dependency on `prev_virt`, which is needed by Zicfilp to determine the `xLPE` state prior to the trap, so I have to place the update to `sstatus` later than all updates to `prev_virt`, which includes that read from `hstatus` in HS-mode. Though that being the reason, I know your concern, and that's why I did not move down other accesses to `sstatus` to keep them together, like this:

```c++
bool prev_virt = STATE.v;
if (!STATE.v) {
  if (p->extension_enabled('H')) {
    prev_virt = get_field(prev_hstatus, HSTATUS_SPV);
    reg_t new_hstatus = set_field(prev_hstatus, HSTATUS_SPV, 0);
    STATE.hstatus->write(new_hstatus);
  }

  STATE.mstatus->write(set_field(STATE.mstatus->read(), MSTATUS_MPRV, 0));
}
reg_t s = STATE.sstatus->read();
reg_t prev_prv = get_field(s, MSTATUS_SPP);
s = set_field(s, MSTATUS_SIE, get_field(s, MSTATUS_SPIE));
s = set_field(s, MSTATUS_SPIE, 1);
s = set_field(s, MSTATUS_SPP, PRV_U);
if (p->extension_enabled(EXT_ZICFILP)) {
  if (ZICFILP_READ_XLPE(prev_virt, prev_prv)) {
    STATE.elp = static_cast<elp_t>(get_field(s, SSTATUS_SPELP));
  } else {
    STATE.elp = elp_t::NO_LP_EXPECTED;
  }
  s = set_field(s, SSTATUS_SPELP, elp_t::NO_LP_EXPECTED);
}
STATE.sstatus->write(s);
```

However, since you took this issue up, perhaps I can ask if you would like me to change the code to something like the above in this case? Maybe for better readability?

riscv/csrs.h
mylai-mtk: Yeah, I think you're right. I find out that I did add `const` to these parameters in the method definition. It's weird that the program got linked successfully.

Fixed.

riscv/insns/lpad.h
mylai-mtk: I did a test run with my compiler (clang 17) and it did not issue a warning. 
Still, it's better to make the sign coherent and avoid possible warnings.
Thanks for the note :)

Fixed.

mylai-mtk:
Update: Misc fix on dret

riscv/decode_macros.h
scottj97: I'm good with this, especially now that the macro name is directly from the spec.

riscv/csrs.h
scottj97: It's not weird: `const` in this context affects only the method body, not the caller, so the language doesn't require the declaration's const-ness to match the definition.

riscv/insns/sret.h
scottj97: No, it's good like it is. Thank you for the explanation.

riscv/decode_macros.h
scottj97: See my [three commits here](https://github.com/scottj97/riscv-isa-sim/commits/pr-1582-my-xenvcfg/)

scottj97 (changes requested):
I tried rebasing this (squashing the fixups) and got a compile failure on "Set ELP state when executing indirect jumps":
```
In file included from c_jalr.cc:12:
.../riscv-isa-sim/riscv/insns/c_jalr.h: In function ‘reg_t fast_rv32i_c_jalr(processor_t*, insn_t, reg_t)’:
.../riscv-isa-sim/riscv/insns/c_jalr.h:8:7: error: ‘ZICFILP_xLPE’ was not declared in this scope
   if (ZICFILP_xLPE(STATE.v, STATE.prv)) {
       ^~~~~~~~~~~~
.../riscv-isa-sim/riscv/insns/c_jalr.h:8:7: note: suggested alternative: ‘ZICFILP_XLPE’
   if (ZICFILP_xLPE(STATE.v, STATE.prv)) {
       ^~~~~~~~~~~~
       ZICFILP_XLPE

File: riscv/trap.h (hunk 15 above)
Comment (ved-rivos) [reply riscv/trap.h:123]: Yes to match the ISA.

File: riscv/execute.cc [hunk 21]
```diff
@@ -291,6 +292,7 @@ void processor_t::step(size_t n)
         // Main simulation loop, fast path.
         for (auto ic_entry = _mmu->access_icache(pc); ; ) {
           auto fetch = ic_entry->data;
+          do_execute_insn_prehook(fetch.insn);
```
ved-rivos: nit. `do_execute_insn_prehook` -> `execute_insn_prehook`.

riscv/insns/c_jalr.h
```diff
@@ -3,3 +3,9 @@ require(insn.rvc_rs1() != 0);
 reg_t tmp = npc;
 set_pc(RVC_RS1 & ~reg_t(1));
 WRITE_REG(X_RA, tmp);
+
+if (p->extension_enabled(EXT_ZICFILP)) {
+  if (ZICFILP_xLPE(STATE.v, STATE.prv)) {
+    STATE.elp = ZICFILP_IS_LP_EXPECTED(insn.rvc_rs1());
```
ved-rivos: The `if (p->extension_enabled(EXT_ZICFILP))` can be merged into `ZICFILP_xLPE`.

riscv/encoding.h
```diff
@@ -346,6 +346,9 @@
 #define SRMCFG_RCID  0x00000FFF
 #define SRMCFG_MCID  0x0FFF0000
 
+/* software check exception xtval codes */
+#define SCE_XTVAL_LANDING_PAD_FAULT 2
+
```
ved-rivos, line 351: `SCE_XTVAL_LANDING_PAD_FAULT` -> `LANDING_PAD_FAULT`

riscv/execute.cc
mylai-mtk: Fixed.

riscv/insns/c_jalr.h
mylai-mtk: Fixed.

riscv/encoding.h
mylai-mtk: Fixed.

riscv/trap.h
mylai-mtk: Fixed.

riscv/decode_macros.h
mylai-mtk: Fixed.

mylai-mtk:
Update: Address comments

---

@scottj97 

> I tried rebasing this (squashing the fixups) and got a compile failure on "Set ELP state when executing indirect jumps":
> 
> ```
> In file included from c_jalr.cc:12:
> .../riscv-isa-sim/riscv/insns/c_jalr.h: In function ‘reg_t fast_rv32i_c_jalr(processor_t*, insn_t, reg_t)’:
> .../riscv-isa-sim/riscv/insns/c_jalr.h:8:7: error: ‘ZICFILP_xLPE’ was not declared in this scope
>    if (ZICFILP_xLPE(STATE.v, STATE.prv)) {
>        ^~~~~~~~~~~~
> .../riscv-isa-sim/riscv/insns/c_jalr.h:8:7: note: suggested alternative: ‘ZICFILP_XLPE’
>    if (ZICFILP_xLPE(STATE.v, STATE.prv)) {
>        ^~~~~~~~~~~~
>        ZICFILP_XLPE
> ```

I tried to reproduce the failure, and noticed that there were conflicts when squashing the fixups due to my misplacement of some particular lines in the commit history. After resolving the conflicts correctly, I did not encounter build failures on any single commit.
I think it'll be fine after we resolve all the comments, and I rebase the commits on `master` and squash the fixups, which will give us the final, clean commits ready for merge.

scottj97:
> I think it'll be fine after we resolve all the comments, and I rebase the commits on `master` and squash the fixups, which will give us the final, clean commits ready for merge.

I think we're ready for this before we can continue reviewing.

mylai-mtk:
Update: Rebase on master and squash fixups

riscv/encoding.h
```diff
@@ -4,7 +4,7 @@
 
 /*
  * This file is auto-generated by running 'make' in
- * https://github.com/riscv/riscv-opcodes (a014979)
+ * https://github.com/riscv/riscv-opcodes (0a1a87a)
```
scottj97: This hash is for a fixup commit on riscv-opcodes, so this should not be merged like this. Let's get that PR merged first and make sure this hash still refers to an actual commit.

scottj97 (changes requested):
CI is failing:
```
./check-opcode-overlap-utst default
Instruction lpad (17) overlaps instruction auipc (17, mask 7f)
make: *** [Makefile:347: check-opcode-overlap-utst.out] Error 255
```

I'm afraid I don't know anything about this check. There must be some additional step needed for overlapping opcodes.

riscv/processor.cc
```diff
@@ -970,6 +992,13 @@ const char* processor_t::get_symbol(uint64_t addr)
   return sim->get_symbol(addr);
 }
 
+void processor_t::execute_insn_prehook(insn_t insn)
+{
+  if (extension_enabled(EXT_ZICFILP)) {
+    zicfilp_maybe_expect_lpad(state.elp, insn);
+  }
```
ved-rivos, line 990: `zicfilp_maybe_expect_lpad` could be renamed as `zicfilp_check_if_lpad_required` to make it slightly more clear what its doing.

riscv/encoding.h
ved-rivos: I think we need to update the `overlap_list.h` to declare this overlap.

mylai-mtk:
Update: Fix CI failure

riscv/processor.cc
mylai-mtk: Fixed

mylai-mtk:
Update: Address comments

riscv/encoding.h
mylai-mtk: Fixed

ved-rivos:
1. In `mseccfg_csr_t::verify_permissions(insn_t insn, bool write)` needs an update to allow access to CSR when `EXT_ZICFILP` is enabled. Add
```
if (!proc->extension_enabled(EXT_SMEPMP) & 
    !proc->extension_enabled(EXT_ZICFILP))
```
2. In `bool mseccfg_csr_t::unlogged_write` add
```
if ( proc->extension_enabled(EXT_ZICFILP) ) {
  new_val &= ~MSECCFG_MLPE;
  new_val |= (val & MSECCFG_MLPE);
}

```

riscv/insns/lpad.h
```diff
@@ -0,0 +1,6 @@
+if (ZICFILP_xLPE(STATE.v, STATE.prv) && STATE.elp == elp_t::LP_EXPECTED) {
+  software_check(pc % 4 == 0 &&
+                 ((READ_REG(7) >> 12) == static_cast<uint64_t>(insn.u_imm()) || insn.u_imm() == 0LL),
+                 LANDING_PAD_FAULT);
```
ved-rivos, line 4: replace with following suggested code as only bits 31:12 should be compared. Also note that `u_imm()` already returns the value left shifted by 12.
```
 software_check(pc % 4 == 0 &&
                 ((READ_REG(7) & 0xFFFFF000) == (insn.u_imm() & 0xFFFFF000) || insn.u_imm() == 0LL),
                 LANDING_PAD_FAULT);

```

mylai-mtk:
Update: Address comments, re-generate encoding.h with a master commit

riscv/insns/lpad.h
mylai-mtk: Fixed

scottj97 (approved):

ved-rivos (approved):

mylai-mtk:
Update: Rebase on master, fix lpad insn behavior bug

mylai-mtk:
Hi @scottj97 , @ved-rivos . This PR had been left unchanged for 3 weeks. Is there anything else that I need to do so it can be merged? Thanks.

scottj97:
It all looks good to me. I usually allow @aswaterman to do the merge.

aswaterman (approved):
LGTM, too.  Thank you!

PR #1590, opened 2024-02-02
Fix hvip.VSEIP and hvip.VSTIP so they don't observe platform-specific interrupts

YenHaoChen:
The H extension defines that bit VSEIP, VSTIP, and VSSIP of hvip are writable. (The other bits of hvip are read-only 0.) Only the VSSIP of hip (mip) is an alias of the same bit of hvip. The VSEIP and VSTIP of hip are the logical-OR of ones of hvip and other sources.

The current aliasing (proxy) implementation does not provide the desired behavior. An ISA-level behavior difference is that any platform-specific external and timer interrupt signals directed to VS-level should not be observable through the hvip. For instance, the hvip should not observe the virtual timer interrupt signal from the Sstc extension (vstimecmp CSR), which isn't true in the current implementation.

This commit fixes the issue by giving the hvip a specialized class, hvip_csr_t. The hvip_csr_t aliases the hvip.VSSIP to the mip.VSSIP but decouples the hvip.VSEIP and hvip.VSTIP from mip.VSEIP and mip.VSTIP. Additionally, the commit updates the read value of mip to be the logical-OR of hvip.VSEIP and hvip.VSTIP and other sources.

image
image

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -728,6 +728,10 @@ mip_csr_t::mip_csr_t(processor_t* const proc, const reg_t addr):
   mip_or_mie_csr_t(proc, addr) {
 }
 
+reg_t mip_csr_t::read() const noexcept {
+  return val | state->hvip->basic_csr_t::read();
+}
+
 void mip_csr_t::backdoor_write_with_mask(const reg_t mask, const reg_t val) noexcept {
   this->val = (this->val & ~mask) | (val & mask);
 }
@@ -1716,4 +1720,17 @@ void srmcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
 
   if (state->v)
       throw trap_virtual_instruction(insn.bits());
+}
+
+hvip_csr_t::hvip_csr_t(processor_t* const proc, const reg_t addr, const reg_t init):
+  basic_csr_t(proc, addr, init) {
+}
+
+reg_t hvip_csr_t::read() const noexcept {
+  return basic_csr_t::read() | (state->mip->read() & MIP_VSSIP); // hvip.VSSIP is an alias of mip.VSSIP
+}
+
+bool hvip_csr_t::unlogged_write(const reg_t val) noexcept {
+  state->mip->write_with_mask(MIP_VSSIP, val); // hvip.VSSIP is an alias of mip.VSSIP
+  return basic_csr_t::unlogged_write(val & (MIP_VSEIP | MIP_VSTIP));
 }
--- a/riscv/csrs.h
+++ b/riscv/csrs.h
@@ -349,7 +349,7 @@ typedef std::shared_ptr<misa_csr_t> misa_csr_t_p;
 class mip_or_mie_csr_t: public csr_t {
  public:
   mip_or_mie_csr_t(processor_t* const proc, const reg_t addr);
-  virtual reg_t read() const noexcept override final;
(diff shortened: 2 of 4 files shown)


riscv/csrs.cc
```diff
@@ -728,6 +728,10 @@ mip_csr_t::mip_csr_t(processor_t* const proc, const reg_t addr):
   mip_or_mie_csr_t(proc, addr) {
 }
 
+reg_t mip_csr_t::read() const noexcept {
+  return val | (state->csrmap[CSR_HVIP]->read() & (MIP_VSEIP | MIP_VSTIP));
```
scottj97: Save `hvip` as a variable in `state_t` and refer to it by name here, instead of looking it up in the `csrmap` every time.

riscv/csrs.cc
```diff
@@ -1717,3 +1721,12 @@ void srmcfg_csr_t::verify_permissions(insn_t insn, bool write) const {
   if (state->v)
       throw trap_virtual_instruction(insn.bits());
 }
+
+hvip_csr_t::hvip_csr_t(processor_t* const proc, const reg_t addr, const reg_t init):
+  basic_csr_t(proc, addr, init) {
+}
+
+bool hvip_csr_t::unlogged_write(const reg_t val) noexcept {
+  state->mip->backdoor_write_with_mask(MIP_VSSIP, val & MIP_VSSIP); // hvip.VSSIP is an alias of mip.VSSIP
```
scottj97: This seems problematic, as it's not going to log the change to `mip`.

YenHaoChen:
Need help on the CI issue. I can compile the code on my computer.

scottj97 (changes requested):
Please change the commit message (and PR title) to indicate what is changing. It's not clear from the message if hvip is supposed to be an alias of mip or not.

The latest commit 5927541c771fc3da7a97d95f3232fbab2fd7507b is core dumping on every test I attempt. (1bdc833f382cf39a0fe74202e71ccd62484bd794 was not doing this.)

riscv/csrs.cc
YenHaoChen: Use write_with_mask() instead of backdoor_write_with_mask() fucntion.

YenHaoChen:
Update the title and commit message to clarify that the hvip isn't an alias of mip.

scottj97 (changes requested):
Let me explain what I'm asking for w.r.t. the commit message and PR title.

The old message of "hvip isn't an alias..." describes a fact. Is this is a true or false fact? Is hvip supposed to be an alias? Is it a problem if it is an alias? Or is it a problem if it's not an alias? Was it an alias in the old code? Does this commit change that, so it's not an alias anymore? Or maybe it wasn't an alias and needs to be, so this commit makes it an alias? None of those important questions are answered by that commit title.

Even the new phrasing of "shouldn't" describes a problem, not a change. Okay, I accept that it shouldn't be an alias. So what? What does this commit have to do with that? Does this change it so it's not an alias anymore? I don't know. This is a good title for an issue report, but not a pull request or commit message.

The commit message should include a verb explaining what is changing. How about:

    Fix hvip.VSEIP and hvip.VSTIP so they don't observe platform-specific interrupts

Followed by all the rest of the explanation from 63bd8d89353a95921bb9e77c91cb20f7befe1368

scottj97 (approved):

aswaterman (approved):

PR #1591, opened 2024-02-02
Teach Sstc to respect xenvcfg.STCE

YenHaoChen:
When menvcfg.STCE=0, mip.STIP reverts to its defined behavior as if unsupporting Sstc extension. When henvcfg.STCE=0, mip.VSTIP reverts to its defined behavior as if unsupporting Sstc extension.

Reference: https://github.com/riscv/riscv-time-compare/issues/5

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1075,7 +1075,8 @@ void time_counter_csr_t::sync(const reg_t val) noexcept {
   if (proc->extension_enabled(EXT_SSTC)) {
     const reg_t mip_val = (shadow_val >= state->stimecmp->read() ? MIP_STIP : 0) |
       (shadow_val + state->htimedelta->read() >= state->vstimecmp->read() ? MIP_VSTIP : 0);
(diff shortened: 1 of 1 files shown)


scottj97 (changes requested):
Please change the title and commit message to describe what is changing. Explain in the PR what was wrong with the old code, an example of how it could misbehave, and how this fixes it.

I guess the old code is only problematic if `--isa=Sstc` is enabled? But it's not clear.

YenHaoChen:
Add additional paragraphs in the commit message to describe an example of misbehaving and the solution's strategy. 

The title was a shortened, one-sentence summary from the spec, which the commit aims to conform to.
image
I need help developing a more concise but respectful sentence to the spec.

scottj97:
How about this title:

    Teach Sstc to respect xenvcfg.STCE

Followed by the explanation of the problem. The last paragraph in 15cd07d7b4862cecaea6ae90ba20edc317a9f6c4 is unnecessary because it's simply describing the implementation which is already easy to understand from the diff.

scottj97 (approved):

aswaterman (approved):

PR #1592, opened 2024-02-02
Update trigger description in README.md

YenHaoChen:
The number of triggers is configurable since https://github.com/riscv-software-src/riscv-isa-sim/pull/1219.

I am unsure whether we should remove the line completely. The description was for tracking the development of various trigger types. Since we have supported all trigger types, I recommend retiring the description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ Spike supports the following RISC-V ISA features:
   - Svinval extension, v1.0
   - Sdext extension, v1.0-STABLE
   - Sdtrig extension, v1.0-STABLE
-    - 4 triggers support type={2, 3, 4, 5, 6, 15} (mcontrol, icount, itrigger, etrigger, mcontrol6, disabled)
   - Smepmp extension v1.0
   - Smstateen extension, v1.0
   - Sscofpmf v0.5.2


README.md
```diff
@@ -39,7 +39,7 @@ Spike supports the following RISC-V ISA features:
   - Svinval extension, v1.0
   - Sdext extension, v1.0-STABLE
   - Sdtrig extension, v1.0-STABLE
-    - 4 triggers support type={2, 3, 4, 5, 6, 15} (mcontrol, icount, itrigger, etrigger, mcontrol6, disabled)
+    - Support trigger type={2, 3, 4, 5, 6, 15} (mcontrol, icount, itrigger, etrigger, mcontrol6, disabled)
```
scottj97: The grammar doesn't sound right here. How about:

    - Triggers support type={2, 3, 4, 5, 6, 15} (mcontrol, icount, itrigger, etrigger, mcontrol6, disabled)

scottj97: On second thought, I like your idea of simply removing the line altogether. It used to describe the limited implementation of Sdtrig, but if all trigger types are supported now, it's not very useful anymore.

YenHaoChen: Removed the line in https://github.com/riscv-software-src/riscv-isa-sim/pull/1592/commits/6903ff698d2a6469b3773558d760b31a2947ae9f

scottj97 (approved):

aswaterman (approved):

PR #1593, opened 2024-02-06
Agnostic: Add support for "write ones" agnostic policy of vector arithmetic instructions.

Siudya:
Enable "write ones" agnostic policy:

```
autoheader && autoconf
mkdir build
cd build
../configure --enable-varith-agnostic-write-ones --prefix=$RISCV
make
```

**Vector load instructions are not included**

### DIFF
--- a/riscv/insns/vcompress_vm.h
+++ b/riscv/insns/vcompress_vm.h
@@ -16,7 +16,7 @@ VI_GENERAL_LOOP_BASE(1)
   switch (sew) {
   case e8:
     if (1 == P.VU.vta) {
-      P.VU.elt<uint8_t>(rd_num, i) = 0xFF;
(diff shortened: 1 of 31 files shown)


aswaterman:
I’m hesitant to move forward with this for a couple reasons.

First, it’s pretty invasive; I would hope we could come up with a much more DRY way of doing this.

Second, I think the scope might not be broad enough to cover actual implementation behavior. Consider that a popular strategy for register-renamed implementations will be to write 1s up to the next ceil(vl * SEW / VLEN) register boundary, then leave the remaining registers in the register group undisturbed. And that’s just one example.

So, maybe a lot of complexity for modest gain?

Siudya:
Can't deny that these changes are invasive and dirty. These changes are made during cosim with our CPU design. Maybe we should just keep our fork This PR can be closed.

PR #1595, opened 2024-02-07
Interaction: Support until-mem operation on physical memory space

Siudya:
This operation is supposed to be available but not actually.
This PR now permitted followings when running debug interactive interface:
```
until mem 80000000 5a7f
```
Spike will keep running until physical memory at 0x80000000 changed to 0x5a7f

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -721,7 +721,7 @@ void sim_t::interactive_until(const std::string& cmd, const std::vector<std::str
   if (args.size() < 3)
     throw trap_interactive();
 
-  if (args.size() == 3)
+  if (args.size() == 4 || (args[0] == "pc" && args.size() == 3)) //dont check mem with arg len = 3
     get_core(args[1]); // make sure that argument is a valid core number
 
   char *end;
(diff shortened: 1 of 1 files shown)


riscv/interactive.cc
```diff
@@ -732,7 +732,8 @@ void sim_t::interactive_until(const std::string& cmd, const std::vector<std::str
     throw trap_interactive();
 
   // mask bits above max_xlen
-  int max_xlen = procs[strtol(args[1].c_str(),NULL,10)]->get_isa().get_max_xlen();
+  bool until_mem_paddr = args[0] == "mem" && args.size() == 3;
+  int max_xlen = until_mem_paddr? procs[0]->get_isa().get_max_xlen() : procs[strtol(args[1].c_str(),NULL,10)]->get_isa().get_max_xlen();
```
scottj97: Make a separate variable for the proc number, instead of repeating the entire expression. Something like:
```
size_t procnum = until_mem_paddr ? 0 : strtol(args[1].c_str(), NULL, 10);
int max_xlen = procs[procnum]->get_isa().get_max_xlen();
```

Siudya: This has been fixed. New review is needed.

jerryz123 (approved):
I think this is right, but this whole part deserves a refactor. It is getting more and more convoluted.

aswaterman (approved):
> It is getting more and more convoluted.

Yeahhhhh...

PR #1596, opened 2024-02-07
Update CLIC CSR_xINTSTATUS addresses, add CSR_xINTTHRESH

dansmathers:
Refresh encoding.h to update CSR_MINTSTATUS, CSR_SINTSTATUS, CSR_UINTSTATUS addresses to read-only csr addresses and add CSR_MINTTHRESH, CSR_SINTTHRESH, CSR_UINTTHRESH csrs.

spec: https://github.com/riscv/riscv-fast-interrupt


aswaterman (changes requested):
`encoding.h` is an auto-generated file and so shouldn't be edited manually.  See the header: https://github.com/riscv-software-src/riscv-isa-sim/blob/d4726e1cf542471298f261f792eeb9a5c67e5f49/riscv/encoding.h#L5-L8

Please make this change in `riscv-opcodes` and, after that PR is merged, regenerate `encoding.h` and copy it over to here.

dansmathers:
created https://github.com/riscv/riscv-opcodes/pull/226

closing pull until riscv-opcodes PR is merged.

PR #1598, opened 2024-02-10
refactor: Enhance performance of u_imm()

howjmay:


aswaterman:
This is a semantic change that breaks Spike.  (`pk` won't boot, for example.)  It doesn't correctly sign-extend the result when the MSB of the immediate is set.

Regardless, the change is not necessary.  GCC optimizes the original code into the optimal thing on both x86 and ARM: sign extend from 32 to 64 bits, then AND with -4096.

PR #1602, opened 2024-02-16
Fix c.mop.N decoding

YenHaoChen:
The c.mop.N only accepts rd={x1, x3, x5, x7, x9, x11, x13, x15}. The previous implemention incorrectly accepts additional rd={x17, x19, x21, x23, x25, x27, x29, x31}.


aswaterman:
@ved-rivos

ved-rivos (approved):

aswaterman (approved):

PR #1610, opened 2024-02-19
Raise illegal instruction instead of virtual instruction on WFI when TW=1 and VTW=0 in VU-mode

YenHaoChen:
The previous implementation raises virtual instruction on WFI when TW=1 and VTW=0 in VU-mode. According to the recent discussion, we expect illegal instruction in this case.

Reference: https://github.com/riscv/riscv-isa-manual/issues/1234

### DIFF
--- a/riscv/insns/wfi.h
+++ b/riscv/insns/wfi.h
@@ -1,9 +1,7 @@
-if (STATE.v && STATE.prv == PRV_U) {
-  require_novirt();
-} else if (get_field(STATE.mstatus->read(), MSTATUS_TW)) {
+if (get_field(STATE.mstatus->read(), MSTATUS_TW)) {
   require_privilege(PRV_M);
-} else if (STATE.v) { // VS-mode
-  if (get_field(STATE.hstatus->read(), HSTATUS_VTW))
+} else if (STATE.v) {
+  if (STATE.prv == PRV_U || get_field(STATE.hstatus->read(), HSTATUS_VTW))
     require_novirt();
 } else if (p->extension_enabled('S')) {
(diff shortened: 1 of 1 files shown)


aswaterman:
Upon further review of the spec, I think Spike is slightly wrong, but this PR is not the right fix.  The issue is that mstatus.TW should take precedence (and raise an illegal-instruction exception when it takes effect).

I claim the following is right.  Can you see if you ultimately agree, and update the PR as we work through this?

```
if (get_field(STATE.mstatus->read(), MSTATUS_TW)) {
  require_privilege(PRV_M);
} else if (STATE.v) {
  if (STATE.prv == PRV_U || get_field(STATE.hstatus->read(), HSTATUS_VTW))
    require_novirt();
} else if (p->extension_enabled('S')) {
  // When S-mode is implemented, then executing WFI in
  // U-mode causes an illegal instruction exception.
  require_privilege(PRV_S);
}
wfi();

```

YenHaoChen:
You are right. I forget to add the condition ```STATE.prv == PRV_U``` in the ```else if (STATE.v)``` block.

aswaterman (approved):
Thanks for the sanity check, and for pushing this along.

rbuchner-aril:
Not an issue, but just wanted to comment since the only case that was mentioned above was TW=1 and VTW=0; for the case where a 'wfi' is issued in VU mode where TW=1 and VTW=1, Spike was previously also incorrectly (at least from my understanding of the spec) taking a Virtual Instruction Exception. After this change, it takes an Illegal Instruction Exception as expected in that situation. I just wanted to make known this second behavioral change.

PR #1617, opened 2024-03-01
workaround to support custom extensions that use standard prefixes

arromanoff:
RISC-V ISA states (21.1):
> A standard-compatible global encoding can also use standard prefixes for non-standard extensions if the associated standard extensions are not included in the global encoding."

Currently all the instructions (either from standard or custom extensions) are all being inserted into a single std::vector which is then being sorted. An instruction matching process performs linear search on that vector. The problem is that when a custom extension uses the same opcode as standard one (i.e. match and mask are equal to the standard counterparts) it is undefined which instruction will be picked. That is because in std::sort "The order of equal elements is not guaranteed to be preserved". That being said it is impossible to define custom extension (via customext) that would use the prefix of a disabled standard extension.

In this change I separate custom and standard extensions in two separate std::vector's. By default we report an error if they have common elements (There're an additional processor_t constructor's argument that skips this check). If this error is disabled during instruction matching we first trying to find it among custom instructions. If it has been found the search is stopped and custom instruction is executed, otherwise we look for it among standard instructions. Overall this change does not completely fix the problem but at least makes it possible to use the feature of RISC-V ISA.

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -1037,6 +1037,24 @@ reg_t illegal_instruction(processor_t UNUSED *p, insn_t insn, reg_t UNUSED pc)
   throw trap_illegal_instruction(insn.bits() & 0xffffffffULL);
 }
 
+static insn_desc_t
+propagate_instruction_in_vector(std::vector<insn_desc_t> &instructions,
+                                std::vector<insn_desc_t>::iterator it) {
+  assert(it != instructions.end());
+  insn_desc_t desc = *it;
+  if (it->mask != 0 && it != instructions.begin() &&
+      std::next(it) != instructions.end()) {
+    if (it->match != std::prev(it)->match &&
+        it->match != std::next(it)->match) {
+      // move to front of opcode list to reduce miss penalty
+      while (--it >= instructions.begin())
+        *std::next(it) = *it;
+      instructions[0] = desc;
+    }
+  }
+  return desc;
+}
+
 insn_func_t processor_t::decode_insn(insn_t insn)
 {
   // look up opcode in hash table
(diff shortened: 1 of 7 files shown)


arromanoff:
@aswaterman , @timsifive could you please take a look?

riscv/processor.cc
```diff
@@ -61,7 +61,7 @@ processor_t::processor_t(const isa_parser_t *isa, const cfg_t *cfg,
 
   disassembler = new disassembler_t(isa);
   for (auto e : isa->get_extensions())
-    register_extension(find_extension(e.c_str())());
+    register_extension(find_extension(e.c_str())(), true /* is_custom */);
```
jerryz123: I think having separate `register_base_extension` and `register_custom_extension` is cleaner.

aswaterman (commented):
My inclination is to assume the authors of custom extensions know what they're doing, and allow them to override standard behavior at their will.  In that case, we could avoid adding a new command-line argument.  What do you think @jerryz123?

jerryz123:
> In that case, we could avoid adding a new command-line argument.

I agree with this. A warning might be appropriate, but there is an argument that spike should not dump anything to console that is not from the target binary when executed without the debug/logging flags.

aswaterman:
For my part, I'd even forgo the warning.

arromanoff:
> My inclination is to assume the authors of custom extensions know what they're doing, and allow them to override standard behavior at their will. In that case, we could avoid adding a new command-line argument. 

Are you suggesting to move this "overrides standard extension" leverage into `extension_t` so that extension authors can express their intents? I like that idea. There's a problem that behavior will still be undefined if user overrides standard opcode unintentionally but it is probably ok for now.

aswaterman:
I was more straightforwardly suggesting that we get rid of the command-line option and make Spike act as though `--extension-intersection-allowed=true` had been passed.

We can presumably use a `stable_sort` instead so that the custom extensions are deterministically picked over the standard ones (which might involve reversing the vector first or something like that).

arromanoff:
`reverse` + `stable_sort` might not be the best idea since we may have several such extensions and have to perform this sorting (reversing) procedure several times.

riscv/processor.cc
arromanoff: Agreed. Changed.

aswaterman (approved):
OK with me, but I'll give @jerryz123 a chance to weigh in

jerryz123 (approved):
Thanks for adding the testing

aswaterman:
I forgot I need to update the CI test tarball to make the tests pass in CI.  However, I couldn't get the new tests to pass locally.  I get:

`test-customext: /scratch/waterman/riscv-isa-sim/ci-tests/../fesvr/elfloader.cc:45: std::map<std::__cxx11::basic_string<char>, long unsigned int> load_elf(const char*, memif_t*, reg_t*, unsigned int): Assertion IS_ELF_RISCV(*eh64) || IS_ELF_EM_NONE(*eh64) failed.`

Can you get them working so I can build the tarball and upload it?  From the root directory, run `ci-tests/build-spike` then `ci-tests/test-spike`.  (You need to run from the root directory to match the CI behavior.)

arromanoff:
Problem solved. Now tests are working again (at least locally).

jerryz123:
@aswaterman I guess you've updated the tarball if the CI is passing? This can be merged now?

aswaterman:
Yeah, I actually uploaded the tarball yesterday when the tests were still failing.  I guess the problem was not with the tarball contents.  Out of caution, I regenerated it and made sure the tests are still passing locally (they are) and uploaded it again.

PR #1619, opened 2024-03-04
Bump GitHub Actions runners to Ubuntu 22.04

### LINKED ISSUE #1618: Failure during Make: SYS_futex not declared in scope
Hey all,

I'm trying to build spike on Mint. Once I succesfully configure the project with $RISCV=/opt/riscv and run make, I get this error:

```
In file included from /usr/include/c++/11/bits/atomic_base.h:41,
                 from /usr/include/c++/11/bits/shared_ptr_atomic.h:33,
                 from /usr/include/c++/11/memory:78,
                 from ../fesvr/context.h:12,
                 from ../fesvr/dtm.h:5,
                 from ../fesvr/dtm.cc:1:
/usr/include/c++/11/bits/atomic_wait.h: In function ‘void std::__detail::__platform_wait(const _Tp*, std::__detail::__platform_wait_t)’:
/usr/include/c++/11/bits/atomic_wait.h:108:29: error: ‘SYS_futex’ was not declared in this scope
  108 |         auto __e = syscall (SYS_futex, static_cast<const void*>(__addr),
      |                             ^~~~~~~~~
/usr/include/c++/11/bits/atomic_wait.h: In function ‘void std::__detail::__platform_notify(const _Tp*, bool)’:
/usr/include/c++/11/bits/atomic_wait.h:121:18: error: ‘SYS_futex’ was not declared in this scope
  121 |         syscall (SYS_futex, static_cast<const void*>(__addr),
      |                  ^~~~~~~~~
At global scope:
cc1plus: note: unrecognized command-line option ‘-Wno-nonportable-include-path’ may have been intended to silence earlier diagnostics
make: *** [Makefile:347: dtm.o] Error 1
```

I did some digging, and I can't replicate this same error in other programs that import the same headers. The above directory contains all RISCV toolchains - I don't know if that's actually important. All help appreciated.

aswaterman:
The problem was our use of `-I` to add subprojects' headers to the include path.  Our `syscall.h` was interfering with libstdc++'s dependence on the system's version of the same header (which is why we were getting `SYS_futex`-not-defined errors).

Resolves #1618

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -18,7 +18,7 @@ on:
 jobs:
   test:
     name: Test Spike build (Ubuntu)
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/debug-smoke.yml
+++ b/.github/workflows/debug-smoke.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   test:
     name: Test debug (Ubuntu)
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v2
 
--- a/Makefile.in
+++ b/Makefile.in
@@ -69,7 +69,7 @@ install_exes_dir := $(INSTALLDIR)/bin
 sprojs         := @subprojects@
 sprojs_enabled := @subprojects_enabled@
 
-sprojs_include := -I. -I$(src_dir) $(addprefix -I$(src_dir)/, $(sprojs_enabled))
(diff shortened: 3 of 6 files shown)


aswaterman:
BTW @jerryz123 if you have a cleaner fix to the makefiles in mind, I'm all ears... the handling of libfdt is a little hacky, but at least the blast radius is small.

jerryz123 (approved):
I think the final clean solution is to eventually stop relying on the in-tree libfdt, assuming that the packaged libfdt doesn't have bugs (#1536 ).

But this is fine for now...  its better than the alternative of editing all the fdt sources.

I think it is expected for the tests to fail. I believe the CI distro is set from the CI config at the HEAD of this branch, so the non-working commits will still get tested (and fail) under 22.04.

aswaterman:
Agreed re: libfdt.

I swapped the commit order so the critical fix comes first, so hopefully it'll pass this time.

PR #1628, opened 2024-03-19
don't read CSR in csrrw(i) with rd=0

ShengMingTangAndes:
According to
The RISC-V Instruction Set Manual Volume I: Unprivileged ISA Document Version 20191213 Table 9.1,
there are some conditional with rd = 0 that inhibit read on CSRs.
螢幕擷取畫面 2024-03-19 110524

Though original implementation is correct in most cases, this always-read implementation may cause some effects. For example, some CSRs that read-write to it may change the architectural state.


aswaterman (changes requested):
No standard CSRs have read side effects, so this doesn't matter.  It's just extra complexity for no benefit.

PR #1630, opened 2024-03-22
Allow software check exception to be delegated from M mode regardless of Zicfilp being enabled

mylai-mtk:
Sorry it's a bit late, but I noticed this bug when casually viewing my just-merged Zicfilp #1582 :

Software check is a new exception which is not dependent on Zicfilp, so it should be delegable regardless of Zicfilp enabled or not.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -906,7 +906,7 @@ bool medeleg_csr_t::unlogged_write(const reg_t val) noexcept {
     | (1 << CAUSE_LOAD_PAGE_FAULT)
     | (1 << CAUSE_STORE_PAGE_FAULT)
     | (proc->extension_enabled('H') ? hypervisor_exceptions : 0)
-    | (proc->extension_enabled(EXT_ZICFILP) ? (1 << CAUSE_SOFTWARE_CHECK_FAULT) : 0)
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

omerguzelelectronicguy:
@mylai-mtk hello, I am running a baremetal test on spike and notice the software check delegate bit opened. I searched and udnerstand that it is the only related extension with `Zicfilp`. However, I saw your command and I confused. Why it should be delegable even if `Zcfilp` closed?

mylai-mtk:
@omerguzelelectronicguy 

> I searched and udnerstand that it is the only related extension with Zicfilp.

While only Zicfilp uses software check exception for now, the exception is defined to be a generic one allowing multiple extensions to utilize it:

> A Software Check exception is a synchronous exception that is triggered when there are violations of checks and assertions defined by ISA extensions that aim to safeguard the integrity of software assets, ...
> -- The RISC-V Instruction Set Manual: Volume II, Privileged Architecture, Version 20241101

 So given its generic status, the bit is implemented regardless of whether Zicfilp is implemented (i.e. enabled in spike).

omerguzelelectronicguy:
I see. Thank you for clarification.

PR #1631, opened 2024-03-22
Implement syscall readlinkat

mylai-mtk:
I implemented this syscall since it's used in my libc testsuite, and think it may be useful others using spike, so I'm contributing it.


aswaterman (approved):
LGTM at a casual glance.

Issue #1557, opened 2024-01-05
Question about permission check of a supervisor physical address in a hlvx instruction 


viktoryou:
As the Priv Spec said
> For the supervisor physical address that results from address translation, the supervisor physical
memory attributes must **grant both execute and read permissions**. (The supervisor physical memory
attributes are the machine’s physical memory attributes as modified by physical memory protection,
Section 3.7, for supervisor level.)

However, when I check the physical memory protection in Spike for this case, only the read permisson is required.
What is expected when the supervisor physical address of a hlvx request is only granted with read permission without execute permission?

viktoryou:
Here is some furthur descrption. The content I quoted is added by https://github.com/riscv/riscv-isa-manual/pull/679. To my understanding, the physical memory atttibute modified by physical memory protection should give the requirement of X on the basis of R permission to hlvx instruction. However, it seems the spike hasn't fit this change. From https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/csrs.cc#183, we can see this feature is not added into  `pmpaddr_csr_t::access_ok`. I am not sure if I missed anything. I suspect that maybe a bug in spike and my reproduction testcase convinces me so.

scottj97:
For HLVX, Spike checks the PTEs for execute permissions ([here](https://github.com/riscv-software-src/riscv-isa-sim/blob/d6f622e84d7b202f9f1ca23411efca24ea2b5955/riscv/mmu.cc#L442) and [here](https://github.com/riscv-software-src/riscv-isa-sim/blob/d6f622e84d7b202f9f1ca23411efca24ea2b5955/riscv/mmu.cc#L536)).

But I concur that I cannot find a similar check for PMP.

viktoryou:
Thanks for the check, Scott. 

Since the issue has been open for a while,  I think it is necessary to mention it again.
@avpatel Could you offer some help on this issue? For a hlvx instruction, what consequence would the lack of execute permission in PMP cause? Is spike simply forgotten to add the check or the check would never added under some consideration?

Any help would be appreciated.

gfavor:
Per the current email thread "Permission check when using hlvx" in tech-priv ...

_The HLVX spec says: HLVX cannot override machine-level physical memory protection (PMP), so attempting to read memory that PMP designates as execute-only still results in an access-fault exception._

_This sentence implies that PMP must grant both read and execute permissions, otherwise an access fault results._

[...]

PR #1641, opened 2024-04-10
add hlvx pmp protect to fix issue 1557

xinyuwang-starfive:
for fix issue #1557, add hlvx check x check for pmp

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -180,7 +180,7 @@ bool pmpaddr_csr_t::subset_match(reg_t addr, reg_t len) const noexcept {
   return !(is_tor ? tor_homogeneous : napot_homogeneous);
 }
 
-bool pmpaddr_csr_t::access_ok(access_type type, reg_t mode) const noexcept {
+bool pmpaddr_csr_t::access_ok(access_type type, reg_t mode, bool hlvx) const noexcept {
   const bool cfgx = cfg & PMP_X;
   const bool cfgw = cfg & PMP_W;
   const bool cfgr = cfg & PMP_R;
@@ -191,7 +191,7 @@ bool pmpaddr_csr_t::access_ok(access_type type, reg_t mode) const noexcept {
   const bool typer = type == LOAD;
   const bool typex = type == FETCH;
   const bool typew = type == STORE;
-  const bool normal_rwx = (typer && cfgr) || (typew && cfgw) || (typex && cfgx);
+  const bool normal_rwx = (typer && cfgr && (!hlvx || cfgx)) || (typew && cfgw) || (typex && cfgx);
   const bool mseccfg_mml = state->mseccfg->get_mml();
 
   if (mseccfg_mml) {
--- a/riscv/csrs.h
+++ b/riscv/csrs.h
@@ -102,7 +102,7 @@ class pmpaddr_csr_t: public csr_t {
   bool subset_match(reg_t addr, reg_t len) const noexcept;
 
   // Is the specified access allowed given the pmpcfg privileges?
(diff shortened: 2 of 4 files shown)


riscv/csrs.cc
```diff
@@ -189,7 +189,7 @@ bool pmpaddr_csr_t::access_ok(access_type type, reg_t mode) const noexcept {
   const bool prvm = mode == PRV_M;
 
   const bool typer = type == LOAD;
-  const bool typex = type == FETCH;
+  const bool typex = (type == FETCH ) || hlvx;
```
aswaterman: no space before `)`

aswaterman (commented):
Please use a space before `,` in general.

aswaterman (changes requested):
This doesn't look right to me, as I think that HLVX should still request only read permissions on PMP/PMA.  I'll move the discussion to #1557.

aswaterman (changes requested):
Nevermind, I think I agree with the conclusion that both R and X are required.  But I don't think that's what this implementation does.  It looks like this implementation makes HLVX will succeed if the PMP offers R _or_ X permissions, rather than requiring R _and_ X.  See this line:

https://github.com/riscv-software-src/riscv-isa-sim/blob/5576090eb011f5cfcfb641bdd2a61d050d2b0074/riscv/csrs.cc#L194

xinyuwang-starfive:
i fix it,now i think it's r && x

xinyuwang-starfive:
use a space before ,  now and rebase into one fix

aswaterman (changes requested):
Are you sure you got it right? This version seems to allow R permissions but still seems not to require X.

How are you testing this?

xinyuwang-starfive:
const bool normal_rwx = ((typer && cfgr) && (!hlvx || cfgx)) || (typew && cfgw) || (typex && cfgx);
I think hlvx is typer.
if hlvx is low , the logic is same with old, and if hlvx is high, it need cfgx is high .
if PMP not have the X ,the cfgx ||!hlvx is 0 and the normal_rwx will be 0 and go to access fault.
and i run a small case which only have  R to test it .

aswaterman (approved):
Thanks, this version looks right to me.  Can I get a second pair of eyes from either @scottj97 or @jerryz123?

riscv/csrs.cc
```diff
@@ -191,7 +191,7 @@ bool pmpaddr_csr_t::access_ok(access_type type, reg_t mode) const noexcept {
   const bool typer = type == LOAD;
   const bool typex = type == FETCH;
   const bool typew = type == STORE;
-  const bool normal_rwx = (typer && cfgr) || (typew && cfgw) || (typex && cfgx);
+  const bool normal_rwx = ((typer && cfgr) && (!hlvx || cfgx)) || (typew && cfgw) || (typex && cfgx);
```
jerryz123: Minor detail, but omitting the parentheses around `typer && cfgr`:

```
(typer && cfgr && (!hlvx || cfgx)) || (typew && cfgw) || (typex && cfgx)
```

is slightly clearer to me. Otherwise this seems to be right, HLVX requires both R and X PMP attributes.

Maybe we should add some assertion that `hlvx` implies `typer`? Or at least a comment on it.

jerryz123 (changes requested):

jerryz123 (approved):

aswaterman:
Can you squash those commits into one commit?  Then I think this is ready to go.

xinyuwang-starfive:
> Can you squash those commits into one commit? Then I think this is ready to go.

i do it

PR #1648, opened 2024-04-23
Smstateen: Ignore writes to read-only hstateen*[n] bits when mstateen*[n]=0

YenHaoChen:
The specification states that writes to read-only bits in a RW CSR are ignored.
image

The hstateen*[n] bits are read-only when mstateen*[n]=0.
image

This PR propose ignoring writes to read-only hstateen*[n] bits when mstateen*[n]=0 instead of writing the bits to 0.


jerryz123 (approved):
This looks correct to me

aswaterman (approved):

PR #1649, opened 2024-04-23
Assert misa.B bit through --isa=...B...

YenHaoChen:
This PR provides misa.B=1 by --isa=...B...

Reference: https://github.com/riscv-software-src/riscv-isa-sim/pull/1559

### DIFF
--- a/disasm/isa_parser.cc
+++ b/disasm/isa_parser.cc
@@ -29,7 +29,7 @@ static void bad_priv_string(const char* priv)
 isa_parser_t::isa_parser_t(const char* str, const char *priv)
 {
   isa_string = strtolower(str);
-  const char* all_subsets = "mafdqcpvh";
+  const char* all_subsets = "mafdqcpvhb";
 
   if (isa_string.compare(0, 4, "rv32") == 0)
     max_xlen = 32;


aswaterman (changes requested):
Seems like we are becoming inconsistent about how we handle these different cases, in multiple ways:

- For P, we use the [switch statement](https://github.com/riscv-software-src/riscv-isa-sim/blob/3192ee4d31f481e84281a24d55bb6130e3743668/disasm/isa_parser.cc#L73) to populate the Zp extensions, rather than the if statements at the bottom.  (This is only a stylistic concern; we don't need to fix it right now, but I thought it was worth mentioning.)
- For A, we make it so that A implies Zaamo and Zalrsc, but we don't make Zaamo + Zalrsc imply A.  This is inconsistent with what you propose we do for B.  We should either delete the logic that makes Zba + Zbb + Zbs imply B, or add the logic so that Zaamo + Zalrsc imply A.  I think people would prefer the latter option, so that `misa` will contain A if Zaamo+Zalrsc is specified.

YenHaoChen:
Thank you. I was not aware of the inconsistency. I targeted the missing character 'b' in the variable 'all_subsets,' which supports --isa=...B...

Let's delete the logic that makes Zba + Zbb + Zbs imply B in this PR. I will provide another PR that discusses the implications of A and B.

aswaterman:
Does this version still make B imply Zba+Zbb+Zbs? It isn’t clear to me how it works.

YenHaoChen:
> Does this version still make B imply Zba+Zbb+Zbs? It isn’t clear to me how it works.

Yes, we have B implying Zba+Zbb+Zbs (disasm/isa_parser.cc:333). The previous PR (https://github.com/riscv-software-src/riscv-isa-sim/pull/1559) provided the code for implication, which aims at the B extension. Unfortunately, it does not offer --isa=...B... functionality from the optional flag. This PR catches up with the missing.

aswaterman:
Ah, got it.

aswaterman (approved):

PR #1650, opened 2024-04-23
Make Zaamo + Zalrsc (Zba + Zbb + Zbs) imply A (B) in misa

YenHaoChen:
The PR makes Zaamo + Zalrsc imply A in misa.
Additionally, it makes Zba + Zbb + Zbs imply B in misa.

Reference: https://github.com/riscv-software-src/riscv-isa-sim/pull/1649#pullrequestreview-2018427770


aswaterman (approved):
LGTM.

Issue #1652, opened 2024-04-24
Counter-intuitive MMIO device arguments behavior


liuyu81:
PR #1522 introduced a feature to pass down extra arguments to device plugins.

The original PR motivated this feature with the following use case,

> The ucb-bar block device is implemented as an MMIO device.
We would like to pass an argument to the block device such that it can read a file and initialize its contents at startup.

Our intuition was that, device arguments should apply to device instances (`abstract_device_t`). Say, if we specify two devices each taking a different set of arguments, i.e.,

```
spike --extlib=myplugin.so --device=mydevice,args=first_input_file --device=mydevice,args=second_input_file ...  
```

we would expect `spike` to register *one* device factory (`device_factory_t`) under the name `mydevice`, and *two* device instances (`abstract_device_t`) specialized with different sets of arguments.

Surprisingly, based on what we observed from the implementation (`spike_main/spike.c:L357-377`), device arguments are actually per-factory (as opposed to per-device). And, if specified more than once, the *last* set of arguments takes effect for *all* device instances. So, in the above example, we end up with *two* `mydevice`'s, both wired to the `second_input_file`.

Of course, to work around, one can register the same device factory multiple times under different names. But still, the CLI `--device` option is a bit counter-intuitive here, and, IMHO, when the arity is always one device per factory, it probably invalidates the use of the factory pattern in the first place.

jerryz123:
Yes, this is indeed a limitation with the current system. I agree it is far more intuitive to support instantiating multiple instances of the same device.

- `device_factory_t::parse_from_fdt` would have to return a vector of device pointers
- devices for which there should never be multiple instances (CLINT/PLIC) should assert this
- `device_factory_t::set_sargs` should be `add_sargs`, I think.
- `device_factory_t::sargs` should be a 2D object - a list of argument-lists.
- `device_parser` in `spike_main` should use `add_sargs` instead of `set_sargs`

We'd probably want a address-map sanity checker across all instantiated devices too.

I'd be happy to review this PR if it were made, and make any modifications to our `ucb-bar/spike-devices` to conform to the updated APIs.

liuyu81:
I would be happy to help with refactoring the MMIO device API, but would propose a slightly different design.

[...]

PR #1655, opened 2024-04-26
Support per-device arguments and device factory reuse

liuyu81:
Device arguments (or `sargs`) is a new feature introduced in #1522. It allows passing extra `sargs` from CLI `--device=factory[,sargs]` option down to MMIO device plugins.

While the use case is definitely sound, previous implementation does *not* really support instantiating multiple devices from the same factory with different sets of arguments. Another limitation (of previous implementation), is that -- due to the injection of per-device `sargs` globally into `device_factory_t` -- it prohibits device factory resue across multiple `sim_t` instances.

As proposed in #1652, we have created this PR to address the above limitations. The fix is three fold,

- we removed `sargs` from `device_factory_t`, and introduced a new type alias `device_factory_sargs_t` to capture <`device_factory_t *`, `sargs`> pairs, this is used to instantiate `sim_t` instances;
- we changed the signature of `device_factory_t::generate_fdt` / `device_factory_t::parse_from_fdt` to take on an extra `sargs` argument, for instantiating devices with per-device arguments;
- we made `device_factory_t` const and potentially resuable across multiple `sim_t` instances.

### DIFF
--- a/riscv/clint.cc
+++ b/riscv/clint.cc
@@ -117,7 +117,7 @@ void clint_t::tick(reg_t rtc_ticks)
 }
 
 clint_t* clint_parse_from_fdt(const void* fdt, const sim_t* sim, reg_t* base,
-    const std::vector<std::string>& sargs) {
+    const std::vector<std::string>& UNUSED sargs) {
   if (fdt_parse_clint(fdt, base, "riscv,clint0") == 0 || fdt_parse_clint(fdt, base, "sifive,clint0") == 0)
     return new clint_t(sim,
                        sim->CPU_HZ / sim->INSNS_PER_RTC_TICK,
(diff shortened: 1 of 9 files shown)


riscv/clint.cc
```diff
@@ -126,7 +126,7 @@ clint_t* clint_parse_from_fdt(const void* fdt, const sim_t* sim, reg_t* base,
     return nullptr;
 }
 
-std::string clint_generate_dts(const sim_t* sim) {
+std::string clint_generate_dts(const sim_t* sim, const std::vector<std::string>& sargs) {
```
jerryz123: We should mark the `sargs` argument of clint/ns16550/plic as `UNUSED`.

jerryz123 (changes requested):
This overall looks good to me. This repo uses a rebase flow for PRs. Rebase your changes on master, instead of merging master into your feature branch.

jerryz123 (approved):

liuyu81:
Thanks for the review. I have marked the `sargs` of clint/ns16550/plic as `UNUSED`, and rebased my changes on master.

aswaterman:
@jerryz123 I don't have time to review this right now, but feel free to merge if you think it's good to go.

PR #1660, opened 2024-05-01
Remove old P extension

aswaterman:
The new Zpa extension is coming along, to the point of having a proposed set of instructions and an encoding.  I expect it will make it to the freeze milestone this year.  Alas, it bears little resemblance to the old Zp extensions: the mnemonics, encodings, and set of included instructions are completely different.  There's obviously some semantic overlap, but there isn't an obvious incremental path to take advantage of it.  So, in preparation, we'll rip everything out.

I expect a subset of the macros and instruction definitions will be useful, but only on a case-by-case basis.  They'll live on in the git history so that whoever does the Spike implementation can take advantage of them at their discretion.


jerryz123 (approved):
Support for the pre-ratification P-extension could still be provided via extension library, for anyone who needs it. They would have to implement that library themselves outside of Spike, but that should be a simple exercise.

aswaterman:
Yep. (Or, in many cases, by just using an older version of Spike.)

rbuchner-aril:
@aswaterman I noticed part of this MR was this change:
```
  void vxsat_csr_t::verify_permissions(insn_t insn, bool write) const {
    require_vector_vs;
 -   if (!proc->extension_enabled('V') && !proc->extension_enabled(EXT_ZPN))
 -     throw trap_illegal_instruction(insn.bits());
    masked_csr_t::verify_permissions(insn, write);
  }
```

Seems to me that we still should check that the vector extension is enabled.
```
void vxsat_csr_t::verify_permissions(insn_t insn, bool write) const {
  require_vector_vs;
  if (!proc->extension_enabled('V'))
  masked_csr_t::verify_permissions(insn, write);
}
```

aswaterman:
@rbuchner-aril Indeed, that was an inadvertent code change on my part, but I believe it's still correct by accident.  `require_vector_vs` will trap if VS is nonzero, and VS can only be nonzero if V is enabled.  Right?

rbuchner-aril:
I don't think `require_vector_vs` will trap in this case, but regardless we should add the explicit check back in.

```
#define require_vector_vs   require(STATE.sstatus->enabled(SSTATUS_VS))
```

Importantly, there is the statement ` if (!orig_sstatus->field_exists(which))` within `enabled()`
```
bool sstatus_csr_t::enabled(const reg_t which) {
  if ((orig_sstatus->read() & which) != 0) {
    if (!state->v || (virt_sstatus->read() & which) != 0)
      return true;
  }

  // If the field doesn't exist, it is always enabled. See #823.
  if (!orig_sstatus->field_exists(which))
    return true;

  return false;
}
```

So the `SSTATUS_VS` field doesn't exist when `!proc->extension_enabled('V')`. Thus, `require_vector_vs` will not throw an illegal instruction trap.
```
reg_t base_status_csr_t::compute_sstatus_write_mask() const noexcept {
  // If a configuration has FS bits, they will always be accessible no
  // matter the state of misa.
  const bool has_fs = (proc->extension_enabled('S') || proc->extension_enabled('F')
              || proc->extension_enabled('V')) && !proc->extension_enabled(EXT_ZFINX);
  const bool has_vs = proc->extension_enabled('V');
  return 0
    | (has_fs ? SSTATUS_FS : 0)
    | (proc->any_custom_extensions() ? SSTATUS_XS : 0)
    | (has_vs ? SSTATUS_VS : 0)
    | (proc->extension_enabled(EXT_ZICFILP) ? SSTATUS_SPELP : 0)
    ;
}
```

rbuchner-aril:
I opened #1678 as a fix.

PR #1661, opened 2024-05-02
config: add Zicond for Xiangshan

cyyself:
Since Zicond has been supoorted in Xiangshan, we add zicond in Xiangshan ISA string for difftest.


cyyself:
Sorry, I create this PR in the wrong place. This should be in a fork repo.

PR #1663, opened 2024-05-04
Add Zawrs extension

ved-rivos:
Tests: https://github.com/riscv-non-isa/riscv-arch-test/pull/432


aswaterman (commented):
Thanks for following through with this. I’ll try to do a code review today or tomorrow.

disasm/isa_parser.cc
```diff
@@ -385,6 +387,9 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
   if (extension_table[EXT_ZABHA] && !extension_table['A'] && !extension_table[EXT_ZAAMO]) {
     bad_isa_string(str, "'Zabha' extension requires either the 'A' or the 'Zaamo' extension");
   }
+  if (extension_table[EXT_ZAWRS] && !extension_table['A'] && !extension_table[EXT_ZALRSC]) {
+    bad_isa_string(str, "'Zabha' extension requires either the 'A' or the 'Zalrsc' extension");
```
aswaterman, line 392: We can delete the dependence on `A` here, since at this point, because of https://github.com/riscv-software-src/riscv-isa-sim/blob/e2834ca9e7ae1a3f97cf2f68f185c604831285ae/disasm/isa_parser.cc#L341

aswaterman: And we can do the same thing for `Zabha` just above, for the same reason

aswaterman (approved):
LGTM, but would you mind addressing the one nit in the ISA parser?

disasm/isa_parser.cc
ved-rivos: @aswaterman updated. Applied same reason to `Zacas` as well.

PR #1670, opened 2024-05-16
add support to load ET_DYN elf

clementleger:
When compiled as PIE, executable can be loaded at any memory address. Lately, OpenSBI switched to such behavior and spike was not able to load it anymore. This patch add an additional load_offset parameter for load_elf(). This load_offset value is passed as DRAM_BASE and used only for ET_DYN elfs.

### DIFF
--- a/fesvr/elfloader.cc
+++ b/fesvr/elfloader.cc
@@ -18,7 +18,8 @@
 #include <map>
 #include <cerrno>
 
-std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry, unsigned required_xlen = 0)
+std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* entry,
+                                         reg_t load_offset, unsigned required_xlen = 0)
 {
   int fd = open(fn, O_RDONLY);
   struct stat s;
@@ -41,9 +42,13 @@ std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t* 
     throw incompat_xlen(required_xlen, xlen);
   }
   assert(IS_ELFLE(*eh64) || IS_ELFBE(*eh64));
-  assert(IS_ELF_EXEC(*eh64));
+  assert(IS_ELF_EXEC(*eh64) || IS_ELF_DYN(*eh64));
   assert(IS_ELF_RISCV(*eh64) || IS_ELF_EM_NONE(*eh64));
   assert(IS_ELF_VCURRENT(*eh64));
+
+  if (IS_ELF_EXEC(*eh64)) {
+    load_offset = 0;
+  }
 
   std::vector<uint8_t> zeros;
   std::map<std::string, uint64_t> symbols;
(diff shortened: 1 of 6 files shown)


fesvr/htif.h
```diff
@@ -27,6 +27,7 @@ class htif_t : public chunked_memif_t
   bool done();
   int exit_code();
   void set_expected_xlen(unsigned int m) { expected_xlen = m; }
+  void set_load_offset(reg_t offset) { load_offset = offset; }
```
aswaterman: The setter method seems like YAGNI to me, since we set it once to a constant.  Can we instead make `load_offset` a `const reg_t` that's initialized to `DRAM_BASE`?

clementleger: Yes of course, Idid this since I wasn't sure if htif would need to be independent from sim or not.

aswaterman (approved):
I'd like to give @jerryz123 a chance to opine, but it's OK with me now (assuming you've tested that it suffices for OpenSBI).

fesvr/htif.h
```diff
@@ -7,6 +7,7 @@
 #include "syscall.h"
 #include "device.h"
 #include "byteorder.h"
+#include "platform.h"
```
jerryz123: To support usage of FESVR as a library, this should be `#include "../riscv/platform.h"`

jerryz123 (changes requested):

clementleger:
> I'd like to give @jerryz123 a chance to opine, but it's OK with me now (assuming you've tested that it suffices for OpenSBI).

Yes, that is enough for OpenSBI to work (sha1 d4d2582eef7aac442076f955e4024403f8ff3d96 is the first one after the switch from EXEC to ET_DYN elf).

fesvr/elfloader.cc
```diff
@@ -41,30 +42,34 @@ std::map<std::string, uint64_t> load_elf(const char* fn, memif_t* memif, reg_t*
...
 #define LOAD_ELF(ehdr_t, phdr_t, shdr_t, sym_t, bswap)                         \
   do {                                                                         \
     ehdr_t* eh = (ehdr_t*)buf;                                                 \
     phdr_t* ph = (phdr_t*)(buf + bswap(eh->e_phoff));                          \
-    *entry = bswap(eh->e_entry);                                               \
+    *entry = bswap(eh->e_entry) + load_offset;                                 \
     assert(size >= bswap(eh->e_phoff) + bswap(eh->e_phnum) * sizeof(*ph));     \
     for (unsigned i = 0; i < bswap(eh->e_phnum); i++) {                        \
       if (bswap(ph[i].p_type) == PT_LOAD && bswap(ph[i].p_memsz)) {            \
+        uint32_t load_addr = bswap(ph[i].p_paddr) + load_offset;               \
```
jerryz123: This should be a `reg_t`, not `uint32_t`. Good catch by the debug tests.

jerryz123 (changes requested):

jerryz123 (approved):

PR #1675, opened 2024-05-23
zicflip: fix [ms]ret behavior

chihminchao:
Based on spec chapter 3.5
"An MRET or SRET instruction is used to return from a trap in M-mode or S-mode, respectively. When executing an xRET instruction, if xPP holds the value y, then ELP is set to the value of xPELP if yLPE is 1; otherwise, it is set to NO_LP_EXPECTED; xPELP is set to NO_LP_EXPECTED."

The change follows the last statement after semicolon "xPELP is set to NO_LP_EXPECTED"

### DIFF
--- a/riscv/insns/mret.h
+++ b/riscv/insns/mret.h
@@ -11,8 +11,8 @@ s = set_field(s, MSTATUS_MPP, p->extension_enabled('U') ? PRV_U : PRV_M);
 s = set_field(s, MSTATUS_MPV, 0);
 if (ZICFILP_xLPE(prev_virt, prev_prv)) {
   STATE.elp = static_cast<elp_t>(get_field(s, MSTATUS_MPELP));
-  s = set_field(s, MSTATUS_MPELP, elp_t::NO_LP_EXPECTED);
 }
+s = set_field(s, MSTATUS_MPELP, elp_t::NO_LP_EXPECTED);
 STATE.mstatus->write(s);
 if (STATE.mstatush) STATE.mstatush->write(s >> 32); // log mstatush change
 p->set_privilege(prev_prv, prev_virt);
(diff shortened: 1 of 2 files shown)


chihminchao:
@mylai-mtk You may be interested at this.

riscv/insns/mret.h
```diff
@@ -11,8 +11,8 @@ s = set_field(s, MSTATUS_MPP, p->extension_enabled('U') ? PRV_U : PRV_M);
 s = set_field(s, MSTATUS_MPV, 0);
 if (ZICFILP_xLPE(prev_virt, prev_prv)) {
   STATE.elp = static_cast<elp_t>(get_field(s, MSTATUS_MPELP));
-  s = set_field(s, MSTATUS_MPELP, elp_t::NO_LP_EXPECTED);
 }
+s = set_field(s, MSTATUS_MPELP, elp_t::NO_LP_EXPECTED);
```
mylai-mtk, line 15: Should have a guard of `p->extension_enabled(EXT_ZICFILP)`

riscv/insns/sret.h
```diff
@@ -25,7 +25,7 @@ if (!STATE.v) {
 }
 if (ZICFILP_xLPE(prev_virt, prev_prv)) {
   STATE.elp = static_cast<elp_t>(get_field(s, SSTATUS_SPELP));
-  s = set_field(s, SSTATUS_SPELP, elp_t::NO_LP_EXPECTED);
 }
+s = set_field(s, SSTATUS_SPELP, elp_t::NO_LP_EXPECTED);
```
mylai-mtk, line 29: Should have a guard of `p->extension_enabled(EXT_ZICFILP)`

mylai-mtk:
I think you're right, but I don't have write access to the repo 😁

riscv/insns/mret.h
aswaterman: I don't think that guard is needed, because the bit is hardwired to 0 when Zicfilp is not implemented, so the statement has no effect.

riscv/insns/sret.h
aswaterman: Same as above

aswaterman (approved):

PR #1677, opened 2024-05-24
vector: Not logging write of reduction instructions when vl = 0

YenHaoChen:
The spec says: "If vl=0, no operation is performed and the destination register is not updated." in Section 14. Vector Reduction Operations.

The commit proposes setting the variable is_write to false when vl = 0, which means not logging the write (instead of logging a write of the same value).


aswaterman (approved):
This looks right to me.  cc @chihminchao

PR #1678, opened 2024-05-27
Require vector extension when attempting vxsat writes

rbuchner-aril:
This requirement was accidentally removed in c9468f6e02.

See pull request #1660


aswaterman (commented):
Thanks for the fix.

aswaterman (approved):

PR #1679, opened 2024-05-27
Updated README with supported Vector Cryptography Extensions

akifejaz:
Spike supports these below cryptography extensions; I've updated readme accordingly. 

Zvbb, Zvbc, Zvkg, Zvkned, Zvknha, Zvknhb , Zvksed, Zvksh, Zvkt, Zvkn, Zvknc, Zvkng, Zvks, Zvksc, Zvksg 

@aswaterman Hi Andrew, this is my first PR here, I'll really appreciate any suggestion. Thanks.


aswaterman:
I believe these are all 1.0, with no third digit.  See https://github.com/riscv/riscv-isa-manual/blob/main/src/vector-crypto.adoc

But yes, we should include them in the README, and probably audit the README for other missing extensions.  I'm sure there are many.

akifejaz:
> I believe these are all 1.0, with no third digit. See https://github.com/riscv/riscv-isa-manual/blob/main/src/vector-crypto.adoc
> 
> But yes, we should include them in the README, and probably audit the README for other missing extensions. I'm sure there are many.

I've corrected the version.

akifejaz:
Hi @aswaterman, Can we accept this at this point?

aswaterman (approved):
Yes, thanks for the ping.

PR #1682, opened 2024-05-29
Adding Zilsd and Zcmlsd extensions (Load/store pair for RV32)

christian-herber-nxp:
Zilsd & Zcmlsd: https://github.com/riscv/riscv-zilsd

This is a PR the currently unratified load/store pair extension.


riscv/decode_macros.h
```diff
@@ -42,6 +42,15 @@
   })
 #define WRITE_VSTATUS STATE.log_reg_write[3] = {0, 0};
 
+/* the value parameter needs to be evaluated before writing to the registers */
+#define WRITE_REG_PAIR(reg, value) \
+  if (reg != 0) { \
+    require(reg % 2 == 0); \
+    uint64_t val = value; \
+    WRITE_REG(reg, sext32(val)); \
+    WRITE_REG(reg + 1, (sreg_t(val)) >> 32); \
```
aswaterman: Please wrap macro parameters in parens when used in expressions (unless it's otherwise unambiguous, e.g. when separated by the comma operator or by other parens).  For example, writing `if (reg)` would be OK, as would be `if ((reg) != 0)`.

riscv/insns/ld.h
```diff
@@ -1,2 +1,7 @@
-require_rv64;
-WRITE_RD(MMU.load<int64_t>(RS1 + insn.i_imm()));
+require((xlen == 64) || p->extension_enabled(EXT_ZILSD));
+if (xlen == 32) {
+    WRITE_RD_PAIR(MMU.load<int64_t>(RS1 + insn.i_imm()));
```
aswaterman: Use two spaces to indent.

riscv/insns/c_flwsp.h
```diff
@@ -1,7 +1,13 @@
...
-  require_extension(EXT_ZCF);
-  require_fp;
-  WRITE_FRD(f32(MMU.load<uint32_t>(RVC_SP + insn.rvc_lwsp_imm())));
+  require_either_extension(EXT_ZCF, EXT_ZCMLSD);
+  if (p->extension_enabled(EXT_ZCF)) {
+    require_fp;
+    WRITE_FRD(f32(MMU.load<uint32_t>(RVC_SP + insn.rvc_lwsp_imm())));
+  } else { // c.ldsp - Load pair RV32
+    require_extension(EXT_ZCMLSD);
+    require(insn.rvc_rd() != 0);
```
aswaterman: This is redundant with the `require` inside of `WRITE_RD_PAIR`.  It's best to handle this in one place, so delete this one.

riscv/insns/c_fsw.h
```diff
@@ -1,7 +1,13 @@
...
-  require_extension(EXT_ZCF);
-  require_fp;
-  MMU.store<uint32_t>(RVC_RS1S + insn.rvc_lw_imm(), RVC_FRS2S.v[0]);
+  require_either_extension(EXT_ZCF, EXT_ZCMLSD);
+  if (p->extension_enabled(EXT_ZCF)) {
+    require_fp;
+    MMU.store<uint32_t>(RVC_RS1S + insn.rvc_lw_imm(), RVC_FRS2S.v[0]);
+  } else {  // c.sd - Store pair RV32
+    require_extension(EXT_ZCMLSD);
+    require(insn.rvc_rs2() % 2 == 0);
```
aswaterman: Redundant with the `require` in READ_REG_PAIR; delete.

riscv/insns/c_fswsp.h
```diff
@@ -1,7 +1,13 @@
...
-  require_extension(EXT_ZCF);
-  require_fp;
-  MMU.store<uint32_t>(RVC_SP + insn.rvc_swsp_imm(), RVC_FRS2.v[0]);
+  require_either_extension(EXT_ZCF, EXT_ZCMLSD);
+  if (p->extension_enabled(EXT_ZCF)) {
+    require_fp;
+    MMU.store<uint32_t>(RVC_SP + insn.rvc_swsp_imm(), RVC_FRS2.v[0]);
+  } else { // c.sdsp - Store pair RV32
+    require_extension(EXT_ZCMLSD);
+    require(insn.rs2() % 2 == 0);
```
aswaterman: Redundant with the `require` in READ_REG_PAIR; delete.

riscv/insns/sd.h
```diff
@@ -1,2 +1,9 @@
-require_rv64;
-MMU.store<uint64_t>(RS1 + insn.s_imm(), RS2);
+if (xlen == 32)
+{ // sd - Store pair RV32
+    require_extension(EXT_ZILSD);
+    require(insn.rs2() % 2 == 0);
```
aswaterman: Redundant; delete.

riscv/insns/sd.h
```diff
@@ -1,2 +1,9 @@
...
+} else {
+    require(xlen == 64);
+}
+
+STORE_D;
```
aswaterman: Please delete the `STORE_D` macro and move that code back into this position.

aswaterman: (Or, rather, move its two pieces into the respective branches of the preceding if/else statement.)

christian-herber-nxp: done in 02fbb0d

riscv/insns/sd.h
christian-herber-nxp: done in 02fbb0d

riscv/insns/c_fswsp.h
christian-herber-nxp: done in 02fbb0d

riscv/insns/c_fsw.h
christian-herber-nxp: done in 02fbb0d

riscv/insns/c_flwsp.h
christian-herber-nxp: done in 02fbb0d

riscv/insns/ld.h
christian-herber-nxp: done in 02fbb0d

riscv/decode_macros.h
christian-herber-nxp: done in 02fbb0d

disasm/disasm.cc
```diff
@@ -1412,6 +1412,12 @@ void disassembler_t::add_instructions(const isa_parser_t* isa)
     DISASM_INSN("c.fsw", c_fsw, 0, {&rvc_fp_rs2s, &rvc_lw_address});
     DISASM_INSN("c.fswsp", c_fswsp, 0, {&rvc_fp_rs2, &rvc_swsp_address});
   }
+  else if (isa->extension_enabled(EXT_ZCMLSD)) {
```
aswaterman: move the `else` onto the same line as the previous closing brace

disasm/disasm.cc
```diff
@@ -1412,6 +1412,12 @@ void disassembler_t::add_instructions(const isa_parser_t* isa)
...
+      DISASM_INSN("c.ld", c_ld, 0, {&rvc_rs2s, &rvc_ld_address});
```
aswaterman: two spaces for indentation

disasm/isa_parser.cc
```diff
@@ -239,6 +239,10 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
       extension_table[EXT_ZICOND] = true;
     } else if (ext_str == "zihpm") {
       extension_table[EXT_ZIHPM] = true;
+    } else if (ext_str == "zilsd") {
+      extension_table[EXT_ZILSD] = true;
+    } else if (ext_str == "zcmlsd") {
+      extension_table[EXT_ZCMLSD] = true;
```
aswaterman, line 247: Towards the end of this function, do you need some logic to issue an error if both Zcmlsd and Zcf are selected?

aswaterman (commented):
This is starting to look pretty good to me, but we'll give others a chance to review.

disasm/disasm.cc
christian-herber-nxp: fixed in 9b27b67

disasm/disasm.cc
christian-herber-nxp: fixed in 9b27b67

disasm/isa_parser.cc
christian-herber-nxp: absolutely. fixed in 9b27b67.
In addition, added checks to 
- ensure Zilsd is only selected with RV32
- Made sure Zcmlsd is only selected when both Zilsd and Zca are there

disasm/isa_parser.cc
```diff
@@ -324,6 +330,14 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
     bad_isa_string(str, ("can't parse: " + std::string(p)).c_str());
   }
 
+  if (extension_table[EXT_ZCMLSD] && extension_table[EXT_ZCF]) {
+    bad_isa_string(str, ("'Zcmlsd' extension conflicts with 'Zcf' extensions"));
```
aswaterman: No need for parens around the string.

disasm/isa_parser.cc
```diff
@@ -324,6 +330,14 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
...
+  }
+
+  if (extension_table[EXT_ZCMLSD] && (!extension_table[EXT_ZCA] || !extension_table[EXT_ZILSD])) {
+    bad_isa_string(str, ("'Zcmlsd' extension requires 'Zca' and 'Zilsd' extensions"));
```
aswaterman: ditto

aswaterman:
Can you squash these into a single commit?

I think this is ready for @jerryz123 to review.

jerryz123:
Why don't these instructions use the `overlap_list` mechanism?

aswaterman:
I assume @christian-herber-nxp is following the pattern of e.g. https://github.com/riscv-software-src/riscv-isa-sim/blob/3a70f84b8a2249c92d35c2229b48ca5735a543fa/riscv/insns/c_fsw.h


But we probably should not proliferate that pattern.  I agree `overlap_list` is cleaner and more self-documenting.

disasm/isa_parser.cc
christian-herber-nxp: fixed in 2b08ffe9

christian-herber-nxp: fixed in 2b08ffe9

disasm/isa_parser.cc
christian-herber-nxp: 2b08ffe9

christian-herber-nxp:
> Can you squash these into a single commit?
> 
> I think this is ready for @jerryz123 to review.

I rebased and squashed. Well I see now that there are conflicts, so I need to revisit
Edit: looks ok now.

christian-herber-nxp:
> I assume @christian-herber-nxp is following the pattern of e.g. https://github.com/riscv-software-src/riscv-isa-sim/blob/3a70f84b8a2249c92d35c2229b48ca5735a543fa/riscv/insns/c_fsw.h
> 
> But we probably should not proliferate that pattern. I agree `overlap_list` is cleaner and more self-documenting.

Correct. I do not have experience with Spike so I just tried to adopt what is there. Had a look at the overlap_list. As this is based on  an unordered set, it does sound like a slight overkill to me. Did you ever check the performance compared to a small if-else?
Of course, it is your choice how you would like to have this integrated. As this would also affect the existing c.flw/c.ld overlaps etc., maybe this goes beyond the scope of this PR.

aswaterman:
The performance is actually neutral: neither the overlap list nor the if-statement has a runtime performance impact.  (The overlap list is only used when setting up the simulation, not when running; and the if-statement is compiled out.)  So it's more about aesthetics.

Let me prepare a PR for the existing C instructions (like c.flw/c.ld) and see if we like it.

aswaterman:
@christian-herber-nxp Now that #1687 has been merged, I think you can proceed with reworking this to move the code into `c_ld[sp].h` and `c_sd[sp].h`.

In this case, you shouldn't add anything to the overlap list, I think.

riscv/insns/ld.h
```diff
@@ -1,2 +1,7 @@
-require_rv64;
-WRITE_RD(MMU.load<int64_t>(RS1 + insn.i_imm()));
+require((xlen == 64) || p->extension_enabled(EXT_ZILSD));
```
christian-herber-nxp, line 1: @aswaterman I noticed that the require checking is inconsistent. I do not know the implications of it, so wanted to check what the best solution is. The alternative to this would be to check in each branch of the if statement. See e.g. sd.h.
The compressed instructions miss any require check in the XLEN != 32 branch. I assume this must be added there. Will do if I know your preference.

aswaterman: I think the strategy you've used for `ld`, using a single `require` at the top, is clean.

For e.g. `c.ld`, it looks like I did screw up by not requiring RV64.  For those, I think we want

    require_extension(EXT_ZCA);
    require((xlen == 64) || p->extension_enabled(EXT_ZILSD));

Thanks for fixing it.

christian-herber-nxp: fixed in 0ddd72a1

aswaterman (approved):
As part of merging, I'm going to make the disassembler code more DRY, but otherwise LGTM.  Thanks for your patience @christian-herber-nxp!

aswaterman:
Manually merged via 70d26d64e6ba2da329357a88dc313277fff6c22c

PR #1684, opened 2024-05-31
Avoid checking ELP before every instruction fetch

aswaterman:
Serialize after setting ELP.  That way, we can hoist the ELP check outside of the main simulation loop.  This keeps the special case isolated to one location and gives a modest speedup.


jerryz123 (approved):
I take it `mret/sret/dret` already serialize execution, so force the ELP check anyways?

aswaterman:
Yup.

PR #1687, opened 2024-06-11
Separate RV32 and RV64 C instructions into separate files

aswaterman:
Pursuant to https://github.com/riscv-software-src/riscv-isa-sim/pull/1682#issuecomment-2138462509, separate RV32 and RV64 C instructions (e.g. c.flw/c.fld) into separate files.

This turned out to be a bit of a rabbit hole.  I found that the semantics of the overlap list are ill-specified and insufficient to handle overlap that results from different base ISAs.  So, I rewrote the instruction-decoding scheme to make the overlap list priority-ordered.  Instructions not on the list are given lowest priority (but need not be explicitly included).

The simplest way to do so was to get rid of the perf optimization that bubbles instruction-list elements to the front of the list.  This resulted in a noticeable perf loss (around 2x running 403.gcc), but it was easily recouped by statically sorting the list (via the Makefile) and improving the hit rate of the opcode cache by making it set-associative.


aswaterman:
I should add: although this turned out to be more complex than I had anticipated, I'm happy with the results.  The overlap list makes more sense now, and the move-to-front scheme always seemed a little hacky to me.

riscv/processor.cc
```diff
@@ -1196,11 +1173,29 @@ void processor_t::register_base_instructions()
         logged_rv32i_##name, \
         logged_rv64i_##name, \
         logged_rv32e_##name, \
-        logged_rv64e_##name}); \
-    }
+        logged_rv64e_##name});
+  #include "overlap_list.h"
+  #undef DECLARE_OVERLAP_INSN
+
+  // add all other instructions, in arbitrary order
```
jerryz123: This comment is misleading right? If performance relies on the sortedness of the extensions.

jerryz123 (approved):
Nice result that the sorting works.

riscv/processor.cc
aswaterman: Yeah.  I wrote that when I was thinking about the functional implications and before I looked into the performance.

PR #1688, opened 2024-06-11
triggers: implement tcontrol

YenHaoChen:
Implement Debug spec Section 5.7.6. Trigger Control (tcontrol). This commit lets tcontrol be read-only 0 if the number of triggers is 0.

The spec says the optional tcontrol is useful only if medeleg[3] (breakpoint) is read-only 0. Please note that Spike lets the medeleg[3] be read-write.
image

### DIFF
--- a/riscv/insn_template.h
+++ b/riscv/insn_template.h
@@ -8,4 +8,5 @@
 #include "specialize.h"
 #include "tracer.h"
 #include "v_ext_macros.h"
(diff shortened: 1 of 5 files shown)


rtwfroody (approved):
Looks correct to me. Thanks for implementing this so quickly.

aswaterman (approved):
Thanks.  (And thanks @rtwfroody for the review)

PR #1695, opened 2024-06-19
Add several BF16 ops to SoftFloat

aswaterman:
I recently got a request to add basic BF16 arithmetic to SoftFloat.  Although I don't object, I want to keep the implementations as simple as possible, so I've implemented them in terms of FP32 ops.

This is straightforward enough for add/sub/mul/div/sqrt, where double rounding is correct.  It's tricker for FMA, where we need to perform the addition using round-to-odd to account for double rounding.  (We then need to correct for -0.0 outputs when rounding down.)

I've added these to Spike first because it was easier for me to test them this way.  If the PR is accepted, I'll backport them to https://github.com/ucb-bar/berkeley-softfloat-3.

### DIFF
--- a/softfloat/bf16_add.c
+++ b/softfloat/bf16_add.c
@@ -1,0 +1,50 @@
+
+/*============================================================================
+
+This C source file is part of the SoftFloat IEEE Floating-Point Arithmetic
+Package, Release 3d, by John R. Hauser.
+
+Copyright 2011, 2012, 2013, 2014, 2015, 2016 The Regents of the University of
+California.  All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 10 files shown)


jerryz123 (approved):

nadime15:
Hi @aswaterman are there any plans to push your changes back to https://github.com/ucb-bar/berkeley-softfloat-3?

The [Sail RISC-V model](https://github.com/riscv/sail-riscv/) is about to update the softfloat library, and I think it would be nice if we had easy access to the same code.

aswaterman:
@nadime15 I haven't had the time, but if someone else wants to do that, I'd support it.

Note there has been at least one bug fix since this PR; whoever backports it should use the most recent source code.

PR #1701, opened 2024-06-20
Correctly determine vector capability from v/zve/zvl ISA strings, remove --varch

jerryz123:
Previously, Spike did not implement ISA parsing for Zve/Zvl. Furthermore, V was being used as a catch-all to enable all vector features, instead of its actual intended use to signify application-profile vector support.

This PR updates the V/Zve/Zvl parsing to properly capture the meaning of those extension strings, and removes the old `--varch` settings. Non-V, but with-vectors (Ex: `rv32imfd_zvl64b_zve32f`) implementations can now be simulated.

This also removes the `vstartalu` command-line option. Is this useful to keep around?

Requests riscv-tests update https://github.com/riscv-software-src/riscv-tests/pull/567

### DIFF
--- a/disasm/isa_parser.cc
+++ b/disasm/isa_parser.cc
@@ -457,6 +457,10 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
   }
 #endif
 
+  if (vlen > 4096) {
+    bad_isa_string(str, "Spike does not currently support VLEN > 4096b");
+  }
+
   if ((vlen != 0) ^ (elen != 0)) {
     bad_isa_string(str, "Invalid Zvl/Zve configuration");
   }
--- a/disasm/isa_parser.cc
+++ b/disasm/isa_parser.cc
@@ -101,9 +101,6 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
       if (ext_str == "zfh")
         extension_table[EXT_ZFH] = true;
     } else if (ext_str == "zvfh" || ext_str == "zvfhmin") {
-      if (!extension_table['V'])
-        bad_isa_string(str, ("'" + ext_str + "' extension requires 'V'").c_str());
-
       extension_table[EXT_ZVFHMIN] = true;
 
       if (ext_str == "zvfh") {
--- a/disasm/isa_parser.cc
+++ b/disasm/isa_parser.cc
@@ -38,6 +38,11 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
   else
     bad_isa_string(str, "ISA strings must begin with RV32 or RV64");
 
+  vlen = 0;
+  elen = 0;
+  zvf = false;
+  zvd = false;
+
   switch (isa_string[4]) {
     case 'g':
       // G = IMAFD_Zicsr_Zifencei, but Spike includes the latter two
@@ -70,7 +75,8 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
     }
 
     switch (*p) {
-      case 'v': // even rv32iv implies double float
+      case 'v': vlen = 128; elen = 64; zvf = true; zvd = true;
+                // even rv32iv implies double float
       case 'q': extension_table['D'] = true;
                 // Fall through
       case 'd': extension_table['F'] = true;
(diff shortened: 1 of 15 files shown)


riscv/v_ext_macros.h
```diff
@@ -1948,17 +1948,17 @@ reg_t index[P.VU.vlmax]; \
   switch (P.VU.vsew) { \
     case e16: \
       { VI_VFP_CVT_LOOP(CVT_INT_TO_FP_PARAMS(16, 16, sign), \
-        { p->extension_enabled(EXT_ZVFH); }, \
+        { require(p->extension_enabled(EXT_ZVFH)); },   \
```
jerryz123, line 1951: I think this was a bug in the old code, the macro expansion wouldn't wrap the `p->extension_enabled` with a `require`, from what I could tell.

aswaterman: Yeah, looks that way.

disasm/isa_parser.cc
```diff
@@ -314,6 +320,31 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
...
     } else if (ext_str == "zicfiss") {
       extension_table[EXT_ZICFISS] = true;
+    } else if (ext_str.substr(0, 3) == "zvl") {
+      try {
+        vlen = std::stol(ext_str.substr(3, ext_str.size() - 4));
+      } catch (std::logic_error& e) {
+        vlen = 0;
+      }
+      if ((vlen & (vlen - 1)) != 0 || vlen < 32)
+        bad_isa_string(str, ("Invalid Zvl string: " + ext_str).c_str());
```
aswaterman, line 328: Strictly speaking, something like `V_Zvl64b` is legal.  Zvl64b means VLEN >= 64, not VLEN = 64; similarly, Zve64f doesn't mean that Zve64d is absent.  I think we can use Zvl to increase VLEN, rather than overriding it, as in something like

```
      int new_vlen = 0;
      try {
        new_vlen = std::stol(ext_str.substr(3, ext_str.size() - 4));
      } catch (std::logic_error& e) {}
      if ((new_vlen & (new_vlen - 1)) != 0 || new_vlen < 32)
        bad_isa_string(str, ("Invalid Zvl string: " + ext_str).c_str());
      vlen = std::max(vlen, new_vlen);
```

disasm/isa_parser.cc
```diff
@@ -314,6 +320,31 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
...
+    } else if (ext_str.substr(0, 3) == "zve") {
+      try {
+        elen = std::stol(ext_str.substr(3, ext_str.size() - 4));
+      } catch (std::logic_error& e) {
+        elen = 0;
+      }
+      if (ext_str.substr(5) == "d") {
+        zvd = true; zvf = true;
+      } else if (ext_str.substr(5) == "f") {
+        zvd = false; zvf = true;
```
aswaterman: Similarly, Zve64f doesn't imply Zve64d is absent, so these could become `|=` instead of `=`.

disasm/isa_parser.cc
```diff
@@ -455,6 +455,31 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
 		     "extensions are incompatible with WORDS_BIGENDIAN setups.");
   }
 #endif
+
+  if (extension_table['V'] && (vlen < 128 || elen != 64 || !zvd)) {
+    bad_isa_string(str, "V extension requires VLEN >= 128 and Zve64d");
+  }
```
aswaterman, line 462: With the changes I mentioned on the previous commit, I think this can go away.

disasm/isa_parser.cc
jerryz123: I fixed this in a later commit in the series, I'll just merge the fix with the original change.

aswaterman: Ah, okay, obviously I was reviewing the commits chronologically.

riscv/csrs.cc
```diff
@@ -1428,8 +1428,6 @@ vector_csr_t::vector_csr_t(processor_t* const proc, const reg_t addr, const reg_
 
 void vector_csr_t::verify_permissions(insn_t insn, bool write) const {
   require_vector_vs;
-  if (!proc->extension_enabled('V'))
```
aswaterman, line 1432: I would've expected this to be replaced with `!has_any_vector()`, but I'm guessing you left it out because some other mechanism has the same effect?

riscv/decode_macros.h
```diff
@@ -171,7 +171,6 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
 #define require_vector(alu) \
   do { \
     require_vector_vs; \
-    require_extension('V'); \
```
aswaterman, line 174: Same question as https://github.com/riscv-software-src/riscv-isa-sim/pull/1701#discussion_r1648223166

...but I guess the mechanism might be different here?

aswaterman (approved):
LGTM, provided my last couple questions have already been addressed.

Thanks for handling this one.

riscv/csrs.cc
jerryz123: I think `require_vector_vs` would do it, since `status_csr_t` masks the vs bits with `has_any_vector()` now. It's a roundabout check

aswaterman: I was thinking that VS might be implemented even when `!has_any_vector()`, which is permitted by the ISA.  I forgot we don't do that in Spike.  So this is fine.

chihminchao:
"This also removes the vstartalu command-line option. Is this useful to keep around?"
It is implementation-defined behavior and allow spike to match the resul with differenet implemenation.   Spike is one of RISCV implementations.  It is ok to keep it on one of the legal choices.

riscv/processor.cc
```diff
@@ -55,7 +55,10 @@ processor_t::processor_t(const isa_parser_t *isa, const cfg_t *cfg,
   }
 #endif
 
-  parse_varch_string(cfg->varch);
+  VU.VLEN = isa->get_vlen();
+  VU.ELEN = isa->get_elen();
+  VU.vlenb = isa->get_vlen() / 8;
+  VU.vstart_alu = 1;
```
aswaterman: Thinking about this some more, I'm questioning whether this is the right choice to pick.  There are cases where supporting vstart>0 adds complexity, and it's safe to say that there will be more over time.  So, why pick the option that makes Spike harder to maintain?

aswaterman (changes requested):
(Placing a hold on this until we resolve the vstart_alu matter.)

jerryz123:
I actually missed this commit when going back into the history: https://github.com/riscv-software-src/riscv-isa-sim/commit/c6ec6f3582eb2ae106dff95e569b7da927910629, yeah the default to 0 makes more sense.

Is there value in preserving support in the code for supporting `start_alu=1`?.
If so, then does that support need to be exposed via cmdline flag?  I'm not sure what the use case @chihminchao is envisioning for the flag, certainly there are many other implementation choices regarding vector execution that Spike does not support or expose flags for.

riscv/processor.cc
jerryz123: Updated to set `start_alu=0`

aswaterman (approved):
I don't personally see much value in continuing to support `vstart_alu`=1, but I'd rather leave it around in the codebase for the time being, so we can see whether any users are upset by the loss of functionality.  If, after a few months, no one seems to care, then we should remove it.

PR #1703, opened 2024-06-21
Fix build of softfloat.h when threads.h is not available.

dybv-sc:
softfloat.h happens to be included both in C and C++ modules and these languages have different keywords for thread local attribute.


aswaterman:
I agree this approach seems cleanest, but it does not work in Mac OS, which does not support linkage between `_Thread_local` and `thread_local`.

But I agree it _should_ work on Mac OS, and so maybe the right fix is to special-case the broken OS, and use the cleaner approach for Linux etc.

aswaterman:
I've created an alternate PR #1704, so I'll close this one.  Thanks for raising the issue, though.

PR #1704, opened 2024-06-21
Fix C/C++ thread-local linkage differently

aswaterman:
Just admit Mac OS is broken, so explicitly special-case it.

See #1689 and #1703


jerryz123 (approved):
Seems to work.

PR #1709, opened 2024-06-26
Add insn cmd to interactive debug mode

abejgonzalez:
This PR adds a new interactive debug mode cmd of `insn` with the same API as the `pc` command to print out the instruction associated with the PC of a specific hart. This command can be used with the `until*` commands to run a binary "until" the "insn" value is a specific value.

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -277,6 +277,7 @@ void sim_t::interactive()
   funcs["fregs"] = &sim_t::interactive_fregs;
   funcs["fregd"] = &sim_t::interactive_fregd;
   funcs["pc"] = &sim_t::interactive_pc;
+  funcs["insn"] = &sim_t::interactive_insn;
   funcs["priv"] = &sim_t::interactive_priv;
   funcs["mem"] = &sim_t::interactive_mem;
   funcs["str"] = &sim_t::interactive_str;
@@ -367,6 +368,7 @@ void sim_t::interactive_help(const std::string& cmd, const std::vector<std::stri
     "fregd <core> <reg>              # Display double precision <reg> in <core>\n"
     "vreg <core> [reg]               # Display vector [reg] (all if omitted) in <core>\n"
     "pc <core>                       # Show current PC in <core>\n"
+    "insn <core>                     # Show current instruction corresponding to PC in <core>\n"
     "priv <core>                     # Show current privilege level in <core>\n"
     "mem [core] <hex addr>           # Show contents of virtual memory <hex addr> in [core] (physical memory <hex addr> if omitted)\n"
     "str [core] <hex addr>           # Show NUL-terminated C string at virtual address <hex addr> in [core] (physical address <hex addr> if omitted)\n"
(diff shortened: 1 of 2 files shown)


aswaterman (commented):
This looks like a reasonable addition to me, but do you think it makes sense to enhance it by printing out the disassembled instruction after printing out its opcode?  I guess it doesn't matter much since the main purpose is the `until` aspect, not the printing aspect, so I'll leave it up to you.

abejgonzalez:
> This looks like a reasonable addition to me, but do you think it makes sense to enhance it by printing out the disassembled instruction after printing out its opcode? I guess it doesn't matter much since the main purpose is the `until` aspect, not the printing aspect, so I'll leave it up to you.

It would be nice. I have a commit ready to add it but I'm unsure of the format for output. Right now it looks like the following:

```bash
(spike) insn 0
0x0202859300000297 auipc   t0, 0x0
```

If there is no preference on the format (i.e. `<HEX> <DISASM>`) then I can commit it.

abejgonzalez:
> > This looks like a reasonable addition to me, but do you think it makes sense to enhance it by printing out the disassembled instruction after printing out its opcode? I guess it doesn't matter much since the main purpose is the `until` aspect, not the printing aspect, so I'll leave it up to you.
> 
> It would be nice. I have a commit ready to add it but I'm unsure of the format for output. Right now it looks like the following:
> 
> ```shell
> (spike) insn 0
> 0x0202859300000297 auipc   t0, 0x0
> ```
> 
> If there is no preference on the format (i.e. `<HEX> <DISASM>`) then I can commit it.

Done.

riscv/interactive.cc
```diff
@@ -448,6 +452,50 @@ void sim_t::interactive_pc(const std::string& cmd, const std::vector<std::string
...
+    case 2:
+    case 6:
+      val = mmu->load<uint16_t>(addr);
+      break;
+    default:
+      val = mmu->load<uint8_t>(addr);
+      break;
+  }
+  return val;
+}
```
abejgonzalez: Currently this is a bit duplicated with `get_pc` and `get_mem`, but this is significantly faster since reusing those functions requires you to re-create a `vector<string>`'s to use them.

aswaterman: I think it's OK... no more sinful than everything else in `interactive.cc`.

aswaterman:
Yeah, that's fine.

Can you squash these commits into one?

aswaterman:
Also note that CI is broken until #1710 is merged.

abejgonzalez:
> Yeah, that's fine.
> 
> Can you squash these commits into one?

Will do.

riscv/interactive.cc
aswaterman: Actually, can you do some light refactoring?  Hoist this out as its own function `static reg_t load(mmu_t* mmu, reg_t addr)` so you can share the code between the two places without perf loss.

abejgonzalez: Done.

aswaterman (approved):
LGTM.

PR #1710, opened 2024-06-26
Fix riscv-tests commit for CI

jerryz123:
https://github.com/riscv-software-src/riscv-tests/pull/567 was merged by squash accidentally, so the referenced commit was lost


aswaterman (approved):
Oops, I should've realized that.

PR #1713, opened 2024-06-27
Add flag to disable ns16550 device

abejgonzalez:
Sometimes, it makes sense to use HTIF (and only HTIF) as a console device (for example when booting Linux). This allows users to disable the default console device (removing it from the DTS).

### DIFF
--- a/riscv/dts.cc
+++ b/riscv/dts.cc
@@ -13,6 +13,7 @@
 #include <sys/types.h>
 
 std::string make_dts(size_t insns_per_rtc_tick, size_t cpu_hz,
+                     bool disable_ns16550,
                      reg_t initrd_start, reg_t initrd_end,
                      const char* bootargs,
                      size_t pmpregions,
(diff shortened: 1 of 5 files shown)


riscv/dts.cc
```diff
@@ -30,8 +31,11 @@ std::string make_dts(size_t insns_per_rtc_tick, size_t cpu_hz,
          "  #size-cells = <2>;\n"
          "  compatible = \"ucbbar,spike-bare-dev\";\n"
          "  model = \"ucbbar,spike-bare\";\n"
-         "  chosen {\n"
-         "    stdout-path = &SERIAL0;\n";
+         "  chosen {\n";
+  if (!disable_ns16550)
+    s << "    stdout-path = &SERIAL0;\n";
+  else
+    s << "    stdout-path = &HTIF;\n";
```
abejgonzalez, line 38:
```suggestion
  else
    s << "    stdout-path = &HTIF;\n";
```

spike_main/spike.cc
```diff
@@ -64,6 +64,7 @@ static void help(int exit_code = 1)
   fprintf(stderr, "  --dump-dts            Print device tree string and exit\n");
   fprintf(stderr, "  --dtb=<path>          Use specified device tree blob [default: auto-generate]\n");
   fprintf(stderr, "  --disable-dtb         Don't write the device tree blob into memory\n");
+  fprintf(stderr, "  --disable-ns16550     Disable default ns16550 UART device\n");
```
abejgonzalez, line 67: I'm not sure this should be called "UART device" (maybe just serial or console device)?

jerryz123:
Is configuring spike via DTS insufficient for disabling the ns16550?

abejgonzalez:
> Is configuring spike via DTS insufficient for disabling the ns16550?

It is sufficient, I just figured this was a nicer way to disable the device without having to provide a modified DTS. If the preferred way is to use a custom DTS then I'll close this PR.

jerryz123:
The preferred way is to pass a modified DTS, to avoid blowing up the number of flags to keep track of.

PR #1714, opened 2024-06-28
Fix insn interactive command (catch/print trap, use proper access func)

abejgonzalez:
#1709 Added the `insn` interactive mode command. However, if fetching the instruction failed it would print "0xBad or missing arguments for command" since the exception logic blindly catches any exception (without distinguishing what type it is). This adds a `try` `catch` around the printout for more descriptive text. Additionally, this uses the `access_icache` `mmu` function instead of blindly accessing memory (which seems to be the main way to fetch an instruction). Finally, this inlines the `load()` function introduced in #1709 (originally this was added to dedup some shared code between the `get_mem` and `get_insn` functions that is no longer needed).

### DIFF
--- a/riscv/interactive.cc
+++ b/riscv/interactive.cc
@@ -452,37 +452,16 @@ void sim_t::interactive_pc(const std::string& cmd, const std::vector<std::string
       << zext(get_pc(args), max_xlen) << std::endl;
 }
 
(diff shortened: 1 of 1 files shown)


jerryz123 (approved):
LGTM

PR #1715, opened 2024-06-28
Don't print vregs in interactive mode if no V extension exists

abejgonzalez:
#1701 changed Spike's ISA parsing to avoid setting a default VLEN/... if no V extension was found. This broke the interactive `vreg N` command (which assumed VLEN and ELEN would be set by default and would be non-zero) by throwing a `0/0` FP exception. This wraps existing logic in an `if` and prints a nice message saying no V extensions supported.


jerryz123 (approved):
Oops, good catch

PR #1718, opened 2024-07-02
Implement pointer masking

YenHaoChen:
This PR implements pointer masking.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -613,8 +613,8 @@ void mmu_t::register_memtracer(memtracer_t* t)
   tracer.hook(t);
 }
 
-reg_t mmu_t::get_pmlen(bool effective_virt, reg_t effective_priv) const {
-  if (!proc || proc->get_xlen() != 64 || (in_mprv() && (proc->state.sstatus->read() & MSTATUS_MXR)))
+reg_t mmu_t::get_pmlen(bool effective_virt, reg_t effective_priv, xlate_flags_t flags) const {
+  if (!proc || proc->get_xlen() != 64 || (in_mprv() && (proc->state.sstatus->read() & MSTATUS_MXR)) || flags.hlvx)
     return 0;
 
   reg_t pmm = 0;
@@ -644,7 +644,7 @@ mem_access_info_t mmu_t::generate_access_info(reg_t addr, access_type type, xlat
       mode = get_field(proc->state.hstatus->read(), HSTATUS_SPVP);
     }
   }
-  reg_t pmlen = get_pmlen(virt, mode);
+  reg_t pmlen = get_pmlen(virt, mode, xlate_flags);
   reg_t satp = proc->state.satp->readvirt(virt);
   bool is_physical_addr = mode == PRV_M || get_field(satp, SATP64_MODE) == SATP_MODE_OFF;
   reg_t transformed_addr = is_physical_addr ? zext(addr, 64 - pmlen) : sext(addr, 64 - pmlen);
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -5,6 +5,7 @@
 #include "arith.h"
 #include "simif.h"
 #include "processor.h"
+#include "decode_macros.h"
 
 mmu_t::mmu_t(simif_t* sim, endianness_t endianness, processor_t* proc)
  : sim(sim), proc(proc),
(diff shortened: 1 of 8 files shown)


YenHaoChen:
The pointer masking takes effect on all loads and stores. People may want to evaluate the performance impact.

aswaterman (changes requested):
We want to handle this via the TLB rather than by adding more instructions to the critical code path.

(A single address would then occupy multiple entries in the TLB, one for each pointer-masking value that is used, but that is preferable.)

YenHaoChen:
> We want to handle this via the TLB rather than by adding more instructions to the critical code path.
> 
> (A single address would then occupy multiple entries in the TLB, one for each pointer-masking value that is used, but that is preferable.)

I am working on this and need a few days.

I aim to have a single entry for each address at runtime and flush TLB upon updating the pointer-masking value. This way, we don't need to obtain the pointer-masking value before looking up TLB.

(We must obtain the pointer-masking value for TLB lookup if we allow each address to occupy multiple entries. Obtaining the value before TLB lookup would add more instructions to the critical code path and hamper system performance.)

Please let me know if this proposal would satisfy your requirements.

aswaterman:
My requirement is that the code sequence for TLB hits is completely unchanged, and all of the new code is on the TLB refill path.

YenHaoChen:
@aswaterman I revised the PR and included two performance commits that moved the code out of the code sequence for TLB hits.

riscv/mmu.cc
```diff
@@ -224,39 +225,41 @@ void mmu_t::load_slow_path_intrapage(reg_t len, uint8_t* bytes, mem_access_info_
 void mmu_t::load_slow_path(reg_t addr, reg_t len, uint8_t* bytes, xlate_flags_t xlate_flags)
```
aswaterman: I think there is too much risk of someone accidentally using the short name `addr` instead of using `transformed_addr` within the body of this function.  So, maybe rename `addr` to `original_addr` or something?

Same comment applies to `store_slow_path`.

aswaterman (commented):
Looks OK to me, but I would appreciate a sanity check from @jerryz123.

riscv/mmu.h
```diff
@@ -4,6 +4,7 @@
 #define _RISCV_MMU_H
 
 #include "decode.h"
+#include "decode_macros.h"
```
jerryz123: `mmu.h` should not depend on `decode_macros.h`, as `mmu.h` is a installed header, while `decode_macros.h` is not, and cannot be.

I think `generate_access_info` can probably be implemented in `mmu.cc`, which does include `decode_macros.h`

riscv/mmu.h
```diff
@@ -71,9 +73,34 @@ class mmu_t
   std::map<reg_t, reg_t> alloc_cache;
   std::vector<std::pair<reg_t, reg_t >> addr_tbl;
 
+  reg_t get_pmlen(bool effective_virt, reg_t effective_priv, xlate_flags_t flags) const {
```
jerryz123: How does this handle the case of `MPRV && MXR`?

jerryz123 (changes requested):

riscv/mmu.h
aswaterman: @jerryz123 I guess `ci-tests/testlib.c` should include all of the installed headers to catch this sort of error in CI?  I'll create a PR for this.

riscv/mmu.h
YenHaoChen: Thanks for the catch. I overlooked the condition.
image

riscv/mmu.h
YenHaoChen: I will revise the PR after https://github.com/riscv-software-src/riscv-isa-sim/pull/1730 for the CI test.

aswaterman: #1730 is merged, so I suggest rebasing from master.

riscv/mmu.cc
YenHaoChen: I added https://github.com/riscv-software-src/riscv-isa-sim/pull/1718/commits/04eb65437bf3f6c7a32b7a276a46a74c8cc3bba6 and https://github.com/riscv-software-src/riscv-isa-sim/pull/1718/commits/4a4ab850de34a99489ac9c47702ef815e13b6889 to rename the function parameter ```addr``` to ```original_addr``` in ```load_slow_path``` and ```store_slow_path```.

I don't know if we should also rename the local variable ```addr``` in ```load_slow_path_intrapage``` and ```store_slow_path_intrapage```. (Additionally, the variable ```vaddr``` in ```mem_access_info_t```.)

riscv/mmu.h
YenHaoChen: Commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1718/commits/95b65c6c5e59af2b204a374410c332eeae7d0510 addresses the ```MPRV && MXR``` condition.

jerryz123 (approved):

aswaterman (approved):

aswaterman:
@YenHaoChen please ping me after you resolve the conflict, and I'll merge the PR.

YenHaoChen:
@aswaterman I rebased the PR and resolved the conflict.

aswaterman (approved):

PR #1726, opened 2024-07-15
Deprecate dcsr.halt

YenHaoChen:
The debug spec 1.0.0-rc3 deprecates the dcsr.halt and lets the bit become read-only dcsr.nmip.

To my knowledge, the dcsr_csr_t::halt variable is for halt_on_reset, i.e., the -H flag for the debugger, meanwhile the dcsr.nmip is to indicate a hardware error condition. The halt_on_reset debugging feature does not necessarily mean a hardware error condition. Thus, Spike should decouple the halt variable and the dcsr.nmip. 
@rtwfroody Would you share some words about this?

This commit separates the halt variable from the dcsr.nmip and designates it as an internal variable for halt_on_reset (-H flag). Additionally, this commit removes the notifying comment about the deprecated dcsr.halt in the main loop.

Update 2024.07.17:
@rtwfroody clarified that the debugger should not depend the writability of dcsr.halt (dcsr.nmip) in https://github.com/riscv-software-src/riscv-isa-sim/issues/1727#issuecomment-2231328997.


YenHaoChen:
The PR does not alter any behavior to the user due to:
+ The dcsr.nmip is read-only 0 in the dcsr_csr_t::read().
+

aswaterman (approved):

PR #1729, opened 2024-07-17
Fix: Vector CSRs exist without any vector extension since a484f6e

YenHaoChen:
The require_vector_vs, i.e., sstatus_csr_t::enabled(SSTATUS_VS), was expected to provide has_any_vector() check [1]. Unfortunately, a previous commit [2] made the sstatus_csr_t::enabled(SSTATUS_VS) true without any vector extension.

The previous commit [2] was for P-extension, which has been removed from Spike [3].

This commit reverts the commit [2] and corrects require_vector_vs [1].

[1] https://github.com/riscv-software-src/riscv-isa-sim/pull/1701/commits/a484f6efc5f50836bb8d846180dfbb9786d09ae2
[2] https://github.com/riscv-software-src/riscv-isa-sim/commit/66c4853bdc5b22bf4c4b364218c713e3f1e487f3
[3] https://github.com/riscv-software-src/riscv-isa-sim/pull/1660

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -642,11 +642,6 @@ bool sstatus_csr_t::enabled(const reg_t which) {
     if (!state->v || (virt_sstatus->read() & which) != 0)
       return true;
   }
-
-  // If the field doesn't exist, it is always enabled. See #823.
-  if (!orig_sstatus->field_exists(which))
-    return true;
-
   return false;
 }
 
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1446,7 +1446,7 @@ vector_csr_t::vector_csr_t(processor_t* const proc, const reg_t addr, const reg_
 }
 
 void vector_csr_t::verify_permissions(insn_t insn, bool write) const {
-  require_vector_vs;
+  require(proc->any_vector_extensions() && STATE.sstatus->enabled(SSTATUS_VS));
   basic_csr_t::verify_permissions(insn, write);
 }
 
@@ -1467,7 +1467,7 @@ vxsat_csr_t::vxsat_csr_t(processor_t* const proc, const reg_t addr):
 }
 
 void vxsat_csr_t::verify_permissions(insn_t insn, bool write) const {
-  require_vector_vs;
+  require(proc->any_vector_extensions() && STATE.sstatus->enabled(SSTATUS_VS));
   masked_csr_t::verify_permissions(insn, write);
 }
 
--- a/riscv/v_ext_macros.h
+++ b/riscv/v_ext_macros.h
@@ -1202,10 +1202,10 @@ reg_t index[P.VU.vlmax]; \
 
 #define VI_LD(stride, offset, elt_width, is_mask_ldst) \
   const reg_t nf = insn.v_nf() + 1; \
+  VI_CHECK_LOAD(elt_width, is_mask_ldst); \
   const reg_t vl = is_mask_ldst ? ((P.VU.vl->read() + 7) / 8) : P.VU.vl->read(); \
   const reg_t baseAddr = RS1; \
   const reg_t vd = insn.rd(); \
-  VI_CHECK_LOAD(elt_width, is_mask_ldst); \
   for (reg_t i = 0; i < vl; ++i) { \
     VI_ELEMENT_SKIP; \
     VI_STRIP(i); \
@@ -1220,12 +1220,12 @@ reg_t index[P.VU.vlmax]; \
 
 #define VI_LD_INDEX(elt_width, is_seg) \
   const reg_t nf = insn.v_nf() + 1; \
+  VI_CHECK_LD_INDEX(elt_width); \
   const reg_t vl = P.VU.vl->read(); \
   const reg_t baseAddr = RS1; \
   const reg_t vd = insn.rd(); \
   if (!is_seg) \
     require(nf == 1); \
-  VI_CHECK_LD_INDEX(elt_width); \
   VI_DUPLICATE_VREG(insn.rs2(), elt_width); \
   for (reg_t i = 0; i < vl; ++i) { \
     VI_ELEMENT_SKIP; \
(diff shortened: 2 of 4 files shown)


riscv/csrs.cc
```diff
@@ -638,11 +638,7 @@ void sstatus_csr_t::dirty(const reg_t dirties) {
 }
 
 bool sstatus_csr_t::enabled(const reg_t which) {
-  if ((orig_sstatus->read() & which) != 0) {
-    if (!state->v || (virt_sstatus->read() & which) != 0)
-      return true;
-  }
-  return false;
+  return ((orig_sstatus->read() & which) != 0) && (!state->v || (virt_sstatus->read() & which) != 0);
```
aswaterman: IMO the original version of this function is easier to read (even though it is verbose).

YenHaoChen: Remove the second commit.

aswaterman (commented):
I'd like @jerryz123's opinion on how to proceed, since he authored #1701.  An alternative might be to redefine `require_vector_vs` to call `any_vector_extensions()`.  I don't know the implications of that choice, though.

YenHaoChen:
I vote for explicitly calling ```any_vector_extensions()``` (instead of the current PR), which could improve clarity. I got confused tracing the ```require_vector_vs```.

jerryz123:
> An alternative might be to redefine require_vector_vs to call any_vector_extensions(). I don't know the implications of that choice, though.

I don't think this would be right, this would cause `require_vector_vs` to succeed even when VS is unset in a with-vectors configuration.

The two places where I see `require_vector_vs` being used are in vector CSRs, and in `require_vector` decode macro.

For the former, is the root cause that vector CSRs are being added to the csrmap even when `~any_vector_extensions()`? If so, I think we can fix that by just gating `VU.reset()` with `any_vector_extensions()`.

For the latter, is there any problem? The only way the `require_vector` macro is called is through the vector instruction functions. If those instructions exist in the instruction table, then `field_exists(SSTATUS_VS)` should be true.

YenHaoChen:
@jerryz123 You are correct that gating ```VU.reset()``` with ```any_vector_extensions()``` can raise an exception on accessing vector CSRs through CSR instructions. Unfortunately, I got a core dump on executing a vle8.v instruction after conditionalizing the ```VU.reset()```. Currently, vector instructions do not explicitly check ```require_vector``` before using the vector CSRs. I haven't done the coding yet, but adding require_vector to all vector instructions could fix the issue.

I believe @aswaterman proposed making ```require_vector_vs``` successful when both ```any_vector_extensions()``` and VS are true, i.e.,
```#define require_vector_vs   require(p->any_vector_extensions() && STATE.sstatus->enabled(SSTATUS_VS))```.
This looks promising and descriptive. (I believe this provides the exact behavior of reverting commit https://github.com/riscv-software-src/riscv-isa-sim/commit/66c4853bdc5b22bf4c4b364218c713e3f1e487f3) However, I cannot compile the above code after adding the ```p->any_vector_extensions()```. That is because both ```riscv/csrs.cc``` and ```riscv/insns/v*.h``` use the ```require_vector_vs```, where the variable names for the processor_t are different, i.e., ```proc``` and ```p```. Again, I haven't done the coding but feel duplicating the macro definition could solve the compilation error.

I want to invite comments from @scottj97 if applicable. To my understanding, the https://github.com/riscv-software-src/riscv-isa-sim/commit/66c4853bdc5b22bf4c4b364218c713e3f1e487f3 was for the condition that any extensions that need vector CSRs exist but not the general V extension. The new function, ```any_vector_extensions()```, provides a more flexible check for this condition. If so, do we still need the old commit?

**Comment by scottj97** [comment]
> I want to invite comments from @scottj97 if applicable.

Sorry, I'm not involved with this project anymore.

**Comment by jerryz123** [comment]
> `#define require_vector_vs   require(p->any_vector_extensions() && STATE.sstatus->enabled(SSTATUS_VS))`

This seems correct to me.

When the P-ext gets re-implemented, the handling of shared vxsat/VS can be done differently than before. Instantiation of all vector CSRs should be hoisted out of `vectorUnit_t` into `processor_t`, where adding `vxsat` can be a special case `if (any_vector_extensions() || any_p_extensions())`.  Similarly for `has_vs` in `base_status_csr_t`. I believe that would allow for reverting https://github.com/riscv-software-src/riscv-isa-sim/commit/66c4853bdc5b22bf4c4b364218c713e3f1e487f3. 

Since no P-exts are implemented now, reverting https://github.com/riscv-software-src/riscv-isa-sim/commit/66c4853bdc5b22bf4c4b364218c713e3f1e487f3 now should be ok too.

>  I haven't done the coding but feel duplicating the macro definition could solve the compilation error.

I think just explicitly writing out `require_vector_vs` in `csrs.cc` is the clearest. Relying on `decode_macros.h` in that context seems to be an accidental pattern.

**Comment by YenHaoChen** [comment]
Commits https://github.com/riscv-software-src/riscv-isa-sim/pull/1729/commits/cefa747e9a241e1fe7a7dcd32108d3a889cbed6f and https://github.com/riscv-software-src/riscv-isa-sim/pull/1729/commits/67933ec0c397124aa27f558c973673a5d74a39e5 check ```any_vector_extensions()``` in ```require_vector_vs``` and vector CSRs separately as suggested by @jerryz123 .

**Comment by YenHaoChen** [comment]
Commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1729/commits/cc8c2fc923c62132f208d29af745b2db4aecc75a provides the prototype by @jerryz123, where instantiation of all vector CSRs depends on ```any_vector_extensions()```. The commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1729/commits/9fd40e8bc8788074f37e850183eaab8f585fe646 prevents core dump by checking the extension's existence before using the vector CSRs.

**Review by jerryz123 — APPROVED** [review]
This LGTM, awaiting @aswaterman 's approval

File: riscv/v_ext_macros.h [hunk 2]
```diff
@@ -1202,10 +1202,10 @@ reg_t index[P.VU.vlmax]; \
 
 #define VI_LD(stride, offset, elt_width, is_mask_ldst) \
   const reg_t nf = insn.v_nf() + 1; \
+  VI_CHECK_LOAD(elt_width, is_mask_ldst); \
   const reg_t vl = is_mask_ldst ? ((P.VU.vl->read() + 7) / 8) : P.VU.vl->read(); \
   const reg_t baseAddr = RS1; \
   const reg_t vd = insn.rd(); \
-  VI_CHECK_LOAD(elt_width, is_mask_ldst); \
```
aswaterman, line 1208: Don't you need to make the analogous change here?  https://github.com/riscv-software-src/riscv-isa-sim/blob/cc8c2fc923c62132f208d29af745b2db4aecc75a/riscv/v_ext_macros.h#L1313

aswaterman: And here? https://github.com/riscv-software-src/riscv-isa-sim/blob/cc8c2fc923c62132f208d29af745b2db4aecc75a/riscv/v_ext_macros.h#L1223

aswaterman: And here? https://github.com/riscv-software-src/riscv-isa-sim/blob/cc8c2fc923c62132f208d29af745b2db4aecc75a/riscv/v_ext_macros.h#L1277

YenHaoChen: @aswaterman 
I updated PR accordingly.

aswaterman (approved):
LGTM now, thanks.

PR #1732, opened 2024-07-17
Fix zkr

chihminchao:


riscv/csrs.cc
```diff
@@ -1440,6 +1448,17 @@ void seed_csr_t::verify_permissions(insn_t insn, bool write) const {
...
   csr_t::verify_permissions(insn, write);
+
+  unsigned priv = state->prv == PRV_S && !state->v ? PRV_HS : state->prv;
+  if (state->v) {
+    if (state->mseccfg->get_sseed() && write)
+      throw trap_virtual_instruction(insn.bits());
+    else
+      throw trap_illegal_instruction(insn.bits());
+  } else if ((priv == PRV_U && !state->mseccfg->get_useed()) ||
+      ((priv == PRV_S || priv == PRV_HS) && !state->mseccfg->get_sseed())) {
```
aswaterman: `(priv == PRV_S || priv == PRV_HS)` is equivalent to `state->prv == PRV_S`.

It looks like you can delete the new variable `priv` and simply use `state->prv` everywhere.

binno: Yes, I will revise it later, thanks

binno:
@aswaterman , please help to review it again, thanks

aswaterman (approved):

PR #1734, opened 2024-07-17
sstc: only check henvcfg only in V mode for SSTC extension

chihminchao:


aswaterman (changes requested):
I don't follow the logic here.  This looks like we will only recognize VS-mode timer interrupts when we are in VS-/VU-mode.  But what if the timer interrupt becomes pending while we are in HS-mode?  Of course we should not take the interrupt, but we still need to record it, so that we will take the interrupt once we return to VS-/VU-mode.

@ved-rivos can you check my logic?

ved-rivos:
Agree. It should reflect in hip.

chihminchao:
It is not necessary change.

PR #1740, opened 2024-07-22
fcvtmod.w.d: Not update fflags if no exception flag, e.g., exp == frac == 0

YenHaoChen:
The commit proposes not updating the fflags if there is no exception flag at all. For instance, when exp == frac == 0, there is no exception flag. 

Other floating-point instructions do not update the fflags when there is no exception. This commit follows the habit.

### DIFF
--- a/riscv/decode_macros.h
+++ b/riscv/decode_macros.h
@@ -211,10 +211,12 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
     } \
   } while (0);
 
-#define set_fp_exceptions ({ if (softfloat_exceptionFlags) { \
-                               STATE.fflags->write(STATE.fflags->read() | softfloat_exceptionFlags); \
-                             } \
-                             softfloat_exceptionFlags = 0; })
+#define raise_fp_exceptions(flags) do { if (flags) STATE.fflags->write(STATE.fflags->read() | (flags)); } while (0);
+#define set_fp_exceptions \
+  do { \
+    raise_fp_exceptions(softfloat_exceptionFlags); \
+    softfloat_exceptionFlags = 0; \
+  } while (0);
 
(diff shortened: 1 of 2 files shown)


aswaterman (approved):
Indeed `fcvtmod.w.h` is the only instruction that directly accesses the `fflags`.  But I'd propose a further improvement: define a new macro `raise_fp_exceptions(flags)` that performs the operation `if (flags) STATE.fflags->write(STATE.fflags->read() | (flags))`.  Then invoke `raise_fp_exceptions((inexact ? softfloat_flag_inexact : 0) | (invalid ? softfloat_flag_invalid : 0))`.  This is cleaner than messing with `softfloat_exceptionFlags`.

We can also redefine `set_fp_exceptions` to use `raise_fp_exceptions`: specifically, `set_fp_exceptions` can perform `do { raise_fp_exceptions(softfloat_exceptionFlags); softfloat_exceptionFlags = 0; } while (0)` or something.

aswaterman (changes requested):
LMK once you've considered my counter-proposal.

YenHaoChen:
@aswaterman 
The first commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1740/commits/1342c687f2b0429b3b185c9f11080300ca14529e is a refactoring commit, which updates the ```set_fp_exceptions``` with the new ```raise_fp_exceptions(flags)``` in ```riscv/decode_macros.h```.
The second commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1740/commits/e86e653ef35fcd6de8bb2ebe701dccdb9f30a6c7 makes the ```riscv/insns/fcvtmod_w_d.h``` not update the ```fflags``` using the ```raise_fp_exceptions(flags)```.

aswaterman (approved):
Thanks, this version LGTM.

PR #1746, opened 2024-07-25
svpbmt: don't reset [mh]envcfg.pbmt to 1

chihminchao:
The part is introducd by ea70a93 to keep backward compatiable but the behavior is not mentioned and defined in spec. The patch remove this initialization part of pbmt in [mh]envcfg


aswaterman (approved):
LGTM.

PR #1749, opened 2024-07-31
vnclip.wx: Check if there is any vector extension before using vector CSRs

YenHaoChen:
Since https://github.com/riscv-software-src/riscv-isa-sim/pull/1729/commits/707b1f24841fbe80453f5978ac298df14c54546d, we must check if there is any vector extension before using vector CSRs.

This is a miss of https://github.com/riscv-software-src/riscv-isa-sim/pull/1729/commits/af90d427af03fe11ba84d962fb539e8d33976d69. Sorry for the mistake.


aswaterman (approved):
Thanks @YenHaoChen.  I assume you audited the code to check that this is the only problematic case?

YenHaoChen:
Unfortunately, I did miss other vector instructions. We are working on a direct test for all vector instructions.

PR #1750, opened 2024-07-31
vector: Check if there is any vector extension before using vector CSRs

YenHaoChen:
Basically, the same issue of https://github.com/riscv-software-src/riscv-isa-sim/pull/1749.

I grep xrm to enumerate the problemetic instructions.


aswaterman (approved):
Thanks for following up.

PR #1753, opened 2024-07-31
Fix segfault accessing menvcfg when U-mode doesn't exist

### LINKED ISSUE #1752: Segmentation Fault when interrupt handled while interrupts are disabled in machine mode
When running in machine mode `PRV_M`, and an interrupt is handled by `take_interrupt` while interrupts are disabled, the current implementation attempts to access `sstatus`, which only exists in supervisor mode.  This results in a segmentation fault.

An example use case for this would be running in machine mode, interrupts disabled, then a WFI which waits for a configured timer to trigger an interrupt by setting `mtimecmp`.

Code snippet from processor.cc:
```C++
void processor_t::take_interrupt(reg_t pending_interrupts)
{
  // Do nothing if no pending interrupts
  if (!pending_interrupts) {
    return;
  }

  // Exit WFI if there are any pending interrupts
  in_wfi = false;

  // M-ints have higher priority over HS-ints and VS-ints
  const reg_t mie = get_field(state.mstatus->read(), MSTATUS_MIE);
  const reg_t m_enabled = state.prv < PRV_M || (state.prv == PRV_M && mie);
  reg_t enabled_interrupts = pending_interrupts & ~state.mideleg->read() & -m_enabled;
  if (enabled_interrupts == 0) {
    // HS-ints have higher priority over VS-ints
    const reg_t deleg_to_hs = state.mideleg->read() & ~state.hideleg->read();
    const reg_t sie = get_field(state.sstatus->read(), MSTATUS_SIE);
    const reg_t hs_enabled = state.v || state.prv < PRV_S || (state.prv == PRV_S && sie);
    enabled_interrupts = pending_interrupts & deleg_to_hs & -hs_enabled;
    if (state.v && enabled_interrupts == 0) {
      // VS-ints have least priority and can only be taken with virt enabled
      const reg_t deleg_to_vs = state.hideleg->read();
      const reg_t vs_enabled = state.prv < PRV_S || (state.prv == PRV_S && sie);
      enabled_interrupts = pending_interrupts & deleg_to_vs & -vs_enabled;
    }
  }
```

A simple fix for this issue might be to just qualify this `if` with `PRV_S` like this:

```C++
   if (enabled_interrupts == 0 && state.prv == PRV_S) {
```

In the described example test case, this resolved the issue.

Previously cloned versions of spike (approximately late last year) did not hit this issue, but I could not find what broke the functionality from some looking through git history.

aswaterman:
The simplest fix is to create the CSRs even if they don't need to exist, and just skip adding them to the CSR map to prevent the target machine from being able to access them.

It looks like there are other place we should be following this pattern: e.g. why does sstatus exist if S-mode does not exist?  But that's a matter for another day.

Resolves #1752

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -381,39 +381,43 @@ void state_t::reset(processor_t* const proc, reg_t max_isa)
   csrmap[CSR_MVENDORID] = std::make_shared<const_csr_t>(proc, CSR_MVENDORID, 0);
   csrmap[CSR_MHARTID] = std::make_shared<const_csr_t>(proc, CSR_MHARTID, proc->get_id());
   csrmap[CSR_MCONFIGPTR] = std::make_shared<const_csr_t>(proc, CSR_MCONFIGPTR, 0);
+  const reg_t menvcfg_mask = (proc->extension_enabled(EXT_ZICBOM) ? MENVCFG_CBCFE | MENVCFG_CBIE : 0) |
+                            (proc->extension_enabled(EXT_ZICBOZ) ? MENVCFG_CBZE : 0) |
+                            (proc->extension_enabled(EXT_SMNPM) ? MENVCFG_PMM : 0) |
+                            (proc->extension_enabled(EXT_SVADU) ? MENVCFG_ADUE: 0) |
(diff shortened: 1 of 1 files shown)


jerryz123 (approved):
Seems reasonable

PR #1754, opened 2024-08-01
vcompress.vm: Check if there is any vector extension before using vector CSRs

YenHaoChen:
The commit resolves the same issue of https://github.com/riscv-software-src/riscv-isa-sim/pull/1750 and https://github.com/riscv-software-src/riscv-isa-sim/pull/1749. We executed all vector instructions; this is the last one that missed out.

### DIFF
--- a/riscv/insns/vcompress_vm.h
+++ b/riscv/insns/vcompress_vm.h
@@ -1,13 +1,14 @@
 // vcompress vd, vs2, vs1
-require(P.VU.vstart->read() == 0);
-require_align(insn.rd(), P.VU.vflmul);
-require_align(insn.rs2(), P.VU.vflmul);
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
LGTM, thanks again.

xinyuwang-starfive:
@YenHaoChen  for this code change if vl=0 but vstart not equal 0,the instruction will not rise illegal and it different from before

chihminchao:
agree with @xinyuwang-starfive

YenHaoChen:
@xinyuwang-starfive @chihminchao 
Thank you for pointing out this issue. I apologize for overlooking the case where vl=0. A PR has been created to resolve this issue: https://github.com/riscv-software-src/riscv-isa-sim/pull/1904

PR #1758, opened 2024-08-01
Only supply CSRs if corresponding extensions are enabled

aswaterman:
There were a whole bunch of CSRs that we were providing even when they weren't necessary, e.g. S-mode CSRs when S-mode is not present.  Fix that.

This PR is structured as a sequence of refactorings, which clean up the way CSRs are added and remove boilerplate.  The final commit then makes the functional change.


jerryz123 (approved):
This is much nicer

PR #1760, opened 2024-08-02
Let MXR not affect implicit memory access for VS-stage address translation

YenHaoChen:
The behavior of MXR is clarified in https://github.com/riscv/riscv-isa-manual/pull/1543.


aswaterman (commented):
It might be cleaner to pass `mxr` as an argument, instead of passing `is_for_vs_pt_addr`, and just make sure that `mxr` is false when doing an implicit memory access.

aswaterman (approved):

PR #1769, opened 2024-08-11
Fix ordering of B single-letter extension

jerryz123:
The canonical order is IMAFDQLCBKJTPVH


YenHaoChen:
It makes sense to follow the canonical order. However, I was unable to find the desired order. Is there a document or guideline for the ordering? (https://github.com/riscv-software-src/riscv-isa-sim/pull/1649)

jerryz123:
Yes, unprivileged spec here: 
image

YenHaoChen:
Thank you. Thus, we want to make ```B``` between ```C``` and ```P```.

aswaterman (approved):

PR #1721, opened 2024-07-03
Enable more configuration using the DTB

abejgonzalez:
Addresses parts of #1716 

Thanks @jerryz123 for the main work on this.

- A DTB that is passed in is used to set the number of harts, ISA, PMPs, MMU, and more of Spike.
  - Note: Device extensions that generate a DTS string do not affect the output of `--dump-dts` if the DTB is given.
  - Note: DTB provided memory regions are not used.
- Correctly catches interactive traps v.s. other traps in interactive mode.

PR #1770, opened 2024-08-12
Fix a typo in f11bd7b (Support parsing procs fully from DTS)

YenHaoChen:
Fix a typo in https://github.com/riscv-software-src/riscv-isa-sim/pull/1721/commits/f11bd7b511d7909f0291589e3aaab720ededdc8a


jerryz123 (approved):
Good catch, thanks

PR #1771, opened 2024-08-12
Fix mcontrol6 mask low/high operations.

rtwfroody:
I doubt this code was ever tested, and this change isn't tested either, because OpenOCD doesn't use this trigger type.

This problem was reported in
https://github.com/riscv/riscv-debug-spec/issues/1057

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -195,13 +195,17 @@ bool mcontrol_common_t::simple_match(unsigned xlen, reg_t value) const {
       return value < tdata2;
     case MATCH_MASK_LOW:
       {
-        reg_t mask = tdata2 >> (xlen/2);
-        return (value & mask) == (tdata2 & mask);
+        reg_t tdata2_high = tdata2 >> (xlen/2);
+        reg_t tdata2_low = tdata2 & ((reg_t(1) << (xlen/2)) - 1);
+        reg_t value_low = value & ((reg_t(1) << (xlen/2)) - 1);
+        return (value_low & tdata2_high) == tdata2_low;
       }
     case MATCH_MASK_HIGH:
       {
-        reg_t mask = tdata2 >> (xlen/2);
-        return ((value >> (xlen/2)) & mask) == (tdata2 & mask);
+        reg_t tdata2_high = tdata2 >> (xlen/2);
+        reg_t tdata2_low = tdata2 & ((reg_t(1) << (xlen/2)) - 1);
+        reg_t value_high = value >> (xlen/2);
+        return (value_high & tdata2_high) == tdata2_low;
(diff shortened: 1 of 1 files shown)


rtwfroody:
@YenHaoChen, can you review this?

riscv/triggers.cc
```diff
@@ -195,13 +195,17 @@ bool mcontrol_common_t::simple_match(unsigned xlen, reg_t value) const {
       return value < tdata2;
     case MATCH_MASK_LOW:
       {
-        reg_t mask = tdata2 >> (xlen/2);
-        return (value & mask) == (tdata2 & mask);
+        reg_t tdata2_high = tdata2 >> (xlen/2);
+        reg_t tdata2_low = tdata2 & ((1 << (xlen/2)) - 1);
+        reg_t value_low = value & ((1 << (xlen/2)) - 1);
```
YenHaoChen: I think we need ```reg_t(1)```.
```
        reg_t tdata2_low = tdata2 & ((reg_t(1) << (xlen/2)) - 1);
        reg_t value_low = value & ((reg_t(1) << (xlen/2)) - 1);
```

riscv/triggers.cc
```diff
@@ -195,13 +195,17 @@ bool mcontrol_common_t::simple_match(unsigned xlen, reg_t value) const {
...
+        return (value_low & tdata2_high) == tdata2_low;
       }
     case MATCH_MASK_HIGH:
       {
-        reg_t mask = tdata2 >> (xlen/2);
-        return ((value >> (xlen/2)) & mask) == (tdata2 & mask);
+        reg_t tdata2_high = tdata2 >> (xlen/2);
+        reg_t tdata2_low = tdata2 & ((1 << (xlen/2)) - 1);
```
YenHaoChen: I think we need ```reg_t(1)```.
```
        reg_t tdata2_low = tdata2 & ((reg_t(1) << (xlen/2)) - 1);
```

YenHaoChen:
The old code was from the 0.13 version, and the patch provides the behavior of the ratified version. After the patch, the code makes more sense to me.

Reference:
image

rtwfroody:
@aswaterman I think this change is good to go.

YenHaoChen (commented):
I think we need ```reg_t(1)```.

(Sorry that I forget to submit the review.)

aswaterman:
I think @YenHaoChen is right about the out-of-bounds shifts by 32 when `int` is 32 bits.  Ping me again when that's done and I'll merge.

rtwfroody:
I've added reg_t() where necessary. Thanks.

aswaterman (approved):

PR #1776, opened 2024-08-16
pointer masking: Fix: Let transformed_addr of fetching be unchanged

YenHaoChen:
The transformation does not apply to implicit accesses such as instruction fetches.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -643,6 +643,7 @@ mem_access_info_t mmu_t::generate_access_info(reg_t addr, access_type type, xlat
     return {addr, addr, 0, false, {}, type};
   bool virt = proc->state.v;
   reg_t mode = proc->state.prv;
+  reg_t transformed_addr = addr;
   if (type != FETCH) {
     if (in_mprv()) {
       mode = get_field(proc->state.mstatus->read(), MSTATUS_MPP);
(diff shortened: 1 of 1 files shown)


riscv/mmu.cc
```diff
@@ -653,10 +654,10 @@ mem_access_info_t mmu_t::generate_access_info(reg_t addr, access_type type, xlat
...
     }
+    reg_t pmlen = get_pmlen(virt, mode, xlate_flags);
+    reg_t satp = proc->state.satp->readvirt(virt);
+    bool is_physical_addr = mode == PRV_M || get_field(satp, SATP64_MODE) == SATP_MODE_OFF;
+    transformed_addr = is_physical_addr ? zext(addr, 64 - pmlen) : sext(addr, 64 - pmlen);
   }
-  reg_t pmlen = get_pmlen(virt, mode, xlate_flags);
-  reg_t satp = proc->state.satp->readvirt(virt);
-  bool is_physical_addr = mode == PRV_M || get_field(satp, SATP64_MODE) == SATP_MODE_OFF;
-  reg_t transformed_addr = is_physical_addr ? zext(addr, 64 - pmlen) : sext(addr, 64 - pmlen);
```
aswaterman, line 660: Hardcoded use of 64 here is correct but sketchy.  I think it would be clearer to use `auto xlen = proc->get_const_xlen();` and replace 64 with `xlen`.

riscv/mmu.cc
```diff
@@ -653,10 +654,10 @@ mem_access_info_t mmu_t::generate_access_info(reg_t addr, access_type type, xlat
       virt = true;
       mode = get_field(proc->state.hstatus->read(), HSTATUS_SPVP);
     }
+    reg_t pmlen = get_pmlen(virt, mode, xlate_flags);
+    reg_t satp = proc->state.satp->readvirt(virt);
+    bool is_physical_addr = mode == PRV_M || get_field(satp, SATP64_MODE) == SATP_MODE_OFF;
+    transformed_addr = is_physical_addr ? zext(addr, 64 - pmlen) : sext(addr, 64 - pmlen);
```
aswaterman: Hardcoded use of 64 here is correct but sketchy.  I think it would be clearer to use `auto xlen = proc->get_const_xlen();` and replace 64 with `xlen`.

aswaterman (changes requested):
I believe this fix is incomplete; you also need to disable pointer masking when MXR is in effect.

aswaterman (approved):
Ah, I see `get_pmlen` checks MXR.  Nevermind.

Can you fix the coding style about 64 vs. xlen I mentioned in the other comment and ping me afterwards?  Then, I'll merge.

YenHaoChen:
@aswaterman 
Commit https://github.com/riscv-software-src/riscv-isa-sim/pull/1776/commits/0648ab40fc0742a8dbdbfbf82d71bbe0a3385317 fixes the coding style about 64 vs. xlen.

aswaterman (approved):

PR #1779, opened 2024-08-20
For mcontrol6, default to BEFORE timing.

rtwfroody:
The existing implementation would end up using AFTER even for instruction execute and data store triggers, which is not desirable.

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -265,7 +265,14 @@ bool mcontrol_common_t::legalize_timing(reg_t val, reg_t timing_mask, reg_t sele
     return TIMING_AFTER;
   if (get_field(val, execute_mask))
     return TIMING_BEFORE;
-  return get_field(val, timing_mask);
+  if (timing_mask) {
+    // Use the requested timing.
+    return get_field(val, timing_mask);
+  } else {
+    // For mcontrol6 you can't request a timing. Default to before since that's
+    // most useful to the user.
+    return TIMING_BEFORE;
+  }
 }
 
 reg_t mcontrol6_t::tdata1_read(const processor_t * const proc) const noexcept {


rtwfroody:
@YenHaoChen can you look at this?

YenHaoChen:
The ratifed spec removes mcontrol6.timing. Spike still implements mcontrol6.timing, which should be read-only-0 in the ratifed spec. That might be the root cause of this issue.

I think we can remove the macro CSR_MCONTROL6_TIMING (let it be 0) to resolve the issue.

@rtwfroody Do we still have CSR_MCONTROL6_TIMING=0 in the newest debug_defines.h? If so, we can update the file to resolve the issue.

YenHaoChen:
Sorry, I don't get this one.

If ```timing_mask``` is 0 (false), ```get_field(val, timing_mask)``` would be 0 (```TIMING_BEFORE```). Thus, the modification does not change any behavior. Do I miss anything?

rtwfroody:
get_field(reg, 0) expands to:
```
  (((reg) & (std::remove_cv<decltype(reg)>::type)(0)) / ((0) & ~((0) << 1)))
  ((reg & (0)) / (0 & ~(0)))
  ((reg & 0) / (0 & ~0))
  ((0) / (0))
  0 / 0
```
which is undefined.

YenHaoChen:
I see. Thus, ```get_field(val, 0)``` does not guarantee a value ```0```. In that case, this PR is indeed an improvement.

rtwfroody:
@YenHaoChen Can you approve this change?

YenHaoChen (approved):
LGTM. Given that ```get_field(val, 0)``` does not result in ```0```.

rtwfroody:
Hmm... I guess @YenHaoChen's review isn't good enough. @aswaterman, are you the main reviewer here?

aswaterman:
Yeah, someone with write permissions needs to approve.  Let me take a look and do so right now.

aswaterman (approved):

PR #1780, opened 2024-08-22
Only implement tcontrol to handle reentrant triggers.

rtwfroody:
The spec says to only implement a single option. Option 2 gives software more flexibility, and we already have an implementation. Let's use that one. See also #1777.

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -110,21 +110,6 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


rtwfroody:
Maybe it's better to delete the code altogether?

@YenHaoChen, please take a look.

YenHaoChen:
I would vote for deleting the code altogether if we want to go with Solution 2.

rtwfroody:
> I would vote for deleting the code altogether if we want to go with Solution 2.

Done.

rtwfroody:
We can't do this because spike has S-Mode. See discussion in #1777

PR #1784, opened 2024-08-26
pointer masking: Pointer masking does not apply when MXR=1 regardless of MPRV in v1.0.0-rc2

YenHaoChen:
RC2 revises the following wordings and removes the effectiveness of MPRV to MXR.

image

Reference: https://github.com/riscv/riscv-j-extension/issues/70


aswaterman (approved):

aswaterman:
Thanks for the fix @YenHaoChen

Issue #1785, opened 2024-08-26
Discrepancy with respect to maskmax field in mcontrol (spec vs spike)


RaamyPi:
From the spec (Version 1.0.0-rc3, Revised 2024-05-15), Page 72,

> A value of 0 indicates match 1 is not supported

but in spike's implementation (latest), maskmax is tied to zero but match == 1 (NAPOT) is also supported at the same time.

aswaterman:
cc @YenHaoChen @rtwfroody

YenHaoChen:
@RaamyPi Thanks for reporting the issue. I provided a PR for the problem. Please let me know if you have any further concerns.

raamypi:
Thanks for the help!

PR #1786, opened 2024-08-27
triggers: Let mcontrol.match be default (0/equal) if maskmax is 0

YenHaoChen:
This PR fixes the issue in https://github.com/riscv-software-src/riscv-isa-sim/issues/1785.

Specifically, the spec requires unsupporting ```mcontrol.match=NAPOT``` if ```mcontrol.maskmax=0```.
image

### DIFF
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -146,7 +146,7 @@ reg_t mcontrol_t::tdata1_read(const processor_t * const proc) const noexcept {
   auto xlen = proc->get_xlen();
   v = set_field(v, MCONTROL_TYPE(xlen), CSR_TDATA1_TYPE_MCONTROL);
   v = set_field(v, CSR_MCONTROL_DMODE(xlen), dmode);
-  v = set_field(v, MCONTROL_MASKMAX(xlen), 0);
+  v = set_field(v, MCONTROL_MASKMAX(xlen), maskmax);
   v = set_field(v, CSR_MCONTROL_HIT, hit);
   v = set_field(v, MCONTROL_SELECT, select);
   v = set_field(v, MCONTROL_TIMING, timing);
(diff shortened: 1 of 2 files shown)


riscv/triggers.cc
```diff
@@ -294,13 +294,14 @@ void mcontrol6_t::tdata1_write(processor_t * const proc, const reg_t val, const
   auto xlen = proc->get_const_xlen();
   assert(get_field(val, CSR_MCONTROL6_TYPE(xlen)) == CSR_TDATA1_TYPE_MCONTROL6);
   dmode = get_field(val, CSR_MCONTROL6_DMODE(xlen));
+  const reg_t maskmax6 = xlen - 1;
```
YenHaoChen, line 297: @rtwfroody I didn't find an explicit upper bound for maskmax6 in the Debug spec. Does the spec intend ```maskmax6=31``` in RV32 and ```maskmax6=63``` in RV64 for maximum?

riscv/triggers.cc
```diff
@@ -244,11 +244,11 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
   return std::nullopt;
 }
 
-mcontrol_common_t::match_t mcontrol_common_t::legalize_match(reg_t val) noexcept
+mcontrol_common_t::match_t mcontrol_common_t::legalize_match(reg_t val, reg_t maskmax) noexcept
 {
   switch (val) {
+    case MATCH_NAPOT: if (maskmax == 0) return MATCH_EQUAL;
     case MATCH_EQUAL:
-    case MATCH_NAPOT:
```
YenHaoChen, line 251: This seems tricky. Invite comment on this modification.

riscv/triggers.cc
rtwfroody: Yes. It doesn't make sense for maskmax6 to be any larger than that, because then you're just doing an equal comparison.

rtwfroody (approved):
This looks OK to me.

To be clear, it explicitly removes napot support from mcontrol, and makes it work for mcontrol6.

aswaterman (approved):
Code looks OK; I didn't review the functionality in detail.

PR #1787, opened 2024-08-27
Use cmdline --priv flag when parsing proc configurations from DTB

### LINKED ISSUE #1781: Wrong mstatus.[SXL|UXL] for RV64I in M-mode only config
I'm runnig some simple asm code using Spike (current version - today's build) in RV64I, M mode only config:
```
spike -p1  -m0x80000000:1568768,0x80180000:131072,0x801a0000:4096 -l --isa=rv64i --pmpregions=0 --priv=m --hartids=0 --log-commits --log=tmp/log.txt --triggers=0 --dm-sba=64 output.elf
```
At some point code reads mstatus (from log):
```
core   0: 0x0000000080000104 (0x300026f3) csrr    a3, mstatus
core   0: 3 0x0000000080000104 (0x300026f3) x13 0x0000000a00001800
```
Read value is `0x0000000a00001800` which seems wrong assuming `--priv=m` flag. It seems that spike assigns mstatus.SXL|UXL as misa.MXL. Correct value should be: `0x0000000000001800` .  Accroding to spec:

> When MXLEN=64, if S-mode is not supported, then SXL is read-only zero. [...] When MXLEN=64, if U-mode is not supported, then UXL is read-only zero.
(_Volume II, 3.1.6.2. Base ISA Control in mstatus Register, p. 27_)

jerryz123:
Resolves #1781 .

### DIFF
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -195,7 +195,7 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
       exit(1);
     }
 
-    procs.push_back(new processor_t(isa_str, DEFAULT_PRIV,
+    procs.push_back(new processor_t(isa_str, cfg->priv,
                                     cfg, this, hartid, halted,
                                     log_file.get(), sout_));
     harts[hartid] = procs[cpu_idx];


aswaterman (approved):
Ah, that makes sense.

PR #1789, opened 2024-08-28
pointer masking: Consider effective v bit instead of current v bit

YenHaoChen:
I am sorry that I overlooked an implication of the modification in RC2.

A previous commit removes the effectiveness of MPRV to MXR. (https://github.com/riscv-software-src/riscv-isa-sim/pull/1784)
However, the removal implies the MPRV affects point masking individually, and the MXR should consider the effective v bit.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -614,7 +614,7 @@ void mmu_t::register_memtracer(memtracer_t* t)
 }
 
 reg_t mmu_t::get_pmlen(bool effective_virt, reg_t effective_priv, xlate_flags_t flags) const {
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

aswaterman:
@YenHaoChen Wait, I'm not sure this version is correct, either.  Don't you want to do the same thing as in `mmu_t::walk`, i.e. 

    bool mxr = (proc->state.sstatus->readvirt(false) | proc->state.sstatus->readvirt(effective_virt)) & MSTATUS_MXR;

i.e. sstatus.MXR always applies regardless of effective V, and vsstatus.MXR additionally applies when effective V=1.

YenHaoChen:
I did not aware that asserting ```mstatus.MXR``` overrides ```vsstatus.MXR``` to 1.

image

PR #1791, opened 2024-08-29
pointer masking: Always apply sstatus.MXR regardless of effective V

YenHaoChen:
ISA spec says "Setting MXR at HS-level overrides both VS-stage and G-stage execute-only permissions."

Related issue: https://github.com/riscv-software-src/riscv-isa-sim/pull/1789#issuecomment-2316200325


aswaterman (approved):
Thanks.  Hopefully we got it right this time, @YenHaoChen 😓

PR #1793, opened 2024-08-29
Only implement one solution for native triggers.

rtwfroody:
When S mode is present, use option 1 (disable triggers in M-Mode unless MIE is set) from the Debug Spec. When S mode is not present, use option 2 (implement mte and mpte bits in tcontrol).

See discussion in #1777.

### DIFF
--- a/riscv/csr_init.cc
+++ b/riscv/csr_init.cc
@@ -205,13 +205,14 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
     add_csr(CSR_TDATA2, tdata2 = std::make_shared<tdata2_csr_t>(proc, CSR_TDATA2));
     add_csr(CSR_TDATA3, std::make_shared<tdata3_csr_t>(proc, CSR_TDATA3));
     add_csr(CSR_TINFO, std::make_shared<tinfo_csr_t>(proc, CSR_TINFO));
-    add_csr(CSR_TCONTROL, tcontrol = std::make_shared<masked_csr_t>(proc, CSR_TCONTROL, CSR_TCONTROL_MPTE | CSR_TCONTROL_MTE, 0));
+    if (!proc->extension_enabled_const('S')) {
+      add_csr(CSR_TCONTROL, tcontrol = std::make_shared<masked_csr_t>(proc, CSR_TCONTROL, CSR_TCONTROL_MPTE | CSR_TCONTROL_MTE, 0));
+    }
   } else {
     add_csr(CSR_TDATA1, std::make_shared<const_csr_t>(proc, CSR_TDATA1, 0));
     add_csr(CSR_TDATA2, tdata2 = std::make_shared<const_csr_t>(proc, CSR_TDATA2, 0));
     add_csr(CSR_TDATA3, std::make_shared<const_csr_t>(proc, CSR_TDATA3, 0));
     add_csr(CSR_TINFO, std::make_shared<const_csr_t>(proc, CSR_TINFO, 0));
-    add_csr(CSR_TCONTROL, tcontrol = std::make_shared<const_csr_t>(proc, CSR_TCONTROL, 0));
   }
   unsigned scontext_length = (xlen == 32 ? 16 : 32); // debug spec suggests 16-bit for RV32 and 32-bit for RV64
   add_supervisor_csr(CSR_SCONTEXT, scontext = std::make_shared<masked_csr_t>(proc, CSR_SCONTEXT, (reg_t(1) << scontext_length) - 1, 0));
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -117,8 +117,9 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
   return true;
 }
 
-bool trigger_t::allow_action(const state_t * const state) const
+bool trigger_t::allow_action(processor_t * const proc) const
 {
+  const state_t *state = proc->get_state();
   if (get_action() == ACTION_DEBUG_EXCEPTION) {
     const bool mstatus_mie = state->mstatus->read() & MSTATUS_MIE;
     const bool sstatus_sie = state->sstatus->read() & MSTATUS_SIE;
@@ -242,7 +243,7 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
     value &= 0xffffffff;
   }
 
-  if (simple_match(xlen, value) && allow_action(proc->get_state())) {
+  if (simple_match(xlen, value) && allow_action(proc)) {
     /* This is OK because this function is only called if the trigger was not
      * inhibited by the previous trigger in the chain. */
     set_hit(timing ? HIT_IMMEDIATELY_AFTER : HIT_BEFORE);
@@ -331,7 +332,7 @@ void mcontrol6_t::tdata1_write(processor_t * const proc, const reg_t val, const 
 
 std::optional<match_result_t> icount_t::detect_icount_fire(processor_t * const proc) noexcept
 {
-  if (!common_match(proc) || !allow_action(proc->get_state()))
+  if (!common_match(proc) || !allow_action(proc))
     return std::nullopt;
 
   std::optional<match_result_t> ret = std::nullopt;
@@ -346,7 +347,7 @@ std::optional<match_result_t> icount_t::detect_icount_fire(processor_t * const p
 
 void icount_t::detect_icount_decrement(processor_t * const proc) noexcept
 {
-  if (!common_match(proc) || !allow_action(proc->get_state()))
+  if (!common_match(proc) || !allow_action(proc))
     return;
 
   if (count >= 1) {
@@ -438,7 +439,7 @@ std::optional<match_result_t> trap_common_t::detect_trap_match(processor_t * con
   bool interrupt = (t.cause() & ((reg_t)1 << (xlen - 1))) != 0;
   reg_t bit = t.cause() & ~((reg_t)1 << (xlen - 1));
   assert(bit < xlen);
-  if (simple_match(interrupt, bit) && allow_action(proc->get_state())) {
+  if (simple_match(interrupt, bit) && allow_action(proc)) {
     hit = true;
     return match_result_t(TIMING_AFTER, action);
   }
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -55,11 +55,18 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
   sselect = (sselect_t)((proc->extension_enabled_const('S') && get_field(val, CSR_TEXTRA_SSELECT(xlen)) <= SSELECT_MAXVAL) ? get_field(val, CSR_TEXTRA_SSELECT(xlen)) : SSELECT_IGNORE);
 }
 
+static reg_t tcontrol_value(const state_t * state) {
+  if (state->tcontrol)
+    return state->tcontrol->read();
+  else
+    return 0;
+}
+
 bool trigger_t::common_match(processor_t * const proc, bool use_prev_prv) const noexcept {
   auto state = proc->get_state();
   auto prv = use_prev_prv ? state->prev_prv : state->prv;
   auto v = use_prev_prv ? state->prev_v : state->v;
-  auto m_enabled = get_action() != 0 || (state->tcontrol->read() & CSR_TCONTROL_MTE);
+  auto m_enabled = get_action() != 0 || (tcontrol_value(state) & CSR_TCONTROL_MTE);
   return (prv < PRV_M || m_enabled) && mode_match(prv, v) && textra_match(proc);
 }
 
--- a/riscv/triggers.cc
+++ b/riscv/triggers.cc
@@ -66,8 +66,30 @@ bool trigger_t::common_match(processor_t * const proc, bool use_prev_prv) const 
   auto state = proc->get_state();
   auto prv = use_prev_prv ? state->prev_prv : state->prv;
   auto v = use_prev_prv ? state->prev_v : state->v;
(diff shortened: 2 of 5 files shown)


rtwfroody:
@YenHaoChen can you look at this?

riscv/csr_init.cc
```diff
@@ -205,7 +205,11 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
     add_csr(CSR_TDATA2, tdata2 = std::make_shared<tdata2_csr_t>(proc, CSR_TDATA2));
     add_csr(CSR_TDATA3, std::make_shared<tdata3_csr_t>(proc, CSR_TDATA3));
     add_csr(CSR_TINFO, std::make_shared<tinfo_csr_t>(proc, CSR_TINFO));
-    add_csr(CSR_TCONTROL, tcontrol = std::make_shared<masked_csr_t>(proc, CSR_TCONTROL, CSR_TCONTROL_MPTE | CSR_TCONTROL_MTE, 0));
+    if (proc->extension_enabled_const('S')) {
+      add_csr(CSR_TCONTROL, tcontrol = std::make_shared<masked_csr_t>(proc, CSR_TCONTROL, 0, 0));
+    } else {
+      add_csr(CSR_TCONTROL, tcontrol = std::make_shared<masked_csr_t>(proc, CSR_TCONTROL, CSR_TCONTROL_MPTE | CSR_TCONTROL_MTE, 0));
+    }
```
YenHaoChen: This only partially consists of my reading on the spec

The spec says that triggers with action=0 do not match/fire in M-mode if ```tcontrol.mte=0```. Although the spec says that only Solution 2 implements ```tcontrol.mte```, I don't feel software doing Solution 1 expects ```tcontrol.mte``` to be read-only 0. (In such a case, I would be confused if the trigger fires with a read-only ```tcontrol.mte=0```.)

Nevertheless, I agree that the behavior does not violate the wording of the spec and would like to let others decide.

YenHaoChen (approved):
I am unsure whether we should implement a read-only 0 tcontorl in Solution 1 and could not find a direct answer from the spec.

The spec may want to keep the flexibility for hardware implementation. However, software would need a straight method to understand the hardware features for development. Additionally, from the hardware perspective, implementing an optional read-only 0 CSR increases costs without providing any beneficial features, and this does not make sense.

After another reading of the spec, I am convinced the modification does not violate the words of the spec.

aswaterman:
Ping me once the two of you are in agreement, then I'll merge the PR.

rtwfroody:
Good point. The spec is a bit messy here, and can use some clarification. The easy way to make it unambiguous what spike is doing is for tcontrol to not exist at all when S is implemented.

rtwfroody:
Just force pushed no change to get the checks to start again. I'm pretty sure they failed because of some network issue.

aswaterman:
@rtwfroody Not a network problem; there's now a segfault when running a basic test on Spike.  I checked this by rerunning CI locally.  Here's what `gdb` has to say; I didn't spend any time looking into it:

```
$ gdb ../install/bin/spike 
GNU gdb (GDB) Red Hat Enterprise Linux 8.2-20.el8
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ../install/bin/spike...run done.
(gdb) run --isa=rv64gc pk hello
Starting program: /scratch/waterman/riscv-isa-sim/ci-tests/build/install/bin/spike --isa=rv64gc pk hello
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[Detaching after fork from child process 1192447]
[Detaching after fork from child process 1192448]

fast_rv32i_mret (p=0x7ffff7fa3010, insn=..., pc=<optimized out>) at /scratch/waterman/riscv-isa-sim/ci-tests/../riscv/processor.h:264
264	  state_t* get_state() { return &state; }
Missing separate debuginfos, use: yum debuginfo-install glibc-2.28-236.el8_9.13.x86_64 libgcc-8.5.0-20.el8.x86_64 libstdc++-8.5.0-20.el8.x86_64
(gdb) bt
#0  fast_rv32i_mret (p=0x7ffff7fa3010, insn=..., pc=<optimized out>) at /scratch/waterman/riscv-isa-sim/ci-tests/../riscv/processor.h:264
#1  0x0000000000929ba4 in execute_insn_fast (fetch=..., pc=2147497700, p=0x7ffff7fa3010) at /scratch/waterman/riscv-isa-sim/ci-tests/../riscv/execute.cc:306
#2  processor_t::step (this=0x7ffff7fa3010, n=908, n@entry=5000) at /scratch/waterman/riscv-isa-sim/ci-tests/../riscv/execute.cc:306
#3  0x000000000054b386 in sim_t::step (this=this@entry=0x7fffffffca80, n=n@entry=5000) at /usr/include/c++/8/bits/stl_vector.h:930
#4  0x000000000054b4d5 in sim_t::idle (this=0x7fffffffca80) at /scratch/waterman/riscv-isa-sim/ci-tests/../riscv/sim.cc:432
#5  sim_t::idle (this=0x7fffffffca80) at /scratch/waterman/riscv-isa-sim/ci-tests/../riscv/sim.cc:424
#6  0x0000000000957985 in htif_t::run (this=this@entry=0x7fffffffca80) at /scratch/waterman/riscv-isa-sim/ci-tests/../fesvr/byteorder.h:12
#7  0x000000000054bdb4 in sim_t::run (this=this@entry=0x7fffffffca80) at /scratch/waterman/riscv-isa-sim/ci-tests/../riscv/sim.cc:277
#8  0x0000000000509d09 in main (argc=<optimized out>, argv=<optimized out>) at /scratch/waterman/riscv-isa-sim/ci-tests/../spike_main/spike.cc:556
(gdb) 
```

YenHaoChen:
@rtwfroody It looks like the issue might be related to writing tcontrol in ```mret.h```. We can access the tcontrol only when it exists, i.e., ```if (!proc->extension_enabled_const('S'))```.

rtwfroody:
I wonder why that's not happening for me. I rebased off of master just in case. Running spike with no arguments under gdb or valgrind report no issues whatsover. Anyway, I'll figure it out.

rtwfroody:
Got it to reproduce. Typical user error...

riscv/triggers.cc
```diff
@@ -55,11 +55,18 @@ void trigger_t::tdata3_write(processor_t * const proc, const reg_t val) noexcept
...
+  else
+    return 0;
+}
+
 bool trigger_t::common_match(processor_t * const proc, bool use_prev_prv) const noexcept {
   auto state = proc->get_state();
   auto prv = use_prev_prv ? state->prev_prv : state->prv;
   auto v = use_prev_prv ? state->prev_v : state->v;
-  auto m_enabled = get_action() != 0 || (state->tcontrol->read() & CSR_TCONTROL_MTE);
+  auto m_enabled = get_action() != 0 || (tcontrol_value(state) & CSR_TCONTROL_MTE);
```
YenHaoChen: The previous commit https://github.com/riscv-software-src/riscv-isa-sim/commit/db762327efe1c754081f7cbfa1f1cae168fedf17 introduced this code. Now, the ```allow_action()``` is responsible for checking the ```tcontrol.mte```. Do we still need this? Or, should we obliterate this line?

riscv/triggers.cc
```diff
@@ -110,17 +117,29 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
   return true;
 }
 
-bool trigger_t::allow_action(const state_t * const state) const
+bool trigger_t::allow_action(processor_t * const proc) const
```
YenHaoChen: We could have a refactoring commit, which only changes the parameter of the ```allow_action()``` function to break down the commit.

riscv/triggers.cc
```diff
@@ -110,17 +117,29 @@ bool trigger_t::textra_match(processor_t * const proc) const noexcept
...
+      const bool vsstatus_sie = state->vsstatus->read() & MSTATUS_SIE;
+      const bool medeleg_breakpoint = (state->medeleg->read() >> CAUSE_BREAKPOINT) & 1;
+      const bool hedeleg_breakpoint = (state->hedeleg->read() >> CAUSE_BREAKPOINT) & 1;
+      return (state->prv != PRV_M || mstatus_mie) &&
+            (state->prv != PRV_S || state->v || !medeleg_breakpoint || sstatus_sie) &&
+            (state->prv != PRV_S || !state->v || !medeleg_breakpoint || !hedeleg_breakpoint || vsstatus_sie);
+    } else {
+      // mte and mpte in tcontrol is implemented. medeleg [3] is hard-wired to 0.
+      return (state->prv != PRV_M) || (tcontrol_value(state) & CSR_TCONTROL_MTE);
+    }
```
YenHaoChen: How about using the existence of ```tcontrol``` to decide which solution to use, such that guarantees the existence of ```tcontrol``` in Solution 2 (required by the spec)?
```
if (state->tcontrol) {
  ... Solution 1 ...
} else {
  ... Solution 2 ...
  return (state->prv != PRV_M) || (state->tcontrol->read() & CSR_TCONTROL_MTE);
}
```

YenHaoChen (commented):
Overall LGTM. The functionality seems correct from my tests.

1. The commit is big and could be broken down a little bit. Specifically, we could have a refactoring commit, which only changes the  ```allow_action()``` function parameter.
2. I wonder whether we still need to check the tcontrol.mte in ```common_match()```.
3. Solution 2 requires the ```tcontrol```. We could use the existence of ```tcontrol``` to decide which solution to use and guarantee that the ```tcontrol``` exists in Solution 2.

riscv/triggers.cc
rtwfroody: Hmm... I'd say it looks like a bug even to have it. We don't want to inhibit a trigger match because tcontrol isn't implemented at all.

riscv/triggers.cc
rtwfroody: I prefer to keep it as-is. The spec is a little vague about whether tcontrol can exist without implementing solution 2 (since the bits are listed as WARL), and ideally we solve that in a way where tcontrol is definitely allowed to exist. If we need more bits in the future, they should be in tcontrol and not in yet another CSR.

riscv/triggers.cc
rtwfroody: Done.

riscv/triggers.cc
YenHaoChen: Yes, we should test whether ```tcontrol``` is implemented before checking the ```tcontrol.mte```.

From the coding tracing, trigger detection functions always call the ```common_match()``` and ```allow_action()``` together. Since this PR checks the ```tcontrol.mte``` in ```allow_actions()```, it does not need to check the ```tcontrol.mte``` again in ```common_match()```. In other words, removing the checking on ```tcontrol.mte``` from ```common_match()``` will not change any behavior. 

That is:
```
bool trigger_t::common_match(processor_t * const proc, bool use_prev_prv) const noexcept {
  auto state = proc->get_state();
  auto prv = use_prev_prv ? state->prev_prv : state->prv;
  auto v = use_prev_prv ? state->prev_v : state->v;
  return mode_match(prv, v) && textra_match(proc);
}
```
Could you please help verify this?

rtwfroody: Removing this line caused some debug trigger tests to fail. I don't entirely understand why, which is probably because it's not clear to me exactly what common_match() takes care of.

rtwfroody:
Should the allow_action() logic really just be in common_match()?

riscv/triggers.cc
YenHaoChen: After another code tracing, although ```trap_common_t::detect_trap_match()``` always calls the ```common_match()``` and ```allow_action()```, it turns out they check the tcontrol.mte depending on different privilege modes. Precisely, the ```common_match()``` matches if the _previous_ privilege mode is not M-mode, and the ```allow_action()``` allows firing if the current privilege mode is not M-mode. This might be the root cause of the reported issue in https://github.com/riscv-software-src/riscv-isa-sim/pull/1777.

That is another clarification that needs to be made. The itrigger/etrigger detects the _previous_ privilege mode from which the trap is taken. In such a case, which privilege mode does tcontrol.mte (Solution 2) consider? I have an intuition that Solution 1 and Solution 2 consider the same privilege mode since they both avoid the reentancy problem to the current privilege mode. (The current Spike lets Solution 1 consider the current privilege mode but Solution 2 the _previous_ one. This PR makes Solution 2 consider both _previous_ and current privilege modes.)

We should figure out the root cause of failures before merging the PR. Would you please provide more information about the debug trigger tests?

rtwfroody: The tests I ran were the OpenOCD ones at https://github.com/riscv-software-src/riscv-tests/tree/master/debug.

The debug spec rules are about the mode an exception is taken *from*. For etrigger that means spike needs to look at the previous mode because the trigger code isn't called until after the privilege has been changed.

I still think allow_action() should just be part of common_match(). It doesn't look like one is ever called without the other. I'll try that and see if it simplifies things.

aswaterman (approved):
I'm OK with this version, but I'll let @YenHaoChen sign off and click the merge button when he's ready.

YenHaoChen:
The PR has addressed all of my comments.

PR #1796, opened 2024-09-01
add support for mcountinhibit CSR

cyyself:
We hardwired mcountinihibit to 0 previously. Now, we implemented it.


aswaterman (changes requested):
Please don't edit `encoding.h` directly, as this is an auto-generated file from https://github.com/riscv/riscv-opcodes

Note also when you update encoding.h over there, you should use `1U` rather than `1ULL` for these constants, which will minimize friction when using them in places where the underlying datatype is not `long long`.  After all, they are very small numbers.

Otherwise, the changes LGTM.

cyyself:
> Please don't edit `encoding.h` directly, as this is an auto-generated file from https://github.com/riscv/riscv-opcodes
> 

Fixed.

> Note also when you update encoding.h over there, you should use `1U` rather than `1ULL` for these constants, which will minimize friction when using them in places where the underlying datatype is not `long long`. After all, they are very small numbers.
> 

Fixed. I checked the ISA spec. The mcounteren and mcountinhibit are 32-bit wide CSR, so using 1U is enough. 

> Otherwise, the changes LGTM.

Thanks!

aswaterman (approved):
LGTM now, thanks for the contribution.

cyyself:
Is this PR ready to merge?

aswaterman:
Yes, thanks for the ping.

PR #1800, opened 2024-09-09
README.md: update "Debugging With Gdb"

hirooih:
- OpenOCD v0.11.0-2 released Oct 17, 2021 or later supports address translation.

- we can use pk (and system calls)
- we can use stack.
- we don't need a special linker script


hirooih:
I withdraw this PR. 
I've opened #2234 which uses semihosting instead of pk.

PR #1801, opened 2024-09-10
Fix dtc_output read when interrupted by EINTR

MarcoBonino:
The read() call may be interrupted by a signal, resulting in an EINTR error. In this case, read() will return -1 and set errno to EINTR. This is a common issue when dealing with I/O in a multi-process environment.

Check the error code returned by errno.
If it is EINTR, we can safely retry the read().

### DIFF
--- a/riscv/dts.cc
+++ b/riscv/dts.cc
@@ -158,9 +158,11 @@ static std::string dtc_compile(const std::string& dtc_input, bool compile)
 
   int got;
   char buf[4096];
-  while ((got = read(dtc_output_pipe[0], buf, sizeof(buf))) > 0) {
-    dtc_output.write(buf, got);
-  }
+  do {
+    while ((got = read(dtc_output_pipe[0], buf, sizeof(buf))) > 0) {
+      dtc_output.write(buf, got);
+    }
+  } while (got == -1 && errno == EINTR); // Retry the read() if it was interrupted by a signal
   if (got == -1) {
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
There are myriad I/O calls within Spike.  I'm not yet convinced we need to solve this problem--can't we just be OK with it being undefined to send signals to Spike?--but if we do solve it, we should solve it systematically rather than providing a point solution to this one `read` call.

MarcoBonino:
> There are myriad I/O calls within Spike. I'm not yet convinced we need to solve this problem--can't we just be OK with it being undefined to send signals to Spike?--but if we do solve it, we should solve it systematically rather than providing a point solution to this one `read` call.

The reason I write this fix was that when running spike using lldb debugger I was always getting -1 from the read, preventing the model to actually run.
Need to get more knowledge on the model before thinking if we can have the same in other parts of codebase.
For now that seems not creating problems (and it fixes the issue I had). Probably this issue happens when we interact with the DTC process.
An easy thing could be to wrap the call and not use it anymore directly, but I think that may be too big change and possibly may impact the model run.
In my opinion is better to fix it rather that waiting for a better fix that may not arrive soon. What about creating an "issue" to track that we would like to extend it to the whole codebase ?

aswaterman:
Yeah, probably best to file an issue for the time being.

PR #1804, opened 2024-09-13
Fix error in reading right sstatus

### LINKED ISSUE #1799: Ssdbltrp - VS mode double trap check
In Spike's double trap check for VS/S mode, the current code accesses `state.nonvirtual_sstatus` in VS mode and `state.sstatus` in S mode.
    [processor.cc#L447-L453](https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/processor.cc#L447-L453)
I think this reverses the intended access to the VS/S status registers, preventing the correct execution of double traps in VS mode. 
Moreover, I think it might not be necessary to use curr_virt to select between VS and S status registers. 
Spike inherently supports accessing `sstatus` based on the current virtualization mode, reading from either `sstatus` or `vsstatus` as appropriate.
Therefore, we might consider adjusting line 466 to: `reg_t s = state.sstatus->read();`

ved-rivos:
closes #1799

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -446,7 +446,7 @@ void processor_t::take_trap(trap_t& t, reg_t epc)
   // An unexpected trap - a trap when SDT is 1 - traps to M-mode
   if ((state.prv <= PRV_S && bit < max_xlen) &&
       (((vsdeleg >> bit) & 1)  || ((hsdeleg >> bit) & 1))) {
-    reg_t s = curr_virt ? state.nonvirtual_sstatus->read() : state.sstatus->read();
+    reg_t s = state.sstatus->read();
     supv_double_trap = get_field(s, MSTATUS_SDT);
     if (supv_double_trap)
       vsdeleg = hsdeleg = 0;


ved-rivos:
closes #1799

riscv/processor.cc
```diff
@@ -446,7 +446,7 @@ void processor_t::take_trap(trap_t& t, reg_t epc)
   // An unexpected trap - a trap when SDT is 1 - traps to M-mode
   if ((state.prv <= PRV_S && bit < max_xlen) &&
       (((vsdeleg >> bit) & 1)  || ((hsdeleg >> bit) & 1))) {
-    reg_t s = curr_virt ? state.nonvirtual_sstatus->read() : state.sstatus->read();
+    reg_t s = curr_virt ? state.sstatus->read() : state.nonvirtual_sstatus->read();
```
aswaterman: `sstatus->read()` evaluates to `curr_virt ? vsstatus : sstatus`.

`nonvirtual_sstatus->read()` evaluates to `sstatus`.

So I think you just want `reg_t s = state.sstatus->read();`.

ved-rivos: Thanks. Updated.

aswaterman (approved):

Issue #1813, opened 2024-09-22
CSRRWI Shouldn't be updating CSR if "uimm" field is 0


FluxByt:
According to spec:
```
if the uimm[4:0] field is zero, then these instructions will not write to the CSR,
and shall not cause any of the side effects that might otherwise occur on a CSR write, nor raise illegal-
instruction exceptions on accesses to read-only CSRs.
```

However within [riscv/insns/csrrwi.h](../blob/master/riscv/insns/csrrwi.h), This check isn't done therefore always writing to the csr's.

DavidYu360:
The behavior you mentioned is only for CSRRSI and CSRRCI.
Also in the Table 8, CSRRWI always writes to CSR.
![snapshot_table_8](https://github.com/user-attachments/assets/6daabe16-cf87-4cc4-aea4-a3465269fe2c)

FluxByt:
I see, I merged the two sentences while reading, sounds good thank you!

PR #1814, opened 2024-09-22
Fix for CSRRWI writing to CSR when uimm = 0.

FluxByt:
Proposed fix for #1813


aswaterman:
(For anyone who stumbles across this PR, the author closed it because the original code was correct.)

PR #1816, opened 2024-09-26
Only enter debug mode once with -H flag (halt_on_reset)

YenHaoChen:
The dcsr->halt indicates whether to enter debug mode due to halt_on_reset. The variable only depends on whether the -H flag exists or not.

This commit proposes entering the debug mode only once and resetting the dcsr->halt. (Without the commit, the step() will always enter the debug mode with the -H flag.)

I need help confirming that the debugging software expects this behavior. @rtwfroody

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -217,6 +217,7 @@ void processor_t::step(size_t n)
     } else if (halt_request == HR_GROUP) {
       enter_debug_mode(DCSR_CAUSE_GROUP, 0);
     } else if (state.dcsr->halt) {
+      state.dcsr->halt = false;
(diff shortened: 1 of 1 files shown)


aswaterman:
Deferring review to @rtwfroody.

rtwfroody (approved):
This change is good. At one point halt was a bit in dcsr, so when the debugger does practically anything it would end up clearing the bit. Now the bit isn't there anymore so it needs to be cleared some other way. Probably the bit should be moved out of the dcsr structure here altogether.

aswaterman (approved):

YenHaoChen:
An other PR is provided to move dcsr::halt out of the dcsr structure.

PR #1820, opened 2024-09-27
refactor: Remove dcsr::halt variable

YenHaoChen:
This is a refactoring PR, which removes dcsr::halt.

The dcsr::halt was for dcsr.halt in debug spec v0.13. The debug spec v1.0.0 removes the dcsr.halt.

Reference: https://github.com/riscv-software-src/riscv-isa-sim/pull/1816#pullrequestreview-2331806142

@rtwfroody


aswaterman (approved):
Definitely an improvement.

PR #1822, opened 2024-09-29
fix typos

howjmay:


aswaterman (changes requested):
`configure` is an auto-generated file; please don't modify it unless as a result of running `autoconf`.

`fdt` is a copy of an open-source library that is maintained elsewhere; we should not modify it here.

The other changes are OK.

howjmay:
fixed. Thank you for review

aswaterman (approved):
Thanks.

howjmay:
thank you too

PR #1823, opened 2024-09-30
Change -H flag into --halted

YenHaoChen:
There is a comment about aiming at --halted but failing to achieve so. This commit provides the behavior.

Reference: https://github.com/riscv-software-src/riscv-isa-sim/commit/bd17c7e5a80c21b8b7dd6be8cd24a6f89dd63bf0

@rtwfroody


YenHaoChen:
BTW, do we still want the -H flag? It is kind of a conflict with the -h (--help) flag. @rtwfroody

aswaterman (commented):
I suggest supporting the long name only.

rtwfroody (approved):
--halted is better for an obscure feature like this one.

aswaterman (approved):

PR #1824, opened 2024-09-30
ADD: a basic BTM N-trace spec compliant trace encoder model

iansseijelly:
This PR adds a trace encoder model to spike, making it capable of dumping encoded n-trace packets in the slow path, which a separate decoder can later decode. It is enabled by a trace controller device (see: https://github.com/ucb-bar/spike-devices/pull/8) and implements the basic trace encoding mechanisms. Adding this to spike will make the software development of the decoder easier, and a reference for RTL encoders to check upon. 

* Known issues: adding a separate command trigger may be a good idea.


riscv/processor.h
```diff
@@ -381,6 +385,7 @@ class processor_t : public abstract_device_t
   bool histogram_enabled;
   bool log_commits_enabled;
   FILE *log_file;
+  trace_encoder_n *trace_encoder;
```
jerryz123: I think the member here does not need to be a pointer.

aswaterman: Yes, the current code has a memory leak, and the easiest fix is to make this field a `trace_encoder_n` rather than a `trace_encoder_n *`.

aswaterman:
@YenHaoChen can you review this one?

riscv/execute.cc
```diff
@@ -173,6 +175,13 @@ static inline reg_t execute_insn_logged(processor_t* p, reg_t pc, insn_fetch_t f
 
   try {
     npc = fetch.func(p, fetch.insn, pc);
+
+    if (p->get_log_commits_enabled()) {
+      hart_to_encoder_ingress_t packet;
```
jerryz123: Can you use [designated initializers](https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html) here?

riscv/execute.cc
```diff
@@ -173,6 +175,13 @@ static inline reg_t execute_insn_logged(processor_t* p, reg_t pc, insn_fetch_t f
...
+      hart_to_encoder_ingress_init(p, &packet, &fetch.insn, npc);
+      p->get_trace_encoder()->trace_encoder_push_commit(&packet);
```
jerryz123: This is pushing a reference to a stack variable - a memory leak

iansseijelly:
Hi, thanks for the review. I have fixed (I think) the memory leak regarding `trace_encoder_push_commit()`. I have also, using this PR + the controller device + a simple decoder implementation (https://github.com/iansseijelly/rust-trace-deocder), decoded a hello world trace correctly. Please let me know what are the future things to fix and add.

riscv/processor.h
iansseijelly: this is closely matching the implementation of some other processor fields like MMU. Also, using the cpp class construction `new trace_encoder_n();` returns a pointer, so I think this makes sense.

riscv/execute.cc
iansseijelly: `hart_to_encoder_ingress_init` contains some more lines of logic than just designated initialization and I think it probably deserves a separate function.

riscv/processor.h
jerryz123: I don't think the MMU pattern is necessarily the right one to follow here. We aren't consistent at all - note that `vectorUnit_t`, `entropy_source`, and `triggers::module_t` are not pointers.

Generally, you should use pointers if the member object is "given" to the class, rather than "created" by the class. We are obviously breaking this guidline for the mmu.

riscv/trace_encoder_n.cc
```diff
@@ -0,0 +1,211 @@
...
+trace_encoder_n::trace_encoder_n() {
+  this->trace_sink= fopen("trace_n.bin", "wb");
+  this->debug_reference = fopen("trace_n_ref_debug.log", "wb");
+  this->active = true;
+  this->enabled = false;
+  this->src = 0;
+  this->state = TRACE_ENCODER_N_IDLE;
+  this->icnt = 0;
+  this->packet_0 = (hart_to_encoder_ingress_t*) malloc(sizeof(hart_to_encoder_ingress_t)); // create empty packet
+  this->packet_1 = (hart_to_encoder_ingress_t*) malloc(sizeof(hart_to_encoder_ingress_t)); // create empty packet
```
jerryz123: These don't need to be pointers again, these can just be member objects. I suspect the overhead of your repeated `malloc`/`free` in the inner loop will offset any benefit from the pass-by-reference, if performance is what you are concerned about.

riscv/trace_encoder_n.cc
```diff
@@ -0,0 +1,211 @@
...
+}
+
+void trace_encoder_n::set_enable(bool enabled) {
+  bool was_enabled = this->enabled;
+  printf("[trace_encoder_n] setting enable to %d\n", enabled);
```
jerryz123: These printfs should be removed.

riscv/execute.cc
```diff
@@ -173,6 +175,13 @@ static inline reg_t execute_insn_logged(processor_t* p, reg_t pc, insn_fetch_t f
 
   try {
     npc = fetch.func(p, fetch.insn, pc);
+
+    if (p->get_log_commits_enabled()) {
```
jerryz123: Can you add an additional flag `--trace`, for example, instead of overloading `log-commits`.

riscv/trace_encoder_n.cc
```diff
@@ -0,0 +1,211 @@
+#include "trace_encoder_n.h"
+
+trace_encoder_n::trace_encoder_n() {
+  this->trace_sink= fopen("trace_n.bin", "wb");
```
jerryz123: This file name should be different for each hartId .. a simple way to avoid clobbering the same file in a multi-core with-trace simulation.

riscv/execute.cc
jerryz123: You can flatten that entire method into this block. This is the only place its called, and its not doing much else besides assigning the fields of the struct. 
You don't even have to use designated initializers, just inlining it would make this more readable.

riscv/processor.h
iansseijelly: Actually, the trace_encoder_ctrl device (https://github.com/ucb-bar/spike-devices/blob/46ba028b63ded301f8e2ad3bf1b9c51d5f46ab4e/src/trace_encoder_ctrl.cc#L44) needs a reference to the processor's trace encoder. Would this change the design decision here?

riscv/execute.cc
```diff
@@ -173,6 +177,20 @@ static inline reg_t execute_insn_logged(processor_t* p, reg_t pc, insn_fetch_t f
...
+      hart_to_encoder_ingress_t packet {
+        .i_type = _get_insn_type(&fetch.insn, npc != p->get_state()->pc + insn_length(fetch.insn.bits())),
+        .exc_cause = 0,
+        .tval = 0,
+        .priv = P_M, // TODO: check for processor privilege level
```
jerryz123, line 186: This `priv` should be fixed.

Should this `trace_encoder_push_commit` also be added to the `take_trap` and `take_interrupt` methods?

riscv/processor.cc
```diff
@@ -39,7 +40,7 @@ processor_t::processor_t(const char* isa_str, const char* priv_str,
   log_file(log_file), sout_(sout_.rdbuf()), halt_on_reset(halt_on_reset),
   in_wfi(false), check_triggers_icount(false),
   impl_table(256, false), extension_enable_table(isa.get_extension_table()),
-  last_pc(1), executions(1), TM(cfg->trigger_count)
+  last_pc(1), executions(1), TM(cfg->trigger_count), trace_encoder()
```
jerryz123: The default constructor for the `trace_encoder` member object is called by default in the constructor of `processor_t`, adding stuff to the initializer list is done to use a non-default constructor.

So you don't need to add `trace_encoder()` here... the default constructor (which is what you are using) will still be executed.

riscv/trace_encoder_n.h
```diff
@@ -0,0 +1,119 @@
...
+#define MSEO_LAST 0b11
+
+#define MAX_ICNT_BITS 15
+#define MAX_ICNT 1 << MAX_ICNT_BITS
+
+class trace_encoder_n {
+public:
+
+  trace_encoder_n() {
+    this->trace_sink= fopen("trace_n.bin", "wb");
```
jerryz123, line 80: Perhaps the constructor should include `hartId` as a parameter, and then `trace_n.bin` can be `hart<id>_trace_n.bin` here?

riscv/trace_encoder_n.h
```diff
@@ -0,0 +1,119 @@
...
+    this->debug_reference = fopen("trace_n_ref_debug.log", "wb");
+    this->active = true;
+    this->enabled = false;
+    this->src = 0;
+    this->state = TRACE_ENCODER_N_IDLE;
+    this->icnt = 0;
+    this->packet_0 = hart_to_encoder_ingress_t(); // create empty packet
+    this->packet_1 = hart_to_encoder_ingress_t(); // create empty packet
+  }
+  void trace_encoder_push_commit(hart_to_encoder_ingress_t packet);
```
jerryz123: Stylistically, this should be `push_packet` and `generate_packet`. The `trace_encoder` part of the method name is unnecessary.

riscv/trace_encoder_n.h
```diff
@@ -0,0 +1,119 @@
...
+  bool active;
+  bool enabled;
+  uint16_t src;
+  trace_encoder_n_state_t state;
+  uint8_t buffer[MAX_TRACE_BUFFER_SIZE];
+  uint16_t icnt;
+  uint64_t prev_addr;
+
+  void _set_direct_branch_packet(trace_encoder_n_packet_t* packet);
+  void _set_indirect_branch_packet(trace_encoder_n_packet_t* packet);
```
jerryz123: Why are some method names prefixed with underscore?

riscv/trace_ingress.cc
```diff
@@ -0,0 +1 @@
+#include "trace_ingress.h"
```
jerryz123: This file can be removed

riscv/trace_encoder_n.h
jerryz123: You should implement a destructor for this class that `fcloses` the file.

riscv/trace_encoder_n.h
```diff
@@ -0,0 +1,119 @@
...
+  void _set_program_trace_sync_packet(trace_encoder_n_packet_t* packet);
+
+  int packet_to_buffer(trace_encoder_n_packet_t* packet);
+  int _packet_to_buffer_program_trace_sync(trace_encoder_n_packet_t* packet);
+  int _packet_to_buffer_direct_branch_packet(trace_encoder_n_packet_t* packet);
+  int _packet_to_buffer_indirect_branch_packet(trace_encoder_n_packet_t* packet);
+
+  inline hart_to_encoder_ingress_t* _get_packet_0() { return &packet_0; }
```
jerryz123: Is this level of indirection necessary? This is a private method, so any usage of `_get_packet_0` can just directly reference `this->packet_0`

riscv/trace_encoder_n.h
iansseijelly: Feels like multi-core needs to be handled more carefully. According to the spec there needs to be a spec funnel, and the decoder also needs to spawn multiple workers and discern sources. Maybe we should add the multi-core support later.

riscv/trace_encoder_n.h
iansseijelly: They are encapsulated and called inside the push or generate functions. Makes things clearer this way.

jerryz123: I personally don't feel this is necessary, and it stylistically conflicts with the rest of this codebase.
The `private` access specifier should be sufficient at indicating that this method is only accessible to other internal methods of `trace_encoder_n`.

riscv/trace_encoder_n.cc
```diff
@@ -0,0 +1,201 @@
+#include "trace_encoder_n.h"
+
```
jerryz123, line 2: I noticed this file is a mix of 2-space and 4-space indentation. Can you standardize it all to 2-space, matching the rest of this repo?

PR #1828, opened 2024-10-04
Add instruction limit

Timmmm:
Adds an optional --instructions=N CLI argument which will stop the simulation after N instructions.

This is useful for benchmarking and profiling and sometimes debugging.

### DIFF
--- a/fesvr/htif.h
+++ b/fesvr/htif.h
@@ -24,6 +24,9 @@ class htif_t : public chunked_memif_t
   virtual void start();
   virtual void stop();
 
+  // Cause the simulation to exit with the given exit code.
+  void htif_exit(int exit_code);
+
   int run();
   bool done();
   int exit_code();
@@ -76,6 +79,10 @@ class htif_t : public chunked_memif_t
   // Given an address, return symbol from addr2symbol map
   const char* get_symbol(uint64_t addr);
 
+  // Return true if the simulation should exit due to a signal,
+  // or end-of-test from HTIF, or an instruction limit.
+  bool should_exit() const;
+
  private:
   void parse_arguments(int argc, char ** argv);
   void register_devices();
@@ -93,7 +100,8 @@ class htif_t : public chunked_memif_t
   addr_t sig_len; // torture
   addr_t tohost_addr;
   addr_t fromhost_addr;
-  int exitcode;
+  // Set to a value by htif_exit() when the simulation should exit.
+  std::optional<int> exitcode;
   bool stopped;
 
   device_list_t device_list;
--- a/riscv/sim.cc
+++ b/riscv/sim.cc
@@ -44,13 +44,15 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
              const char *log_path,
              bool dtb_enabled, const char *dtb_file,
              bool socket_enabled,
-             FILE *cmd_file) // needed for command line option --cmd
+             FILE *cmd_file, // needed for command line option --cmd
+             std::optional<unsigned long long> instruction_limit)
   : htif_t(args),
     cfg(cfg),
     mems(mems),
     dtb_enabled(dtb_enabled),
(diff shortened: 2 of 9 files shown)


jerryz123:
I wonder if the desired behavior can be achieved through using the `--debug-cmd` flag, which lets you specify a file of interactive debug commands. Does setting the file to be 
```
rs <some number of cycles>
q
```
achieve the same effect?

Timmmm:
Doesn't seem to - it does stop the simulation but it just drops you into the REPL rather than exiting.

In any case it's a far less convenient and discoverable interface.

Timmmm:
Any chance of a review for this?

aswaterman (commented):
The changes to the control flow make me a little nervous, but they are probably harmless.  @jerryz123?

fesvr/htif.h
```diff
@@ -93,7 +100,8 @@ class htif_t : public chunked_memif_t
   addr_t sig_len; // torture
   addr_t tohost_addr;
   addr_t fromhost_addr;
-  int exitcode;
+  // Set to a value by exit() when the simulation should exit.
+  std::optional<int> exitcode;
```
aswaterman, line 104: Make this `uint64_t` or `unsigned long long`; after all, it takes Spike less than a minute to execute enough instructions to overflow `int`.

riscv/sim.cc
```diff
@@ -414,8 +416,19 @@ void sim_t::idle()
...
     interactive();
-  else
+  else {
+    if (instruction_limit.has_value()) {
+      if (*instruction_limit < INTERLEAVE) {
+        // Final step.
+        step(*instruction_limit);
+        exit(0);
+        *instruction_limit = 0;
+        return;
```
aswaterman: I don't follow the logic here.  `exit(0)` shouldn't return, so can we delete the last two statements?

fesvr/htif.h
Timmmm: This is just the exit code written to htif, not the number of instructions. It was already `int`.

The number of instructions is `instruction_limit` below which is `unsigned long long`.

riscv/sim.cc
Timmmm: Err yeah that is confusing, but it's actually calling `htif_t::exit()`, not `std::exit()`. There are a couple of calls in the constructor that do the same.

Probably should rename the method to avoid confusion, or we could do `this->exit(0);` but I kind of hate that.

Timmmm (commented):
Thanks for the review!

fesvr/htif.h
aswaterman: Sorry, I read through the code too quickly.

riscv/sim.cc
aswaterman: Ohh, I see.  Yeah, `this->exit(0)` would make this a lot clearer.

jerryz123: Wouldn't `htif::exit(0)` work here?

aswaterman: Or that.

Timmmm: Changed to `htif_t::exit(0)`. I didn't touch the ones in the constructor - that code looks a bit wrong actually, probably due to exactly this confusion.

aswaterman (approved):

aswaterman:
Looks like you need to update the ci-tests to account for the API change.

riscv/sim.cc
Timmmm: Sorry, it's been so long since I wrote this... double checking it was actually me that added the `exit()` method, which is obviously a terrible idea (e.g. since it breaks the calls in the constructor! fun C++ footgun). I've renamed it to `htif_exit()`.

aswaterman: SGTM.

Timmmm:
Thanks!

PR #1830, opened 2024-10-08
Fix non-standard interrupt start position

demin-han:


jerryz123 (changes requested):
Instead of the magic number, `IRQ_M_EXT` should be `IRQ_LCOF`

jerryz123 (approved):

PR #1834, opened 2024-10-14
update encoding.h to get rid of erroneous define

aap-sc:


aap-sc:
@aswaterman I've encountered a minor warning during the build. Looks like you've already fixed this in https://github.com/riscv/riscv-opcodes/pull/288 , so this is just a minor update

aswaterman (approved):
Yep, thanks for making the corresponding Spike update.

PR #1835, opened 2024-10-15
Fix ignored-attributes warning for unique_ptr declaration

joe-rivos:
The attribute `__nonnull` was added to `fclose` in glibc 2.38, which causes a warning when using its `decltype` on a template argument.


aswaterman (approved):
LGTM, thanks.

PR #1839, opened 2024-10-18
add missing sdt/sie interaction when writing mstatus directly

### LINKED ISSUE #1838: mstatus can update mstatus.sdt and mstatus.sie at the same time, but sstatus cannot, is this match the expection?
image
for writing sstatus, there is a logic to ensure when writing sstatus.sdt to 1, sstatus.sie cannot be write to 1.
but when writing mstatus, there is no according logic, which makes it possible to write mstatus.sdt and mstatus.sie to 1 at the same time, Is the expected or a bug?

ved-rivos:
close #1838


aswaterman (approved):
LGTM, thank you.

PR #1842, opened 2024-10-23
Print values of involved registers in an instruction before and after of its execution

ahmedshakill:
print register values before and after of execution (with -l parameter).


jerryz123 (changes requested):
I'm not sure the approach used here is ideal. There is already a "logging" API in `log_reg_write/log_mem_read/log_mem_write`.

I thought `--enable-commitlog` would dump all the modified registers/memory after each instruction? Why is that insufficient?

ahmedshakill:
Closing this as workaround already exists

PR #1815, opened 2024-09-25
[#1548][vector crypto] adding register index LMUL alignement checks

nibrunieAtSi5:
This pull request implement two things:
- (a) fix for issue #1548: adding missing LMUL alignment checks for vector crypto instructions. All extensions from the following extensions are affected (that should be all instructions manipulating element groups): **Zvkned**, **Zvknh[a/b]**, **Zvksed**, **Zvksh**, **Zvkg**.
- (b) implementing precise LMUL alignment checks for the scalar-element-group operand of `.vs` instruction

(a) the EMUL alignment of vector operand/destination (`=EMUL` for the vector element group operands and the destination) is specified in the main RVV spec; when a vector register group is not EMUL aligned, the behavior is listed as "reserved" and spike has implemented this check to trigger illegal instruction exception when the condition was not met (e.g. https://github.com/riscv-software-src/riscv-isa-sim/blob/de5094a1a901d77ff44f89b38e00fefa15d4018e/riscv/v_ext_macros.h#L128) 

(b) is not specified clearly in the vector crypto specification (current version https://github.com/riscv/riscv-isa-manual/blob/7023c601443adf21850772e02e3d920bc68b2237/src/vector-crypto.adoc)

The vector crypto specification lists multiple times the intent to allow the vector register group for the scalar-element-group to have `EMUL = EGW / VLEN` (and to be aligned with this `EMUL` and not with the global `LMUL` associated with other vector operands/destination of the instruction).

PR #1843, opened 2024-10-24
zvk: Check VR index alignment constraint and vm constraint

YenHaoChen:
The vector specification requires aligning the VR index regarding the LMUL. Also, it requires that vd cannot be v0 when masking.

image

image

Although the vector crypto specification does not explicitly state those two constraints, the deriving extension should include the general constraints from the vector specification.

This commit checks the VR index and vm constraints on Zvk instructions.

### DIFF
--- a/riscv/zvk_ext_macros.h
+++ b/riscv/zvk_ext_macros.h
@@ -276,6 +276,7 @@
 //
 #define VI_ZVK_VD_VS1_VS2_EGU32x4_NOVM_LOOP(PRELUDE, EG_BODY) \
   do { \
+    VI_CHECK_SSS(true); \
     require_element_groups_32x4; \
     require_no_vmask; \
     const reg_t vd_num = insn.rd(); \
@@ -331,6 +332,7 @@
 //
 #define VI_ZVK_VD_VS1_VS2_EGU32x8_NOVM_LOOP(PRELUDE, EG_BODY) \
   do { \
+    VI_CHECK_SSS(true); \
     require_element_groups_32x8;; \
     require_no_vmask; \
     const reg_t vd_num = insn.rd(); \
@@ -397,6 +399,7 @@
                                                                PRELOOP, \
                                                                EG_BODY) \
   do { \
+    VI_CHECK_SSS(true); \
     require_element_groups_32x4; \
     require_no_vmask; \
     const reg_t vd_num = insn.rd(); \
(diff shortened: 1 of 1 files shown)


aswaterman:
@nibrunieAtSi5 @kdockser

nibrunieAtSi5:
I think I tried to cover some of the missing alignment constraints in https://github.com/riscv-software-src/riscv-isa-sim/pull/1815.

I don't think this PR handles the scalar-element-group source register index alignment for `.vs` the way it should be (see https://github.com/riscv/riscv-isa-manual/pull/1653): the `vs2` operand alignment should not be checked against LMUL.

None of the vector crypto instruction working on element groups support masking.

nibrunieAtSi5:
@YenHaoChen , could you help review https://github.com/riscv-software-src/riscv-isa-sim/pull/1815 ? It should fix the issues you intended to fix there (excluding the mask overlap which I believe is not an issue for any instruction working with element groups in vector crypto).

YenHaoChen:
@nibrunieAtSi5 
Before reviewing the PR, I need to familiarize myself with the vector crypto specification. Nevertheless, the behavior seems correct at a glance.

There are non-behavioral concerns in https://github.com/riscv-software-src/riscv-isa-sim/pull/1815. Reviewers would be much more willing to approve the PR if it could make the following modifications.
1. Usually, we expect each PR to deal with one thing. The PR implements two things. Perhaps, we can split it into two PRs, and reviewers could approve them separately.
2. Reviewers would be thankful if the contributor could squash and rebase their PR. Specifically, the second commit deletes all the codes from the first commit and adds different codes, confusing the reviewers. We may delete the first commit to help reviewers understand the PR. (Does the later commit delete all the code from the second commit again? I am unsure which commit contributes to the final code.)
3. Comments should go along with the code. Commits https://github.com/riscv-software-src/riscv-isa-sim/pull/1815/commits/61ad5bdedad24e38befa703e6bc5f9d0dfca2b4c and https://github.com/riscv-software-src/riscv-isa-sim/pull/1815/commits/0f1ef5c9bcbe53cd5b6af003b4650bc49aafcaa6 modify the comments only. Please squash the modification to the corresponding commits.
4. PR should not include typo correction of the PR itself. Please squash commits https://github.com/riscv-software-src/riscv-isa-sim/pull/1815/commits/636a5f3600a73d2647154164424d2ca75a940237 and https://github.com/riscv-software-src/riscv-isa-sim/pull/1815/commits/494e21942e595deff58d465e36a1d36119da9c7c.

kdockser:
As I just mentioned in another discussion. We need to make a blanket statement that _Vector Crypto inherits all constraints from RVV._

PR #1847, opened 2024-10-31
Add Shlcofideleg support

demin-han:


riscv/csrs.cc
```diff
@@ -834,21 +834,23 @@ generic_int_accessor_t::generic_int_accessor_t(state_t* const state,
 }
 
 reg_t generic_int_accessor_t::ip_read() const noexcept {
-  return (state->mip->read() & deleg_mask() & read_mask) >> shiftamt;
+  const reg_t val = state->mip->read() & deleg_mask() & read_mask;
+  return ((val & 0xfff) >> shiftamt) | (val & ~0xfff);
```
aswaterman: The magic constant `0xfff` is not OK.  Same comment applies throughout the rest of this file.

demin-han: magic number removed in new commit

PR #1866, opened 2024-12-05
mmu: raise the prioity of misaligned superpage

chihminchao:
Based on the change in ttps://github.com/riscv/riscv-isa-manual/pull/1742 The priority of misaligned superpage is higher than Zicfiss cases


aswaterman (approved):

aswaterman:
Merging despite MacOS CI failure, since the failure is due to a GitHub Actions change, not due to this PR.

PR #1877, opened 2024-12-16
Save only usefull bits of pmpaddr

### LINKED ISSUE #1876: Bug with pmpaddrX if pmpgran > 4
https://github.com/riscv-software-src/riscv-isa-sim/blob/a2dcf1fd70f359e2cf0a953986c9fe222a6d53c2/riscv/csrs.cc#L122

According to RISC-V priv spec, in the NAPOT mode pmpaddrX's `log2(region_size)-3`th bit is zero and the following bits are ones.

When storing pmpaddr (`pmpaddr_csr_t::unlogged_write`), spike doesn't clear the least significant bits.
When reading pmpaddr (`pmpaddr_csr_t::read`), spike ORs the ones but doesn't clear the `log2(region_size)-3`th bit either.

This causes the following bug:
Let pmpgran be 8
Let pmpcfg0 be 0
Let pmpaddr0 be 0b00..0111
Let pmpcfg0 be 0x11 (NA4)
pmpcfg0 is now 0x19 (NAPOT) due to pmpgran > 4
pmpaddr must be 0b00..0110 now, i.e. the protected range is [0b011000..0b100000), region size = 8 bytes
actual pmpaddr is 0b00..0111, i.e. the protected range is [0..0b1000000), region size = 64 bytes

marcfedorow:
Fixes #1876

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -120,6 +120,7 @@ bool pmpaddr_csr_t::unlogged_write(const reg_t val) noexcept {
 
   if (pmpidx < proc->n_pmp && !locked && !next_locked_and_tor()) {
     this->val = val & ((reg_t(1) << (MAX_PADDR_BITS - PMP_SHIFT)) - 1);
+    this->val &= proc->pmp_tor_mask();
   }
   else
     return false;


aswaterman (changes requested):
discussion continues on #1876

PR #1880, opened 2024-12-17
Set dcsr.xdebugver to 4, as it ought to be

### LINKED ISSUE #1878: dcsr.debugver field default value
Spike’s dcsr.xdebugver is 1 while spec is 0/4/15
7f44697577187f009e1594e76444cfdd

aswaterman:
See discusson on #1878.  This restores the behavior prior to ec292be4fdac9685c9d6f2f209b73743c2e573f7, which inadvertently changed the value to 1.

Resolves #1878


aswaterman:
@jerryz123 I'm going to yolo-merge this since you're on vacation.

jerryz123 (approved):

PR #1881, opened 2024-12-18
Add spike-dasm --strict option 

### LINKED ISSUE #1879: Can spike-dasm disassemble specified extensions through the --isa option?
When I run spike-dasm to disassemble instructions, it disassembled vector extension. When I specify the extensions, it ignored. It seems that this tool could disassemble any instruction.

image

I have read the code of [spike-dasm.cc](https://github.com/riscv-software-src/riscv-isa-sim/blob/master/spike_dasm/spike-dasm.cc), it supports the DEFAULT_ISA, which is `rv64imafdc_zicntr_zihpm`.  In the code, spike-dasm will parse `isa` option, but it fails to disassemble instructions of specified extensions. **Is this normal?**

**version**: latest(a2dcf1f), 1.1.1-dev

aswaterman:
Prevents disassembling instructions that are not enabled by the `--isa` setting.

Resolves #1879

### DIFF
--- a/disasm/disasm.cc
+++ b/disasm/disasm.cc
@@ -809,7 +809,12 @@ static void NOINLINE add_unknown_insns(disassembler_t* d)
   #undef DECLARE_INSN
 }
 
-void disassembler_t::add_instructions(const isa_parser_t* isa)
+#define ext_enabled_strict(x) (isa->extension_enabled(x))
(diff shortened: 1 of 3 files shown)


aswaterman:
I'm going to yolo-merge this because it's likely to fall victim to merge conflicts, but I'm happy to address any concerns in a follow-up PR.

PR #1889, opened 2025-01-09
(Guest)Page fault cause by Svnapot occur before A/D update

### LINKED ISSUE #1884: Will page fault exception still update PTE AD bits when svadu extension
image
from above code in mmu.cc:
After pte ad bits were updated when svadu extension enabled， it is still possible to enter the second red box and thus trigger a page fault exception.
Is this bahavior correct? I couldn't find any relevant description of this behavior.

ved-rivos:
Page fault or Guest Page Fault due to Svnapot rules should be higher priority than setting A and/or D bit in the PTE.

closes #1884


aswaterman (approved):

aswaterman:
Thanks, @ved-rivos.

Issue #1893, opened 2025-01-14
Probable bugs in the Smcsrind/Sscsrind implementation


JJ-Gaisler:
There are a few issues with the smcsrind/sscsrind extension. First of all (m/h)stateen.csrind is not checked for either the select registers or the ireg registers. This could be solved by adding a couple of new classes (csrind_select & csrind_virt_select).

https://github.com/riscv-software-src/riscv-isa-sim/blob/49727e8c61e44ebf10bae5bb7a488752137ac75f/riscv/csrs.cc#L1746

The second issues is that mireg has the same type as (v)sireg*. This currently means that a virtual instruction exception could be raised for an M mode CSR if the proxy csr is not implemented. Again this could be solved by either creating a new class or by explicitly checking the csr privilege in the current (sscsrind) class.
https://github.com/riscv-software-src/riscv-isa-sim/blob/49727e8c61e44ebf10bae5bb7a488752137ac75f/riscv/csr_init.cc#L348

The third potential issue is the following, the current priv spec says the following:
```
The behavior upon accessing vsireg* from M-mode or HS-mode, or accessing sireg* (really vsireg*) from VS-mode, 
while vsiselect holds a value that is not implemented at HS level, is UNSPECIFIED. 
(I) It is recommended that implementations raise an illegal instruction exception for such accesses, 
to facilitate possible emulation (by M-mode) of these accesses.
```
If proxy_csr == nullptr (proxy csr not available to HS mode?) then it is possible to get a virtual instruction exception even if the spec recommends illegal.

I am willing to submit a PR to fix these issues if someone is willing to confirm that I have understood things correctly.

aswaterman:
cc @atulkharerivos, who provided the implementation.

@JJ-Gaisler I suggest you go ahead with a PR, and hopefully @atulkharerivos can review it.  Thank you for your help.

PR #1896, opened 2025-01-20
Smcsrind exception fixex

JJ-Gaisler:
Fixes for: https://github.com/riscv-software-src/riscv-isa-sim/issues/1893
I have implemented smcdeleg in another branch and tested this implementation with smcdeleg and from the way I interpret the spec this should fix the issues I was seeing before. Let me know if something is poorly explained or incorrect.


aswaterman:
Haven't had the chance to review yet, but please clean up the commit history to get rid of the merge commit and the commits with debug prints.  In this case, squashing everything into a single commit is probably appropriate.

JJ-Gaisler:
I was just extending my testing and I noticed that checks for the s/hcontext bits is also missing, I'll be adding these soon. Should I also add hedelegh as a read-only zero CSR and incorporate the check for p1p13 while at it?

aswaterman:
May as well, thank you!

aswaterman:
You can add them to this PR if you prefer.  Up to you.

JJ-Gaisler:
I couldn't find mstateen0_priv113 nor csr_hedelegh in the encoding.h file. I have added these constants in the code inline for now, I am waiting for https://github.com/riscv/riscv-opcodes/pull/333

I have a test case which loops over CSRs that are affected by stateen, it checks that the correct trap is raised in all privilege modes. The test can be run with:
`spike -m0x80000000:0x10000000,0x20000000:0x1000,0x4000:0x1000,0xB000:0x1000 --isa rv64gh_svinval_zba_zbb_zbc_zbs_zfh_zfa_zbkb_zbkx_zicsr_zifencei_sscofpmf_smstateen_smepmp_zicbom_zicntr_zihpm_svadu_sstc_zicond_smcsrind_sscsrind -p1 csr_test.txt`

[csr_test.txt](https://github.com/user-attachments/files/18570383/csr_test.txt)

aswaterman:
riscv-opcodes PR has been merged.

PR #1901, opened 2025-01-23
CFI fixes

binno:


chihminchao:
@aswaterman  @ved-rivos 
Just a mention there are some CFI-related work.

aswaterman (approved):
Sorry we didn't see this earlier.

@ved-rivos this looks correct to me, but can you provide a sanity check before I merge it?

aswaterman:
@ved-rivos I'm going to go ahead and merge this, but LMK if you have any comments and we can fix them via a follow-up PR.

ved-rivos:
@aswaterman these updates look good to me.

aswaterman:
thank you!

PR #1902, opened 2025-01-23
Extend bf16

chihminchao:


softfloat/fall_neg.c
```diff
@@ -0,0 +1,63 @@
...
+{
+    return signF16UI(a.v);
+}
+
+bool f32_neg( float32_t a )
+{
+    return signF32UI(a.v);
+}
+
+bool f64_neg( float64_t a )
```
aswaterman: `neg` is a confusing name for these functions; it sounds like a function that negates (i.e. returns `-a`), especially since RISC-V has pseudoinstructions named `fneg` that perform the negation operation.  `sign` is a better name.

softfloat/i32_to_bf16.c
```diff
@@ -0,0 +1,47 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+#include "platform.h"
+#include "internals.h"
+#include "softfloat.h"
+
+bfloat16_t i32_to_bf16( int32_t a )
+{
+    return f32_to_bf16(i32_to_f32(a));
```
aswaterman: I would like @nibrunieAtSi5's confirmation, but I believe this implementation can incorrectly double-round.  You need to perform the i32 to f32 conversion under round-to-odd, then restore the rounding mode before converting to bf16.

softfloat/fall_reciprocal.c
```diff
@@ -262,6 +299,49 @@ static inline uint64_t recip7(uint64_t val, int e, int s, int rm, bool sub,
...
+        break;
+    case 0x100: // sNaN
+        softfloat_exceptionFlags |= softfloat_flag_invalid;
+    case 0x200: //qNaN
+        uA.ui = defaultNaNBF16UI;
+        break;
+    case 0x004: // -subnormal
+    case 0x020: //+ sub
+        sub = true;
+    default: // +- normal
```
aswaterman, line 333: @nibrunieAtSi5 should review this one, since IIRC there is a special case for BF16 subnormals.

softfloat/ui32_to_bf16.c
```diff
@@ -0,0 +1,46 @@
...
+=============================================================================*/
+
+#include <stdint.h>
+#include "platform.h"
+#include "internals.h"
+#include "softfloat.h"
+
+float16_t ui32_to_bf16( uint32_t a )
+{
+    return f32_to_bf16(ui32_to_f32(a));
```
aswaterman: Same comment as https://github.com/riscv-software-src/riscv-isa-sim/pull/1902/files#r1928221609

aswaterman (changes requested):
Looks like it's on the right track, but please address the few issues I raised, possibly in consultation with @nibrunieAtSi5.

softfloat/i32_to_bf16.c
nibrunieAtSi5: That sounds about right,
for example converting 0b1_0000_0001_0000_0000_0001_0000 (a.k.a. 0x1010011) from integer to BF16 going through FP32 will result in an error in RNE mode:

- 1st conversion will result in 0b1_0000_0001_0000_0000_000 (since the input is a mid-point for FP32, it is rounded to the even mantissa)
- 2nd conversion from FP32 to BF16 will round the mantissa to 0b1_0000_000 (same reason)

but in fact 0b1_0000_001 was expected but the sticky information encompassed in the round bit of the first conversion was lost

softfloat/fall_reciprocal.c
nibrunieAtSi5: I don't think this is mentioned in the proposal yet, but indeed we need to settle on how to round subnormal numbers since there are cases where this can happen for BF16.
My proposal was
> The approximation can exceed BFloat16 available precision for subnormal results. In such a case, the result is rounded towards zero and no flags (neither underflow nor inexact) are raised. Overflow cases continue to be rounded according to the current rounding mode.

I think this is compliant with how `recip7` handles things today.
https://github.com/riscv-software-src/riscv-isa-sim/blob/18f4d0ff13ce8d6933ddaf080dfbc5aa2217d2ab/softfloat/fall_reciprocal.c#L251-L260
```
    int idx = sig >> (s-p);
    uint64_t out_sig = (uint64_t)(table[idx]) << (s-p);
    uint64_t out_exp = 2 * make_mask64(0, e - 1) + ~exp;
    if (out_exp == 0 || out_exp == UINT64_MAX) {
        out_sig = (out_sig >> 1) | make_mask64(s - 1, 1);
        if (out_exp == UINT64_MAX) {
            out_sig >>= 1;
            out_exp = 0;
        }
    }
```

For BF16, both `s=7` and `p=7`, so `out_sig` is directly the unshifted table look-up result.

```
        out_sig = (out_sig >> 1) | make_mask64(s - 1, 1);
```
might discard a significant LSB (for a table entry that is odd).
Mandating round-toward-zero for `vfrec7.v` for BF16 (`vsew=1` and `altfmt=1`) is compatible with this behavior.

aswaterman: Right, I think the plan was to round towards zero, so this code does the right thing.

I'll update the ISA spec to state this explicitly.

softfloat/fall_neg.c
chihminchao: renamed in updated version

softfloat/ui32_to_bf16.c
chihminchao: follow the suggestion to do round_to_odd first

softfloat/i32_to_bf16.c
chihminchao: follow the suggestion to do round_to_odd first

aswaterman (approved):
LGTM now, thank you!

PR #1904, opened 2025-01-23
vcompress.vm: Check vstart value even if vl = 0

YenHaoChen:
A previous commit inadvertently included an incorrect behavior: when vl=0, the instruction incorrectly failed to check if vstart!=0. (https://github.com/riscv-software-src/riscv-isa-sim/pull/1754#issuecomment-2609155665)

This PR aims to achieve the same goal as https://github.com/riscv-software-src/riscv-isa-sim/pull/1754 while correcting this incorrect behavior.

### DIFF
--- a/riscv/insns/vcompress_vm.h
+++ b/riscv/insns/vcompress_vm.h
@@ -1,4 +1,5 @@
 // vcompress vd, vs2, vs1
+require(P.any_vector_extensions());
 require(P.VU.vstart->read() == 0);
 require_align(insn.rd(), P.VU.vflmul);
 require_align(insn.rs2(), P.VU.vflmul);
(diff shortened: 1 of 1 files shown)


riscv/insns/vcompress_vm.h
```diff
@@ -1,14 +1,14 @@
 // vcompress vd, vs2, vs1
+P.any_vector_extensions();
```
aswaterman: Looks like a statement with no effect; do you mean `require(P.any_vector_extensions());`?

YenHaoChen: @aswaterman
Yes, we need the `require()`. Thank you for pointing out.

aswaterman (approved):
LGTM now, thanks.

PR #1907, opened 2025-01-27
[riscv|insns] Fix zbs immediate instructions: bclri, bexti, binvi, an…

kassasAndes:
[riscv|insns] Fix zbs immediate instructions: bclri, bexti, binvi, and bseti


aswaterman (approved):
LGTM, thanks.

PR #1915, opened 2025-02-10
cosim: right shift can't be applied on negative number

chihminchao:
referece from:  ISO C99 (6.5.7/4)
"The result of E1 << E2 is E1 left-shifted E2 bit positions; vacated
 bits are ﬁlled with zeros. If E1 has an unsigned type, the value of the
 result is E1 × 2E2, reduced modulo one more than the maximum value
 representable in the result type. If E1 has a signed type and
 nonnegative value, and E1 × 2E2 is representable in the result type,
 then that is the resulting value; otherwise, the behavior is undeﬁned."

### DIFF
--- a/riscv/insns/vnsra_wi.h
+++ b/riscv/insns/vnsra_wi.h
@@ -1,5 +1,5 @@
 // vnsra.vi vd, vs2, zimm5
 VI_VI_LOOP_NSHIFT
 ({
-  vd = vs2 >> (zimm5 & (sew * 2 - 1) & 0x1f);
+  vd = vs2 >> (zimm5 & (sew * 2 - 1));
 })
--- a/riscv/insns/vsll_vi.h
+++ b/riscv/insns/vsll_vi.h
@@ -1,5 +1,5 @@
 // vsll.vi  vd, vs2, zimm5
-VI_VI_LOOP
+VI_VI_ULOOP
 ({
-  vd = vs2 << (simm5 & (sew - 1) & 0x1f);
+  vd = vs2 << (zimm5 & (sew - 1));
 })
--- a/riscv/insns/vsll_vv.h
+++ b/riscv/insns/vsll_vv.h
@@ -1,5 +1,5 @@
 // vsll
-VI_VV_LOOP
+VI_VV_ULOOP
 ({
   vd = vs2 << (vs1 & (sew - 1));
 })
--- a/riscv/insns/vsll_vx.h
+++ b/riscv/insns/vsll_vx.h
@@ -1,5 +1,5 @@
 // vsll
-VI_VX_LOOP
+VI_VX_ULOOP
 ({
   vd = vs2 << (rs1 & (sew - 1));
 })
--- a/riscv/insns/vsrl_vi.h
+++ b/riscv/insns/vsrl_vi.h
@@ -1,5 +1,5 @@
 // vsrl.vi vd, vs2, zimm5
 VI_VI_ULOOP
 ({
-  vd = vs2 >> (zimm5 & (sew - 1) & 0x1f);
+  vd = vs2 >> (zimm5 & (sew - 1));
 })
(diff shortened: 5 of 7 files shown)


aswaterman (changes requested):
LGTM so far, but I have two requests:

- Get rid of the `& 0x1f`, which is redudant because `zimm5` is already limited to the range [0, 0x1f].
- There are several shift instructions written in this style: `vsll`, `vsra`, `vsrl`, `vnsra`, `vnsrl`, `vssra`, `vssrl`, `vwsll`.  Please audit all of them and fix them as part of this PR.  I see `simm5` used in some of them.

chihminchao:
> LGTM so far, but I have two requests:
> 
> * Get rid of the `& 0x1f`, which is redudant because `zimm5` is already limited to the range [0, 0x1f].
> * There are several shift instructions written in this style: `vsll`, `vsra`, `vsrl`, `vnsra`, `vnsrl`, `vssra`, `vssrl`, `vwsll`.  Please audit all of them and fix them as part of this PR.  I see `simm5` used in some of them.

I check all of vector shift-related and include the fix in the same commit.   The list is described in commit log.

chihminchao:
ci failed with fetch failure
```
Cloning into '/home/runner/work/riscv-isa-sim/riscv-isa-sim/riscv-openocd/tools/git2cl'...
fatal: unable to access 'https://git.savannah.nongnu.org/git/git2cl.git/': Failed to connect to git.savannah.nongnu.org port 443 after 134990 ms: Connection timed out
fatal: clone of 'https://git.savannah.nongnu.org/git/git2cl.git' into submodule path '/home/runner/work/riscv-isa-sim/riscv-isa-sim/riscv-openocd/tools/git2cl' failed
Failed to clone 'tools/git2cl'. Retry scheduled
Cloning into '/home/runner/work/riscv-isa-sim/riscv-isa-sim/riscv-openocd/tools/git2cl'...
fatal: unable to access 'https://git.savannah.nongnu.org/git/git2cl.git/': Failed to connect to git.savannah.nongnu.org port 443 after 135154 ms: Connection timed out
fatal: clone of 'https://git.savannah.nongnu.org/git/git2cl.git' into submodule path '/home/runner/work/riscv-isa-sim/riscv-isa-sim/riscv-openocd/tools/git2cl' failed
Failed to clone 'tools/git2cl' a second time, aborting
Error: Process completed with exit code 1.
```

chihminchao:
use forced push to trigger CI

aswaterman (approved):
LGTM now, thanks @chihminchao.

emlaprise2358:
This change is causing a problem with the vssra.vi instruction 
because the vssra.vi instruction is using simm5 instead of zimm5, the immediate is getting sign extended instead of zero-extended. When the &0x1f was removed, it gives the wrong shift value when sew=64. For example if 0x13 is the intended shift value, it is actually giving 0x33.

  int sh = simm5 & (sew - 1) & 0x1f;  was changed to:
  int sh = simm5 & (sew - 1);

aswaterman:
Thanks for pointing that out, @emlaprise2358.  I'll get into it.

aswaterman:
Fixed.

emlaprise2358:
Thank you for the fix
I've had another problem with the vssra.vi instruction.
It looks like the assignment
int128_t val = vs2 
is now not sign extending the vs2 value when setting val
So 0x8fffffff is getting set to 0x8fffffff instead of 0xffffffff8fffffff
This is causing the wrong result instead of fffe4000 it is giving 00024000
I don't quite understand how changing simm5 to zimm5 could be causing this change in behavior
Is there a way to explicitly tell the compiler to sign extend the value?

emlaprise2358:
@aswaterman  I have found the problem
vssra_vi function macro loop was changed from VI_VI_LOOP to VI_VI_ULOOP -- but the instruction is supposed to sign extend the input vs2 because it is an arithmetic shift and not a logical shift. At least that is how the instruction used to behave and how I had intepreted how it should behave.

emlaprise2358:
Interesting -- and it won't compile with the VI_VI_LOOP because zimm5 is not defined -- because one argument needs to be zero-extended and the other sign extended

aswaterman:
Ah, I see.

PR #1918, opened 2025-02-11
Autotools infra

sevan:
Update extensions from autoconf-archive & regen autotools infra.


sevan:
On the CI side, the failure regarding "install.sh" is more my fault, as its removal crept in with a prior commit.
Other failures on actions seem to be related to infrastructure issues at Github itself? (cloning a repo times out?)

aswaterman:
@sevan Our CI flow tests that every commit is functional, so you'll need to fix up the PR history, rather than causing the install.sh error in an earlier commit then fixing it in a later commit.

sevan:
@aswaterman thank you for the heads up.
Hopefully I've managed to fix it now & not create a big mess. :)
Please fire off a new CI run.

aswaterman:
@sevan no prob, let's see if it works this time around.

sevan:
@aswaterman looks like it worked. :)

aswaterman (approved):
Thanks for seeing it through!

PR #1923, opened 2025-02-23
Fesvr non htif bootloader support

fhunstock-semron:
These changes are helpful in scenarios where a bootup mechanism outside of HTIF is used, but HTIF is still wanted as a communication interface. (e.g. with libgloss-htif)

FESVR already allowed to specify `"none"` as the binary, which skipped the loading.
But it still tried to reset the system, which would interfere with the function of a bootloader on the simulated system, which was still loading a binary over a different channel.

As an example, a simulator (with libfesvr) using these changes might have a command line invocation like this:
`sim +custom_boot_mechanism_binary=foo +symbol-elf=foo none`

To my knowledge these changes do not break any behaviour likely to be relied upon by any program using libfesvr.

### DIFF
--- a/fesvr/htif.cc
+++ b/fesvr/htif.cc
@@ -84,12 +84,17 @@ htif_t::~htif_t()
 
 void htif_t::start()
 {
-  if (!targs.empty() && targs[0] != "none") {
-    try {
-      load_program();
-    } catch (const incompat_xlen & err) {
-      fprintf(stderr, "Error: cannot execute %d-bit program on RV%d hart\n", err.actual_xlen, err.expected_xlen);
-      exit(1);
+  if (!targs.empty()) {
+    if (targs[0] != "none") {
+      try {
+        load_program();
+      } catch (const incompat_xlen & err) {
+        fprintf(stderr, "Error: cannot execute %d-bit program on RV%d hart\n", err.actual_xlen, err.expected_xlen);
+        exit(1);
+      }
+    } else {
+      auto empty_symbols = std::map<std::string, uint64_t>();
+      load_symbols(empty_symbols);
     }
   }
 
@@ -150,21 +155,8 @@ std::map<std::string, uint64_t> htif_t::load_payload(const std::string& payload,
   }
 }
 
-void htif_t::load_program()
+void htif_t::load_symbols(std::map<std::string, uint64_t>& symbols)
 {
-  std::map<std::string, uint64_t> symbols = load_payload(targs[0], &entry, load_offset);
-
-  // detect torture tests so we can print the memory signature at the end
-  if (symbols.count("begin_signature") && symbols.count("end_signature")) {
-    sig_addr = symbols["begin_signature"];
-    sig_len = symbols["end_signature"] - sig_addr;
-  }
-
-  for (auto payload : payloads) {
-    reg_t dummy_entry;
-    load_payload(payload, &dummy_entry, 0);
-  }
-
   class nop_memif_t : public memif_t {
    public:
     nop_memif_t(htif_t* htif) : memif_t(htif), htif(htif) {}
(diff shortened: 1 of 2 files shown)


jerryz123 (changes requested):
Thanks for breaking up these changes. The only one I'm hesitant about is "Don't send reset from HTIF, if no binary is given". This changes the existing behavior when `none` is passed as the target.

Can I ask what your use case of FESVR/HTIF looks like?  Are you using it only through the Spike ISS? Or do you have your own target? Which implementation of HTIF are you using, are you providing your own, or using TSI/DTM? 

For reference, in Chipyard sims, we add the `+no_hart0_msip` to our extension of `tsi_t`, to prevent FESVR from kicking hart0 alive.

fhunstock-semron:
The simulations have a custom serial interface, with a custom bootloader capable of loading the binary through a (simulated) host. (could be e.g. an FPGA, not unlike TSI, just not with DMA)
In the current simulations we either use the standard rocketchip sim dtm, or a custom wrapper that provides htif over jtag. (no TSI)

This modified behaviour is useful for when we want to run binaries with HTIF functionality (for debug), AND also load them through (and test) the custom boot mechanism.

I think there are 2 ways of addressing the change in reset behaviour.
- Adding a plusarg (very much like `+no_hart0_msip`)
- Adding a virtual function to `htif_t`. (`is_address_preloaded()` already provides a precedent for something like this)
- Adding an argument to the constructors of `htif_t`

The thing that confused me is that the only scenario I could think of where anyone would want to put `none` as a binary and still have it reset to `0x80000000` is when you have all of that preloaded, but that's what `is_address_preloaded()` is already for, so that case (in my mind) wouldn't be broken by these changes. (you would still specify the binary, it's just skipped)

At least in the cases of `dtm_t`&`tsi_t` that makes 0 sense to me, because you'd have to specify `none` as a binary, and then expect it to jump to your code at `DRAM_BASE`. However in that case the `to_host` and `from_host` symbols would have to be unknown (without the changes in this PR). So the scenario would be "I want htif to be running, but not load anything through it, and also not use the runtime interface", i.e. not use htif at all.

That all is to say, whatever option is the preferred one, I'd argue that _not_ resetting when `none` is provided, should be the default behaviour.
Since the only way to rely on this behaviour is not through `dtm_t` or `tsi_t`, I think adding a pure virtual function, or an additional mandatory argument to `dtm_t()` would at least cause compile errors for code bases that might rely on the current behaviour.

Either way, whichever way you prefer, I'd be happy to add it.

jerryz123:
> At least in the cases of dtm_t&tsi_t that makes 0 sense to me, because you'd have to specify none as a binary, and then expect it to jump to your code at DRAM_BASE. However in that case the to_host and from_host symbols would have to be unknown (without the changes in this PR). So the scenario would be "I want htif to be running, but not load anything through it, and also not use the runtime interface", i.e. not use htif at all.

Ah you are right, this is a bit unintuitive. I think the use case we have, is that our HTIF frontend allows users to append arbitrary MMIO reads/writes ahead of binary loading and core wakeup. 

So a potential use case is - "write this memory on the SoC, don't load a ELF, wakeup the cores, and then do nothing". Actually, now that I type this out, I can think of a time when we did use this flow to debug a a system. We wanted to test what memory regions a core could write to very precisely, so we used the MMIO writes to write a very simple 16-byte program into the first few bytes above `0x8000_0000`, instead of generating a ELF.

However, this is a very unusual use case on our end, an one that could have also been accomplished by just writing the CLINT with an additional MMIO write to manually wakeup hart0, instead of relying on TSI to do it. 

I think changing the default behavior of `none`, then, is acceptable. These changes look good then.

jerryz123 (approved):

fhunstock-semron:
Well, apparently messed something up during the cherry-pick, I'll fix it.

fhunstock-semron:
Fixed, I applied the cherry-pick from before https://github.com/riscv-software-src/riscv-isa-sim/commit/a316a37331aaaa4c8905ba37a86bde42a2c75343 was merged, which meant that I missed that argument.

PR #1926, opened 2025-02-26
Check for NAPOT PTE validity before raising SS-related faults

### LINKED ISSUE #1925: Memory access exception priority
Image
    From the documentation, the priority of exception “reserved” should be higher than that of exception “shadow stack protection rules”，
but from the code, it is just the opposite.

Image

aswaterman:
Resolves #1925

@ved-rivos can you provide a sanity check?

Note the change to `s2xlate` is not a functional change; it's just to keep the code symmetric with `walk`.


ved-rivos:
@aswaterman lgtm.

aswaterman:
Thanks!

PR #1937, opened 2025-03-17
Shadow stack inst should cause samo access fault

### LINKED ISSUE #1933: Shadow stack instructions exception error
Image
When instruction 'shadow stack' trap an exception, it is required to throw an exception according to the store type.
 
Image

ved-rivos:
closes #1933


aswaterman (approved):
Thanks!

PR #1942, opened 2025-03-24
Add Svadu/Svade Support

trdthg:
keep consistent with the qemu implementation

see https://gitlab.com/qemu-project/qemu/-/blob/master/target/riscv/cpu_helper.c#L1407
```cpp
    bool svade = riscv_cpu_cfg(env)->ext_svade;
    bool svadu = riscv_cpu_cfg(env)->ext_svadu;
    bool adue = svadu ? env->menvcfg & MENVCFG_ADUE : !svade;

    /*
     * If ADUE is enabled, set accessed and dirty bits.
     * Otherwise raise an exception if necessary.
     */
    if (adue) {
        updated_pte |= PTE_A | (access_type == MMU_DATA_STORE ? PTE_D : 0);
    } else if (!(pte & PTE_A) ||
               (access_type == MMU_DATA_STORE && !(pte & PTE_D))) {
        return TRANSLATE_FAIL;
    }
```


aswaterman (changes requested):
We have the flexibility to make profile decisions in Spike, and I'm of the fairly firm opinion that we should state that Spike _always_ supports the Svade extension (when virtual memory is supported, anyway).  This is what is required by the RVA and RVB profiles, after all.

So, I think we should incorporate your README and ISA parser changes, but the MMU changes should be backed out.

aswaterman:
BTW, sorry for the delay in reviewing this PR; it slipped through the cracks.

trdthg:
Updated

> Spike always supports the Svade extension

Then should we mark this in README? Like:

```diff
-Svadu extension, v1.0 
+Svadu extension, v1.0 (always enabled)
```

aswaterman:
I see where you're coming from, but I think it's better without the "(always enabled)".  For one, when it isn't actually enabled when virtual memory is disabled, or when Svadu is enabled by software.

aswaterman (approved):
Thank you!

PR #1946, opened 2025-04-03
Add `size` method to `abstract_device_t`; make `bus_t` more robust

aswaterman:
Check for address validity wrt. device sizes; check for device address overlap.

### DIFF
--- a/riscv/devices.cc
+++ b/riscv/devices.cc
@@ -55,26 +55,26 @@ reg_t bus_t::size()
   return 0;
 }
 
-std::pair<reg_t, abstract_device_t*> bus_t::find_device(reg_t addr)
-{
-  // Obtain iterator to device immediately after the one that might match
-  auto it = devices.upper_bound(addr);
-  if (devices.empty() || it == devices.begin()) {
-    // No devices with base address <= addr
-    return std::make_pair((reg_t)0, (abstract_device_t*)NULL);
-  }
-
-  // Rewind to device that will match if its size suffices
-  it--;
-
-  return std::make_pair(it->first, it->second);
-}
-
 std::pair<reg_t, abstract_device_t*> bus_t::find_device(reg_t addr, size_t len)
 {
-  if (auto [base, dev] = find_device(addr); addr - base + len - 1 < dev->size())
-    return std::make_pair(base, dev);
+  if (!len || addr + len - 1 < addr)
+    return std::make_pair(0, nullptr);
 
+  // Obtain iterator to device immediately after the one that might match
+  auto it_after = devices.upper_bound(addr);
+  reg_t base, size;
+  if (it_after != devices.begin()) {
+    // Obtain iterator to device that might match
+    auto it = std::prev(it_after);
+    base = it->first;
+    size = it->second->size();
+    if (addr - base + len - 1 < size) {
+      // it fully contains [addr, addr + len)
+      return std::make_pair(it->first, it->second);
+    }
+  }
+
+  // No matching device
   return std::make_pair(0, nullptr);
 }
 
--- a/riscv/devices.cc
+++ b/riscv/devices.cc
@@ -10,6 +10,27 @@ mmio_device_map_t& mmio_device_map()
 
 void bus_t::add_device(reg_t addr, abstract_device_t* dev)
 {
+  // Allow empty devices by omitting them
+  auto size = dev->size();
+  if (size == 0)
+    return;
+
+  // Reject devices that overflow address size
+  if (addr + size - 1 < addr) {
+    fprintf(stderr, "device at [%" PRIx64 ", %" PRIx64 ") overflows address size\n",
+            addr, addr + size);
+    abort();
+  }
+
+  // Reject devices that overlap other devices
+  if (auto it = devices.upper_bound(addr);
+      (it != devices.end() && addr + size - 1 >= it->first) ||
+      (it != devices.begin() && (it--, it->first + it->second->size() - 1 >= addr))) {
+    fprintf(stderr, "devices at [%" PRIx64 ", %" PRIx64 ") and [%" PRIx64 ", %" PRIx64 ") overlap\n",
+            it->first, it->first + it->second->size(), addr, addr + size);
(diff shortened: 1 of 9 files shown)


riscv/devices.cc
```diff
@@ -10,53 +10,74 @@ mmio_device_map_t& mmio_device_map()
...
+
   return std::make_pair(it->first, it->second);
 }
 
+std::pair<reg_t, abstract_device_t*> bus_t::find_device(reg_t addr, size_t len)
+{
+  if (auto [base, dev] = find_device(addr); addr - base + len - 1 < dev->size())
+    return std::make_pair(base, dev);
+
+  return std::make_pair(0, nullptr);
```
mslijepc: `find_device(addr)` can return null device pointer, so I propose we add this check first

```
  auto [base, dev] = find_device(addr);
  if (dev && addr - base + len - 1 < dev->size())
    return std::make_pair(base, dev);

  return std::make_pair(0, nullptr);
  ```

File: riscv/devices.cc [hunk 2]
```diff
@@ -10,53 +10,74 @@ mmio_device_map_t& mmio_device_map()
...
-    return false;
-  }
-  it--;
-  return it->second->store(addr - it->first, len, bytes);
+  if (auto [base, dev] = find_device(addr, len); dev)
+    return dev->store(addr - base, len, bytes);
+  return false;
+}
+
+reg_t bus_t::size()
```
mslijepc, line 63: what is the semantics of _bus_ `size`?

mslijepc:
`abstract_mem_t` has a method `virtual reg_t size() = 0;` which should be removed since base class already has it

riscv/sim.cc
```diff
@@ -406,10 +406,9 @@ void sim_t::set_rom()
 char* sim_t::addr_to_mem(reg_t paddr) {
   if (!paddr_ok(paddr))
     return NULL;
-  auto desc = bus.find_device(paddr);
+  auto desc = bus.find_device(paddr, 1);
```
mslijepc: maybe size 4 is more appropriate? or to include size in `addr_to_mem` ?

riscv/devices.cc
aswaterman: Oops, good catch.

riscv/devices.cc
aswaterman: It fits the general pattern, where a device owns addresses in the range `[base, base + size)`.  The base address of a bus is implicitly 0 (i.e. these are relative addresses, not absolute addresses), and so returning the last device's address plus its size corresponds to the extent of addresses owned by the bus.

riscv/sim.cc
aswaterman: I used `1` because it preserves the current semantics.

~I agree with your second proposal.  When `mmu_t` calls `addr_to_mem`, it should be checking that the entire 4K page surrounding the address is valid, because that is the assumption its TLB makes.~

Or, more simply, I'm making it part of the `addr_to_mem` API that memory accesses to the entire surrounding 4K page must be legal.  This matches the assumptions made by the MMU.

riscv/sim.cc
```diff
@@ -406,10 +406,9 @@ void sim_t::set_rom()
 char* sim_t::addr_to_mem(reg_t paddr) {
   if (!paddr_ok(paddr))
     return NULL;
-  auto desc = bus.find_device(paddr);
+  auto desc = bus.find_device(paddr >> PGSHIFT << PGSHIFT, PGSIZE);
   if (auto mem = dynamic_cast<abstract_mem_t*>(desc.second))
```
mslijepc, line 410: afaik `dynamic_cast` of a `nullptr` has an undefined behavior. 
maybe is more safe to do check before, e.g. 
```
if (desc.second && (auto mem = dynamic_cast<abstract_mem_t*>(desc.second)))

File: riscv/devices.cc [hunk 5]
```diff
@@ -8,53 +8,92 @@ mmio_device_map_t& mmio_device_map()
...
 }
 
-std::pair<reg_t, abstract_device_t*> bus_t::find_device(reg_t addr)
+std::pair<reg_t, abstract_device_t*> bus_t::find_device(reg_t addr, size_t len)
 {
-  // See comments in bus_t::load
-  auto it = devices.upper_bound(addr);
-  if (devices.empty() || it == devices.begin()) {
-    return std::make_pair((reg_t)0, (abstract_device_t*)NULL);
+  if (!len || addr + len - 1 < addr)
```
mslijepc: maybe we can put here `unlikely`

riscv/devices.cc
```diff
@@ -8,53 +8,92 @@ mmio_device_map_t& mmio_device_map()
...
-  return std::make_pair(it->first, it->second);
+
+  if ((it_after != devices.end() && addr + len - 1 >= it_after->first)
+      || (it_after != devices.begin() && addr - base < size)) {
+    // it_after or it contains part of, but not all of, [addr, add + len)
+    return std::make_pair(0, nullptr);
+  }
+
+  // No matching device
+  return std::make_pair(0, fallback);
```
mslijepc, line 96: I am just wandering if we should return _addr_ `0`. Most probably is fine, but looks a bit suspicious

riscv/sim.cc
aswaterman: Pretty sure this is actually well defined to return a `nullptr` of the new type as long as the return type is a pointer to a derived class of the input pointer.  See e.g. https://en.cppreference.com/w/cpp/language/dynamic_cast

"target-type is [...] pointer to Base and the type of expression is pointer to Derived such that Base is a base class of Derived, the result is
- a null pointer value if expression is a null pointer value, or [...]"

riscv/devices.cc
aswaterman: Agreed, also annotated the other paths.

riscv/devices.cc
aswaterman: I think 0 is the most self-consistent result, since logically `fallback` is a single device that covers all holes in the address map.

jerryz123 (approved):

PR #1949, opened 2025-04-07
refactor: make extensions accept processor as an argument

arromanoff:
As discussed in https://github.com/riscv-software-src/riscv-isa-sim/issues/1863 we consider a good idea to refactor custom extensions and make their handling of processor more explicit. In this commit I make all extensions's methods accept processor as an argument.

This change also makes it possible to configure instructions and disasms depending on isa info (e.g. now it is possible to add certain instructions only if we are running RV32). Previously it was not possible due to `extension_t::p` being set only after requesting the instructions from extension.

### DIFF
--- a/ci-tests/custom-csr.cc
+++ b/ci-tests/custom-csr.cc
@@ -20,15 +20,15 @@ class dummycsr_t: public csr_t {
 
 // dummy extension with dummy CSRs. Nice.
 struct xdummycsr_t : public extension_t {
-  const char *name() { return "dummycsr"; }
+  const char *name() const override { return "dummycsr"; }
 
   xdummycsr_t() {}
 
-  std::vector<insn_desc_t> get_instructions() override {
+  std::vector<insn_desc_t> get_instructions(const processor_t &) override {
     return {};
   }
 
-  std::vector<disasm_insn_t *> get_disasms() override {
+  std::vector<disasm_insn_t *> get_disasms(const processor_t *) override {
     return {};
   }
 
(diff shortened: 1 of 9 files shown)


arromanoff:
@jerryz123 Can you take a look please

ci-tests/custom-csr.cc
```diff
@@ -24,11 +24,11 @@ struct xdummycsr_t : public extension_t {
 
   xdummycsr_t() {}
 
-  std::vector<insn_desc_t> get_instructions() override {
+  std::vector<insn_desc_t> get_instructions(processor_t &) override {
```
jerryz123: Should/can these be `const`?

riscv/extension.h
```diff
@@ -11,21 +11,19 @@
...
-  virtual std::vector<insn_desc_t> get_instructions() = 0;
-  virtual std::vector<disasm_insn_t*> get_disasms() = 0;
-  virtual std::vector<csr_t_p> get_csrs ([[maybe_unused]] processor_t &proc) const { return {}; };
-  virtual const char* name() = 0;
-  virtual void reset() {};
-  virtual void set_debug(bool UNUSED value) {}
-  virtual ~extension_t();
+   virtual std::vector<insn_desc_t> get_instructions(processor_t &proc) = 0;
+   virtual std::vector<disasm_insn_t *>
+   get_disasms(processor_t *proc = nullptr) = 0;
```
jerryz123: Can you split the formatting into a different commit? Or omit it altogether from this PR (This didn't seem to do the right thing here I think)

jerryz123 (changes requested):

jerryz123 (approved):

aswaterman (approved):

PR #1950, opened 2025-04-08
Fix implementation of instreth writes

aswaterman:
Decrementing the counter in advance of incrementing it produces incorrect results when there's a carry-out.  Fix by getting rid of this scheme altogher: just skip the increment if there was an expicit write.

Looks like Spike tried to do the right thing here, but misimplemented it--i.e. it wasn't a misreading of the spec.

See test case https://github.com/riscv-software-src/riscv-tests/commit/162d3e7b61559dc47d001c6a42bb6c14982b7866 and ISA spec discussion https://github.com/riscv/riscv-isa-manual/issues/1255

cc @Timmmm


jerryz123 (approved):

Timmmm:
Thanks! I guess an official test and Spike agreeing is enough.

PR #1953, opened 2025-04-10
Enabling using Spike as a library

mslijepc:
This PR should introduce API to enable some external simulator connect to Spike.

Steps:
1. External simulator (e.g. wrapper of a simulator, or a bridge) has be derived from `abstract_sim_if_t` class and has to implement abstract functions `do_load` and `do_store`.
2. `cfg_t` is now extended with `std::optional<abstract_sim_if_t*> external_simulator` field. This field should be filled with pointer to simulator object from Step 1. 
3. `external_sim_device_t` should be constructed with pointer from step 2, and create `bus_t` with `fallback` parameter of `external_sim_device_t`. 

For all addresses which are not in Spike address space, bus will issue `load/store` requests to `fallback` branch, i.e. `external simulator`

### DIFF
--- a/riscv/devices.h
+++ b/riscv/devices.h
@@ -71,6 +71,13 @@ class mem_t : public abstract_mem_t {
   reg_t sz;
 };
 
+class abstract_sim_if_t {
+public:
+  virtual ~abstract_sim_if_t() = default;
+  virtual bool do_load(reg_t addr, size_t len, uint8_t* bytes) = 0;
+  virtual bool do_store(reg_t addr, size_t len, const uint8_t* bytes) = 0;
+};
+
 class clint_t : public abstract_device_t {
  public:
   clint_t(const simif_t*, uint64_t freq_hz, bool real_time);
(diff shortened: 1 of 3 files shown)


riscv/devices.h
```diff
@@ -71,6 +71,25 @@ class mem_t : public abstract_mem_t {
   reg_t sz;
 };
 
+class abstract_sim_if_t {
+public:
+  virtual ~abstract_sim_if_t() = default;
+  virtual bool do_load(reg_t addr, size_t len, uint8_t* bytes) = 0;
+  virtual bool do_store(reg_t addr, size_t len, const uint8_t* bytes) = 0;
```
aswaterman: Does the `do_` prefix actually help us out here?  Wouldn't `load`/`store` be equally appropriate?

mslijepc: Yes, but I guess that most of simulators already have `load/store` , maybe not with the same signature. I can rename it

aswaterman: Noted, but I think we should go ahead and rename them.

mslijepc: Renamed

aswaterman (approved):

PR #1955, opened 2025-04-10
mmu: fix pointer masking with mstatus.MXR in M-mode

radimkrcmar:
Pointer masking is disabled if sstatus.MXR is in effect, but M-mode doesn't use paging and hence sstatus.MXR should not be in effect for effective M-mode loads.

Do not consider mstatus.MXR when deciding the pointer mask length for effective M-mode loads.

vsstatus.MXR wasn't an issue as no effective M-mode load could happen with V=1, but move it below M-mode as well to simplify the conditions.

Fixes: 71bdc3bbd13a ("pointer masking: Pointer masking does not apply when MXR=1 regardless of MPRV in v1.0.0-rc2")


aswaterman (approved):
Looks correct to me.  Thanks, @radimkrcmar.

PR #1961, opened 2025-04-16
Avoid appending chars to result `strtolower`

i3abghany:
Avoids reallocations.

BTW, the function is not used in the processor.cc file (along with other static util functions). Should we remove them as they are unused?


aswaterman (approved):
I'll make a follow-up PR that deletes the unused functions and turns on `-Werror` for this case in the CI flow.

PR #1963, opened 2025-04-16
Make compilation with -Wextra -Wall -Werror work

aswaterman:
...at least for current GCC.

Turning on these errors for regular builds is generally ill-advised, since compiler upgrades (or downgrades) can result in spurious compilation failures.  So, only enable them in CI.

I gave up on `-Wpedantic`, as we use the GNUism `({ ... })` ubiquitously.


aswaterman:
@jerryz123 I'm going to yolo-merge this because it's a minefield for merge conflicts.  But I'm happy to address any concerns after the fact.

PR #1966, opened 2025-04-29
Substantially improve uncommon-case memory-system performance

aswaterman:
Several commingled improvements, but no API changes:

- Allow TLB to be used for MMIO accesses.  This speeds up simple MMIO accesses (e.g. reading the boot ROM) by >2x when VM is disabled or 5x when VM is enabled.
- Skip checking the debug triggers for MMIO accesses and misaligned-but-intrapage memory accesses unless necessary, speeding up these cases by another >2x.
- Allow TLB to be used even when a memtracer is registered on an address.  This speeds up D$ simulation by 6x.
- Give the load/store/fetch TLBs their own tags, since they already had their own data anyway.  This eliminates a common source of conflict misses and simplifies the implementation.
- Also fix an ubsan error associated with outside-of-object memory accesses when computing host addresses.
- Also improve common-case performance by moving uncommon-case checks (commit log, matched debug trigger) off the critical code path.

### DIFF
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -342,10 +342,6 @@ void processor_t::step(size_t n)
     }
     catch (triggers::matched_t& t)
     {
-      if (mmu->matched_trigger) {
-        delete mmu->matched_trigger;
-        mmu->matched_trigger = NULL;
-      }
       take_trigger_action(t.action, t.address, pc, t.gva);
     }
     catch(trap_debug_mode&)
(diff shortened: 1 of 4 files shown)


jerryz123 (approved):
This is quite a lot of code change, I think I'd need to reread the fetch/load/store path in its entirety to appreciate the new flow, but the individual commits here look good.

How much faster is this now when running in a bus-functional model? I'm interpreting a 4x speedup from your changelog.

aswaterman:
Yeah, if I knew what I was getting into beforehand, I would've split this up into multiple PRs over multiple days...

Yeah, aggregate speedup is in the 4x range when half of instructions are loads/stores and use of the functional I$ is allowed.  I didn't write down the measurements for the case where use of the I$ is disallowed, but it's at least another 2x on top of that.

PR #1968, opened 2025-04-29
exposing riscv and softfloat as libraries

mslijepc:
this PR will make `riscv` and `softfloat` exposed and easily linked in case we want to link `riscv-isa-sim` with some other simulation environment.

### DIFF
--- a/riscv/riscv.mk.in
+++ b/riscv/riscv.mk.in
@@ -13,6 +13,8 @@ riscv_install_shared_lib = yes
 
 riscv_install_pcs = yes
 
+riscv_install_lib = yes
+
 riscv_install_prog_srcs = \
 
 riscv_install_hdrs = \
--- a/softfloat/softfloat.mk.in
+++ b/softfloat/softfloat.mk.in
@@ -247,6 +247,8 @@ softfloat_c_srcs = \
 
 softfloat_install_shared_lib = yes
 
+softfloat_install_lib = yes
+
 softfloat_test_srcs =
 
 softfloat_install_hdrs = \


aswaterman (dismissed):

jerryz123:
Hang on, softfloat should be installed through softfloat, right?

aswaterman:
I think we will have to learn from other users whether this causes annoyance, namely as a result of overwriting `libsoftfloat` with the Spike verison.  If so, we will need to rename this one.  But we can cross that bridge when we come to it.

aswaterman:
OK, since @jerryz123 and I had similar reactions, let's hold off on this one.

jerryz123:
Also this change wasn't necessary for my flow to link with libriscv at the installed path. I'll need to check how this works

aswaterman:
Spike's softfloat has its own augmentations, so we should avoid naming it the same as public softfloat.

We might want to namespace both of these libraries (e.g. `libspike-riscv` and `libspike-softfloat`).  But let's give Jerry time to opine before we take further action.

Furthermore, I'm not sure that `libspike-softfloat` should actually be distinct from `libspike-riscv`.  It seems like Spike-as-library should lump everything into something called `libspike` (or `libriscv` in Jerry's flow, or whatever).  This might require more Makefile hacking, but it's better to figure that out now, rather than creating an ABI that we'll soon break.

jerryz123:
As it turns out, the Softfloat repo provides no flow for installing libsoftfloat and its headers... only this repo does. Given that, I don't think there's any risk from installing libsoftfloat.a. This repo already installs all the softfloat headers, and the .so as well. Installing libriscv.a in addition to libriscv.so is also acceptable.

If someone were to untangle this mess properly, I think they should:

 - Add a proper build flow to berkeley-softfloat-3. This build flow should install variants of Softfloat specialized for different architectures (libsoftfloat-x86, libsoftfloat-armvfp2, libsoftfloat-riscv, etc.)
 - Excise all softfloat sources here, and add `-lsoftfloat-riscv` when linking.
 
I don't expect ^ to get done anytime soon, so this PR should be good as it stands. @mslijepc , you have a use case which requires the static libraries to be installed instead of the dynamic libs?

mslijepc:
@jerryz123 Here is a demo project which I made for our usecase (using Spike as a library) https://github.com/sifive/riscv-isa-sim-lib-demo 
You can check [Makefile](https://github.com/sifive/riscv-isa-sim-lib-demo/blob/main/src/Makefile). It wasn't possible to link w/o installing `riscv` and `softfloat` libs. 
If you have some solution for this please let me know

jerryz123:
You are linking with the static lib. It should be possible to do dynamic linkage (which is what I do)

`-lriscv` is all you need, and then either 
 - set the env variable `LD_LIBRARY_PATH=/path/to/lib` or 
 - pass to the linker `-L /path/to/lib`
 
 In my flow, I install all the RISCV related stuff (including spike) to a sysroot at `$RISCV`. Perhaps your flow is similar? In which case `-L $RISCV/lib` should be sufficient.

jerryz123:
The ci-tests/ directory here also contains some code + scripts (`test-spike`) demonstrating how to compile simple programs that link with libriscv.

mslijepc:
Hi @jerryz123 , thanks for the info and help! I adjusted my flow, looks ok. 
Let me do some more tests, I think I will close this PR w/o being merged

Issue #1956, opened 2025-04-11
Errors with VLEN == 32 configuration


gregdavill:
## Summary

The riscv vector specification lists an embedded configuration: zve32x, and lists that it has a minimum VLEN of 32. Spike accepts this as options on the isa string, and some vector operation work, But many cause an assertion error.
```
spike: ../riscv/vector_unit.cc:71: T& vectorUnit_t::elt(reg_t, reg_t, bool) [with T = long unsigned int; reg_t = long unsigned int]: Assertion `(VLEN >> 3)/sizeof(T) > 0' failed.
```

It seems that in the vector implementation internal working copies and masks currently create 64bit types, this causes an assertion to trip.

I think this is the underlying issue faced in #1685. Here is a test-case which activates this with vanilla spike.

## Test case

vlen32.S
```asm
.globl _start
.section .text
_start:
    // Enable vector instructions
    li x10, 0x200
    csrs mstatus, x10

    li x11, 2
    vsetvli x11, x0, e32, m1, ta, ma
    vadd.vx v24, v28, x12, v0.t
    ret
```

link.ld
```
OUTPUT_ARCH("riscv")
ENTRY(_start)

SECTIONS
{
    . = 0x80000000;
    .text : { *(.text) }
}
```

Commands to build and run
```bash
$ riscv-none-elf-gcc -nostdlib -nostartfiles -march=rv32imac_zve32x_zicsr -T link.ld vlen32.S -o vlen32.elf
$ spike --isa=rv32imac_zve32x_zvl32b -l --log-commits --instructions=32 vlen32.elf
core   0: 0x00001000 (0x00000297) auipc   t0, 0x0
core   0: 3 0x00001000 (0x00000297) x5  0x00001000
core   0: 0x00001004 (0x02028593) addi    a1, t0, 32
core   0: 3 0x00001004 (0x02028593) x11 0x00001020
core   0: 0x00001008 (0xf1402573) csrr    a0, mhartid
core   0: 3 0x00001008 (0xf1402573) x10 0x00000000
core   0: 0x0000100c (0x0182a283) lw      t0, 24(t0)
core   0: 3 0x0000100c (0x0182a283) x5  0x80000000 mem 0x00001018
core   0: 0x00001010 (0x00028067) jr      t0
core   0: 3 0x00001010 (0x00028067)
core   0: 0x80000000 (0x20000513) li      a0, 512
core   0: 3 0x80000000 (0x20000513) x10 0x00000200
[...]

PR #1969, opened 2025-04-29
Fix vlen == 32 vector configuration

gregdavill:
Following from discussion in #1956 

This PR adjusts the logic to pass the assert when VLEN == 32, and clamps the elts_per_reg to 1 when a 64bit mask element is created.


aswaterman:
Now that I've looked at this more carefully, I'm concerned that suppressing this assertion will make it harder to find other VLEN=32 bugs that we haven't yet run across.  I'd rather fix this in a more disciplined way.  I'll follow up with an alternate PR.

aswaterman:
Superseded by #1972

PR #1987, opened 2025-05-13
Fix regression in vssra.vi instruction

aswaterman:
See https://github.com/riscv-software-src/riscv-isa-sim/pull/1915#issuecomment-2874313019


aswaterman:
cc @chihminchao

jerryz123 (approved):

PR #1990, opened 2025-05-14
changing type of external_simulator member of external_sim_device_t

mslijepc:
Even though we allowed `void*` data member of `external_simulator`, in `load/store` we need do `static cast`. I.e. `static_cast<abstract_sim_if_t*>(external_simulator)`. I think it is better to have proper data member type which also prevents some bugs which can happen with `static_cast`

### DIFF
--- a/riscv/devices.h
+++ b/riscv/devices.h
@@ -80,14 +80,14 @@ public:
 
 class external_sim_device_t : public abstract_device_t {
 public:
-  external_sim_device_t(void* sim);
-  void set_simulator(void* sim);
+  external_sim_device_t(abstract_sim_if_t* sim);
+  void set_simulator(abstract_sim_if_t* sim);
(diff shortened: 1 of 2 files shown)


riscv/devices.h
```diff
@@ -80,14 +80,14 @@ class abstract_sim_if_t {
...
-  void set_simulator(void* sim);
+  external_sim_device_t(abstract_sim_if_t* sim);
+  void set_simulator(abstract_sim_if_t* sim);
   bool load(reg_t addr, size_t len, uint8_t* bytes) override;
   bool store(reg_t addr, size_t len, const uint8_t* bytes) override;
   reg_t size() override;
 
 private:
-  void* external_simulator;
+abstract_sim_if_t* external_simulator;
```
aswaterman: Restore the two spaces.

aswaterman (approved):

PR #1993, opened 2025-05-15
feat: move cache block size initialization to constructor

arromanoff:
Previously cache block size had to be initialized via special set_chache_blocksz setter function and was uninitialized if you didn't call it. This caused us some issues when we forgot to call a setter function and everything worked fine until first Zicbom instruction where spike went into an infinite loop (on uninitialized value). This commit was made as a measure to prevent similar issues in the future. In this commit I move initialization of the block size to the mmu_t's constructor. The value of this member goes through cfg_t struct. This way the value will always be initialized without loosing ability to be reset via mentioned setter function if one needs to do so.

### DIFF
--- a/riscv/cfg.cc
+++ b/riscv/cfg.cc
@@ -47,4 +47,5 @@ cfg_t::cfg_t()
   explicit_hartids = false;
   real_time_clint  = false;
   trigger_count    = 4;
+  cache_blocksz    = 64;
 }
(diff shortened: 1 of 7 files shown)


aswaterman (commented):
Shouldn't we get rid of the `set_cache_blocksz` method?

arromanoff:
I agree. Removed `set_cache_blocksz` method

aswaterman (approved):
Thanks, LGTM.

PR #1997, opened 2025-05-21
Partially revert #1987 to fix regrssion in vsra.vi and vssra.vi 

### LINKED ISSUE #1996: vssra.vi and vsra.vi don't sign extend vs2
In https://github.com/riscv-software-src/riscv-isa-sim/pull/1915 the vssra_vi and vsra_vi function macro loop were changed from VI_VI_LOOP to VI_VI_ULOOP -- but the instruction is supposed to sign extend the input vs2 because it is an arithmetic shift and not a logical shift. The loop was changed because the immediate needs to be zero extended and the zimm5 variable was not available when the VI_VI_LOOP was used. But changing the loop makes the vs2 operand zero-extended instead of sign extended.

When val is set, the vs2 operand is now not sign extended
So 0x8fffffff is getting set to 0x8fffffff instead of 0xffffffff8fffffff
This is causing the wrong result instead of fffe4000 it is giving 00024000

From the 1915 pull request:
This change is causing a problem with the vssra.vi instruction
because the vssra.vi instruction is using simm5 instead of zimm5, the immediate is getting sign extended instead of zero-extended. When the &0x1f was removed, it gives the wrong shift value when sew=64. For example if 0x13 is the intended shift value, it is actually giving 0x33.

int sh = simm5 & (sew - 1) & 0x1f; was changed to:
int sh = simm5 & (sew - 1);

aswaterman:
Resolves #1996.

Use arithmetic shifts, not logical shifts.

I broke these instructions while fixing #1915.

### DIFF
--- a/riscv/insns/vsra_vi.h
+++ b/riscv/insns/vsra_vi.h
@@ -1,5 +1,5 @@
 // vsra.vi vd, vs2, zimm5
-VI_VI_ULOOP
+VI_VI_LOOP
 ({
-  vd = vs2 >> (zimm5 & (sew - 1));
+  vd = vs2 >> (insn.v_zimm5() & (sew - 1));
 })
--- a/riscv/insns/vssra_vi.h
+++ b/riscv/insns/vssra_vi.h
@@ -1,8 +1,8 @@
 // vssra.vi vd, vs2, zimm5
-VI_VI_ULOOP
+VI_VI_LOOP
 ({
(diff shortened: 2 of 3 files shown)


jerryz123 (approved):
Ah right, the mix of signed/unsigned operands doesn't work with VI_VI_(U)LOOP.

aswaterman:
Yeah, oops.

PR #1998, opened 2025-05-24
Update README with Zimop extension

mingtians:


aswaterman (approved):
Alas, this list is woefully out of date.

PR #1999, opened 2025-06-06
Implement IMSIC and APLIC

i2h2:
Boots SMP Linux as kvm guest with VS IMSIC interrupts. Dependent on https://github.com/riscv/riscv-opcodes/pull/354

@binno @glg-rv

### DIFF
--- a/riscv/aplic.cc
+++ b/riscv/aplic.cc
@@ -1,0 +1,328 @@
+// See LICENSE for license details.
+#include "devices.h"
+#include "processor.h"
+#include "arith.h"
+#include "sim.h"
+#include "dts.h"
+
+#define DOMAINCFG       0x0
+#define SOURCECFG_BASE  0x4
+#define MMSIADDRCFG     0x1bc0
+#define MMSIADDRCFGH    0x1bc4
+#define SMSIADDRCFG     0x1bc8
+#define SMSIADDRCFGH    0x1bcc
+#define SETIP_BASE      0x1c00
+#define SETIPNUM        0x1cdc
+#define IN_CLRIP_BASE   0x1d00
+#define CLRIPNUM        0x1ddc
+#define SETIE_BASE      0x1e00
+#define SETIENUM        0x1edc
+#define CLRIE_BASE      0x1f00
+#define CLRIENUM        0x1fdc
+#define GENMSI          0x3000
+#define TARGET_BASE     0x3004
+#define IDC             0x4000
+
+#define DOMAINCFG_IE_MASK       0x100
+#define DOMAINCFG_DM_MASK       0x4
+
+#define SOURCECFG_D_MASK        0x400
+#define SOURCECFG_CHILD_MASK    0x3ff
+#define SOURCECFG_SM_MASK       0x7
+
+#define MMSIADDRCFGH_L_MASK     0x80000000
+
+#define GENMSI_HART_MASK        0xfffc0000
+#define GENMSI_BUSY_MASK        0x1000
+#define GENMSI_EIID_MASK        0x7ff
+
+#define TARGET_HART_MASK        0xfffc0000
+#define TARGET_GUEST_MASK       0x3f000
+#define TARGET_EIID_MASK        0x7ff
+
+aplic_t::aplic_t(const simif_t *simif, aplic_t *parent) : simif(simif), parent(parent), child(nullptr), domaincfg((0x80 << 24) | DOMAINCFG_DM_MASK), sourcecfg(), mmsiaddrcfgh(MMSIADDRCFGH_L_MASK), ie(), target(), level(), genmsi(0), deleg_mask()
+{
+}
+
+void aplic_t::delegate(uint32_t id, bool dm)
+{
+  if (!id || id >= APLIC_MAX_DEVICES)
+    return;
+
+  if (!child) {
+    // child's sourcecfg = 0 when interrupt is changed to delegated
+    if (!parent->delegated(id) && dm)
+      sourcecfg[id] = 0;
+  } else {
+    uint32_t mask = (1 << (id % 32));
+    deleg_mask[id / 32] &= ~mask;
+    deleg_mask[id / 32] |= dm ? mask : 0;
+    child->delegate(id, dm);
+  }
+}
+
+bool aplic_t::delegated(uint32_t id)
+{
+  if (!id || id >= APLIC_MAX_DEVICES)
+    return false;
+  return (sourcecfg[id] & SOURCECFG_D_MASK);
+}
+
+uint32_t aplic_t::get_deleg_mask(uint32_t idx) {
+    if (idx >= APLIC_MAX_DEVICES / 32)
+      return 0;
+    if (parent)
+      return parent->get_deleg_mask(idx);
+    return deleg_mask[idx];
+  }
+
+bool aplic_t::interrupt_enabled(uint32_t id)
+{
+  if (!id || id >= APLIC_MAX_DEVICES)
+    return false;
+  if (parent && !parent->delegated(id))
+    return false;
+  // Domain interrupt enabled, source IE and source mode != inactive (0)
+  return (domaincfg & DOMAINCFG_IE_MASK) && (ie[id / 32] & (1 << (id % 32))) && (sourcecfg[id] & SOURCECFG_SM_MASK);
+}
+
+bool aplic_t::accessible(uint32_t id)
+{
+  if (!id || id >= APLIC_MAX_DEVICES)
+    return false;
+  if (!parent)
+    return true;
+  return parent->delegated(id);
+}
+
+void aplic_t::set_interrupt_level(uint32_t id, int lvl)
+{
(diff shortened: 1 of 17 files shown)


aswaterman:
@binno @arrv-sc @chihminchao can y'all lend a hand reviewing this large but useful PR?

riscv/aplic.cc
```diff
@@ -0,0 +1,328 @@
...
+
+  update_interrupt(id);
+}
+
+void aplic_t::send_msi(uint32_t proc_id, uint32_t guest, uint32_t eiid)
+{
+  auto &procs = simif->get_harts();
+  auto it = procs.find(proc_id);
+  if (!eiid || it == procs.end())
+    return;
```
arromanoff, line 120: Shouldn't this be an internal error/assert instead of a NOP?

riscv/aplic.cc
```diff
@@ -0,0 +1,328 @@
...
+    val = ie[idx];
+  } else if (addr == GENMSI) {
+    val = genmsi;
+  } else if (addr >= TARGET_BASE && addr < IDC) {
+    auto idx = (addr - TARGET_BASE) / 4 + 1;
+    if (accessible(idx))
+      val = target[idx];
+  }
+
+  memcpy(bytes, (uint8_t *)&val, len);
```
arromanoff, line 194: [nit] `uint8_t *` cast inside memcpy is unnecessary as both source and destination accept `void *` and any pointer implicitly casts to `void *`

riscv/aplic.cc
```diff
@@ -0,0 +1,328 @@
...
+        delegate(idx, val & SOURCECFG_D_MASK);
+
+      // force SM to be 6 if non-zero
+      if (val & SOURCECFG_SM_MASK)
+        val = set_field(val, SOURCECFG_SM_MASK, 6);
+
+      sourcecfg[idx] = val;
+    }
+  } else if (addr >= SETIP_BASE && addr < SETIPNUM) {
+    auto idx = (addr - SETIP_BASE) / 4;
```
arromanoff, line 230: This `4` keeps appearing throughout this function. Can we name it somehow? constexpr variable or define would be nice.

riscv/csr_init.cc
```diff
@@ -215,8 +205,13 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
...
   add_hypervisor_csr(CSR_HSTATUS, hstatus = std::make_shared<hstatus_csr_t>(proc, CSR_HSTATUS));
-  add_hypervisor_csr(CSR_HGEIE, std::make_shared<const_csr_t>(proc, CSR_HGEIE, 0));
-  add_hypervisor_csr(CSR_HGEIP, std::make_shared<const_csr_t>(proc, CSR_HGEIP, 0));
+  if (proc->extension_enabled_const(EXT_SSAIA)) {
+    add_hypervisor_csr(CSR_HGEIE, hgeie = std::make_shared<hgeie_csr_t>(proc, CSR_HGEIE, proc->geilen));
+    add_hypervisor_csr(CSR_HGEIP, hgeip = std::make_shared<hgeip_csr_t>(proc, CSR_HGEIP));
+  } else {
+    add_hypervisor_csr(CSR_HGEIE, std::make_shared<const_csr_t>(proc, CSR_HGEIE, 0));
+    add_hypervisor_csr(CSR_HGEIP, std::make_shared<const_csr_t>(proc, CSR_HGEIP, 0));
+  }
```
arromanoff, line 214: Q: Why don't we set hgeie/hgeip members under `else` branch?

riscv/csrs.h
```diff
@@ -998,4 +998,61 @@ class aia_csr_t: public masked_csr_t {
...
+  bool vs;
+  csrmap_t_p csrmap;
+};
+typedef std::shared_ptr<aia_ireg_proxy_csr_t> aia_ireg_proxy_csr_t_p;
+
+class imsic_file_t;
+typedef std::shared_ptr<imsic_file_t> imsic_file_t_p;
+class topei_csr_t: public csr_t {
+ public:
+  topei_csr_t(processor_t* const proc, const reg_t addr, imsic_file_t_p const imsic);
```
arromanoff, line 1039: [nit][here and around] I am not a fan of moving trivial constructors to the `.cc` files. It does not reduce number of dependencies but negatively affects readability and prevents this function from being inlined. I don't think it's critical, however, as we already do that a lot.

riscv/devices.h
```diff
@@ -159,6 +159,48 @@ class plic_t : public abstract_device_t, public abstract_interrupt_controller_t
...
+  void set_interrupt_level(uint32_t id, int lvl);
+  reg_t size() { return APLIC_SIZE; }
+  void set_child(aplic_t *c) { child = c; }
+  void delegate(uint32_t id, bool dm);
+  bool delegated(uint32_t id);
+  uint32_t get_deleg_mask(uint32_t idx);
+ private:
+  const simif_t *simif;
+  class aplic_t *parent;
+  class aplic_t *child;
```
arromanoff, line 177:
```suggestion
  aplic_t *parent;
  aplic_t *child;
```

riscv/devices.h
```diff
@@ -159,6 +159,48 @@ class plic_t : public abstract_device_t, public abstract_interrupt_controller_t
...
+  uint32_t mmsiaddrcfgh;
+  uint32_t ie[APLIC_MAX_DEVICES / 32];
+  uint32_t target[APLIC_MAX_DEVICES];
+  uint32_t level[APLIC_MAX_DEVICES / 32];
+  uint32_t genmsi;
+  // deleg_mask can be directly applied to other packed registers
+  uint32_t deleg_mask[APLIC_MAX_DEVICES / 32];
+
+  bool interrupt_enabled(uint32_t id);
+  bool accessible(uint32_t id);
```
arromanoff, line 189:
```suggestion
  bool accessible(uint32_t id) const;
```

riscv/devices.h
```diff
@@ -159,6 +159,48 @@ class plic_t : public abstract_device_t, public abstract_interrupt_controller_t
...
+  uint32_t sourcecfg[APLIC_MAX_DEVICES];
+  uint32_t mmsiaddrcfgh;
+  uint32_t ie[APLIC_MAX_DEVICES / 32];
+  uint32_t target[APLIC_MAX_DEVICES];
+  uint32_t level[APLIC_MAX_DEVICES / 32];
+  uint32_t genmsi;
+  // deleg_mask can be directly applied to other packed registers
+  uint32_t deleg_mask[APLIC_MAX_DEVICES / 32];
+
+  bool interrupt_enabled(uint32_t id);
```
arromanoff, line 188:
```suggestion
  bool interrupt_enabled(uint32_t id) const;
```

riscv/dts.cc
```diff
@@ -441,3 +441,58 @@ int fdt_parse_hartid(const void *fdt, int cpu_offset, uint32_t *hartid)
...
   return 0;
 }
+
+int fdt_parse_imsics(const void *fdt, reg_t *imsic_m_addr, reg_t *imsic_s_addr, const char *compatible)
+{
+  int nodeoffset = -1, len, rc;
+  const fdt32_t *p;
+  int val;
+
+  for (int i = 0; i < 2; i++) {
```
arromanoff, line 451: What does this `2` represent? It is hard to read/maintain this code without knowing every part of APLIC spec. Please add a constexpr variable/define. Such constants deserve to be named.

riscv/imsic.cc
```diff
@@ -0,0 +1,177 @@
...
+  }
+  return true;
+}
+
+imsic_t::imsic_t(processor_t *proc, unsigned geilen) {
+  if (proc->extension_enabled_const(EXT_SMAIA))
+    m = std::make_shared<imsic_file_t>(proc, MIP_MEIP, IMSIC_M_FILE_REGS);
+  if (proc->extension_enabled_const(EXT_SSAIA)) {
+    s = std::make_shared<imsic_file_t>(proc, MIP_SEIP, IMSIC_S_FILE_REGS);
+    assert(geilen <= 63);
```
arromanoff, line 124: [nitpick] This would be much more readable:
```suggestion
    assert(geilen < 64);
```

riscv/sim.cc
```diff
@@ -262,6 +300,15 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
         assert(!plic);
         plic = std::static_pointer_cast<plic_t>(dev_ptr);
       }
+
+      if (dynamic_cast<aplic_m_t*>(&*dev_ptr)) {
+        assert(!aplic_m);
+        aplic_m = std::static_pointer_cast<aplic_m_t>(dev_ptr);
+      }
```
arromanoff, line 307:
```suggestion
      if (auto *aplic_m_new = dynamic_cast<aplic_m_t*>(&*dev_ptr)) {
        assert(!aplic_m);
        aplic_m = aplic_m_new;
      }
```

arromanoff (changes requested):
I would appreciate if someone else would also take a look at this. As I am not very familiar with APLIC

PR #2004, opened 2025-06-11
Bugfix cm.jalt

jmonesti:
Hello @aswaterman & al.

Completed my SystemC/TLM stuff, but when sync'ing upstream, I noticed lots of changes in mmu_t,
including a rewrite of `mmu_t::fetch_jump_table()`

Unfortunately, this breaks cm.jalt on my side -even on my 'non TLM' model.

I suspect a typo in `mmu_t::fetch_jump_table()` (note the `* 8`)

Could you please confirm my understanding ?

Thanks again
Jean-François


aswaterman (approved):
Yes, thanks, sorry about that.

PR #2008, opened 2025-06-18
Bypass masked-off elements for viota

HaochenGui:
This patch bypasses masked-off elements for viota.  Thus these elements keep untouched.  It's helpful to detect inactive elements.


aswaterman (changes requested):
This PR does more than what the description says it does.  Please strip out the commit related to widening.

HaochenGui:
I reset my branch and remove the widening patch. Sorry for the confusing.

aswaterman (approved):

PR #2010, opened 2025-06-18
Performance: processor_t::take_interrupt() should check EXT_SSAIA

jmonesti:
Hello @aswaterman & @YenHaoChen 

Following my latest upstream resync,
our Jenkins CI/CD non-reg. platform shows a noticeable performance drop

After some valgrind & kcachegrind digging,
it turns out that the implementation of SMAIA and SSAIA extensions incurs a significant performance penalty.

In particular, in method `processor_t::take_interrupt()`, I'd suggest to check whether EXT_SSAIA is enabled before accessing the considered CSR's

Your feedback is welcome

All the best meanwhile,
Jean-François

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -303,9 +303,15 @@ bool processor_t::is_handled_in_vs()
 
 void processor_t::take_interrupt(reg_t pending_interrupts)
 {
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
Fine with me.

PR #2014, opened 2025-06-23
Adjust vsdeleg from VSSIP to SSIP interrupt occurred

binno:
As title

### DIFF
--- a/riscv/processor.cc
+++ b/riscv/processor.cc
@@ -447,6 +447,7 @@ void processor_t::take_trap(trap_t& t, reg_t epc)
   bool supv_double_trap = false;
   if (interrupt) {
     vsdeleg = (curr_virt && state.prv <= PRV_S) ? state.hideleg->read() : 0;
+    vsdeleg >>= 1;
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

i2h2:
Something seems off with this change - if there's an HS interrupt while VS is running (e.g. SEIP), it would get trapped into VS (i.e. as if VSEIP was asserted). It seems like https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/processor.cc#L313C5-L313C25 just takes VS interrupts from vstopi and reports/throws them as S-mode interrupts.

I pretty much had to rework VS interrupt handling in https://github.com/i2h2/riscv-isa-sim/tree/vs-interrupts

aswaterman:
Not sure if this is related to #2031

i2h2:
> Not sure if this is related to #2031

Unrelated.

aswaterman:
Then, @i2h2, can you file an issue?  Comments on merged PRs tend to get lost.

i2h2:
https://github.com/riscv-software-src/riscv-isa-sim/pull/2033

PR #2022, opened 2025-06-28
remove unused EGS constant from vsm4r_vs handler

aap-sc:
The macro require_vsm4_constraints defines EGS once again (shadowing the definition of this constant, since it introduces a new scope). This renders the original definition as "unused", causing warning during compilation.


aap-sc:
@tsewei-lin would you kindly take a look?

tsewei-lin:
That is a mistake. Thank you for fixing this.

aswaterman (approved):

PR #2025, opened 2025-07-02
added jtag_dtm and remote_bitbang to riscv_install_hdrs

mslijepc:
Need to add those `jtag_dtm.h` and `remote_bitbang.h` to the list of files which will be installed in case I want to use Spike as a library with `GDB` attachment

### DIFF
--- a/riscv/riscv.mk.in
+++ b/riscv/riscv.mk.in
@@ -33,11 +33,13 @@ riscv_install_hdrs = \
 	entropy_source.h \
 	extension.h \
 	isa_parser.h \
+	jtag_dtm.h \
 	log_file.h \
 	memtracer.h \
 	mmu.h \
 	platform.h \
 	processor.h \
+	remote_bitbang.h \
 	rocc.h \
 	sim.h \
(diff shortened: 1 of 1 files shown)


aswaterman (commented):
This is surprising to me.  Normally we only need to install headers that are part of the API, or are included by headers that are part of the API.  But no headers include `remote_bitbang.h`, and the only header that includes `jtag_dtm.h` is `remote_bitbang.h`.  So why do they need to be installed?

mslijepc:
It is true that those two files don't belong to Spike API, and maybe shouldn't be installed. 

But in case I want to connect GDB to Spike wrapper (and link with installed Spike on the machine _w/o specifying `SPIKE_SOURCE_DIR` folder_) I need to include those files. 
Forward declaration is not enough, neither installing only `remote_bitbang.h` 
Here is a use-case example [changes needed to connect GDB](https://github.com/sifive/riscv-isa-sim-lib-demo/pull/3)

aswaterman (approved):
That explanation works for me.

PR #2033, opened 2025-07-15
AIA: vstopi should only be used for AIA-extended interrupts

i2h2:
VSEIP/VSTIP/VSSIP should come in through the non-SSAIA path. vstopi will be used to throw VTI and IID > 13 interrupts.

@binno - let me know if I missed something.


binno:
LGTM. Thanks for your correction, @i2h2

aswaterman (approved):

PR #2036, opened 2025-07-28
Fix pmpaddr read

trdthg:
As the spec said

- When G>=1, the NA4 mode is not selectable.
- When G>=2 and pmpcfg[i]A[1] is set, i.e. the mode is NAPOT,
  then bits pmpaddr[i][G-2:0] read as all ones.
- When G>=2 and pmpaddr[i]A[1] is clear, i.e. the mode is OFF or TOR,
  then bits pmpaddr[i][G-1:0] read as all zeros.

and [63, 54] bits of pmpaddr are WARL

---

When I was comparing spike and sail, I found the return value of read_pmpaddr are different.

I think the issue should came from spike, and this PR mainly references the [sail implementation](https://github.com/riscv/sail-riscv/blob/87eb6fb8f5db66f52cc570948acd78a1ffec0f1d/model/riscv_pmp_regs.sail).

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -103,9 +103,19 @@ void pmpaddr_csr_t::verify_permissions(insn_t insn, bool write) const {
 }
 
 reg_t pmpaddr_csr_t::read() const noexcept {
-  if ((cfg & PMP_A) >= PMP_NAPOT)
-    return val | (~proc->pmp_tor_mask() >> 1);
-  return val & proc->pmp_tor_mask();
+  reg_t G = proc->lg_pmp_granularity;
+  uint8_t mode = cfg & PMP_A;
+  if (G >= 2 && mode == PMP_NAPOT) {
+    uint64_t mask =
+        make_mask64(0, std::min((unsigned int)(G - 1), proc->get_xlen()));
+    return val | mask;
+  } else if (G >= 1 && (mode == PMP_TOR | mode == PMP_OFF)) {
+    uint64_t mask =
+        make_mask64(0, std::min((unsigned int)(G), proc->get_xlen()));
+    return val & ~mask;
+  } else {
+    return val;
+  }
 }
 
 bool pmpaddr_csr_t::unlogged_write(const reg_t val) noexcept {
@@ -121,7 +131,7 @@ bool pmpaddr_csr_t::unlogged_write(const reg_t val) noexcept {
   const bool locked = !lock_bypass && (cfg & PMP_L);
 
   if (pmpidx < proc->n_pmp && !locked && !next_locked_and_tor()) {
-    this->val = val & ((reg_t(1) << (MAX_PADDR_BITS - PMP_SHIFT)) - 1);
+    this->val = val & ((reg_t(1) << PMP_ADDR_BITS) - 1);
   }
   else
     return false;
--- a/riscv/encoding.h
+++ b/riscv/encoding.h
@@ -358,7 +358,9 @@
 #define PMP_A     0x18
 #define PMP_L     0x80
 #define PMP_SHIFT 2
+#define PMP_ADDR_BITS  54
 
+#define PMP_OFF   0x00
 #define PMP_TOR   0x08
 #define PMP_NA4   0x10
 #define PMP_NAPOT 0x18


aswaterman:
It looks like this PR is making multiple changes.  I need to understand each change individually.  For each change, can you provide an example code sequence where Spike and the SAIL model differ, and explain what Spike does and what the SAIL model does, and why you think Spike is wrong?

riscv/encoding.h
```diff
@@ -358,7 +358,9 @@
 #define PMP_A     0x18
 #define PMP_L     0x80
 #define PMP_SHIFT 2
+#define PMP_ADDR_BITS  54
 
+#define PMP_OFF   0x00
```
aswaterman, line 363: `encoding.h` is an auto-generated file and so shouldn't be manually modified.  Instead of introducing `PMP_OFF` to write `mode == PMP_TOR || mode == PMP_OFF`, you could write `!(mode == PMP_NA4 || mode == NAPOT)`, for example.

riscv/csrs.cc
```diff
@@ -121,7 +131,7 @@ bool pmpaddr_csr_t::unlogged_write(const reg_t val) noexcept {
   const bool locked = !lock_bypass && (cfg & PMP_L);
 
   if (pmpidx < proc->n_pmp && !locked && !next_locked_and_tor()) {
-    this->val = val & ((reg_t(1) << (MAX_PADDR_BITS - PMP_SHIFT)) - 1);
+    this->val = val & ((reg_t(1) << PMP_ADDR_BITS) - 1);
```
aswaterman, line 134: I don't think this is right.  Nothing in the spec limits PMP addresses to 56 bits.  That might be a quirk of the SAIL model.

trdthg:
Here are more detailed explanations.

I found during testing that when writing all 1 to `PMPADDR0`, the value read by spike is `0x3fffffffffffffff` (62 bits), while the value read by sail is `0x3ffffffffffff` (54 bits)

The reason for the difference is that when sail writes to pmpaddr, it only retains 54 bits
```
function pmpWriteAddr(locked: bool, tor_locked: bool, reg: xlenbits, v: xlenbits) -> xlenbits =
  ...
  else { if (locked | tor_locked) then reg else zero_extend(v[53..0]) }
```

But spike retains 64 - 2 = 62 bits.

```
if (pmpidx < proc->n_pmp && !locked && !next_locked_and_tor()) {
    this->val = val & ((reg_t(1) << (MAX_PADDR_BITS - PMP_SHIFT)) - 1);
}
```

For pmpaddr, the specification (2025.7.24) describes that,

> For RV64, each PMP address register encodes bits 55-2 of a 56-bit physical address, as shown in [Figure 33](https://riscv.github.io/riscv-isa-manual/snapshot/privileged/#pmpaddr-rv64). Not all physical address bits may be implemented, and so the pmpaddr registers are WARL.

image

The bits [54, 63] are marked with `0(WARL)`, so I think sail is right and they should be set to 0.

---

Simply modifying the write part can already solve the problem I encountered.

This PR also modified the read part. Because the current spike's read_pmpaddr impl is almost completely different from the spec description if I understood correctly? I can split it into another PR.

trdthg:
What does the `0(WARL)` mean?

- 0 is just a meaningless placeholder, and although it's WARL, it's not marked as Reserved, meaning it can be read and written freely.
- 0 means that the Legal value is 0.

riscv/encoding.h
trdthg: 🆗

aswaterman:
Ah, you're right.  Spike supports the extra address bits as a nonstandard extension.  I'm a little worried that removing it will break some users' use cases.  So maybe a compromise is to automatically size the PMP registers based upon the memory map.

Note that the result will not be that Spike exactly matches SAIL, but for a different (valid) reason: it is legal to support fewer than 56 address bits in PMP, since the address field is WARL.

aswaterman:
Hmm, maybe we should just do the simple, correct thing for now, and if any users complain about removal of support for nonstandard >56b PMPs, we can cross that bridge then.

riscv/encoding.h
aswaterman: I'm working on a new PR that's inspired by yours, so no need to make the code changes.  Once I'm finished, I'll ask you to take a look at my PR and make sure it works for you.

PR #2037, opened 2025-07-29
Only support 56-bit PMP and ATP addresses

aswaterman:
As mandated by the ISA spec.


aswaterman:
@trdthg

trdthg:
Make sense to me.

Issue #2035, opened 2025-07-25
Can't use system installed spike libraries to create loadable extensions


youngar:
I want to be able to compile a extension using a system installed spike, but it is currently not working and gives an error due to doubly registering devices.  It seems like it is more common in this group for people to drop new extensions in the customext directory, and have spike build it directly, which does work currently but a bit strange (details below).  The steps to reproduce my issue are below.

Install spike using:
```
cd build
CC=clang CXX=clang++ ../configure --prefix=$PWD/../install
make -j install
```
Build extension with these commands (run from the root)
```
clang++ -std=c++2a -fPIC -O2 -Wall -MMD -MP -D_GNU_SOURCE -Iinstall/include/riscv -Iinstall/include/fesvr -Iinstall/include/disasm -Iinstall/include/softfloat -c customext/xrockettest.cc -o xrockettest.o
clang++ -shared -o librockettest.so xrockettest.o -Linstall/lib -lriscv -ldisasm -lsoftfloat
```

This gives us a library linked against the `.so`s:
```
LD_LIBRARY_PATH=install/lib ldd ./librockettest.so 
        linux-vdso.so.1 (0x00007ffc2b1f6000)
        libriscv.so => install/lib/libriscv.so (0x00007f6426d1d000)
        libsoftfloat.so => install/lib/libsoftfloat.so (0x00007f6426b06000)
        libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f6426771000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f64263ef000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f64261d7000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f6425e12000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f6425bf2000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f64259ee000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f64279e4000)
```

And finally, trying to run with our plugin gives me the error:
```
LD_LIBRARY_PATH=.:install/lib ./install/bin/spike --isa=RV64IMAFDCV_xrockettest $PWD/../riscv-pk/build/pk hello
terminate called after throwing an instance of 'std::runtime_error'
  what():  Plugin "clint" already registered
```
[...]

PR #2039, opened 2025-07-30
Dynamically link all libraries in spike

youngar:
This PR is not currently complete, but I was hoping to start a discussion of #2035.

This changes spike to dynamically link all libraries. The main benefit of this change is that we can build extension libraries using the system installed spike by linking them against `libriscv.so`.  Without this change, we are [doubly running static initializers/constructors](https://stackoverflow.com/questions/45335499/c-link-library-twice-are-global-constructors-run-twice) when we do this.

The main negatives of this change are:
1. We now need to install more libraries
2. If you want to run spike directly out of the `build` directory instead of the `install` directory, you will have to set `LD_LIBRARY_PATH=<dir>` to the `build` directory

Some ideas follow up work:
- We don't *have* to dynamically link all libraries, its just the build system isn't set up to make that easy right now. This way we don't have to install everything. We will probably have to tread very carefully to avoid triggering the same issue again.
- We can remove the `-Wl,--export-dynamic` flag which exports symbols from the executable
- We can remove the [hack](https://github.com/riscv-software-src/riscv-isa-sim/blob/master/spike_main/spike_main.mk.in#L18) to ensure that certain symbols make it into the final executable 
- If setting `LD_LIBRARY_PATH` is too onerous, we can probably set the `rpath` to the build directory and rewrite when installing.  Things like `libtool` and `cmake` do handle this automagically.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -237,7 +237,7 @@ $(2)_reverse_deps   := $$(call reverse_list,$$($(2)_subproject_deps))
 # Build a library for this subproject
 
 $(2)_lib_libs       := $$($(2)_reverse_deps)
-$(2)_lib_libnames   := $$(patsubst %, lib%.a, $$($(2)_lib_libs))
+$(2)_lib_libnames   := $$(patsubst %, lib%.so, $$($(2)_lib_libs))
 $(2)_lib_libarg     := $$(patsubst %, -l%, $$($(2)_lib_libs))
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
We used to do this and consciously moved away from it because it caused lots of confusion, and also added to the difficulty in building in one place and moving binaries to another.

Building the dynamic versions of these libraries is fine if we find it useful, but the `spike` executable should not depend on them.

PR #2040, opened 2025-07-30
Sireg is not in the sireg_csrs[] list, so needs to be checked outside… #2041

mmhus:
… of the switch-case.

Sireg proxy is skipped since SIREG_CSR is instantiated outside of the switch-case statement. Needs to be fixed.

issue is https://github.com/riscv-software-src/riscv-isa-sim/issues/2041

### DIFF
--- a/riscv/csr_init.cc
+++ b/riscv/csr_init.cc
@@ -430,6 +430,17 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
     auto sireg = std::make_shared<sscsrind_reg_csr_t>(proc, CSR_SIREG, siselect);
     add_ireg_proxy(proc, sireg);
     add_supervisor_csr(CSR_SIREG, std::make_shared<virtualized_indirect_csr_t>(proc, sireg, vsireg));
+    if (proc->extension_enabled(EXT_SSCCFG) || proc->extension_enabled(EXT_SMCDELEG)) {
+      // case CSR_SIREG
+      if (proc->extension_enabled_const(EXT_ZICNTR)) {
+        sireg->add_ireg_proxy(SISELECT_SMCDELEG_START, mcycle);
+        sireg->add_ireg_proxy(SISELECT_SMCDELEG_INSTRET, minstret);
+      }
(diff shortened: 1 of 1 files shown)


riscv/csr_init.cc
```diff
@@ -430,6 +430,14 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
...
     add_ireg_proxy(proc, sireg);
     add_supervisor_csr(CSR_SIREG, std::make_shared<virtualized_indirect_csr_t>(proc, sireg, vsireg));
+    if (proc->extension_enabled_const(EXT_ZICNTR)) {
+      sireg->add_ireg_proxy(SISELECT_SMCDELEG_START, mcycle);
+      sireg->add_ireg_proxy(SISELECT_SMCDELEG_INSTRET, minstret);
+    }
+    if (proc->extension_enabled_const(EXT_ZIHPM)) {
+      for (size_t j = 0; j < (SISELECT_SMCDELEG_END - SISELECT_SMCDELEG_HPMEVENT_3 + 1); j++)
+        sireg->add_ireg_proxy(SISELECT_SMCDELEG_HPMCOUNTER_3 + j, csrmap[CSR_HPMCOUNTER3 + j]);
+    }
```
aswaterman: Shouldn't this code have been moved only two lines up, so that it is still conditioned on `if (proc->extension_enabled(EXT_SSCCFG) || proc->extension_enabled(EXT_SMCDELEG))`?

mmhus: Hi, while you are correct that this should be gated by:
```c
    if (proc->extension_enabled(EXT_SSCCFG) || proc->extension_enabled(EXT_SMCDELEG)) {
```
and I've added a check for this just now, the code cannot be moved just 2 lines up because the scope for `auto sireg` is inside the `for` loop in that suggesstion, which would be incorrect. There are two `auto sireg` variables with different scopes, one refers to CSR_SIREG, and the other refers to CSR_SIREG2, CSR_SIREG3.... and so on.

aswaterman: Makes sense.  Thanks for fixing it.

aswaterman (approved):

PR #2042, opened 2025-07-30
Don't include built-in devices in mmio_device_map

### LINKED ISSUE #2035: Can't use system installed spike libraries to create loadable extensions

aswaterman:
Only use mmio_device_map for plugin devices.

This fixes a collision caused by multiple static initializers if Spike depends on a library that depends on libriscv.so.

Fixes #2035

cc @youngar

### DIFF
--- a/riscv/abstract_device.h
+++ b/riscv/abstract_device.h
@@ -46,4 +46,11 @@ mmio_device_map_t& mmio_device_map();
   std::string generate_dts(const sim_t* sim, const std::vector<std::string>& sargs) const override { return generate(sim, sargs); } \
   }; device_factory_t *name##_factory = new name##_factory_t();
 
+#define REGISTER_BUILTIN_DEVICE(name, parse, generate) \
+  class name##_factory_t : public device_factory_t { \
+  public: \
+  name##_t* parse_from_fdt(const void* fdt, const sim_t* sim, reg_t* base, const std::vector<std::string>& sargs) const override { return parse(fdt, sim, base, sargs); } \
+  std::string generate_dts(const sim_t* sim, const std::vector<std::string>& sargs) const override { return generate(sim, sargs); } \
+  }; device_factory_t *name##_factory = new name##_factory_t();
+
 #endif
--- a/riscv/clint.cc
+++ b/riscv/clint.cc
@@ -145,4 +145,4 @@ std::string clint_generate_dts(const sim_t* sim, const std::vector<std::string>&
   return s.str();
 }
 
-REGISTER_DEVICE(clint, clint_parse_from_fdt, clint_generate_dts)
+REGISTER_BUILTIN_DEVICE(clint, clint_parse_from_fdt, clint_generate_dts)
--- a/riscv/ns16550.cc
+++ b/riscv/ns16550.cc
@@ -361,4 +361,4 @@ ns16550_t* ns16550_parse_from_fdt(const void* fdt, const sim_t* sim, reg_t* base
   }
 }
 
-REGISTER_DEVICE(ns16550, ns16550_parse_from_fdt, ns16550_generate_dts)
+REGISTER_BUILTIN_DEVICE(ns16550, ns16550_parse_from_fdt, ns16550_generate_dts)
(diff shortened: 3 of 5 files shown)


aswaterman:
@jerryz123 Since you authored the current version of this functionality, you should sanity-check it.

youngar:
Thanks! I can confirm that this fixes my issue!

jerryz123 (approved):
I do vaguely recall this was an issue, I'm glad the fix is this simple

aswaterman:
cool!

PR #2043, opened 2025-07-30
For writes + logging, if len > 8 bytes, break up the memory log

### LINKED ISSUE #2038: amocas.q with logging enabled aborts when the condition is true
A simple oversight.  The offending code:

https://github.com/riscv-software-src/riscv-isa-sim/blob/483e01d486dbace0d7ead3087acb062158682e90/riscv/mmu.cc#L372-L373

When `amocas.q` attempts store on a successful comparison, it tries to store 128 bits/16 bytes which `reg_from_bytes` does not support and cannot return a data structure that large.

To avoid introducing a pervasive fix for a potential corner case, thinking about fixing just this call to break up logging into 2 calls.  I have a test case to try out...

klingaard:
Before the fix, `reg_from_bytes` would abort on AMOCAS.Q if the 16 byte write were to occur.

`reg_from_bytes`  only supports up to 8 bytes due to `reg_t` size.

The solution is to break up the 16-byte log entries into two 8-byte log entries.  However, my solution might support
sizes bigger than 16 :grin:.

After the fix, the logger looks like:
```
core   0: 0x0000000080010fa0 (0x2802c92f) amocas.q s2, zero, (t0)
core   0: 3 0x0000000080010fa0 (0x2802c92f) x18 0x11223344dead5678 x19 0x11113333dead5678 mem 0x0000000080012100 mem 0x0000000080012100 0x0000000000000000 mem 0x0000000080012108 0x0000000000000000
```
where `mem 0x0000000080012100 0x0000000000000000 mem 0x0000000080012108 0x0000000000000000` 
represent the full 128-bit write.

Ran `valgrind` on this and it appears happy.

closes #2038

### DIFF
--- a/riscv/abstract_device.h
+++ b/riscv/abstract_device.h
@@ -46,4 +46,11 @@ mmio_device_map_t& mmio_device_map();
   std::string generate_dts(const sim_t* sim, const std::vector<std::string>& sargs) const override { return generate(sim, sargs); } \
   }; device_factory_t *name##_factory = new name##_factory_t();
 
+#define REGISTER_BUILTIN_DEVICE(name, parse, generate) \
+  class name##_factory_t : public device_factory_t { \
+  public: \
+  name##_t* parse_from_fdt(const void* fdt, const sim_t* sim, reg_t* base, const std::vector<std::string>& sargs) const override { return parse(fdt, sim, base, sargs); } \
+  std::string generate_dts(const sim_t* sim, const std::vector<std::string>& sargs) const override { return generate(sim, sargs); } \
+  }; device_factory_t *name##_factory = new name##_factory_t();
+
 #endif
--- a/riscv/clint.cc
+++ b/riscv/clint.cc
@@ -145,4 +145,4 @@ std::string clint_generate_dts(const sim_t* sim, const std::vector<std::string>&
   return s.str();
 }
 
-REGISTER_DEVICE(clint, clint_parse_from_fdt, clint_generate_dts)
+REGISTER_BUILTIN_DEVICE(clint, clint_parse_from_fdt, clint_generate_dts)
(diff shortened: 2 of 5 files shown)


riscv/mmu.cc
```diff
@@ -369,8 +369,17 @@ void mmu_t::store_slow_path(reg_t original_addr, reg_t len, const uint8_t* bytes
...
   }
 
-  if (actually_store && proc && unlikely(proc->get_log_commits_enabled()))
-    proc->state.log_mem_write.push_back(std::make_tuple(original_addr, reg_from_bytes(len, bytes), len));
+  if (actually_store && proc && unlikely(proc->get_log_commits_enabled())) {
+    // #2038 amocas.q sends len == 16, reg_from_bytes only supports up to 8
+    // bytes per conversion.  Make multiple entries in the log
+    reg_t offset = 0;
+    while (unlikely(len > 8)) {
+      proc->state.log_mem_write.push_back(std::make_tuple(original_addr, reg_from_bytes(8, bytes), 8));
```
klingaard: Hmmm... missing offset

aswaterman (commented):
I'm glad this turned out to be straightforward.  Can you make some minor changes for me:
- replace all of the `8` with `sizeof(reg_t)` for better self-documentation.
- get rid of the `#2038`; the house style in Spike is not to include this kind of thing in comments, but instead have people look at the git history
- squash all of the commits into a single one

And then I'll gladly merge this.  Thanks, @klingaard!

klingaard:
>     * replace all of the `8` with `sizeof(reg_t)` for better self-documentation.

Done!

>     * get rid of the `#2038`; the house style in Spike is not to include this kind of thing in comments, but instead have people look at the git history

Gotcha!  Makes sense.  Done!

>     * squash all of the commits into a single one

Done!  I'm used to PR merge squishes, so didn't realize it was different here.

Should be good to go now.  Thanks for accepting!

klingaard:
Ugh, I screwed something up on my rebase.

PR #2044, opened 2025-07-31
For writes + logging, if len > 8 bytes, break up the memory log 

### LINKED ISSUE #2038: amocas.q with logging enabled aborts when the condition is true

klingaard:
Before the fix, `reg_from_bytes` would abort on AMOCAS.Q if the 16 byte write were to occur.

`reg_from_bytes` only supports up to 8 bytes due to reg_t size.

The solution is to break up the 16-byte log entries into two 8-byte log entries. However, my solution might support
sizes bigger than 16 😁.

After the fix, the logger looks like:
```
core   0: 0x0000000080010fa0 (0x2802c92f) amocas.q s2, zero, (t0)
core   0: 3 0x0000000080010fa0 (0x2802c92f) x18 0x11223344dead5678 x19 0x11113333dead5678 mem 0x0000000080012100 mem 0x0000000080012100 0x0000000000000000 mem 0x0000000080012108 0x0000000000000000
```
where `mem 0x0000000080012100 0x0000000000000000 mem 0x0000000080012108 0x0000000000000000`
represent the full 128-bit write.

Ran valgrind on this and it appears happy.

closes https://github.com/riscv-software-src/riscv-isa-sim/issues/2038

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -369,8 +369,18 @@ void mmu_t::store_slow_path(reg_t original_addr, reg_t len, const uint8_t* bytes
     store_slow_path_intrapage(len, bytes, access_info, actually_store);
   }
 
-  if (actually_store && proc && unlikely(proc->get_log_commits_enabled()))
-    proc->state.log_mem_write.push_back(std::make_tuple(original_addr, reg_from_bytes(len, bytes), len));
+  if (actually_store && proc && unlikely(proc->get_log_commits_enabled())) {
+    // amocas.q sends len == 16, reg_from_bytes only supports up to 8
+    // bytes per conversion.  Make multiple entries in the log
+    reg_t offset = 0;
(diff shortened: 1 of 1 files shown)


klingaard:
original PR: https://github.com/riscv-software-src/riscv-isa-sim/pull/2043 before I completely hosed my local fork.  :angry:

aswaterman (approved):
Thanks again!

PR #2048, opened 2025-08-03
ns16550.cc : fix handling clear rx/tx FIFO requests.

hdinc:


aswaterman:
@avpatel as the author of the original code, can you please review this PR?

avpatel:
Looks good to me. I also tried with latest Linux and OpenSBI.

aswaterman (approved):
Thanks!

PR #2054, opened 2025-08-07
Add VLEN >= ELEN validation check

nadime15:
Prevents invalid configurations where VLEN < ELEN, such as combining zvl32b with zve64d.

### DIFF
--- a/disasm/isa_parser.cc
+++ b/disasm/isa_parser.cc
@@ -524,7 +524,7 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
     bad_isa_string(str, "Spike does not currently support VLEN > 4096b");
   }
 
-  if ((vlen != 0) ^ (elen != 0)) {
+  if ((vlen != 0) ^ (elen != 0) || vlen < elen) {
     bad_isa_string(str, "Invalid Zvl/Zve configuration");
   }
 


nadime15:
I encountered the following situation when running:

`spike -d isa=rv64gcv_zvbb_zvbc_zfh_zvfh_zvkg_zvkned_zvknha_zvksed_zvksh_zvfbfmin_zvfbfwma_zvl64b_zve64d_zfh_zfhmin_zvfh
`

In debug mode, I noticed that VLEN was set to 128, even though I explicitly included Zvl64b. This happens because of the following line in the code:

`vlen = std::max(vlen, new_vlen);`

This effectively preserves the default value of 128.

I think it would be better to avoid implicitly setting VLEN to 128 in such cases. Otherwise, users intending to verify a vector unit with VLEN=64 (non VLEN-agnostic code) might unintentionally run tests compiled for VLEN=128, leading to incorrect assumptions or results (Tests that should actually fail pass).

aswaterman (changes requested):
By definition, V includes Zvl128b and Zve64d; see https://github.com/riscv/riscv-isa-manual/blob/518c6ebda95b4a87181e089bfbc5fe9bfdab08c8/src/v-st-ext.adoc#v-vector-extension-for-application-processors

If you want to select VLEN < 128, you need to remove V from the ISA string and use Zve-something instead.  e.g. `--isa=rv64gc_zvl64b_zve32f`

However, I believe your `vlen < elen` change is correct.  If you force-push over the PR with only that change, I'll approve and merge.

nadime15:
Ahh yeah good point, I missed that, thanks!

I reverted my changes except for the `vlen < elen` check.

This triggers an error for invalid ISA combinations, such as:

```
spike -d --isa=rv64gc_zvbb_zvbc_zfh_zvfh_zvkg_zvkned_zvknha_zvksed_zvksh_zvl32b_zve64d  out/v64x64machine/bin/stage2/vsse16_v-0
error: bad --isa option 'rv64gc_zvbb_zvbc_zfh_zvfh_zvkg_zvkned_zvknha_zvksed_zvksh_zvl32b_zve64d'. Invalid Zvl/Zve configuration
```

aswaterman (approved):
Thank you!

PR #2058, opened 2025-08-19
Fix BFloat16 vector extension dependencies and requirements

nadime15:
Update requirements for the Bfloat16 extensions including:

* Enable EXT_INTERNAL_ZFH_MOVE when either Zfbfmin or Zfhmin is enabled
* Zvfbfmin now requires Zve32f instead of the V extension.
* Zvfbfwma now requires both Zvfbfmin and Zfbfmin, not just the V extension.

### DIFF
--- a/disasm/isa_parser.cc
+++ b/disasm/isa_parser.cc
@@ -460,11 +460,11 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
     extension_table[EXT_INTERNAL_ZFH_MOVE] = true;
   }
 
-  if (extension_table[EXT_ZFBFMIN] && (!extension_table['F'] || !extension_table[EXT_INTERNAL_ZFH_MOVE])) {
+  if (extension_table[EXT_ZFBFMIN] && (!extension_table['F'])) {
     bad_isa_string(str, "'Zfbfmin' extension requires 'F' extension");
   }
 
-  if (extension_table[EXT_ZVFBFMIN] && (vlen == 0 || elen == 0 || !zvf)) {
+  if (extension_table[EXT_ZVFBFMIN] && (vlen == 0 || !zvf)) {
     bad_isa_string(str, "'Zvfbfmin' extension requires 'Zve32f' extension");
   }
 
--- a/disasm/isa_parser.cc
+++ b/disasm/isa_parser.cc
@@ -456,16 +456,20 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
     bad_isa_string(str, "'Zclsd' extension requires 'Zca' and 'Zilsd' extensions");
   }
 
-  if (extension_table[EXT_ZFBFMIN] && !extension_table['F']) {
+  if (extension_table[EXT_ZFBFMIN] || extension_table[EXT_ZFHMIN]) {
+    extension_table[EXT_INTERNAL_ZFH_MOVE] = true;
+  }
+
+  if (extension_table[EXT_ZFBFMIN] && (!extension_table['F'] || !extension_table[EXT_INTERNAL_ZFH_MOVE])) {
     bad_isa_string(str, "'Zfbfmin' extension requires 'F' extension");
   }
 
-  if ((extension_table[EXT_ZVFBFMIN] || extension_table[EXT_ZVFBFWMA]) && !extension_table['V']) {
-    bad_isa_string(str, "'Zvfbfmin/Zvfbfwma' extension requires 'V' extension");
+  if (extension_table[EXT_ZVFBFMIN] && (vlen == 0 || elen == 0 || !zvf)) {
+    bad_isa_string(str, "'Zvfbfmin' extension requires 'Zve32f' extension");
   }
 
-  if (extension_table[EXT_ZFBFMIN] || extension_table[EXT_ZVFBFMIN] || extension_table[EXT_ZFHMIN]) {
-    extension_table[EXT_INTERNAL_ZFH_MOVE] = true;
+  if (extension_table[EXT_ZVFBFWMA] && (!extension_table[EXT_ZFBFMIN] || !extension_table[EXT_ZVFBFMIN])) {
+    bad_isa_string(str, "'Zvfbfwma' extension requires 'Zfbfmin' and 'Zvfbfmin' extensions");
   }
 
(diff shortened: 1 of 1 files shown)


nadime15:
I might be missing something, but I dont think there is really a need for `EXT_INTERNAL_ZFH_MOVE`. If asked, I can remove it and move these instructions to Zfh, as defined in the spec. Just let me know!

aswaterman (commented):
> Zfbfmin now requires both F and Zfh extensions.

This isn't what the ISA spec says, though.  It says that Zfbfmin requires F, and that Zfbfmin also includes the loads/stores/moves defined in the Zfh extension.  It doesn't say that Zfbfmin requires Zfh.  This is the very reason that we defined EXT_INTERNAL_ZFH_MOVE.

> Zvfbfmin now requires Zve32f instead of the V extension.

Ack.

> Zvfbfwma now requires both Zvfbfmin and Zfbfmin, not just the V extension.

Sounds right.

Can you update the PR to remove the first change?  (You might want to start from scratch since the premise of the patch is based on messing with EXT_INTERNAL_ZFH_MOVE :))

nadime15:
> This isn't what the ISA spec says, though. It says that Zfbfmin requires F, and that Zfbfmin also includes the loads/stores/moves defined in the Zfh extension. It doesn't say that Zfbfmin requires Zfh. This is the very reason that we defined EXT_INTERNAL_ZFH_MOVE.

Hmm I think the spec is ambiguous here.

> "This extension depends upon the single-precision floating-point extension F, and the FLH, FSH, FMV.X.H,
> and FMV.H.X instructions as defined in the Zfh extension."

I might misinterpreted the spec, but my reading is that `Zfbfmin` depends on `F` **and** on the FLH, FSH, etc. instructions (which are part of `Zfh`). Since we cant check individual instructions, we check for Zfh.

(We depend on these but we dont define them)

But the current code does the opposite, it assumes `Zfbfmin` provides these instructions by setting `EXT_INTERNAL_ZFH_MOVE = true`. The spec does not say Zfbfmin defines these instructions, it says Zfbfmin depends on them 'as defined in the Zfh extension.' If Zfbfmin depends on instructions defined elsewhere, how can it provide them itself? 

It might be clearer if we rephrased it as something like: 

> "This extension depends upon the single-precision floating-point extension F, and **defines** the FLH, FSH, FMV.X.H,
> and FMV.H.X instructions as defined in the Zfh extension if Zfh is not supported."

Because the above is what the code actually does.

> Can you update the PR to remove the first change? (You might want to start from scratch since the premise of the patch is based on messing with EXT_INTERNAL_ZFH_MOVE :))

Sure will do, thanks!

gfavor:
I would generally agree with Andrew.  Checking for Zfh is incorrect (despite there not being an extension name for the subset of Zfh that Zfbfmin requires).  And since these instructions are already defined in a ratified spec, Zfbfmin should not be providing its own definitions.  And yes, Zfbfmin can depend on this subset of instructions "as defined in Zfh".

As far as the suggested "clearer" phrasing, it undesirably or ambiguously leaves the question of what happens if Zfh IS supported.  The reality being that Zfbfmin depends on these instructons irrespective of whether Zfh is supported or not.

Lastly, if the code doesn't match up with the arch spec, then it is the code that needs to be changed.

aswaterman:
I agree with Greg that the "as defined in" phrasing is unambiguous: it specifies where the definition is located, and does not indicate an extension-dependence relationship.  However, I agree with Nadime that it is not helpful to readers to describe extension-dependence relationships and instruction-inclusion relationships in the same breath.  I've made a PR to improve the wording, which I'll shortly merge: https://github.com/riscv/riscv-isa-manual/pull/2220

BTW, I looked back through the mailing-list archives and git history to confirm that my/Greg's interpretation is what was intended at the time.

nadime15:
Thank you all for your feedback! Yes, I think Andrew's change makes it clearer!

I pushed some changes and edited the original PR message. Let me know if anything else needs to be changed. :)

disasm/isa_parser.cc
```diff
@@ -456,16 +456,20 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
     bad_isa_string(str, "'Zclsd' extension requires 'Zca' and 'Zilsd' extensions");
   }
 
-  if (extension_table[EXT_ZFBFMIN] && !extension_table['F']) {
+  if (extension_table[EXT_ZFBFMIN] || extension_table[EXT_ZFHMIN]) {
+    extension_table[EXT_INTERNAL_ZFH_MOVE] = true;
+  }
+
+  if (extension_table[EXT_ZFBFMIN] && (!extension_table['F'] || !extension_table[EXT_INTERNAL_ZFH_MOVE])) {
```
aswaterman: @nadime15 Sorry for the delay in responding.  One more thing.  I think you can change this to `if (extension_table[EXT_ZFBFMIN] && !extension_table['F']) {` -- the ZFH_MOVE check is redundant because of the change you made above, which forces ZFH_MOVE to true when ZFBFMIN is enabled.

disasm/isa_parser.cc
```diff
@@ -456,16 +456,20 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
...
     bad_isa_string(str, "'Zfbfmin' extension requires 'F' extension");
   }
 
-  if ((extension_table[EXT_ZVFBFMIN] || extension_table[EXT_ZVFBFWMA]) && !extension_table['V']) {
-    bad_isa_string(str, "'Zvfbfmin/Zvfbfwma' extension requires 'V' extension");
+  if (extension_table[EXT_ZVFBFMIN] && (vlen == 0 || elen == 0 || !zvf)) {
```
aswaterman: Remove the `elen` check here; it is impossible for `elen` to be 0 when `vlen` is nonzero.  (If you want a paranoid check here, then include above an `assert(!elen || vlen);`.  This case should be treated as an internal error, so an `assert` is more appropriate.)

aswaterman (commented):
OK with  me once my two remaining comments are addressed.  Please ping me again afterwards.  Thanks, @nadime15!

nadime15:
@aswaterman Done!

aswaterman (approved):
Thanks again!

PR #2059, opened 2025-08-21
Remove sign extension from hlvx.wu instruction.

i2h2:
hlvx.wu is not supposed to sign extend the value.

### DIFF
--- a/riscv/insns/hlvx_wu.h
+++ b/riscv/insns/hlvx_wu.h
@@ -1,4 +1,4 @@
 require_extension('H');
 require_novirt();
 require_privilege(get_field(STATE.hstatus->read(), HSTATUS_HU) ? PRV_U : PRV_S);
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
No, this is correct as-is.

`sext_xlen` is a no-op for RV64, so the instruction does what you want for RV64: the value is zero-extended from 32 bits to 64 bits, then that 64-bit value is sign-extended to 64 bits (i.e. a no-op).

For RV64, we _need_ the sign extension because of a Spike implementation detail.  Spike always has 64-bit registers.  Some instructions, e.g. branches, always operate on all 64 bits, even in RV32 mode.  For this to work correctly, we need to sign-extend all 32-bit values to 64 bits, even if they aren't notionally signed.  This is the same reason that e.g. `addi` uses `sext_xlen`.

i2h2:
Thanks!

PR #2069, opened 2025-09-04
Relax VLEN/ELEN checking

### LINKED ISSUE #2068: [isa parser] checking vector
If we in ISA string have vector extension,  do we really need to have `v` at the beginning? 
E.g. with `rv64imafdch_zfbfmin_zve32f` I get 

```
Zvfbfmin/Zvfbfwma' extension requires 'V' extension
```

rv64imafdc***v***h_zfbfmin_zve32f is correct

aswaterman:
We should allow ISA strings like rv64gc_zve32f.  Per the spec, the various Zve extensions imply a minimum VLEN, so rv64gc_zve32f is unambiguously equivalent to rv64gc_zve32f_zvl32b. Similarly, rv64gc_zve64x, rv64gc_zve64x_zvl64b, and rv64gc_zve64x_zvl32b are all unambiguously equivalent.

@nadime15 This is different from what we discussed, but it is supported by the spec and is obviously easier on users.

Fixes #2068

### DIFF
--- a/disasm/isa_parser.cc
+++ b/disasm/isa_parser.cc
@@ -383,7 +383,7 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
       vlen = std::max(vlen, new_vlen);
     } else if (ext_str.substr(0, 3) == "zve") {
       if (ext_str.size() != 6) {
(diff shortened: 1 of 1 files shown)


nadime15 (approved):
Yeah, this makes sense! Looks good to me.

PR #2086, opened 2025-09-22
Implement abstract memory read/write command

fkhaidari:
This commit implements the abstract memory read/write commands in the debug module. The changes include:

- Added support for abstract memory access commands (cmdtype 2) in perform_abstract_command()
- Implemented perform_abstract_memory_access() method to handle memory operations
- Moved abstract register access code to perform_abstract_register_access() method. Code is NOT changed
- Extended opcodes.h with csrrci and csrrsi instructions for memory access control

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -906,13 +906,13 @@ bool debug_module_t::perform_abstract_memory_access() {
     return true;
   }
 
-  if (aamvirtual || aamsize > 3 || aamsize > idx(xlen)) {
+  if (aamsize > idx(xlen)) {
     abstractcs.cmderr = CMDERR_NOTSUP;
     return true;
   }
 
   unsigned offset = 0;
-  generate_initial_sequence(offset);
+  generate_initial_sequence(aamvirtual, offset);
   is_write ? handle_memory_write(xlen, aamsize, offset)
            : handle_memory_read(xlen, aamsize, offset);
 
@@ -927,8 +927,10 @@ bool debug_module_t::perform_abstract_memory_access() {
 }
 
 using handle_memory_func = uint32_t (*)(unsigned rd_src, unsigned base, uint16_t offset);
+using handle_mstatus_func = uint32_t(*)(unsigned rd, unsigned rs1, unsigned csr);
 static constexpr std::array<handle_memory_func, 4> lx = {&lb, &lh, &lw, &ld};
 static constexpr std::array<handle_memory_func, 4> sx = {&sb, &sh, &sw, &sd};
+static constexpr std::array<handle_mstatus_func, 2> csrrx = {&csrrc, &csrrs};
 
 unsigned debug_module_t::arg(unsigned xlen, unsigned idx)
 {
@@ -937,16 +939,21 @@ unsigned debug_module_t::arg(unsigned xlen, unsigned idx)
 
 void debug_module_t::handle_memory_read(size_t xlen, unsigned aamsize, unsigned &offset)
 {
-  write32(debug_abstract, offset++, lx[idx(xlen)](S0, ZERO, arg(xlen, 1)));
-  write32(debug_abstract, offset++, lx[aamsize](S0, S0, 0));
-  write32(debug_abstract, offset++, sx[idx(xlen)](S0, ZERO, arg(xlen, 0)));
+  write32(debug_abstract, offset++, lx[idx(xlen)](S1, ZERO, arg(xlen, 1)));
+  write32(debug_abstract, offset++, lx[aamsize](S1, S1, 0));
+  write32(debug_abstract, offset++, sx[idx(xlen)](S1, ZERO, arg(xlen, 0)));
 }
 
 void debug_module_t::handle_memory_write(size_t xlen, unsigned aamsize, unsigned &offset)
 {
-  write32(debug_abstract, offset++, lx[idx(xlen)](S0, ZERO, arg(xlen, 0)));
-  write32(debug_abstract, offset++, lx[idx(xlen)](S1, ZERO, arg(xlen, 1)));
+  // Use Arg1 as temporary storage for old mstatus value
+  write32(debug_abstract, offset++, lx[idx(xlen)](S1, ZERO, arg(xlen, 1))); // Arg1 -> S1
+  write32(debug_abstract, offset++, sx[idx(xlen)](S0, ZERO, arg(xlen, 1))); // S0 -> Arg1
+  write32(debug_abstract, offset++, lx[idx(xlen)](S0, ZERO, arg(xlen, 0))); // Arg0 -> S0
+
   write32(debug_abstract, offset++, sx[aamsize](S0, S1, 0));
+
+  write32(debug_abstract, offset++, lx[idx(xlen)](S0, ZERO, arg(xlen, 1))); // Restore S0
 }
 
 void debug_module_t::handle_post_increment(size_t xlen, unsigned aamsize, unsigned &offset)
@@ -956,14 +963,21 @@ void debug_module_t::handle_post_increment(size_t xlen, unsigned aamsize, unsign
   write32(debug_abstract, offset++, sx[idx(xlen)](S1, ZERO, arg(xlen, 1)));
 }
 
-void debug_module_t::generate_initial_sequence(unsigned &offset)
+void debug_module_t::generate_initial_sequence(bool aamvirtual, unsigned &offset)
 {
   write32(debug_abstract, offset++, csrw(S0, CSR_DSCRATCH0));
   write32(debug_abstract, offset++, csrw(S1, CSR_DSCRATCH1));
+
+  // Modify mstatus.mprv and save old mstatus
+  write32(debug_abstract, offset++, lui(S0, MSTATUS_MPRV >> 12));
+  write32(debug_abstract, offset++, csrrx[aamvirtual](S0, S0, CSR_MSTATUS));
 }
 
 void debug_module_t::generate_termination_sequence(unsigned &offset)
 {
+  // Restore mstatus
+  write32(debug_abstract, offset++, csrw(S0, CSR_MSTATUS));
+
   write32(debug_abstract, offset++, csrr(S0, CSR_DSCRATCH0));
   write32(debug_abstract, offset++, csrr(S1, CSR_DSCRATCH1));
   write32(debug_abstract, offset++, ebreak());
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -697,11 +697,26 @@ bool debug_module_t::perform_abstract_command()
     return true;
   }
 
-  if ((command >> 24) == 0) {
-    // register access
-    unsigned size = get_field(command, AC_ACCESS_REGISTER_AARSIZE);
-    bool write = get_field(command, AC_ACCESS_REGISTER_WRITE);
-    unsigned regno = get_field(command, AC_ACCESS_REGISTER_REGNO);
+  auto cmdtype = get_field(command, DM_COMMAND_CMDTYPE);
+  constexpr decltype(cmdtype) CMDTYPE_ACCESS_REGISTER = 0ULL;
+  constexpr decltype(cmdtype) CMDTYPE_ACCESS_MEMORY = 2ULL;
+
+  if (cmdtype == CMDTYPE_ACCESS_REGISTER)
+    return perform_abstract_register_access();
+
+  if (cmdtype == CMDTYPE_ACCESS_MEMORY)
+    return perform_abstract_memory_access();
+
+  abstractcs.cmderr = CMDERR_NOTSUP;
+  return true;
+}
+
+bool debug_module_t::perform_abstract_register_access()
+{
+  // register access
+  unsigned size = get_field(command, AC_ACCESS_REGISTER_AARSIZE);
+  bool write = get_field(command, AC_ACCESS_REGISTER_WRITE);
+  unsigned regno = get_field(command, AC_ACCESS_REGISTER_REGNO);
 
     if (!selected_hart_state().halted) {
       abstractcs.cmderr = CMDERR_HALTRESUME;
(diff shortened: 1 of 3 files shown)


fkhaidari:
Looks scary, but it is much better if you push `Hide whitespace` button.
image

fkhaidari:
@aswaterman, would you take a look?

aswaterman:
> Looks scary, but it is much better if you push Hide whitespace button.

You are right on both counts :)

riscv/debug_module.cc
```diff
@@ -649,178 +653,281 @@ bool debug_module_t::perform_abstract_command()
...
-    bool write = get_field(command, AC_ACCESS_REGISTER_WRITE);
-    unsigned regno = get_field(command, AC_ACCESS_REGISTER_REGNO);
+  auto cmdtype = get_field(command, DM_COMMAND_CMDTYPE);
 
-    if (!selected_hart_state().halted) {
-      abstractcs.cmderr = CMDERR_HALTRESUME;
-      return true;
-    }
+  if (cmdtype == DM_COMMAND_CMDTYPE_ACCESS_REGISTER)
+    return perform_abstract_register_access();
```
aswaterman, line 705: This is cleaner than before; thanks.

riscv/debug_module.cc
```diff
@@ -649,178 +653,281 @@ bool debug_module_t::perform_abstract_command()
...
+    handle_post_increment(xlen, aamsize, offset);
+
+  generate_termination_sequence(offset);
+  start_command_execution();
+
+  abstractcs.cmderr = !handled ? CMDERR_NOTSUP : CMDERR_NONE;
+  return true;
+}
+
+using func = uint32_t (*)(unsigned rd_src, unsigned base, uint16_t offset);
```
aswaterman: I suggest using a more descriptive name than `func` since this is in file scope, and it is a large file.  Maybe `memory_access_func`.

riscv/debug_module.cc
```diff
@@ -649,178 +653,281 @@ bool debug_module_t::perform_abstract_command()
...
+static constexpr std::array<func, 4> lx = {&lb, &lh, &lw, &ld};
+static constexpr std::array<func, 4> sx = {&sb, &sh, &sw, &sd};
```
aswaterman: It looks like the `handle_memory_` functions are always called with XLEN, so only `lw`/`ld`/`sw`/`sd` will ever be invoked.  Am I wrong about that?  Are we including `lb` etc. for some future feature addition?  I think it's OK either way; I am just trying to understand.

aswaterman (commented):
LGTM overall; I'll approve once my comments are addressed.

aswaterman:
Also, please dig into the CI failure.

riscv/debug_module.cc
fkhaidari: Addressed

riscv/debug_module.cc
fkhaidari: It may be used if `aamsize == 0`. There are lines like: `lx[aamsize]`

aswaterman (approved):

aswaterman:
Please ping me again once the CI failures are addressed.

fkhaidari:
> Also, please dig into the CI failure.

@aswaterman, looks like there is a bug  in OpenOCD: [here](https://github.com/riscv-collab/riscv-openocd/blob/d4c5d2657074613d429f57f60e939ca151ed4f32/src/target/riscv/riscv-013.c#L1258-L1259) and [here](https://github.com/riscv-collab/riscv-openocd/blob/d4c5d2657074613d429f57f60e939ca151ed4f32/src/target/riscv/riscv-013.c#L1275-L1276) (wrong comparison sign). To make it work I need to patch OpenOCD and Spike debug testing workflow

fkhaidari:
@aswaterman, do you mind if I add an extra configuration option to spike? The motivation follows.

Right now datasize is hardcoded. It would be useful to make it a config parameter like progbufsize.
The spec says datacount should be at least 1 for RV32, 2 for RV64, or 4 for RV128. 
Making this configurable would let us test edge cases - like what happens with the minimal datacount=1 config,
or extended configs with more data registers. This is especially relevant for abstract commands 
that use multiple data registers for 64/128-bit operations. To implement this we need need to:

* Add datacount to debug_module_config_t
* Turn dmdata to a vector to use the config value instead of the hardcoded constant
* Add validation in perform_abstract_command() to check if the requested operation fits in the available data registers

This would help with automated testing of OpenOCD against different DM implementations.

arromanoff:
@fk-sc Configurable `datasize` sounds great to me

aswaterman:
I'm not wild about adding new command-line options, especially for such a niche feature that would really only ever be used for testing OpenOCD.

If we really need to add this as a command-line option, then I would rather define an ISA extension name that describes the behavior.  (That should be standard practice when standards have ISA-visible options like this, anyway.)

fkhaidari:
> I'm not wild about adding new command-line options, especially for such a niche feature that would really only ever be used for testing OpenOCD.
> 
> If we really need to add this as a command-line option, then I would rather define an ISA extension name that describes the behavior. (That should be standard practice when standards have ISA-visible options like this, anyway.)

@aswaterman, I understand the concern about CLI proliferation. Here's why I think `--dm-datacount` fits better as a CLI flag than an ISA extension:

* I believe it's not ISA-visible

`abstractcs` is a Debug Module register accessed via DMI, not a CSR accessible to the hart. The hart cannot execute `csrr x, abstractcs` - there's no such CSR. Only the external debugger accesses it via the DMI protocol.

* Existing Spike precedent

All other DM implementation parameters are CLI flags:
```
--dm-progsize=<words>
--dm-sba=<bits>
--dm-abstract-rti=<n>
--dm-no-hasel
```

These control Debug Module behavior but aren't ISA extensions. `datacount` is the same category.

* ISA extensions define instructions/CSRs

It is unclear what would a "datacount ISA extension" specify. It has no instructions, no CSRs, no encodings - just a DMI register field value. The Debug Spec already defines the legal range (1-12).

My proposal is to keep the existing pattern:
```
--dm-datacount=<n>    # [1-12, default 2]
```

nadime15:
@fk-sc Do you actually test your implementation against downstream riscv-openocd or upstream openocd? I know their behavior isn’t exactly the same and (correct me if I’m wrong) some changes aren’t upstream yet or vice versa. I’m currently working on the debug module and couldn’t even get past the init step of OpenOCD with the riscv-openocd version (I’m using the upstream one).

I didn’t spend any time debugging it because I assume most people using OpenOCD will use the upstream version. So, the bug reports you mentioned, are they also present in upstream? Because you linked to riscv-openocd.

aswaterman:
I forgot that we already have the various `--dm-` flags.  I guess I'm OK with adding another one.

fkhaidari:
> @fk-sc Do you actually test your implementation against downstream riscv-openocd or upstream openocd? I know their behavior isn’t exactly the same and (correct me if I’m wrong) some changes aren’t upstream yet or vice versa. I’m currently working on the debug module and couldn’t even get past the init step of OpenOCD with the riscv-openocd version (I’m using the upstream one).
> 
> I didn’t spend any time debugging it because I assume most people using OpenOCD will use the upstream version. So, the bug reports you mentioned, are they also present in upstream? Because you linked to riscv-openocd.

@nadime15, I have tested against riscv-openocd. Same commit that was used in CI

I have checked upstream openocd source code and it has same problem: [here](https://github.com/openocd-org/openocd/blob/e5888bda38f4952e2ae92f7dc5b25fc9a2d1c2b3/src/target/riscv/riscv-013.c#L1155) and [here](https://github.com/openocd-org/openocd/blob/e5888bda38f4952e2ae92f7dc5b25fc9a2d1c2b3/src/target/riscv/riscv-013.c#L1172)

aswaterman:
@fk-sc I think CI was failing because @arrv-sc and I reworked the CI flow recently.  If it passes after rebasing, I'll merge it.

aswaterman:
OK, well, that's why a subset of the CI tests were failing.  The rest is up to you :)

riscv/debug_module.cc
```diff
@@ -649,178 +653,278 @@ bool debug_module_t::perform_abstract_command()
...
+  bool aamvirtual = get_field(command, AC_ACCESS_MEMORY_AAMVIRTUAL);
+  bool is_write = get_field(command, AC_ACCESS_MEMORY_WRITE);
+  auto xlen = sim->get_harts().at(selected_hart_id())->get_xlen();
+
+  if (!selected_hart_state().halted) {
+    abstractcs.cmderr = CMDERR_HALTRESUME;
+    return true;
+  }
+
+  if (aamvirtual || aamsize > 3 || aamsize > idx(xlen)) {
```
nadime15: So I have a general comment about handling `aamvirtual` because I am not entirely sure myself. aamvirtual = 0 means the address is physical, no matter what, and regardless of dcsr.mprven, mstatus.mprv, or mstatus.mpp. I’m not very familiar with the internals of Spike, but I imagine these bits are already handled correctly for each memory access. However, if the debugger allows address translation as if executed in S-Mode by setting the mentioned bits, the hart would interpret the provided addresses as virtual, even if aamvirtual is 0.

To make it short (and I am not 100% certain about this either), aamvirtual enforces physical memory access. Setting aamvirtual = 1 basically behaves as usual by checking MPRV, in other words, you are already supporting aamvirtual = 1, but not really aamvirtual = 0.

Assuming the above is correct, there are a few ways to work around this. One approach is to clear MPRV and then restore it to its initial state when aamvirtual = 0.

fkhaidari: Thank you for catching this! You're absolutely right that the current implementation 
has issues with `aamvirtual` handling.

The current code doesn't properly support `aamvirtual=1`, which should behave as if 
`MPRV=1`. The current logic doesn't implement this translation mode override.

I'm working on a follow-up commit to add proper `MPRV` bit handling. Probably I have
to merge this commit with current to common PR.

nadime15: So I had another look at the spec. I had assumed that if aamvirtual is set to 1, we still check whether MPRV is set, but after rereading the spec, 

> Addresses are virtual, and translated the way they would be from M-mode, with MPRV set.

It seems to me that in this case we should behave as if mstatus.mprv is set to 1 (even if its at the given moment set to 0) and check mpp.

@fk-sc what I meant is I am not sure if the current code handles aamvirtual=0 correctly. I am thinking about the case when aamvirtual=0 but dcsr.mprven=1, mstatus.MPRV=1, and mstatus.MPP=1 (S-mode) are set. In this situation, I believe the abstract command will do virtual translation instead of physical, which would be wrong(?).
I still dont fully understand how the internals work, but from what I can see, Spike translates abstract commands into instructions that are stored in a dedicated memory region called "debug_abstract". These instructions then execute as normal instructions and go through the full MMU checking and translation pipeline (correct me if I'm wrong here). When the MMU processes these instructions, it will see that dcsr.mprven=1 and mstatus.MPRV=1 are set, which makes the in_mprv() function return true. This causes the MMU to use the privilege mode from mstatus.MPP for translation, resulting in virtual address translation even though the debugger requested physical addresses with aamvirtual=0.

riscv/debug_module.cc
```diff
@@ -844,39 +859,124 @@ bool debug_module_t::perform_abstract_command()
...
+
+  debug_rom_flags[selected_hart_id()] |= 1 << DEBUG_ROM_FLAG_GO;
+  rti_remaining = config.abstract_rti;
+  abstract_command_completed = false;
+
+  abstractcs.busy = true;
+  return true;
+}
+
+static inline unsigned idx(unsigned xlen)
```
arromanoff: Can remove `inline` here please? It doesn't mean anything inside a .cc file (unlike in C)

fkhaidari: Addressed

arromanoff: Thanks!

arromanoff (approved):

riscv/debug_module.cc
```diff
@@ -715,120 +730,120 @@ bool debug_module_t::perform_abstract_command()
...
-              write32(debug_abstract, i++, fsw(fprnum, ZERO, debug_data_start));
-              break;
-            case 3:
-              write32(debug_abstract, i++, fsd(fprnum, ZERO, debug_data_start));
-              break;
-            default:
-              abstractcs.cmderr = CMDERR_NOTSUP;
-              return true;
-          }
+    } else if (regno >= 0x1020 && regno < 0x1040 && config.support_abstract_fpr_access) {
```
nadime15, line 817:
```suggestion
    } else if (regno < 0x1040 && config.support_abstract_fpr_access) {
```

riscv/debug_module.cc
```diff
@@ -715,120 +730,120 @@ bool debug_module_t::perform_abstract_command()
...
-          /*
-           * The exception handler starts out be restoring dscratch to s0,
-           * which was saved before executing the abstract memory region. Since
-           * we just wrote s0, also make sure to write that same value to
-           * dscratch in case an exception occurs in a program buffer that
-           * might be executed later.
-           */
-          write32(debug_abstract, i++, csrw(S0, CSR_DSCRATCH0));
-        }
+    } else if (regno >= 0x1000 && regno < 0x1020) {
```
nadime15, line 785:
```suggestion
    } else if (regno < 0x1020) {
```

It’s just a personal preference, but I find it easier to read and compare against the spec if less than or equal (regno <= 0x101f) is used.

riscv/debug_module.cc
```diff
@@ -715,120 +730,120 @@ bool debug_module_t::perform_abstract_command()
...
+          case 3:
+            write32(debug_abstract, i++, fsd(fprnum, ZERO, debug_data_start));
+            break;
+          default:
+            abstractcs.cmderr = CMDERR_NOTSUP;
+            return true;
+        }
+      }
+
       } else if (regno >= 0xc000 && (regno & 1) == 1) {
```
nadime15, line 847:
```suggestion
      } else if (regno <= 0xffff) {
```

Any reason why only odd-numbered custom CSRs are allowed?

riscv/debug_module.cc
```diff
@@ -844,39 +859,138 @@ bool debug_module_t::perform_abstract_command()
...
+  bool aamvirtual = get_field(command, AC_ACCESS_MEMORY_AAMVIRTUAL);
+  bool is_write = get_field(command, AC_ACCESS_MEMORY_WRITE);
+  auto xlen = sim->get_harts().at(selected_hart_id())->get_xlen();
+
+  if (!selected_hart_state().halted) {
+    abstractcs.cmderr = CMDERR_HALTRESUME;
+    return true;
+  }
+
+  if (aamsize > 3 || aamsize > idx(xlen)) {
```
nadime15:
```suggestion
  if (aamsize > idx(xlen)) {
```
Should be enough, no?

riscv/debug_module.cc
fkhaidari: All the code related to abstract register access was not modified by me. I had to change tabs. I'd be happy to refactor this part, but probably it's too much for this commit

riscv/debug_module.cc
fkhaidari: Addressed

riscv/debug_module.cc
nadime15: Ah ok, fair enough

nadime15 (approved):
Looks great! The handling of aamvirtual has been addressed.

aswaterman:
Thanks for taking the time to review this, @nadime15.  Looks like it still isn't passing, @fk-sc; ping me when it's time to merge.

fkhaidari:
> @fk-sc; ping me when it's time to merge.

@aswaterman, ping

fkhaidari:
@aswaterman, @nadime15, @arrv-sc, thanks for review

riscv/debug_module.cc
aap-sc: wait a minute! this was not in the original scope of changes. I believe that this is require for OpenOCD testing, they had tests targeting this.

aap-sc: There is literally the following comment a few lines below:

```
 // Support odd-numbered custom registers, to allow for debugger testing
```

It is worth investigating why this was not detected by the current CI

fkhaidari: @aap-sc, no, I haven't touched that part. @nadime15, looks like we have justification for why it's done

nadime15: @aap-sc I think thats strange, unless its explicitly specified by the spec. If I define an even-numbered custom register now, I wont be able to access it.

i would like to understand the rationality behind that.

en-sc: @nadime15, please take a look https://github.com/riscv-software-src/riscv-tests/commit/4dddbc79ada7f0a836cf538676c57c8df103ccf6

This is just an arbitrary decision made some time ago by Tim to allow testing both successful and failing accesses.

aap-sc: @nadime15 sorry, somehow I've missed your comment. @en-sc already provided the answer.

PR #2088, opened 2025-09-26
PTE store in s2xlate should use the trap_type instead of type

### LINKED ISSUE #2078: Incorrect access fault is thrown during G-stage page walk
In the situation where a store instruction (eg hsv) is performed and the implicit G-stage walk to support the VS/VU level access, the access type is set to LOAD https://github.com/riscv-software-src/riscv-isa-sim/blob/f51df5d3955a27602a872eaf01492177513baf6f/riscv/mmu.cc#L602

The spec states that:
```
For a memory access made to support VS-stage address translation (such
as to read/write a VS-level page table), permissions and the need to set A and/or D bits at the G-stage level
are checked as though for an implicit load or store, not for the original access type. However, any exception
is always reported for the original access type (instruction, load, or store/AMO).
```

Setting the access type to LOAD is correct but the the problem is that we may not respect the "trap original access type" requirement when loading the PTE. Since the original access type can also be a fetch or a store, there is no code to convert the load access fault into a store/fetch access fault. The way I understand the spec the following code would be needed.

```cpp
reg_t pte_paddr = 0;
try {
  // access type should be LOAD but trap type should be that of original access type
  pte_paddr = s2xlate(addr, base + idx * vm.ptesize, LOAD, type, virt, false, true, insn, true, allow_ad_updates);
} catch (trap_load_access_fault& t){
  if (type == STORE)
    throw trap_store_access_fault(t.has_gva(), t.get_tval(), t.get_tval2(), t.get_tinst());
   else if (type == FETCH)
     throw trap_instruction_access_fault(t.has_gva(), t.get_tval(), t.get_tval2(), t.get_tinst());
   else
     throw t;
   }
```
EDIT: I use a pretty modified version of spike which is why there are some extra parameters in certain places, this should be irrelevant for the core issue.

ved-rivos:
Closes #2078

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -534,7 +534,7 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
         if ((pte & ad) != ad) {
           if (hade) {
             // set accessed and possibly dirty bits
-            pte_store(pte_paddr, pte | ad, gva, virt, type, vm.ptesize);
+            pte_store(pte_paddr, pte | ad, gva, virt, trap_type, vm.ptesize);
           } else {
             // take exception if access or possibly dirty bit is not set.
             break;


aswaterman (approved):
Ahh, makes sense.

PR #2091, opened 2025-09-30
Get rid of leaking pointers in examples for custom extension

aap-sc:
Building of spike-based simulator with memory sanitizer reports leaking pointers if custom extension are used. This is because existing facilities do not have a proper destructor procedure, so the objects representing custom extentions are leaked. This commit implements quick-and-dirty fix for the problem.

### DIFF
--- a/ci-tests/custom-csr.cc
+++ b/ci-tests/custom-csr.cc
@@ -37,7 +37,7 @@ struct xdummycsr_t : public extension_t {
   }
 };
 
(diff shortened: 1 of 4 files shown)


aap-sc:
@aswaterman would you kindly take a look? This is really more a quick-and-dirty workaround than a proper fix, however I'm not sure that we need more than that. The motivations to simplify usage of spike in environments with enabled sanitizers.

aswaterman (approved):
Since by construction these can only execute once, the memory leak doesn't matter functionally, but being sanitizer-clean is a good thing.

Issue #2080, opened 2025-09-10
RFC: Integrate llvm-snippy into  riscv-isa-sim's test suite


arromanoff:
# RFC: Integrate llvm-snippy into riscv-isa-sim's test suite

Hi! I want to propose an idea for improving Spike's testing. I'm the developer of [llvm-snippy](https://github.com/syntacore/snippy), a tool that generates random RISC-V code snippets for testing purposes. It supports most of the RISC-V standard extensions and can create interesting control flow with loops and other branches. We've already used it to find and fix several bugs in Spike and even some in sail-riscv (e.g. https://github.com/riscv-software-src/riscv-isa-sim/pull/1995, https://github.com/riscv-software-src/riscv-isa-sim/pull/1992 and https://github.com/riscv/sail-riscv/pull/1090). I think it could be valuable for this project.

### What is llvm-snippy?
llvm-snippy is an open-source llvm-based tool that generates random but valid RISC-V code snippets. It allows you to specify how many instructions to generate and probabilities of generating each of these instructions. This makes it great for testing a wide range of scenarios, from simple arithmetic instruction sequences to complex control flows, floating-point arithmetic with different rounding modes and atomic memory accesses into different memory regions. One of the snippy's best features in my opinion is that it reuses llvm's backend and instruction descriptions, which means that every instruction supported by LLVM and clang (Which is most of the existing RISC-V extensions) is automatically supported in llvm-snippy. That allows snippy not to worry about generating invalid opcodes.

We've been using it extensively at my company, and it's helped us catch a lot of issues that might have gone unnoticed otherwise. I think it could be a great addition to Spike's test suite.

### Why llvm-snippy?

Random instruction testing is a common approach and many cores/simulators use it. For example [sail-riscv](https://github.com/riscv/sail-riscv) uses [riscv-vector-tests](https://github.com/chipsalliance/riscv-vector-tests) in CI. llvm-snippy, however is a much more capable generator. While riscv-vector-tests (and many others) generate single-instruction tests, llvm-snippy can generate many instructions from different extensions in the same snippet. That allows to cover more complex test cases. For example snippet where result of instruction from V extension is a source operand of double-precision operation from D extension. Or where atomics interleave with strided loads/stores to the same memory, etc.

I will not describe all the features snippy has here. However we recently had [2.0 release](https://github.com/syntacore/snippy/releases/tag/snippy-2.0) with extensive documentation and examples that you can check out.

### The Proposal
[...]

PR #2095, opened 2025-10-01
ci: add testing with llvm-snippy

arromanoff:
This commits adds basic spike testing using llvm-snippy random code generator. This initial testing runs spike on random valid code snippets and checks that it doesn't fail.

This commit addresses https://github.com/riscv-software-src/riscv-isa-sim/issues/2080


arromanoff:
@aswaterman
This commit's test-spike script passes on my machine. It is hard, however, to make sure it will work on github runner. Should we do some other checks (like testing on different machine) before merging?

aswaterman:
@arrv-sc The problem is that the CI tarball needs to be manually uploaded to github.

But I'd like to get rid of the tarball altogether and build the dependencies inside CI.  I'm working on that now.

aswaterman:
@arrv-sc See https://github.com/riscv-software-src/riscv-isa-sim/pull/2096 -- the tarball is gone and you can assume the existence of a RISC-V GCC toolchain within CI.  You'll need to redo your modifications to the build scripts in ci-tests, but it should be much lower friction than before.

arromanoff:
@aswaterman  Thanks a lot! Already rebased changes

aswaterman (approved):
Cool, thanks!  Let's move forward with this, and you can extend it over time.

PR #2103, opened 2025-10-11
fix: move vectorUnit_t::elt defintion to the header

arromanoff:
Prior to this commit vectorUnit_t::elt was a bottleneck for vector instructions performance. This function was not being inlined because the definition was inside a .cc file. After moving the definition to the header file I measured the 20-30% increase in performance on random vector test from CI.

I'd like to also mention that placing a template function's definition outside of its' header file is a very fragile thing from C++'s perspective as compiler neither can generate instantiation for it nor is it aware of any specialization existing elsewhere (as it would be in case of having `extern template<> ...` declaration in the same header)

I found this problem by running spike under callgrind. Here're before and after results when running on vector.yaml test from CI:

#### before the change
Screenshot From 2025-10-10 12-11-08

#### after the change
Screenshot From 2025-10-10 12-18-02

Here you can see that `vectorUnit_t::elt`'s time changed from the most time consuming to basically not affecting runtime at all


aswaterman (commented):
This is something I had been meaning to do for the same reason.  I agree, it's a no-brainer of a performance enhancement.

Two comments:
- Shouldn't we do the same for `elt_group`, for symmetry?
- Can't we delete all of this code as a result? https://github.com/riscv-software-src/riscv-isa-sim/blob/cce97251a8bceb083dd96f3237cd34208a5b29ae/riscv/vector_unit.cc#L145-L161

arromanoff:
@aswaterman I agree with you on both accounts. Updated the commit

aswaterman (approved):
LGTM, thanks.

PR #2105, opened 2025-10-14
Add configurable datacount for debug module

fkhaidari:
This commit adds support for configuring the number of data registers available in the debug module. Previously, the debug module had a fixed datasize of 2, but now users can specify the number of data registers using the --dm-datacount option when running spike.

The changes include:
- Adding a datacount parameter to debug_module_config_t
- Making dmdata a std::vector instead of a fixed array
- Validating that datacount is between 1 and 12
- Updating the debug module to use the configured datacount
- Adding command-line option to set datacount
- Updating documentation in help output

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -1,4 +1,8 @@
+#include <algorithm>
+#include <array>
 #include <cassert>
+#include <iterator>
+#include <limits>
 
 #include "simif.h"
 #include "devices.h"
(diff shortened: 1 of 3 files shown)


fkhaidari:
@aswaterman, @arrv-sc, could you take a look?

fkhaidari:
I've mentioned this feature in #2086

riscv/debug_module.cc
```diff
@@ -122,38 +166,27 @@ void debug_module_t::reset()
...
-
-  if (addr >= debug_data_start && (addr + len) <= (debug_data_start + sizeof(dmdata))) {
-    memcpy(bytes, dmdata + addr - debug_data_start, len);
-    return true;
-  }
+  const auto interval_ptr =
+      std::find_if(debug_memory_regions.begin(), debug_memory_regions.end(),
+                   [addr, len](const auto &range) {
+                     return belongs_to_range(addr, len, range.addr, range.len);
+                   });
```
aswaterman, line 186: This is obviously going to perform somewhat more poorly than the old code, but I think that's OK in this context; Spike should not be spending much time in these routines in normal operation.

aswaterman (approved):
LGTM, thanks.

aswaterman:
@arrv-sc I'll wait for your approval before I merge.

arromanoff:
LGTM!

arromanoff (approved):

PR #2121, opened 2025-11-04
Add Svukte extension support

binno:
This adds support for the Svukte extension, which adds support for address-independent latency of user-mode faults to supervisor addresses.

### DIFF
--- a/disasm/isa_parser.cc
+++ b/disasm/isa_parser.cc
@@ -265,6 +265,10 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
       extension_table[EXT_SVPBMT] = true;
     } else if (ext_str == "svinval") {
       extension_table[EXT_SVINVAL] = true;
+    } else if (ext_str == "svukte") {
+      if (max_xlen != 64)
+        bad_isa_string(str, "'svukte' requires RV64");
+      extension_table[EXT_SVUKTE] = true;
     } else if (ext_str == "zfa") {
       extension_table[EXT_ZFA] = true;
     } else if (ext_str == "zicbom") {
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -53,6 +53,16 @@ void throw_access_exception(bool virt, reg_t addr, access_type type)
   }
 }
 
+[[noreturn]] void throw_page_fault_exception(bool virt, reg_t addr, access_type type)
+{
+  switch (type) {
+    case FETCH: throw trap_instruction_page_fault(virt, addr, 0, 0);
+    case LOAD: throw trap_load_page_fault(virt, addr, 0, 0);
+    case STORE: throw trap_store_page_fault(virt, addr, 0, 0);
+    default: abort();
+  }
+}
+
 reg_t mmu_t::translate(mem_access_info_t access_info, reg_t len)
 {
   reg_t addr = access_info.transformed_vaddr;
@@ -560,6 +570,28 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
   }
 }
 
+bool mmu_t::check_svukte_qualified(reg_t addr, reg_t mode, bool forced_virt)
+{
+  state_t* state = proc->get_state();
+
+  if (mode != PRV_U)
+    return true;
+
+  if (proc->extension_enabled('S') && get_field(state->senvcfg->read(), SENVCFG_UKTE)) {
+    if (forced_virt && state->prv == PRV_U) {
+      bool hstatus_hukte = proc->extension_enabled('H') && (get_field(state->hstatus->read(), HSTATUS_HUKTE) == 1);
+      return !hstatus_hukte;
+    }
+
+    assert(proc->get_xlen() == 64);
+    if ((addr >> 63) & 1) {
+      return (state->v || forced_virt) && ((state->vsatp->read() & SATP64_MODE) == 0);
+    }
+  }
+
+  return true;
+}
+
 reg_t mmu_t::walk(mem_access_info_t access_info)
 {
   access_type type = access_info.type;
@@ -581,6 +613,10 @@ reg_t mmu_t::walk(mem_access_info_t access_info)
 
   if (vm.levels == 0)
     return s2xlate(addr, addr & ((reg_t(2) << (proc->xlen-1))-1), type, type, virt, hlvx, false) & ~page_mask; // zero-extend from xlen
+
+  if (proc->extension_enabled(EXT_SVUKTE) && !check_svukte_qualified(addr, mode, access_info.flags.forced_virt)) {
+    throw_page_fault_exception(virt, addr, type);
+  }
 
   bool s_mode = mode == PRV_S;
   bool sum = proc->state.sstatus->readvirt(virt) & MSTATUS_SUM;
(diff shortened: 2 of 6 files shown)


riscv/mmu.cc
```diff
@@ -560,6 +570,35 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
...
+    }
+
+    if (((addr >> SYS_MEM_HEADING_BIT) & 0x1)) {
+      if ((state->v || forced_virt) &&
+          ((proc->get_xlen() == 64) ? ((state->vsatp->read() & SATP64_MODE) == 0) :
+                                      ((state->vsatp->read() & SATP32_MODE) == 0))) {
+        return true;
+      } else {
+        return false;
+      }
```
aswaterman: Replace `if (x) return true; else return false;` with `return x;`.

aswaterman (commented):
@ingallsj can I delegate to you to review this, particularly the various cases in `check_svukte_qualified`?

riscv/mmu.cc
```diff
@@ -560,6 +570,35 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
...
+{
+  state_t* state = proc->get_state();
+
+  if (mode != PRV_U)
+    return true;
+
+  if (proc->extension_enabled('S') && get_field(state->senvcfg->read(), SENVCFG_UKTE)) {
+    if (forced_virt && state->prv == PRV_U) {
+      bool hstatus_hukte = proc->extension_enabled('H') &&
+                           (get_field(state->hstatus->read(), HSTATUS_HUKTE) == 1);
```
aswaterman: This would read better as `bool hstatus_hukte = proc->extension_enabled('H') && get_field(state->hstatus->read(), HSTATUS_HUKTE);`

riscv/mmu.cc
binno: Done

binno (commented):
Hi @aswaterman 
Modification by your suggestion has been done.
Many thanks for review

disasm/isa_parser.cc
```diff
@@ -265,6 +265,8 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
       extension_table[EXT_SVPBMT] = true;
     } else if (ext_str == "svinval") {
       extension_table[EXT_SVINVAL] = true;
+    } else if (ext_str == "svukte") {
+      extension_table[EXT_SVUKTE] = true;
```
aswaterman, line 271: We should have a check that XLEN is 64 and issue an error if not.

riscv/mmu.cc
```diff
@@ -560,6 +570,29 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
...
+  if (mode != PRV_U)
+    return true;
+
+  if (proc->extension_enabled('S') && get_field(state->senvcfg->read(), SENVCFG_UKTE)) {
+    if (forced_virt && state->prv == PRV_U) {
+      bool hstatus_hukte = proc->extension_enabled('H') && (get_field(state->hstatus->read(), HSTATUS_HUKTE) == 1);
+      return !hstatus_hukte;
+    }
+
+    if (((addr >> SYS_MEM_HEADING_BIT) & 0x1)) {
```
aswaterman: I think we should have `assert(proc->get_xlen() == 64)` here.  Then we can simply write `if ((addr >> 63) & 1)`.

riscv/mmu.cc
```diff
@@ -560,6 +570,29 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
...
+      return ((state->v || forced_virt) &&
+          ((proc->get_xlen() == 64) ? ((state->vsatp->read() & SATP64_MODE) == 0) :
+                                      ((state->vsatp->read() & SATP32_MODE) == 0)));
```
aswaterman: And then we can get rid of the XLEN=32 stuff here.

riscv/mmu.h
```diff
@@ -34,6 +34,9 @@ const reg_t PGSIZE = 1 << PGSHIFT;
 #define MMU_OBSERVE_STORE(addr, data, length)
 #endif
 
+// Svukte extension
+#define SYS_MEM_HEADING_BIT 63
```
aswaterman: And we can get rid of this, too.

riscv/mmu.cc
```diff
@@ -53,6 +53,16 @@ void throw_access_exception(bool virt, reg_t addr, access_type type)
   }
 }
 
+[[noreturn]] void throw_page_access_exception(bool virt, reg_t addr, access_type type)
```
aswaterman: Rename this function to `throw_page_fault_exception`.

binno (commented):
Hi @aswaterman
Modification by your suggestion has been done.
Please help to review it again

disasm/isa_parser.cc
```diff
@@ -265,6 +265,10 @@ isa_parser_t::isa_parser_t(const char* str, const char *priv)
       extension_table[EXT_SVPBMT] = true;
     } else if (ext_str == "svinval") {
       extension_table[EXT_SVINVAL] = true;
+    } else if (ext_str == "svukte") {
+      if (max_xlen != 64)
+        bad_isa_string(str, "'Zilsd' requires RV64");
```
aswaterman: One more :) Zilsd -> Svukte

binno: Sorry, thanks for reminder

aswaterman (approved):

riscv/mmu.cc
```diff
@@ -560,6 +570,28 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
...
+{
+  state_t* state = proc->get_state();
+
+  if (mode != PRV_U)
+    return true;
+
+  if (proc->extension_enabled('S') && get_field(state->senvcfg->read(), SENVCFG_UKTE)) {
+    if (forced_virt && state->prv == PRV_U) {
+      bool hstatus_hukte = proc->extension_enabled('H') && (get_field(state->hstatus->read(), HSTATUS_HUKTE) == 1);
+      return !hstatus_hukte;
```
ingallsj, line 583: I feel like there should be an address check before this `return` statement.

riscv/mmu.cc
```diff
@@ -560,6 +570,28 @@ reg_t mmu_t::s2xlate(reg_t gva, reg_t gpa, access_type type, access_type trap_ty
...
+    }
+
+    assert(proc->get_xlen() == 64);
+    if ((addr >> 63) & 1) {
+      return (state->v || forced_virt) && ((state->vsatp->read() & SATP64_MODE) == 0);
```
ingallsj, line 588: Am I understanding this return statement correctly, that we pass the Svukte check if in VU mode and `vsatp.MODE == Bare`?  Where do we check `satp` for when in HU mode?

riscv/mmu.cc
aswaterman: I think there's both an actual bug and a conceptual bug here.  Per the spec, "Svukte-qualified" is a property of an access type, not an address.  Either this function should not accept an address arg, or it should be renamed.  That might be why this bug cropped up.  I'll make a fix.

binno: I think the address check is in line 587

aswaterman: But if you hit the return statement, you never reach like 587, so the address never gets checked for HLV/HSV in U-mode.

I think my PR #2171 fixes this case.

PR #2122, opened 2025-11-04
Add smpmpmt extension support

binno:
This adds support for the Smpmpmt extension which provides a mechanism analogous to Svpbmt but associated with the PMP registers rather than page-table entries.


riscv/encoding.h
```diff
@@ -375,6 +375,7 @@
 #define PMP_W     0x02
 #define PMP_X     0x04
 #define PMP_A     0x18
+#define PMP_MT    0x60
```
chihminchao, line 378: Is it included in riscv-opcodes ?

binno: Not yet

aswaterman: We don't allow PRs that manually edit `encoding.h`.  Please make the change to `riscv-opcodes` first, then re-generate `encoding.h`.

riscv/csrs.cc
```diff
@@ -249,7 +249,9 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
     if (i < proc->n_pmp) {
       const bool locked = (state->pmpaddr[i]->cfg & PMP_L);
       if (rlb || !locked) {
-        uint8_t cfg = (val >> (8 * (i - i0))) & (PMP_R | PMP_W | PMP_X | PMP_A | PMP_L);
+        uint8_t cfg = (val >> (8 * (i - i0))) & (PMP_R | PMP_W | PMP_X | PMP_A |
+            (proc->extension_enabled(EXT_SMPMPMT) ?  PMP_MT : 0) |
+            PMP_L);
```
aswaterman, line 254: Probably best to separate this out into two statements:

```
uint8_t all_cfg_fields = PMP_R | ...;
uint8_t cfg = (val >> (8 * (i - i0))) & all_cfg_fields;
```

riscv/csrs.cc
```diff
@@ -258,6 +260,9 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
         // Disallow A=NA4 when granularity > 4
         if (proc->lg_pmp_granularity != PMP_SHIFT && (cfg & PMP_A) == PMP_NA4)
           cfg |= PMP_NAPOT;
+        // MT value 0x3 is reserved
+        if ((cfg & PMP_MT) == 0x3)
```
aswaterman: This looks wrong because it is missing either a right-shift of the LHS or a left-shift of the RHS.  I think the cleanest expression is `(cfg & PMP_MT) == PMP_MT`

aswaterman (commented):
LGTM once my comments are addressed; ping me when it's ready.

riscv/encoding.h
chihminchao: @binno There is PR for riscv-opcodes, https://github.com/riscv/riscv-opcodes/pull/386. and please also update entire encoding.h after merging.

aswaterman: Merged.

binno: update encoding.h

riscv/csrs.cc
binno: PMP_MT is a WARL field.
Current write value 0x3 is reserved and this kind write value should be ignored.

binno (commented):
Hi @aswaterman
Modification by your suggestion has been done.
Many thanks for review

riscv/csrs.cc
aswaterman: Yeah, I understand that.  But let's say `cfg.MT` is set to 3.  Then the expression `cfg & PMP_MT` will have the value 0x60, not 3.  You need to either right-shift the LHS by 5 bits, or left-shift the RHS by 5 bits, or do the cleaner thing of writing `(cfg & PMP_MT) == PMP_MT`.

aswaterman: Or you could write `get_field(cfg, PMP_MT) == 3`, I guess.  I don't have a preference for either of my two suggestions.

binno: Sorry!!! I make mistake here. It's my fault and I will fix it soon

binno:
Hi @aswaterman,
    Please help to review it again, thanks

aswaterman (approved):

aswaterman:
LGTM now, thanks @binno.  It looks like there is an unrelated CI problem, so please ping me tomorrow once they have been resolved.

PR #2125, opened 2025-11-05
vector: constrain index load/store for eew64 and xlen32

chihminchao:
Section 18.3 says
"The V extension supports all vector load and store instructions (Section Vector Loads and Stores), except the V extension does not support EEW=64 for index values when XLEN=32"

### DIFF
--- a/riscv/v_ext_macros.h
+++ b/riscv/v_ext_macros.h
@@ -89,7 +89,7 @@ static inline bool is_overlapped_widen(const int astart, int asize,
 
 #define VI_CHECK_ST_INDEX(elt_width) \
   require_vector(false); \
-  require(elt_width <= P.VU.ELEN); \
+  require(elt_width <= std::min(P.VU.ELEN, (reg_t)P.get_xlen())); \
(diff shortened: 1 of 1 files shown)


riscv/v_ext_macros.h
```diff
@@ -90,6 +90,7 @@ static inline bool is_overlapped_widen(const int astart, int asize,
 #define VI_CHECK_ST_INDEX(elt_width) \
   require_vector(false); \
   require(elt_width <= P.VU.ELEN); \
+  require(elt_width <= P.get_xlen()); \
```
aswaterman: Maybe `require(elt_width <= std::min(P.VU.ELEN, P.get_xlen()));`?  But it is a matter of opinion.

aswaterman (approved):
LGTM, please rebase to fix the CI failure.

riscv/v_ext_macros.h
chihminchao: Follow the suggestion to merge in one require statement

PR #2126, opened 2025-11-05
csr: disalbe page fault delegation with nommu

chihminchao:


aswaterman (commented):
I am not sure this list is complete.  @ved-rivos can you remind us which other extensions can throw page-fault exceptions when there is no MMU?  Are all of the problematic extensions ones that Spike doesn't support yet?

ved-rivos:
Besides Sv* extensions, the only other extension that can produce a page fault without MMU is SPMP. I believe SPMP is not implemented in spike yet.

aswaterman (approved):
Thanks, @ved-rivos.  I think we can proceed with this PR as-is.

aswaterman:
@chihminchao it looks like the CI failures are unrelated to the code change.  Let me know if you need me to retry CI in a few hours.

Issue #2129, opened 2025-11-06
Disassemble `fcvtmod.w.d` shows "(args unknown)"


liuyu81:
The [`fcvtmod.w.d`](https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/isa_20240411/insts/fcvtmod.w.d) instruction from `zfa` extension is not correctly decoded, i.e.,

```
$ spike-dasm --isa=rv64gc_zfa
DASM(0xc2809553)
fcvtmod.w.d (args unknown)
```

I believe it should be `fcvtmod.w.d  a0, ft1, 1`.

Looking into the source code in `disasm/disasm.cc`, it seems that the instruction is not defined in `disassembler_t::add_instructions`, and there is no `arg_t` formatter for `insn_t::rm` as well.

I am willing to submit a PR if this is worth fixing.

aswaterman:
Sure, go ahead and submit a PR if you have the time.  I'll review it.

Rounding modes should be printed using their strings (rne, rtz, etc.) rather than integers.

aswaterman:
Resolved by #2130

PR #2130, opened 2025-11-06
Decode args for `fcvtmod.w.d` in `zfa` extension

liuyu81:
This should fix the "(args unknown)" issue reported in #2129.

### DIFF
--- a/disasm/regnames.cc
+++ b/disasm/regnames.cc
@@ -31,3 +31,21 @@ const char* csr_name(int which) {
   }
   return "unknown-csr";
 }
+
+const char* frm_name(int which) {
+  switch (which) {
+    case 0:
+      return "rne";
+    case 1:
+      return "rtz";
+    case 2:
+      return "rdn";
+    case 3:
+      return "rup";
+    case 4:
+      return "rmm";
+    case 7:
+      return "dyn";
+  }
+  return "unknown-frm";
+}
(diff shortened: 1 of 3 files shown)


disasm/regnames.cc
```diff
@@ -31,3 +31,19 @@ const char* csr_name(int which) {
   }
   return "unknown-csr";
 }
+
+const char* frm_name(int which) {
+  switch (which) {
+    case 0:
+      return "rte";
```
aswaterman: `rne`

disasm/regnames.cc
```diff
@@ -31,3 +31,19 @@ const char* csr_name(int which) {
...
+    case 1:
+      return "rtz";
+    case 2:
+      return "rdn";
+    case 3:
+      return "rup";
+    case 4:
+      return "rmm";
+  }
```
aswaterman, line 49: Also add case 7, which should return `dyn`.

aswaterman (commented):
Thanks! LGTM after you address my comments.

disasm/regnames.cc
liuyu81: Sure

disasm/regnames.cc
liuyu81: Fixed

aswaterman (approved):

PR #2132, opened 2025-11-10
DMI NOP operation should never cause busy status

en-sc:
See RISC-V Debug Specification Version 1.0 [6.1.5. Debug Module Interface Access (`dmi`, at 0x11)], `op` field description:
> 0 (nop): Ignore data and address.
Don’t send anything over the DMI during Update-DR. This operation should never affect DMI busy or error status.


arromanoff:
LGTM

aswaterman (approved):
Merging on the basis of @arrv-sc's review.

PR #2134, opened 2025-11-10
Gate Zaamo,Zalrsc instructions behind corresponding extensions, not M…

sizn-sc:
…ISA A

This is already done for the disassembler (disasm/disasm.cc:882), but the actual implementations haven't been updated.

A already implies Zaamo,Zalrsc and the other way around (both enabled imply A). Makes the implementation slightly more correct for when only one standard extensions is enabled out of the two.

### DIFF
--- a/riscv/insns/amoadd_d.h
+++ b/riscv/insns/amoadd_d.h
@@ -1,3 +1,3 @@
-require_extension('A');
+require_extension(EXT_ZAAMO);
 require_rv64;
 WRITE_RD(MMU.amo<uint64_t>(RS1, [&](uint64_t lhs) { return lhs + RS2; }));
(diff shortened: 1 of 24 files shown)


aswaterman (commented):
LGTM, but can you also fix `ssamoswap.d` and `ssamoswap.w`?  Ping me afterwards and I'll merge.

sizn-sc:
@aswaterman Done, though I can't find the specifics in the spec (as of 20251030) and/or unified-db. Is that a spec oversight or am I reading it wrong?

aswaterman (approved):
It is probably a spec oversight, but it seems obvious, and the change is benign anyway.

cc @ved-rivos about ssamoswap being predicated on Zaamo.

PR #2135, opened 2025-11-11
Don't hang when invoked as background process

### LINKED ISSUE #2133: Spike hangs as a background process
When loading Spike as a background process, it will hang while waiting to become a foreground process.

"Background process" in the context of Linux, is a process in a group which is not the process group leader of a session. You can easily create a new PG in bash with ampersand:

```
$ hello &
[1] 15229
Hello, world!
```

As you can see, the `hello` program even if sent to the background, starts immediately, prints its result to stdout (which is shared for all PG's in a session), and exits.

However, Spike tries to change the terminal config, which hangs when the process is in the background:

https://github.com/riscv-software-src/riscv-isa-sim/blob/da86ac25be784c7ba51778e0f0f7370dbd8f3513/fesvr/term.cc#L17-L18

```
$ ./out/bin/spike --help &
[1] 15230

(spike doesn't start)
```

I can imagine the call to `tcsetattr` is done to configure the terminal for interactive use, but this is done even for scripted usage.

aswaterman:
It's (normally) only reasonable to launch Spike as a background process when it's used in a mode that does not need stdin.  But we should support those use cases without hanging.  Fix the hang by catching SIGTTOU.

For obvious reasons, `spike -d` will still hang waiting for input until it's in the foreground.

Fixes #2133

### DIFF
--- a/fesvr/term.cc
+++ b/fesvr/term.cc
@@ -1,9 +1,13 @@
 #include "term.h"
+#include "common.h"
 #include <termios.h>
 #include <unistd.h>
 #include <poll.h>
 #include <signal.h>
 #include <stdlib.h>
+#include <string.h>
+
+static int tcsetattr_ttou(int fd, int optional_actions, const struct termios *p);
 
 class canonical_termios_t
 {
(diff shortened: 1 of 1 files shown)


viperML:
Thank you for the swift response!

aswaterman:
You're welcome!

PR #2143, opened 2025-11-13
feat: support zvkb vector cryptography bitmanip subset extension

arromanoff:
This commit gates Zvkb behind Zvkb extension. Previously they were attached to Zvbb which is Zvkb's superset


aswaterman (approved):
LGTM, thanks for the contribution.

PR #2145, opened 2025-11-13
Make abstractauto register optional in debug module

nadime15:
Add --dm-no-abstractauto flag to disable support for the optional abstractauto register. When disabled, writes to the register are ignored and the internal state is kept at 0.

Currently, neither riscv-openocd nor openocd support batch reads or writes for debug modules that don’t implement abstractauto, as both assume every debug module provides abstractauto support. However, work is underway to remove this dependency.

### DIFF
--- a/riscv/debug_module.cc
+++ b/riscv/debug_module.cc
@@ -1101,10 +1101,16 @@ bool debug_module_t::dmi_write(unsigned address, uint32_t value)
         return true;
 
       case DM_ABSTRACTAUTO:
-        abstractauto.autoexecprogbuf = get_field(value,
-            DM_ABSTRACTAUTO_AUTOEXECPROGBUF);
-        abstractauto.autoexecdata = get_field(value,
-            DM_ABSTRACTAUTO_AUTOEXECDATA);
+        if (config.support_abstractauto) {
+          abstractauto.autoexecprogbuf = get_field(value,
+              DM_ABSTRACTAUTO_AUTOEXECPROGBUF);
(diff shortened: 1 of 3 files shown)


riscv/debug_module.cc
```diff
@@ -1101,10 +1101,16 @@ bool debug_module_t::dmi_write(unsigned address, uint32_t value)
...
+        if (config.support_abstractauto) {
+          abstractauto.autoexecprogbuf = get_field(value,
+              DM_ABSTRACTAUTO_AUTOEXECPROGBUF);
+          abstractauto.autoexecdata = get_field(value,
+              DM_ABSTRACTAUTO_AUTOEXECDATA);
+        }
+        else {
+          abstractauto.autoexecprogbuf = 0;
+          abstractauto.autoexecdata = 0;
+        }
```
aswaterman: Delete the `else` clause.  `abstractauto` is initialized to 0 at reset, so overwriting these fields with 0 has no effect.

LGTM otherwise; ping me when it's ready.

nadime15: True but dmcontrol.hasel is doing the same (dmcontrol is also initialized to 0 at reset). You want me to remove the `else` for dmcontrol.hasel as well?

aswaterman: Yeah, consistency is good. Thanks!

nadime15: Done

aswaterman (approved):

PR #2147, opened 2025-11-14
Two Zicfilp improvements

aswaterman:
- Fix an escape where if the I$ was incoherent with main memory, the landing-pad MSBs would not be checked. (For obvious reasons, this was only possible with Ziccid turned off.)
- Significantly speed up function returns by moving most of the ELP overhead off the critical code path.


aswaterman:
FYI @ved-rivos

ved-rivos:
LGTM!

PR #2148, opened 2025-11-15
Fix xtval when illegal instruction is triggered.

5265325:
The new refill_icache function has an implicit conversion bug. It result in the instruction sign expansion. When an illegal instruction exception is triggered, it manifests as an incorrect xtval. As shown in the following log:

```
core   0: 0x00007ff5942a0d80 (0xffffffffc2202573) csrr    a0, vlenb
core   0: exception trap_illegal_instruction, epc 0x00007ff5942a0d80
core   0:           tval 0xffffffffc2202573
```


aswaterman:
Yes, I just caught this myself too, and was going to use the same fix.

aswaterman (approved):

PR #2153, opened 2025-11-18
Enhance zve

chihminchao:
There are some part of V that don't be covered by zveXXX configuration.  The patch set wants to handle 

1. add a zve state to present the vector capability is enabled by V or Zve
2. handle vector widening and narrow conversion with difference floating width
3. handle  vector index load with zve64 and rv32
4. handle widening integer arithmetic which may not be supported on zve64

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -736,6 +736,7 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   const bool prev_h = old_misa & (1L << ('H' - 'A'));
   const reg_t new_misa = (adjusted_val & write_mask) | (old_misa & ~write_mask);
   const bool new_h = new_misa & (1L << ('H' - 'A'));
+  const bool new_v = proc->get_isa().has_any_vector();
 
   proc->set_extension_enable(EXT_ZCA, (new_misa & (1L << ('C' - 'A'))) || !proc->get_isa().extension_enabled('C'));
   proc->set_extension_enable(EXT_ZCF, (new_misa & (1L << ('F' - 'A'))) && proc->extension_enabled(EXT_ZCA) && proc->get_xlen() == 32);
@@ -745,8 +746,8 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   proc->set_extension_enable(EXT_ZCMT, proc->extension_enabled(EXT_ZCA));
   proc->set_extension_enable(EXT_ZFH, new_misa & (1L << ('F' - 'A')));
   proc->set_extension_enable(EXT_ZFHMIN, new_misa & (1L << ('F' - 'A')));
-  proc->set_extension_enable(EXT_ZVFH, (new_misa & (1L << ('V' - 'A'))) && proc->extension_enabled(EXT_ZFHMIN));
-  proc->set_extension_enable(EXT_ZVFHMIN, new_misa & (1L << ('V' - 'A')));
+  proc->set_extension_enable(EXT_ZVFH, new_v && proc->get_isa().get_zvf() && proc->extension_enabled(EXT_ZFHMIN));
+  proc->set_extension_enable(EXT_ZVFHMIN, new_v && proc->get_isa().get_zvf());
   proc->set_extension_enable(EXT_ZAAMO, (new_misa & (1L << ('A' - 'A'))) || !proc->get_isa().extension_enabled('A'));
   proc->set_extension_enable(EXT_ZALRSC, (new_misa & (1L << ('A' - 'A'))) || !proc->get_isa().extension_enabled('A'));
   proc->set_extension_enable(EXT_ZBA, (new_misa & (1L << ('B' - 'A'))) || !proc->get_isa().extension_enabled('B'));
--- a/riscv/insns/vfncvt_f_x_w.h
+++ b/riscv/insns/vfncvt_f_x_w.h
@@ -2,9 +2,9 @@
 VI_NON_ALTFMT_INSN
 
 VI_VFP_NCVT_INT_TO_FP(
-  { vd = i32_to_f16(vs2); },        // BODY32
-  { vd = i64_to_f32(vs2); },        // BODY64
-  { require_extension(EXT_ZVFH); }, // CHECK32
-  { require_extension('F'); },      // CHECK64
-  int                               // sign
+  { vd = i32_to_f16(vs2); },             // BODY32
+  { vd = i64_to_f32(vs2); },             // BODY64
+  { require_extension(EXT_ZVFH); },      // CHECK32
+  { require(p->get_isa().get_zvf()); },  // CHECK64
+  int                                    // sign
 )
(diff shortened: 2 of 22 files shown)


riscv/csrs.cc
```diff
@@ -741,8 +742,8 @@ bool misa_csr_t::unlogged_write(const reg_t val) noexcept {
   proc->set_extension_enable(EXT_ZCMT, proc->extension_enabled(EXT_ZCA));
   proc->set_extension_enable(EXT_ZFH, new_misa & (1L << ('F' - 'A')));
   proc->set_extension_enable(EXT_ZFHMIN, new_misa & (1L << ('F' - 'A')));
-  proc->set_extension_enable(EXT_ZVFH, (new_misa & (1L << ('V' - 'A'))) && proc->extension_enabled(EXT_ZFHMIN));
-  proc->set_extension_enable(EXT_ZVFHMIN, new_misa & (1L << ('V' - 'A')));
+  proc->set_extension_enable(EXT_ZVFH, new_v && proc->get_isa().get_zvf() && proc->extension_enabled(EXT_ZFHMIN));
+  proc->set_extension_enable(EXT_ZVFHMIN, new_v && proc->get_isa().get_zvf());
```
nadime15, line 750:
```suggestion
proc->set_extension_enable(EXT_ZVFHMIN, new_v && proc->get_isa().get_zvf());
proc->set_extension_enable(EXT_ZVFH, new_v && proc->get_isa().get_zvf() && proc->extension_enabled(EXT_ZFHMIN));
```
Does the order matter here? I think that if EXT_ZVFHMIN is currently disabled but you plan to enable it, trying to enable Zvfh first might fail, because Zvfh depends on Zvfhmin being enabled.

nadime15 (commented):
Two more things:

1.  
  Line 101 in (isa_parser.cc) needs to be changed too:
  `case 'v': vlen = 128; elen = 64; zvf = true; zvd = true; zve = true;`

2. 

> Note: If Zve32x is supported then Zvkb or Zvbb provide support for EEW of 8, 16, and 32. If Zve64x is
> supported then Zvkb or Zvbb also add support for EEW 64
> 
> All of the other Vector Crypto Extensions depend on

I have not checked whether Spike currently respects this, but the vector crypto extensions can be enabled via Zve32x and Zve64x.

nadime15:
By the way, we had a similar PR recently for the Sail RISC-V Model (see https://github.com/riscv/sail-riscv/pull/1170) that addressed all these cases, including Vector Crypto. It might be helpful, at least for cross-checking.

chihminchao:
> Two more things:
> 
> 
> Line 101 in (isa_parser.cc) needs to be changed too: `case 'v': vlen = 128; elen = 64; zvf = true; zvd = true; zve = true;`
> 
> 
> > Note: If Zve32x is supported then Zvkb or Zvbb provide support for EEW of 8, 16, and 32. If Zve64x is
> > supported then Zvkb or Zvbb also add support for EEW 64
> > All of the other Vector Crypto Extensions depend on
> 
> I have not checked whether Spike currently respects this, but the vector crypto extensions can be enabled via Zve32x and Zve64x.

1. I don't think it is necessary.  `zve` state is used to tell vector unit is enabled by 'v' or zveXXX.
 `v' includes zve64d but a small part of functionality are only available for 'v'
2. The description in zveXXX doesn't mention anything about vector crypto extension.    I think zveXXX doesn't affect vector crypto  behavior.

riscv/csrs.cc
chihminchao: Yes, you are correct  and the order matters.  I will fix it

chihminchao: @nadime15 correct my previous statement.

`EXT_ZVFH` depend on `EXT_ZFHMIN` but not `EXT_ZVFHMIN`

nadime15:
> I don't think it is necessary. zve state is used to tell vector unit is enabled by 'v' or zveXXX.
> `v' includes zve64d but a small part of functionality are only available for 'v'

The v extension is a superset of Zve64d. If Zve64d sets zve = true, then v should too. Currently, enabling only v leaves zve as false but zvf and zvd true, which seems inconsistent.

> The description in zveXXX doesn't mention anything about vector crypto extension. I think zveXXX doesn't affect vector crypto behavior.

Yes, you are right. It is not mentioned in the Zve section, but it appears in the vector crypto section.

> All of the other Vector Crypto Extensions depend on Zve32x

riscv/csrs.cc
nadime15: Ups, you are right, I read Ext_Zfhmin as Ext_Zvfhmin!

nadime15: This is unrelated to the PR itself, but I noticed this line:

  `proc->set_extension_enable(EXT_ZCF, (new_misa & (1L << ('F' - 'A'))) && proc->extension_enabled(EXT_ZCA));`

Since Zcf is RV32 only, shouldn this check xlen as well (unless its checked somewhere else) ? @aswaterman

aswaterman: > Since Zcf is RV32 only, shouldn this check xlen as well (unless its checked somewhere else) ?

I agree it should be changed to check XLEN == 32.  Please submit a separate PR for that one.

It isn't actually a bug right now, because, while this would cause `register_base_instructions` to do the wrong thing, `register_base_instructions` is only called during the construction of `processor_t`, before `misa` can be changed.  But if we ever add support for dynamically changing XLEN, then we'd need to reinvoke `register_base_instructions`, at which point this would become a problem.

aswaterman:
> The v extension is a superset of Zve64d. If Zve64d sets zve = true, then v should too. Currently, enabling only v leaves zve as false but zvf and zvd true, which seems inconsistent.

Since this is an internal implementation detail, we don't really need to adhere to the principle that V is a superset of Zve, as long as users of Spike can't tell the difference.

With that said, I agree this approach makes the code a bit confusing.  @chihminchao did you consider doing this the other way around: like, in `vmulh`, do `require(p->extension_enabled('V') || P.VU.vsew < e64);` instead of checking `get_zve()` and `ELEN`?  Then we could make V a proper superset of Zve, like @nadime15 suggested.

chihminchao:
> > The v extension is a superset of Zve64d. If Zve64d sets zve = true, then v should too. Currently, enabling only v leaves zve as false but zvf and zvd true, which seems inconsistent.
> 
> Since this is an internal implementation detail, we don't really need to adhere to the principle that V is a superset of Zve, as long as users of Spike can't tell the difference.
> 
> With that said, I agree this approach makes the code a bit confusing. @chihminchao did you consider doing this the other way around: like, in `vmulh`, do `require(p->extension_enabled('V') || P.VU.vsew < e64);` instead of checking `get_zve()` and `ELEN`? Then we could make V a proper superset of Zve, like @nadime15 suggested.

It is reasonable and let me check again

aswaterman:
Cool, ping me when it's time to re-review.

chihminchao:
The `zve` state has been removed by suggestion.

aswaterman (approved):
Yeah, I think this version is cleaner.  Thanks @chihminchao (and @nadime15 for making the suggestion).

PR #2161, opened 2025-11-25
Update trigger behavior for memory accesses to match recommended debug specification behavior

fkhaidari:
According to the debug specification (select bit description):

0 (address): There is at least one compare value and it contains the lowest virtual address of the access. In addition, it is recommended that there are additional compare values for the other accessed virtual addresses match. (E.g. on a 32-bit read from 0x4000, the lowest address is 0x4000 and the other addresses are 0x4001, 0x4002, and 0x4003.)

1 (data): There is exactly one compare value and it contains the data value loaded or stored, or the instruction executed. Any bits beyond the size of the data access will contain 0.

Previously, when select bit was 0, Spike did not follow the recommendation and provided only 1 matching value to the trigger module. This change modifies the behavior to the recommended one.

The implementation follows the debug specification recommendation.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -3,10 +3,14 @@
 #include "config.h"
 #include "mmu.h"
 #include "arith.h"
+#include "memif.h"
 #include "simif.h"
 #include "processor.h"
 #include "decode_macros.h"
 #include "platform.h"
+#include "triggers.h"
+
+#include <cassert>
 
 mmu_t::mmu_t(simif_t* sim, endianness_t endianness, processor_t* proc, reg_t cache_blocksz)
  : sim(sim), proc(proc), blocksz(cache_blocksz),
@@ -109,7 +113,8 @@ mmu_t::insn_parcel_t mmu_t::fetch_slow_path(reg_t vaddr)
   auto access_info = generate_access_info(vaddr, FETCH, {});
 
   if (check_triggers_fetch)
-    check_triggers(triggers::OPERATION_EXECUTE, vaddr, access_info.effective_virt);
+    check_triggers(triggers::OPERATION_EXECUTE, vaddr,
+      access_info.effective_virt, sizeof(insn_parcel_t));
 
   if (!tlb_hit) {
     paddr = translate(access_info, sizeof(insn_parcel_t));
@@ -128,13 +133,16 @@ mmu_t::insn_parcel_t mmu_t::fetch_slow_path(reg_t vaddr)
 
   auto res = perform_intrapage_fetch(vaddr, host_addr, paddr);
 
-  if (check_triggers_fetch)
-    check_triggers(triggers::OPERATION_EXECUTE, vaddr, access_info.effective_virt, from_le(res));
+  if (!check_triggers_fetch)
+    return res;
+
+  check_triggers(triggers::OPERATION_EXECUTE, vaddr,
+    access_info.effective_virt, sizeof(insn_parcel_t), from_le(res));
 
   return res;
 }
 
-reg_t reg_from_bytes(size_t len, const uint8_t* bytes)
+static reg_t reg_from_bytes(size_t len, const uint8_t* bytes)
 {
   switch (len) {
     case 1:
@@ -211,25 +219,41 @@ bool mmu_t::mmio(reg_t paddr, size_t len, uint8_t* bytes, access_type type)
   return true;
 }
 
-void mmu_t::check_triggers(triggers::operation_t operation, reg_t address, bool virt, reg_t tval, std::optional<reg_t> data)
+void mmu_t::check_triggers(triggers::operation_t operation,
+  reg_t addr, bool virt, std::size_t data_size, const std::uint8_t* bytes)
+{
+  assert(data_size > 0);
+  assert(data_size <= sizeof(reg_t));
+  check_triggers(operation, addr, virt,
+    data_size, reg_from_bytes(data_size, bytes));
+}
+
+void mmu_t::check_triggers(triggers::operation_t operation,
+  reg_t addr, bool virt, size_t access_len)
+{
+  check_triggers(operation, addr, virt, access_len, std::nullopt);
+}
+
+void mmu_t::check_triggers(triggers::operation_t operation, reg_t address, bool virt, std::size_t size, std::optional<reg_t> data)
 {
   if (matched_trigger || !proc)
     return;
 
-  auto match = proc->TM.detect_memory_access_match(operation, address, data);
+  auto match = proc->TM.detect_memory_access_match(operation, address, size, data);
+  if (!match.has_value())
+    return;
 
-  if (match.has_value())
-    switch (match->timing) {
-      case triggers::TIMING_BEFORE:
-        throw triggers::matched_t(operation, tval, match->action, virt);
+  switch (match->timing) {
+    case triggers::TIMING_BEFORE:
+      throw triggers::matched_t(operation, address, match->action, virt);
 
-      case triggers::TIMING_AFTER:
-        // We want to take this exception on the next instruction.  We check
-        // whether to do so in the I$ refill slow path, which we can force by
-        // flushing the TLB.
-        flush_tlb();
-        matched_trigger = triggers::matched_t(operation, tval, match->action, virt);
-    }
+    case triggers::TIMING_AFTER:
+      // We want to take this exception on the next instruction.  We check
+      // whether to do so in the I$ refill slow path, which we can force by
+      // flushing the TLB.
+      flush_tlb();
+      matched_trigger = triggers::matched_t(operation, address, match->action, virt);
+  }
 }
 
 inline void mmu_t::perform_intrapage_load(reg_t vaddr, uintptr_t host_addr, reg_t paddr, reg_t len, uint8_t* bytes, xlate_flags_t xlate_flags)
@@ -272,8 +296,11 @@ void mmu_t::load_slow_path_intrapage(reg_t len, uint8_t* bytes, mem_access_info_
   }
 }
 
-void mmu_t::load_slow_path(reg_t original_addr, reg_t len, uint8_t* bytes, xlate_flags_t xlate_flags)
+void mmu_t::load_slow_path(reg_t original_addr, std::size_t len,
+  std::uint8_t* bytes, xlate_flags_t xlate_flags)
 {
+  assert(len > 0);
+
   if (likely(!xlate_flags.is_special_access())) {
     // Fast path for simple cases
     auto [tlb_hit, host_addr, paddr] = access_tlb(tlb_load, original_addr, TLB_FLAGS & ~TLB_CHECK_TRIGGERS);
@@ -289,7 +316,8 @@ void mmu_t::load_slow_path(reg_t original_addr, reg_t len, uint8_t* bytes, xlate
   reg_t transformed_addr = access_info.transformed_vaddr;
 
   if (check_triggers_load)
-    check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt);
+    check_triggers(triggers::OPERATION_LOAD,
+      transformed_addr, access_info.effective_virt, len);
 
   if ((transformed_addr & (len - 1)) == 0) {
     load_slow_path_intrapage(len, bytes, access_info);
@@ -301,7 +329,7 @@ void mmu_t::load_slow_path(reg_t original_addr, reg_t len, uint8_t* bytes, xlate
     if (access_info.flags.lr)
       throw trap_load_access_fault(gva, transformed_addr, 0, 0);
 
-    reg_t len_page0 = std::min(len, PGSIZE - transformed_addr % PGSIZE);
+    reg_t len_page0 = std::min<reg_t>(len, PGSIZE - transformed_addr % PGSIZE);
     load_slow_path_intrapage(len_page0, bytes, access_info);
     if (len_page0 != len) {
       auto tail_access_info = generate_access_info(original_addr + len_page0, LOAD, xlate_flags);
@@ -309,16 +337,14 @@ void mmu_t::load_slow_path(reg_t original_addr, reg_t len, uint8_t* bytes, xlate
     }
   }
 
-  if (check_triggers_load) {
-    while (len > sizeof(reg_t)) {
-        check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt, reg_from_bytes(sizeof(reg_t), bytes));
-      len -= sizeof(reg_t);
-      bytes += sizeof(reg_t);
-    }
-    check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt, reg_from_bytes(len, bytes));
-  }
+  if (!proc)
+    return;
 
-  if (proc && unlikely(proc->get_log_commits_enabled()))
+  if (check_triggers_load)
+    check_triggers(triggers::OPERATION_LOAD,
+      transformed_addr, access_info.effective_virt, len, bytes);
+
+  if (unlikely(proc->get_log_commits_enabled()))
     proc->state.log_mem_read.push_back(std::make_tuple(original_addr, 0, len));
 }
 
@@ -359,7 +385,9 @@ void mmu_t::store_slow_path_intrapage(reg_t len, const uint8_t* bytes, mem_acces
     perform_intrapage_store(vaddr, host_addr, paddr, len, bytes, access_info.flags);
 }
 
-void mmu_t::store_slow_path(reg_t original_addr, reg_t len, const uint8_t* bytes, xlate_flags_t xlate_flags, bool actually_store, bool UNUSED require_alignment)
+void mmu_t::store_slow_path(reg_t original_addr, std::size_t len,
+    const std::uint8_t* bytes, xlate_flags_t xlate_flags,
+    bool actually_store, bool UNUSED require_alignment)
 {
   if (likely(!xlate_flags.is_special_access())) {
     // Fast path for simple cases
@@ -377,16 +405,9 @@ void mmu_t::store_slow_path(reg_t original_addr, reg_t len, const uint8_t* bytes
   auto access_info = generate_access_info(original_addr, STORE, xlate_flags);
   reg_t transformed_addr = access_info.transformed_vaddr;
 
-  if (actually_store && check_triggers_store) {
-    reg_t trig_len = len;
-    const uint8_t* trig_bytes = bytes;
-    while (trig_len > sizeof(reg_t)) {
-      check_triggers(triggers::OPERATION_STORE, transformed_addr, access_info.effective_virt, reg_from_bytes(sizeof(reg_t), trig_bytes));
-      trig_len -= sizeof(reg_t);
-      trig_bytes += sizeof(reg_t);
-    }
-    check_triggers(triggers::OPERATION_STORE, transformed_addr, access_info.effective_virt, reg_from_bytes(trig_len, trig_bytes));
-  }
+  if (actually_store && check_triggers_store)
+    check_triggers(triggers::OPERATION_STORE,
+      transformed_addr, access_info.effective_virt, len, bytes);
 
   if (transformed_addr & (len - 1)) {
     bool gva = access_info.effective_virt;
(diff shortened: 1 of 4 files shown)


fkhaidari:
@aswaterman, could you take a look?

riscv/mmu.cc
```diff
@@ -309,16 +309,21 @@ void mmu_t::load_slow_path(reg_t original_addr, reg_t len, uint8_t* bytes, xlate
...
 
+  if (!proc)
+    return;
+
   if (check_triggers_load) {
-    while (len > sizeof(reg_t)) {
-        check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt, reg_from_bytes(sizeof(reg_t), bytes));
-      len -= sizeof(reg_t);
-      bytes += sizeof(reg_t);
+    auto xlen_bytes = proc->get_xlen();
```
aswaterman: XLEN is measured in bits, not bytes.  You need to divide by 8.

riscv/mmu.cc
```diff
@@ -377,15 +382,18 @@ void mmu_t::store_slow_path(reg_t original_addr, reg_t len, const uint8_t* bytes
...
-  if (actually_store && check_triggers_store) {
+  if (actually_store && check_triggers_store && proc) {
     reg_t trig_len = len;
     const uint8_t* trig_bytes = bytes;
-    while (trig_len > sizeof(reg_t)) {
-      check_triggers(triggers::OPERATION_STORE, transformed_addr, access_info.effective_virt, reg_from_bytes(sizeof(reg_t), trig_bytes));
-      trig_len -= sizeof(reg_t);
-      trig_bytes += sizeof(reg_t);
+    auto trig_addr = transformed_addr;
+    auto xlen_bytes = proc->get_xlen();
```
aswaterman: Same comment as https://github.com/riscv-software-src/riscv-isa-sim/pull/2161/files#r2561412832

riscv/mmu.cc
```diff
@@ -309,16 +309,21 @@ void mmu_t::load_slow_path(reg_t original_addr, reg_t len, uint8_t* bytes, xlate
...
-    while (len > sizeof(reg_t)) {
-        check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt, reg_from_bytes(sizeof(reg_t), bytes));
-      len -= sizeof(reg_t);
-      bytes += sizeof(reg_t);
+    auto xlen_bytes = proc->get_xlen();
+    while (len > xlen_bytes) {
+        check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt, xlen_bytes, reg_from_bytes(xlen_bytes, bytes));
+      len -= xlen_bytes;
+      bytes += xlen_bytes;
+    transformed_addr += xlen_bytes;
```
aswaterman: wrong indentation

riscv/mmu.cc
```diff
@@ -309,16 +309,21 @@ void mmu_t::load_slow_path(reg_t original_addr, reg_t len, uint8_t* bytes, xlate
...
     }
-    check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt, reg_from_bytes(len, bytes));
+    check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt, len, reg_from_bytes(len, bytes));
   }
 
-  if (proc && unlikely(proc->get_log_commits_enabled()))
+  if (unlikely(proc->get_log_commits_enabled()))
```
aswaterman, line 347: this change is unrelated to the PR; undo it

riscv/mmu.cc
```diff
@@ -309,16 +309,21 @@ void mmu_t::load_slow_path(reg_t original_addr, reg_t len, uint8_t* bytes, xlate
...
+    return;
+
   if (check_triggers_load) {
-    while (len > sizeof(reg_t)) {
-        check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt, reg_from_bytes(sizeof(reg_t), bytes));
-      len -= sizeof(reg_t);
-      bytes += sizeof(reg_t);
+    auto xlen_bytes = proc->get_xlen();
+    while (len > xlen_bytes) {
+        check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt, xlen_bytes, reg_from_bytes(xlen_bytes, bytes));
```
aswaterman: This logic doesn't look right to me.  The total number of bytes reported to `check_triggers` will in general be `len` rounded up to the next multiple of XLEN/8, rather than simply `len`.  (Consider e.g. the case of the `lb` instruction, where `len` is 1.)

Don't you want to use `len` instead of `xlen_bytes` in both places?

riscv/mmu.cc
```diff
@@ -377,15 +382,18 @@ void mmu_t::store_slow_path(reg_t original_addr, reg_t len, const uint8_t* bytes
...
     reg_t trig_len = len;
     const uint8_t* trig_bytes = bytes;
-    while (trig_len > sizeof(reg_t)) {
-      check_triggers(triggers::OPERATION_STORE, transformed_addr, access_info.effective_virt, reg_from_bytes(sizeof(reg_t), trig_bytes));
-      trig_len -= sizeof(reg_t);
-      trig_bytes += sizeof(reg_t);
+    auto trig_addr = transformed_addr;
+    auto xlen_bytes = proc->get_xlen();
+    while (trig_len > xlen_bytes) {
+      check_triggers(triggers::OPERATION_STORE, trig_addr, access_info.effective_virt, xlen_bytes, reg_from_bytes(sizeof(reg_t), trig_bytes));
```
aswaterman: Same comment as https://github.com/riscv-software-src/riscv-isa-sim/pull/2161/files#r2561431576

riscv/triggers.cc
```diff
@@ -251,23 +251,25 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
...
   // We need this because in 32-bit mode sometimes the PC bits get sign
   // extended.
   auto xlen = proc->get_xlen();
-  if (xlen == 32) {
-    value &= 0xffffffff;
-  }
+  auto mask = make_mask64(0, xlen);
+  value &= mask;
+
+  for (std::size_t i = 0; i < len; ++i)
```
aswaterman: why in increments of 1 rather than words?

riscv/triggers.cc
```diff
@@ -251,23 +251,25 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
...
+    if (simple_match(xlen, (value + i) & mask)) {
```
aswaterman: I don't understand the logic here.  Why are we adding the current byte offset to the data value?

aswaterman (commented):
@en-sc Can you help review?

riscv/mmu.cc
fkhaidari: Oh, thanks. I've missed that. Addressed

riscv/mmu.cc
fkhaidari: Addressed

riscv/mmu.cc
fkhaidari: Addressed

riscv/mmu.cc
fkhaidari: I did it to avoid repetitions:

```c
  if (!proc)
    return;

  if (check_triggers_load) {
    auto xlen_bytes = proc->get_xlen() / 8;
    // ...
  }

  if (unlikely(proc->get_log_commits_enabled()))
    proc->state.log_mem_read.push_back(std::make_tuple(original_addr, 0, len));
```

Instead of

```c
  if (proc && check_triggers_load) {
    auto xlen_bytes = proc->get_xlen() / 8;
    // ...
  }

  if (proc && unlikely(proc->get_log_commits_enabled()))
    proc->state.log_mem_read.push_back(std::make_tuple(original_addr, 0, len));
```

But I may use 2nd variant if you believe it's better

riscv/mmu.cc
fkhaidari: > The total number of bytes reported to check_triggers will in general be len rounded up to the next multiple of XLEN/8

As I understand it would be equal to `len` exactly. This while loop handles `len` rounded down to the multiple of `XLEN / 8`. Tail is handled after while loop here:
```c
check_triggers(triggers::OPERATION_LOAD, transformed_addr, access_info.effective_virt, len, reg_from_bytes(len, bytes));
```

> Don't you want to use len instead of xlen_bytes in both places?

Here we check data triggers. Comparing loaded data with `tdata2`. Size of `tdata2` is `XLEN`. That's why I am iterating over read memory with stride equal `xlen_bytes`

Did I understand your question correct?

riscv/mmu.cc
fkhaidari: Check: https://github.com/riscv-software-src/riscv-isa-sim/pull/2161#discussion_r2562055851

riscv/triggers.cc
fkhaidari: For address triggers. This for loop checks every address in the accessed range

riscv/triggers.cc
fkhaidari: For data value `len == 1` so there is no problem here

```c
  reg_t value;
  if (select) {
    if (!data.has_value())
      return std::nullopt;
    value = *data;
    len = 1;
  } else {
    value = address;
  }
  ```

**Comment by fkhaidari** [comment]
@aswaterman, this MR may be a little bit confusing. I believe the problem in signature of the original function:
```c
void check_triggers(triggers::operation_t operation, reg_t address, bool virt, std::optional<reg_t> data = std::nullopt);
```

In this parameter to be precise: `std::optional<reg_t> data = std::nullopt`. If `data == std::nullopt` it behaves like address trigger checker and checks whole range. Otherwise it is data trigger checker. Probably it is better to improve this interface:

```c
void check_data_triggers(triggers::operation_t operation, reg_t address, bool virt, reg_t data);
void check_addr_triggers(triggers::operation_t operation, reg_t address, bool virt, reg_t tval);
```

Same for `std::optional<match_result_t> module_t::detect_memory_access_match(operation_t operation, reg_t address, std::optional<reg_t> data)` and `virtual std::optional<match_result_t> trigger_t::detect_memory_access_match(processor_t * const proc, operation_t operation, reg_t address, std::optional<reg_t> data)`.

So I decided to avoid this splitting. I may introduce it in one of the future MRs

riscv/mmu.cc
aswaterman: Ah, OK, I agree.

riscv/mmu.cc
aswaterman: OK, I understand.  I will leave it to @en-sc to review the semantics with that fact in mind.

aswaterman:
@fkhaidari You're probably right that it would be clearer to split `check_triggers` into two `check_data_triggers` and `check_addr_triggers`.

riscv/triggers.cc
aswaterman: That's very subtle.  I agree it appears to be correct, but it probably deserves a comment that the `+ i` only applies in the address case, for the reason you mentioned.

riscv/triggers.cc
```diff
@@ -238,7 +238,7 @@ bool mcontrol_common_t::simple_match(unsigned xlen, reg_t value) const {
   assert(0);
 }
 
-std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(processor_t * const proc, operation_t operation, reg_t address, std::optional<reg_t> data) noexcept {
+std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(processor_t * const proc, operation_t operation, reg_t address, size_t len, std::optional<reg_t> data) noexcept {
```
en-sc:
```suggestion
std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(processor_t * const proc, operation_t operation, reg_t address, std::size_t len, std::optional<reg_t> data) noexcept {
```
For consistency with the index in the loop.

en-sc:
> it would be clearer to split check_triggers into two check_data_triggers and check_addr_triggers.

@fkhaidari, @aswaterman, unfortunately this will not work.

A data trigger can be chained with an address trigger (e.g. a watchpoint on a store of a particular value, see [here](https://github.com/riscv-software-src/riscv-isa-sim/pull/2161/files#diff-c824bb2650fa95877b80dc95aa5ceed5c993b09b9c68e3805e55aa949fb194b1R397)).
For this to work, we will need for both, address and data, to be available when checking the chain.

riscv/triggers.cc
```diff
@@ -251,23 +251,25 @@ std::optional<match_result_t> mcontrol_common_t::detect_memory_access_match(proc
...
   }
 
   // We need this because in 32-bit mode sometimes the PC bits get sign
   // extended.
   auto xlen = proc->get_xlen();
-  if (xlen == 32) {
-    value &= 0xffffffff;
-  }
+  auto mask = make_mask64(0, xlen);
+  value &= mask;
```
en-sc: I'd suggest to either:
1. Move the masking into the `simple_match()`.
2. Add an assertion in the `simple_match()` that the value is masked.

riscv/triggers.cc
en-sc: Perhaps the call to the `simple_match()` can be moved into both branches of `if (select)` (with the loop in the `else` branch), dropping the `value` variable?
I believe this will make the code clearer and will eliminate the hack.

riscv/mmu.cc
en-sc: TBH, I'm not sure this code is correct.

AFAIU, this case can be triggered for a vector memory operation with SEW > XLEN. This case is governed by [5.5.2. Combined Accesses]:
> Some instructions lead a hart to perform multiple memory accesses. This includes vector loads and
stores, as well as cm.push and cm.pop instructions. The Trigger Module should match such accesses as
if they all happened individually. E.g. a vector load should be treated as if it performed multiple loads
of size SEW (selected element width), and cm.push should be treated as if it performed multiple stores
of size XLEN.

It seems like there should be a clarification made for the case of a data value being wider then XLEN. For now the `select` field description only covers sizes less or equal to XLEN.
> 1 (data): There is exactly one compare value and it contains
the data value loaded or stored, or the instruction
executed. Any bits beyond the size of the data access will
contain 0.

On a side note, do cache management operations come through here too? If so, this needs to be addressed. See [5.5.3. Cache Operations] for details. However, I'd suggest to add a TODO and address it sometime later.

en-sc: Please see https://github.com/riscv/riscv-debug-spec/issues/1183

aswaterman:
> For this to work, we will need for both, address and data, to be available when checking the chain.

Ah, right.

riscv/triggers.cc
fkhaidari: Clarified

riscv/mmu.cc
fkhaidari: Added todo

riscv/triggers.cc
fkhaidari: Addressed

riscv/triggers.cc
fkhaidari: Addressed

riscv/triggers.cc
fkhaidari: Addressed 1

fkhaidari:
@aswaterman, @en-sc, could you take a look once again?

aswaterman:
I'm short on time, so I'll defer to @en-sc's expertise.

riscv/mmu.cc
en-sc: @fkhaidari, please consider investigating when this loop is entered.
I feel like vector accesses are already split into elements at this point, though I'm not sure.
I would expect this code is relevant to handling of accesses with an element wider then xlen, e.g.:
1. FPU register store/load with D extension on a RV32 platform.
2. Vector register store/load with element width greater then XLEN (64 bits on RV32).

Please try running a simple binary that executes such accesses and observing when this loop is entered.

The concern is -- if the assumption is true (e.g. (1) enters the loop) the behavior will be erroneous.

riscv/triggers.cc
en-sc: LGTM

riscv/triggers.cc
en-sc: LGTM

aswaterman:
@en-sc I saw you approved of the various sub-changes.  Is the PR OK with you overall now?

en-sc:
> @en-sc I saw you approved of the various sub-changes. Is the PR OK with you overall now?

Not yet. I'm still waiting for a reply here: https://github.com/riscv-software-src/riscv-isa-sim/pull/2161#discussion_r2622832233

I'll post an overall comment outside of any thread once the change is LGTM.

riscv/mmu.cc
fkhaidari: > I feel like vector accesses are already split into elements at this point

You are right, vector insns are using this macro for memory access
```c
#define VI_LD(stride, offset, elt_width, is_mask_ldst) \
  const reg_t nf = insn.v_nf() + 1; \
  VI_CHECK_LOAD(elt_width, is_mask_ldst); \
  const reg_t vl = is_mask_ldst ? ((P.VU.vl->read() + 7) / 8) : P.VU.vl->read(); \
  const reg_t baseAddr = RS1; \
  const reg_t vd = insn.rd(); \
  for (reg_t i = 0; i < vl; ++i) { \
    VI_ELEMENT_SKIP; \
    VI_STRIP(i); \
    P.VU.vstart->write(i); \
    for (reg_t fn = 0; fn < nf; ++fn) { \
      elt_width##_t val = MMU.load<elt_width##_t>( \
        baseAddr + (stride) + (offset) * sizeof(elt_width##_t)); \
      P.VU.elt<elt_width##_t>(vd + fn * emul, vreg_inx, true) = val; \
    } \
  } \
  P.VU.vstart->write(0);
```

Also I have reworked this PR a little so now element size is passed explicitly and trigger doesn't try to match `elem_size > xlen`

riscv/mmu.h
```diff
@@ -217,7 +220,7 @@ class mmu_t
 
     auto base = transformed_addr & ~(blocksz - 1);
     for (size_t offset = 0; offset < blocksz; offset += 1) {
-      check_triggers(triggers::OPERATION_STORE, base + offset, false, transformed_addr, std::nullopt);
+      check_triggers(triggers::OPERATION_STORE, base + offset, false, 1);
```
en-sc: Please, consider moving the call out of the loop.
```
check_triggers(triggers::OPERATION_STORE, base + offset, false, blocksz);
for (size_t offset = 0; offset < blocksz; offset += 1) {
   ...
```
See [5.5.3. Cache Operations] for the reference.

riscv/mmu.h
```diff
@@ -228,7 +231,7 @@ class mmu_t
 
     auto base = transformed_addr & ~(blocksz - 1);
     for (size_t offset = 0; offset < blocksz; offset += 1)
-      check_triggers(triggers::OPERATION_STORE, base + offset, false, transformed_addr, std::nullopt);
+      check_triggers(triggers::OPERATION_STORE, base + offset, false, 1);
```
en-sc: Same as for `cbo_zero()`.

riscv/mmu.cc
```diff
@@ -211,25 +218,40 @@ bool mmu_t::mmio(reg_t paddr, size_t len, uint8_t* bytes, access_type type)
   return true;
 }
 
-void mmu_t::check_triggers(triggers::operation_t operation, reg_t address, bool virt, reg_t tval, std::optional<reg_t> data)
+void mmu_t::check_triggers(triggers::operation_t operation,
+  reg_t addr, bool virt, std::size_t data_size, const std::uint8_t* bytes)
+{
+  assert(data_size > 0);
```
en-sc, line 225: Please, add an assertion that `data_size` is less or equal to `sizeof(reg_t)`

en-sc (changes requested):

riscv/mmu.h
fkhaidari: Addressed

riscv/mmu.h
fkhaidari: Addressed

riscv/mmu.cc
fkhaidari: Addressed

en-sc: LGTM

en-sc (approved):

riscv/mmu.h
en-sc: LGTM

riscv/mmu.h
en-sc: LGTM

riscv/mmu.cc
en-sc: LGTM

en-sc:
LGTM
@aswaterman, I don't have any further concerns. The change seems solid.

aswaterman (approved):
@fkhaidari Thanks again for the PR, and thank you for taking the time to shepherd it, @en-sc.

PR #2162, opened 2025-11-26
zvfofp4min: fix shared contraint with v[zs]ext

chihminchao:
Some rules are not common.


aswaterman (approved):

aswaterman (changes requested):
Actually, I think the `require(((float)P.VU.vflmul / div) >= 0.125 && ((float)P.VU.vflmul / div) <= 8 );` should remain common code.  I agree that the `require(from >= e8 && from <= e64);` needs to be factored out since `SEW / div` equals 4 in the vfext case.

chihminchao:
follow the suggestion to refine the common part.

aswaterman (approved):

aswaterman:
Thanks!

PR #2163, opened 2025-11-26
Fix double trap

chihminchao:


chihminchao:
@ved-rivos  You may be interested.

riscv/csrs.cc
```diff
@@ -544,11 +544,16 @@ mstatus_csr_t::mstatus_csr_t(processor_t* const proc, const reg_t addr):
   val(compute_mstatus_initial_value()) {
 }
 
+reg_t mstatus_csr_t::read() const noexcept {
+  return val & ((state->menvcfg->read() & MENVCFG_DTE) ? ~0ul : ~MSTATUS_SDT);
```
aswaterman: This does the wrong thing if `long` is 32 bits, as it is on a 32-bit host.  It'll clear the 32 MSBs.  I'd rewrite it as `return val & ~reg_t(state->menvcfg->read() & MENVCFG_DTE ? 0 : MSTATUS_SDT);`

riscv/csrs.cc
```diff
@@ -544,11 +544,16 @@ mstatus_csr_t::mstatus_csr_t(processor_t* const proc, const reg_t addr):
...
+}
+
 bool mstatus_csr_t::unlogged_write(const reg_t val) noexcept {
   const bool has_mpv = proc->extension_enabled('H');
   const bool has_gva = has_mpv;
+  const reg_t adj_write_mask = sstatus_write_mask & ((state->menvcfg->read() & MENVCFG_DTE) ? ~0ul : ~SSTATUS_SDT);
```
aswaterman: Same thing here, `sstatus_write_mask & ~reg_t(state->menvcfg->read() & MENVCFG_DTE ? 0 : SSTATUS_SDT)`

aswaterman (commented):
Thanks for splitting the commits to make this PR easy to review.

Ping me once you've addressed my comments and I'll merge it.  (@ved-rivos is on vacation)

riscv/csrs.cc
chihminchao: If all the world was 64bits, what should we have to write.

fix by update

riscv/csrs.cc
chihminchao: fixed by update

chihminchao:
@aswaterman It is ready

riscv/csrs.cc
aswaterman: > If all the world was 64bits

That would be nice...

aswaterman (approved):

PR #2166, opened 2025-11-29
Fix VS-mode check for sireg* (really vsireg*) CSRs

Steven-Li-Xiaogang:
From VS-mode to access sireg* (really vsireg*) will launch permission verification from 'virtualized_indirect_csr_t', instruction code's higher 12-bits should be with value 'sireg*' CSRs, then insn.csr() could not match CSR_VSIREG~CSR_VSIREG6


aswaterman (commented):
I'm deferring review of this one to @mmhus, who contributed the code originally

mmhus:
LGTM. I didn't check everything extensively for the 'H' extension previously.

aswaterman (approved):

PR #2169, opened 2025-11-30
ci: auto-generate per extension snippy tests

arromanoff:
This way adding new extension tests is as easy as adding 1 line script call and it is much harder to leave some instruction untested

This commit removes predefined configs from ci and instead auto-generates them with all of the instructions via `llvm-ie` tool from [snippy-2.1 release](https://github.com/syntacore/snippy/releases/tag/snippy-2.1). This makes it much easier to add tests for new extensions. I added few more tests to prove this (such as Zc* compressed subset extensions and zfh). I also increased timeout for the tests as 5s was too strict on some machines. But no major impact on CI runtime was measured.


arromanoff:
I also wanted to add tests for zalrsc/zaamo extensions but the cross-compiler used doesn't seem to support them. Maybe later

aswaterman:
@arromanoff I'd like to continue using the packaged cross-compiler, even though it's older.  In the meantime, testing only the `A` extension gets us almost all of the benefit of testing `Zaamo` and `Zalrsc` indepenently--the only thing it fails to test is whether the instructions are correctly enabled for a given ISA string.

aswaterman (approved):
This is a nice improvement.

PR #2176, opened 2025-12-06
Fix Q extension on big-endian targets

aswaterman:
The previous routines always stored the low-order bits into the  low-order doubleword, which is only correct on little-endian targets.

@ved-rivos can you sanity-check this?  I spotted it after seeing your comment here: https://github.com/riscv/riscv-isa-manual/issues/2447#issuecomment-3616644330

The idea is to make FLQ/FSQ look more like AMOCAS.Q etc. rather than a special case.


ved-rivos:
@aswaterman looks right to me.

aswaterman:
Thank you!

PR #2180, opened 2025-12-10
softfloat: Add f32_to_i8 and f32_to_ui8 function declarations

mslijepc:
Definitions already exist in `f32_to_i8.c` and `f32_to_ui8.c`, I just put declarations to `softfloat.h`

### DIFF
--- a/softfloat/softfloat.h
+++ b/softfloat/softfloat.h
@@ -221,6 +221,8 @@ bfloat16_t bf16_recip7( bfloat16_t );
 /*----------------------------------------------------------------------------
 | 32-bit (single-precision) floating-point operations.
 *----------------------------------------------------------------------------*/
+int_fast8_t f32_to_i8( float32_t, uint_fast8_t, bool );
+uint_fast8_t f32_to_ui8( float32_t, uint_fast8_t, bool );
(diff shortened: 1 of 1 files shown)


softfloat/softfloat.h
```diff
@@ -221,6 +221,8 @@ bfloat16_t bf16_recip7( bfloat16_t );
 /*----------------------------------------------------------------------------
 | 32-bit (single-precision) floating-point operations.
 *----------------------------------------------------------------------------*/
+int_fast8_t f32_to_i8( float32_t a, uint_fast8_t roundingMode, bool exact );
+uint_fast8_t f32_to_ui8( float32_t a, uint_fast8_t roundingMode, bool exact );
```
aswaterman: It looks like the softfloat style is to omit the argument names in the function prototypes.  Can you change these to match the ones below?  I'll approve afterwards.

mslijepc: Sorry, my bad. fixed.

aswaterman: No prob, thanks for fixing it.

aswaterman (approved):

aswaterman:
CI was broken for unrelated reasons, but it's green now.

PR #2181, opened 2025-12-11
Fix RVV element 0 checking on load fault-only-first

Steven-Li-Xiaogang:
element 0 is not vstart=0, should be vstart value before executing

### DIFF
--- a/riscv/v_ext_macros.h
+++ b/riscv/v_ext_macros.h
@@ -1325,8 +1325,8 @@ VI_VX_ULOOP({ \
         val = MMU.load<elt_width##_t>( \
           baseAddr + (i * nf + fn) * sizeof(elt_width##_t)); \
       } catch (trap_t& t) { \
-        if (i == 0) \
-          throw; /* Only take exception on zeroth element */ \
+        if (i == p->VU.vstart->read()) \
+          throw; /* Only take exception on element 0 */ \
         /* Reduce VL if an exception occurs on a later element */ \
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
That's not right.  The fault-only-first loads are defined to only raise an exception detected on element 0.  If an element is detected on element vstart (and vstart > 0), the correct behavior is to reduce vl to vstart and reset vstart to 0 (which is what Spike does).

Steven-Li-Xiaogang:
> That's not right. The fault-only-first loads are defined to only raise an exception detected on element 0. If an element is detected on element vstart (and vstart > 0), the correct behavior is to reduce vl to vstart and reset vstart to 0 (which is what Spike does).

So element 0 does not indicate first active elements, RVV spec is not clear to indicate element 0 with constraint vstart=0.

aswaterman:
Element 0 always means element 0, rather than the first active element.  This follows from the text in this section: https://github.com/riscv/riscv-isa-manual/blob/9bb5e51fc9eb95e9857c4f3e6e0031e74915af89/src/v-st-ext.adoc#prestart-active-inactive-body-and-tail-element-definitions

Note that all of the functions are indexed by the element number: e.g. if vstart is 2, vl is 3, and the instruction is not masked, then `prestart(0)=1`, `prestart(1)=1`, and `active(2)=1`.  In other words, the first active element is named element 2; element 0 refers to a prestart element.

Issue #2082, opened 2025-09-11
Sireg raises an virtual instruction exception when mstateen[60] = 0 in VU mode


wwwang1111:
**In priv-isa-asciidoc_20240411:**
If extension Smstateen is implemented together with Smcsrind, bit 60 of state-enable register
mstateen0 controls access to siselect, sireg*, vsiselect, and vsireg*. **When mstateen0[60]=0, an
attempt to access one of these CSRs from a privilege mode less privileged than M-mode results in an
illegal instruction exception**. As always, the state-enable CSRs do not affect the accessibility of any
state when in M-mode, only in less privileged modes. For more explanation, see the documentation for
extension Smstateen.
**But in SPIKE in VU mode:**
Sireg raises an virtual instruction exception due to the code in virtualized_indirect_csr_t::verify_permissions.
```
void virtualized_indirect_csr_t::verify_permissions(insn_t insn, bool write) const {
  virtualized_csr_t::verify_permissions(insn, write);
  if (state->v)
    virt_csr->verify_permissions(insn, write);
  else
    orig_csr->verify_permissions(insn, write);
}
```
**Because  virtualized_csr_t's parent class is csr_t, so the verify_permissions(insn, write) as follows. In VU mode, because VU < S，sireg raises an virtual instruction exception. But siselect does not have this issue.**
```
  if (priv < csr_priv) {
    if (state->v && csr_priv <= PRV_HS)
      throw trap_virtual_instruction(insn.bits());
    throw trap_illegal_instruction(insn.bits());
  }
```

aswaterman:
@mmhus since you have worked on this code recently, could you help resolve this issue?

cc @binno @chihminchao

wwwang1111:
The annotation mentions this point, but the actual approach is different. And the permission check of csr_t is repeated.&nbsp;Accessed sireg in VU mode is actually vsireg, insn.csr() is sireg but address is vsireg's address, so&nbsp;this check will not be conducted.The problem is that the order of check has been reversed.
But in Siselect, the order of this check is correct.




&nbsp;


1972043829
***@***.***



&nbsp;




------------------&nbsp;原始邮件&nbsp;------------------
发件人: "Binno ***@***.***&gt;; 
发送时间: 2025年9月17日(星期三) 中午11:15
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
[...]

PR #2182, opened 2025-12-11
#2082 Reorder hstateen0 check for smcdeleg

mmhus:
As mentioned in the following issue, the hstateen0 check should come before verifying base csr_t permissions:
https://github.com/riscv-software-src/riscv-isa-sim/issues/2082

/cc @binno


aswaterman (approved):

aswaterman:
CI was broken for unrelated reasons, but it's green now.

@binno this looks right to me, but if you have any follow-ups, you can reopen #2082.

PR #2186, opened 2025-12-17
fix infinite loop due to address overflow in pmp checking routine

aap-sc:
Before this patch, the code checked addresses as follows: the input address was aligned to the PMP granule boundary, and then it iterated with granule-based increments over the referenced memory chunks. The problem was that when the memory reference corresponded to the last possible granule, the iteration process never ended due to a uint64_t overflow to zero during the address increment.

The patch fixes this problem by changing the iteration from being based on address space to being based on the indexes of the affected granules. This approach also allows us to handle the case where the access wraps around zero (this patch addresses only RV64 case, though)

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -8,6 +8,8 @@
 #include "decode_macros.h"
 #include "platform.h"
 
+#include <limits>
+
 mmu_t::mmu_t(simif_t* sim, endianness_t endianness, processor_t* proc, reg_t cache_blocksz)
  : sim(sim), proc(proc), blocksz(cache_blocksz),
 #ifdef RISCV_ENABLE_DUAL_ENDIAN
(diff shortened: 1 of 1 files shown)


aap-sc:
@aswaterman could you take a look, please?

Current TOT version of spike will hang on the following test case.

- spike.ld
```
ENTRY ( _entry )
SECTIONS { .text 0x80000000 :  { *(.text) } }
```

- test.S
```
.global _entry
.global write_done
.global tohost
.global fromhost

.text
_entry:
  li t2, 0xffffffffffffffff
  sb t2, 0(t2)

  li x1, 1
  la t1, tohost
  sw x1, 0(t1)
1:
  j 1b

.balign 64
tohost:
  .dword 0
.balign 64
fromhost:
  .dword 0
```

- compilation command

```
riscv64-unknown-elf-gcc -march=rv64gc test.S -nostdlib -nostartfiles -T spike.ld -o test.elf
```

- spike invocation:

```
./spike --isa=rv64gc -m0x80000000:0xffffffff80000000 -l test.elf
```

the log:

```
core   0: 0x0000000000001000 (0x00000297) auipc   t0, 0x0
core   0: 0x0000000000001004 (0x02028593) addi    a1, t0, 32
core   0: >>>>  $xrv64i2p1_m2p0_a2p1_f2p2_d2p2_c2p0_zicsr2p0_zifencei2p0_zmmul1p0_zaamo1p0_zalrsc1p0
core   0: 0x0000000080000000 (0x000053fd) c.li    t2, -1
core   0: 0x0000000080000002 (0x00738023) sb      t2, 0(t2
>>> HANG HAPPENS HERE <<<
```

The reason for the hang is described in the commit message/commit description.

aap-sc:
Few additional notes:

- I can add a directed tests if desired to CI. It's just I'm not sure that it is appropriate given that existing tests are more like C programs. Here we'll need to dereference 0xfff....ffff. 
- The commit message states that we can properly handle a case when physical address wraps around zero for RV64. RV32 case is not handled. I can try to handle this case either as part of this MR or in separate one. I'm not sure that this is necessary/desired, though.

Please let me know if these improvements are welcome/desired.

aswaterman:
Oops.  Thanks for catching this and providing a failing test case.  I prefer a simpler fix that I'll PR shortly.  Not sure a unit test dedicated to this case is worth the effort, but if it's part of a bigger test for PMP corner cases in general, I'd certainly appreciate that contribution.

aswaterman:
Fixed by #2187

PR #2191, opened 2025-12-24
indirection CSRs 'iprio0~iprio15' are defined by Smaia/Ssaia extensions

Steven-Li-Xiaogang:
1. register indirection CSRs 'iprio0~iprio15' when Smaia/Ssaia supported
2. Smais/Ssaia requires Smcsrind/Sscsrind with mireg/sirge/vsireg supported but Smcsrind/Sscsrind does not indicates Smaia/Ssaia extensions at the same time

### DIFF
--- a/riscv/csr_init.cc
+++ b/riscv/csr_init.cc
@@ -12,7 +12,7 @@ void state_t::add_csr(reg_t addr, const csr_t_p& csr)
 #define add_supervisor_csr(addr, csr) add_const_ext_csr('S', addr, csr)
 #define add_hypervisor_csr(addr, csr) add_ext_csr('H', addr, csr)
 
-void state_t::add_ireg_proxy(processor_t* const proc, sscsrind_reg_csr_t::sscsrind_reg_csr_t_p ireg)
+void state_t::add_iprio_proxy(processor_t* const proc, sscsrind_reg_csr_t::sscsrind_reg_csr_t_p ireg)
 {
   // This assumes xlen is always max_xlen, which is true today (see
   // mstatus_csr_t::unlogged_write()):
@@ -421,7 +421,8 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
 
     sscsrind_reg_csr_t::sscsrind_reg_csr_t_p mireg;
     add_csr(CSR_MIREG, mireg = std::make_shared<sscsrind_reg_csr_t>(proc, CSR_MIREG, miselect));
-    add_ireg_proxy(proc, mireg);
+    if (proc->extension_enabled_const(EXT_SMAIA))
+      add_iprio_proxy(proc, mireg);
     const reg_t mireg_csrs[] = { CSR_MIREG2, CSR_MIREG3, CSR_MIREG4, CSR_MIREG5, CSR_MIREG6 };
     for (auto csr : mireg_csrs)
       add_csr(csr, std::make_shared<sscsrind_reg_csr_t>(proc, csr, miselect));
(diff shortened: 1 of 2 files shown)


aswaterman (commented):
@mmhus can you review this one?

riscv/processor.h
```diff
@@ -77,7 +77,7 @@ typedef std::vector<std::tuple<reg_t, uint64_t, uint8_t>> commit_log_mem_t;
 // architectural state of a RISC-V hart
 struct state_t
 {
-  void add_ireg_proxy(processor_t* const proc, sscsrind_reg_csr_t::sscsrind_reg_csr_t_p ireg);
+  void add_iprio_proxy(processor_t* const proc, sscsrind_reg_csr_t::sscsrind_reg_csr_t_p ireg);
```
mmhus, line 80: @Steven-Li-Xiaogang 

I do not agree with the function name change As iprio is smaia/ssaia specific, 

1. These proxy registers are also used in the smcdeleg & ssccfg extensions as regs and not iprio. 
2. The `add_ireg_proxy` function is used both for sireg and mireg, not just mireg.
3. Additionally, these registers are also used as proxies for the IMSIC if you were to implement that in your own version of spike or if it is implemented in upstream at some point.

mmhus (changes requested):

riscv/csr_init.cc
```diff
@@ -438,7 +439,8 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
     add_hypervisor_csr(CSR_VSIREG, vsireg);
 
     auto sireg = std::make_shared<sscsrind_reg_csr_t>(proc, CSR_SIREG, siselect);
-    add_ireg_proxy(proc, sireg);
+    if (proc->extension_enabled_const(EXT_SSAIA))
+      add_iprio_proxy(proc, sireg);
```
mmhus: See https://github.com/riscv-software-src/riscv-isa-sim/pull/2191/files#r2657880442

mmhus:
Overall I think you may be thinking these are AIA iprio arrays but these are not iprio arrays, but rather smcdeleg. AIA iprio arrays don't seem to be implemented yet in upstream.

riscv/processor.h
Steven-Li-Xiaogang: 1. original `add_ireg_proxy` in `state_t`  is an encapsulation (attached `iprio` indirect CSRs onto `mireg` and `sireg`, where there is the same name function in class `sscsrind_reg_csr_t`
2. further the` iprio` indirect CSRs have index from 0x30 to 0x3F, which are only defined in AIA spec and under the control of bit 59 of mstateen0 and hstateen0

Steven-Li-Xiaogang:
> Overall I think you may be trying to add AIA iprio arrays but these are not iprio arrays. Those don't seem to be implemented yet in upstream.

1. upstream code has already registered `iprio` registers onto `mireg` and `sireg`, but indirect CSRs 0x30-0x3F are assigned to `smaia/ssaia` extensions.
2. upstream does not have a complete AIA features (IMSIC is not implemented), but it implements some CSRs (`hviprio1, hviprio2, vstopi .etc`) defined in AIA which is controlled by bit 59 and 58 of `mstateen0` and `sstateen0`.
3. I think `iprio` and interrupt file registers should be treated as inaccessible.


> Bit 59 controls access to AIA CSRs siph, sieh, stopi, hidelegh, hvien/hvienh, hviph,
> hvictl, hviprio1/hviprio1h, hviprio2/hviprio2h, vsiph, vsieh, and vstopi, as well as to
> the supervisor-level interrupt priorities accessed through siselect + sireg (the iprio array of
> Section 5.4.1).

> Bit 58 is implemented in mstateen0 only if the hart has an IMSIC. If the hypervisor extension is
> also implemented, this bit does not affect the behavior or accessibility of hypervisor CSRs hgeip
> and hgeie, or field VGEIN of hstatus. In particular, guest external interrupts from an IMSIC
> continue to be visible to HS-mode in hgeip even when bit 58 of mstateen0 is zero.

mmhus:
I think you are correct. The function needs renaming after I took a look.

riscv/processor.h
mmhus: I see that I was confusing `state_t::`
 and `sscsrind_reg_csr_t::`

mmhus (approved):
LGTM

aswaterman (approved):
Thanks @Steven-Li-Xiaogang and @mmhus

PR #2193, opened 2025-12-24
added support for zicclsm

mslijepc:


riscv/mmu.h
```diff
@@ -351,7 +351,7 @@ class mmu_t
 
   int is_misaligned_enabled()
   {
-    return proc && proc->get_cfg().misaligned;
+    return proc && (proc->extension_enabled(EXT_ZICCLSM) || proc->get_cfg().misaligned);
```
aswaterman: I would like to get rid of `cfg_t::misaligned` as part of this PR.  Yes, it is a change to the API, but it is best not to have two ways to do this thing.

mslijepc: make sense. now when extension exist users can add `ZICCLSM` to ISA instead of manual change to `cfg`

riscv/mmu.h
```diff
@@ -351,7 +351,7 @@ class mmu_t
 
   int is_misaligned_enabled()
   {
-    return proc && proc->get_cfg().misaligned;
+    return proc && proc->extension_enabled(EXT_ZICCLSM);
```
mslijepc, line 354: does it make sense to add `private` member `bool processor_t::is_misaligned_enabled` that we don't check this every time is called? 
`is_misaligned_enabled` should be initialized in constructor body

aswaterman: I think it is OK as-is.  `is_misaligned_enabled` is not on a critical performance path, and it is not very expensive to evaluate.

aswaterman (approved):
This version LGTM, thanks for the contribution.

PR #2198, opened 2025-12-27
Fix triggers for accesses wider than XLEN

aswaterman:
I believe that @en-sc's comment here is correct: https://github.com/riscv-software-src/riscv-isa-sim/pull/2161#discussion_r2564958203
    
Nevertheless, failing an assertion when someone sets a trigger on memory accessed by a wide access is not reasonable behavior for Spike.  Better to do something that follows the principle of least surprise, despite the debug spec's lack of clarity on this point.

This PR also fixes some unrelated code-quality issues in adjacent code.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -144,28 +144,12 @@ mmu_t::insn_parcel_t mmu_t::fetch_slow_path(reg_t vaddr)
 
 static reg_t reg_from_bytes(size_t len, const uint8_t* bytes)
 {
-  switch (len) {
-    case 1:
-      return bytes[0];
-    case 2:
-      return bytes[0] |
-        (((reg_t) bytes[1]) << 8);
-    case 4:
-      return bytes[0] |
-        (((reg_t) bytes[1]) << 8) |
-        (((reg_t) bytes[2]) << 16) |
-        (((reg_t) bytes[3]) << 24);
-    case 8:
-      return bytes[0] |
-        (((reg_t) bytes[1]) << 8) |
-        (((reg_t) bytes[2]) << 16) |
-        (((reg_t) bytes[3]) << 24) |
-        (((reg_t) bytes[4]) << 32) |
-        (((reg_t) bytes[5]) << 40) |
-        (((reg_t) bytes[6]) << 48) |
-        (((reg_t) bytes[7]) << 56);
-  }
-  abort();
+  assert(len <= sizeof(reg_t));
+
+  size_t res = 0;
+  for (size_t i = 0; i < len; i++)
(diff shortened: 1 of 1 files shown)


riscv/mmu.cc
```diff
@@ -223,9 +207,12 @@ void mmu_t::check_triggers(triggers::operation_t operation,
...
   assert(data_size > 0);
-  assert(data_size <= sizeof(reg_t));
-  check_triggers(operation, addr, virt,
-    data_size, reg_from_bytes(data_size, bytes));
+
+  for (size_t offset = 0; offset < data_size; offset += sizeof(reg_t)) {
+    auto this_size = std::min(data_size - offset, sizeof(reg_t));
+    auto this_data = reg_from_bytes(this_size, bytes + offset);
+    check_triggers(operation, addr + offset, virt, this_size, this_data);
+  }
```
en-sc: Please, consider the following alternative:
```suggestion
  check_triggers(operation, addr, virt, data_size, reg_from_bytes(std::min(data_size, sizeof(reg_t)), bytes));
```

The motivation is:
1. AFAIU, an access wider then `sizeof(reg_t)` (please note, this is not XLEN) is already split into chunks at this point e.g.:
https://github.com/riscv-software-src/riscv-isa-sim/blob/b92958a9f1689776d90364b9db25920b394b1b3f/riscv/v_ext_macros.h#L1267-L1271
This is why I think the assertion is valid.
2. However, if there is such an access, an alternative aproach that checks a number of "chunks" that are no wider then the size of `reg_t` breaks address matching.
Consider the following:
* 16-byte-wide access on address 0x0.
* There are two triggers in a chain:
  - The first checks whether an address less then 0x1 is accessed.
  - The second checks whether an address greater or equal to 0x8 is accessed.
* If the access is split into two 8-byte wide checks, the chain will not match (the first "chunk" matches the first trigger, the second "chunk" matches the second trigger, neither matches the chain as a whole).

While the scenario is synthetic, IMHO the new behavior better complies with the spec recommendation for address match triggers (`select` is `0 (address)`) to consider all the virtual addresses being accessed.

en-sc (changes requested):

riscv/mmu.cc
aswaterman: @en-sc AMOCAS.Q, FLQ, and FSQ perform accesses wider than `sizeof(reg_t)`.  They will fail the assertion.  With that in mind, what do you recommend?

aswaterman: Ah, I see.  I can apply your suggestion without reinstating the assertion.  I guess that works for me.  Seems like an oversight in the trigger spec to not consider atomic accesses greater than XLEN bits, though.

PR #2199, opened 2025-12-30
Add support for per-hart start PCs (--pcs)

nabudahab:
### Summary
This PR introduces a new command-line argument `--pc-harts` to allow specifying different start addresses (PCs) for different harts. This feature is particularly valuable for RTL co-simulation workflows, where the simulator must exactly match the behavior of the RTL design. In these scenarios, the Boot ROM is often bypassed or disabled, and each hart is released from reset at a specific entry point defined by the hardware configuration.

### Problem
Currently, Spike defaults to setting the PC for all harts to the ELF entry point or the address specified by `--pc`.

However, in many bare-metal verification environments and Asymmetric Multi-Processing (AMP) configurations, harts often have distinct reset vectors. For example, a dual-core system might require:
- Hart 0 to fetch from `0x2000`
- Hart 1 to fetch from `0x4000`

Achieving this previously required modifying the C++ source code or relying on a complex bootloader/bootrom to diverge the harts.

### Solution
I have added the `--pc-harts` flag which parses a comma-separated list of `HartID:Address` pairs.

**Usage Example:**
`spike -p2 --pc-harts=0:0x0,1:0x200 my_program.elf`

### Implementation Details
- Added argument parsing for `--pc-harts` in `spike_main/spike.cc`.
- Logic applies the PC overrides to the respective `processor_t` states after simulation initialization but before the run loop begins.

### Testing
Verified using a dual-core bare-metal assembly test (with the boot ROM disabled) where Hart 0 and Hart 1 must execute distinct code sections located at `0x0` and `0x200` respectively. Confirmed that both harts start at the correct overridden addresses.


nabudahab:
Hello, can someone please review the code so that the CI checks can run? Thanks. I'm always happy to address feedback.

spike_main/spike.cc
```diff
@@ -552,6 +575,14 @@ int main(int argc, char** argv)
   s.configure_log(log, log_commits);
   s.set_histogram(histogram);
 
+  for(size_t i = 0; i < cfg.nprocs(); i++) {
+    size_t hartid = cfg.hartids[i];
+
+    if(hart_start_pcs.count(hartid)){
```
aswaterman: Space after keywords like `for` and `if`.

spike_main/spike.cc
```diff
@@ -346,6 +349,9 @@ int main(int argc, char** argv)
 
   cfg_t cfg;
 
+  //Map to store <hart_id, start_pc>
```
aswaterman: I think the comment can disappear; the statement declaring the map is pretty clear.

spike_main/spike.cc
```diff
@@ -381,6 +387,23 @@ int main(int argc, char** argv)
   parser.option(0, "halted", 0, [&](const char UNUSED *s){halted = true;});
   parser.option(0, "rbb-port", 1, [&](const char* s){use_rbb = true; rbb_port = atoul_safe(s);});
   parser.option(0, "pc", 1, [&](const char* s){cfg.start_pc = strtoull(s, 0, 0);});
+
+  parser.option(0, "pc-harts", 1, [&](const char* s){
```
aswaterman: Not wild about the name of the argument.  I think `pcs` instead of `pc-harts` might be better.

aswaterman (commented):
Can I get a second review from someone else, perhaps @jerryz123?

nabudahab:
@jerryz123 Gentle nudge as it's been a week since the last update. Let me know if you have any thoughts on the implementation or if there's anything you'd like me to address.

nabudahab:
@aswaterman I realized I hadn’t actually applied the spacing/style suggestions earlier — that’s fixed now and pushed. Could you take another quick look / re-approve when you get a chance?

@jerryz123 (or any maintainer) GitHub Actions for this PR is still awaiting maintainer approval to run CI. Could someone approve the workflow runs so checks can execute? Once CI is green, this should be ready to merge.

spike_main/spike.cc
```diff
@@ -381,6 +386,23 @@ int main(int argc, char** argv)
   parser.option(0, "halted", 0, [&](const char UNUSED *s){halted = true;});
   parser.option(0, "rbb-port", 1, [&](const char* s){use_rbb = true; rbb_port = atoul_safe(s);});
   parser.option(0, "pc", 1, [&](const char* s){cfg.start_pc = strtoull(s, 0, 0);});
+
+  parser.option(0, "pcs", 1, [&](const char* s){
+    std::string arg(s);
+    std::stringstream ss(arg);
+    std::string pair;
+    while(std::getline(ss, pair, ',')) {
```
jerryz123:
```suggestion
    while (std::getline(ss, pair, ',')) {
```

spike_main/spike.cc
```diff
@@ -381,6 +386,23 @@ int main(int argc, char** argv)
...
+      size_t delim = pair.find(':');
+      if(delim == std::string::npos) {
```
jerryz123:
```suggestion
      if (delim == std::string::npos) {
```

spike_main/spike.cc
```diff
@@ -552,6 +574,14 @@ int main(int argc, char** argv)
   s.configure_log(log, log_commits);
   s.set_histogram(histogram);
 
+  for (size_t i = 0; i < cfg.nprocs(); i++) {
```
jerryz123: Instead of doing this - which leads to surprising interactions between `cfg.start_pc` and `hart_start_pcs`, recommend you adjust `cfg.start_pc` to handle the hart-specialized start_pc case.

spike_main/spike.cc
```diff
@@ -46,6 +47,8 @@ static void help(int exit_code = 1)
   fprintf(stderr, "  --pmpgranularity=<n>  PMP Granularity in bytes [default 4]\n");
   fprintf(stderr, "  --priv=<m|mu|msu>     RISC-V privilege modes supported [default %s]\n", DEFAULT_PRIV);
   fprintf(stderr, "  --pc=<address>        Override ELF entry point\n");
+  fprintf(stderr, "  --pcs=<H:A,...>  Override start PC for specific harts\n"
```
jerryz123: Fix alignment in help text

jerryz123 (commented):
Whoops, sorry. Need to fix my github notification filtering.

spike_main/spike.cc
nabudahab: I'm working on the implementation of this right now, and I can't define what the interaction between --start-pc and --pcs should be. Would you recommend I combine them into the same command? I would of course maintain backwards compatibility (as in if one PC is set all harts will have that PC).

nabudahab:
@jerryz123 I've implemented your suggested changes and tested --pcs with --pc. The interaction I defined was that -pc sets a global override PC and --pcs can set overrides on top of that for individual harts, --pcs also works without --pc with the unspecified harts simply starting at the PC defined by sim_t::get_entry_point().

Please let me know if there's anything else of concern or if this is ready for merge. Can someone approve the workflows again?

riscv/sim.cc
```diff
@@ -360,7 +360,7 @@ void sim_t::set_rom()
 {
   const int reset_vec_size = 8;
 
-  reg_t start_pc = cfg->start_pc.value_or(get_entry_point());
+  reg_t start_pc = cfg->start_pc.get(0).value_or(get_entry_point());
```
arromanoff, line 397: Pipeline fails. It seems to me that `set_global` should also set pc for hart 0 in order for this to work

arromanoff: After some though I was wrong here. The problem is actually that you set pc in `sim_t::proc_reset`, overriding default ROM start address

nabudahab: I've fixed it, and ensure that the default behavior begins at RSTVEC by modifying proc_reset like you suggested. Thanks!

nabudahab:
Okay, I just got the CI infinite loop issue fixed. I recreated the bug on my local machine (Ubuntu), fixed the bug and checked that it works. I believe that this is ready for another CI run now. Can someone approve that? Thanks!

aswaterman:
Note that our CI flow checks that every commit works, so you might need to flatten this PR into a single commit and force-push over your branch.

nabudahab:
@aswaterman Done! Thanks for the pointer, I've flattened the PR into a single commit.

aswaterman:
Thanks, I'll leave the rest of the review/merging to @jerryz123.

riscv/sim.cc
```diff
@@ -458,9 +516,12 @@ void sim_t::idle()
 
   if (debug || ctrlc_pressed)
     interactive();
-  else {
-    if (instruction_limit.has_value()) {
-      if (*instruction_limit < INTERLEAVE) {
+  else
+  {
```
jerryz123: Can you avoid unrelated formatting changes? 

I realize we aren't consistent in this repo, but should unnecessary churn. 

(And personally, I prefer the style that minimizes LoC)

jerryz123 (commented):
The core logic here looks good, but can you avoid all the unrelated formatting changes?

riscv/sim.cc
nabudahab: My bad, I must've hit whatever key shortcut vscode uses to format files accidentally while edition. I've fixed this now thanks for pointing that out!

nabudahab:
Just bumping this as I think it's ready for CI testing and merge at this point, I fixed the formatting problems @jerryz123

jerryz123:
Thanks, will take a look tonight

Jerry Zhao

riscv/sim.cc
```diff
@@ -106,6 +106,14 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
                                       log_file.get(), sout_));
       harts[cfg->hartids[i]] = procs[i];
     }
+    for (auto& x : this->mems) {
```
jerryz123: Unrelated change here?

riscv/sim.cc
```diff
@@ -246,15 +254,15 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
                           dtb_discovery_plugin_device_factories.end());
 
     //Remove default memories and use dtb discovered memories
-    mems.clear();
```
jerryz123, line 249: Unrelated changes?

riscv/sim.cc
```diff
@@ -246,15 +254,15 @@ sim_t::sim_t(const cfg_t *cfg, bool halted,
...
+    this->mems.clear();
+    dtb_discovery::discover_memory_from_dtb(fdt, this->mems);
   }
   //clint, plic, ns16550 are always discovered via dtb, independently from the --dtb_discovery flag
   device_factories.insert(device_factories.end(),
                           plugin_device_factories.begin(),
                           plugin_device_factories.end());
 
-  for (auto& x : mems)
+  for (auto& x : this->mems)
```
jerryz123: Unrelated change?

jerryz123 (commented):
Please add a comment to the help text in spike.cc that when `--pcs` is used, the builtin bootrom is bypassed - harts will come out reset at the specified PCs, instead of into the bootrom.

Also, I think the only place you need to override the hart PCs is after calling the `.reset()` for each hart.

riscv/sim.cc
nabudahab: Without this loop here after merging the --dtb-discovery PR from a few months ago spike would just infinite loop on startup.

riscv/sim.cc
nabudahab: I've fixed this, part of the debugging process for an infinite loop from a merge conflict with the --dtb-discovery feature.

riscv/sim.cc
nabudahab: fixed this too

nabudahab:
@jerryz123 I've added the comment in spike.cc, please let me know if there's anything else to address.

nabudahab:
@jerryz123 @aswaterman pinging just to get this running through the CI checks and possibly merged unless someone has more suggestions. Thanks!

aswaterman:
Done.  But note that this repo's CI runs the tests on every individual commit.  If CI fails, you might want to consider rebasing against master and flattening the PR into a single commit.

nabudahab:
Hey, is there anything else needed to move this towards a merge? Thank you all for your feedback.

aswaterman:
I think this is good to go.

aswaterman (approved):

Issue #2203, opened 2026-01-02
Smcdeleg needs to be limited to v/siselect values in the range 0x40-0x5F


mmhus:
The priv spec mentions:

```
While the privilege mode is M or S and siselect holds a value in the range 0x40-0x5F, illegal-instruction
exceptions are raised for the following cases:
....
If the hypervisor (H) extension is also implemented, then as specified by extensions Smcsrind/Sscsrind, a
virtual-instruction exception is raised for attempts from VS-mode or VU-mode to directly access vsiselect
or vsireg*, or attempts from VU-mode to access siselect or sireg*. Furthermore, while vsiselect holds
a value in the range 0x40-0x5F
```

aswaterman:
This should be handled by #2204.

Issue #2202, opened 2026-01-02
Smcdeleg references mcycle/minstretcfg instead of cycle/instretcfg


mmhus:
> 
> That sounds right to me. Can you submit a PR that does this?
> 
> cc [@mmhus](https://github.com/mmhus) for comment, since he wrote this code originally.
> 
> Sure, I will do it later.
> 
> There is still another issue, to handle the permission of indirection CSR access, it calls 'verify_permissions' of proxied register.
> 1. for AIA indirection register, it's correct.
> 2. but for others indirection CSRs, such as mcycle/minstreg/mcyclecfg/minstretcfg, which is accessed via sireg* at S-mode, I think it will raise exception by calling 'verify_permissions' directly
> ```
>       // Smcdeleg
>       if (proc->extension_enabled(EXT_SSCCFG) || proc->extension_enabled(EXT_SMCDELEG)) {
>         switch (sireg_csrs[i]) {
>           ......
>           case CSR_SIREG2:
>             if (proc->extension_enabled_const(EXT_ZICNTR)) {
>                 sireg->add_ireg_proxy(SISELECT_SMCDELEG_START, mcyclecfg);
>                 sireg->add_ireg_proxy(SISELECT_SMCDELEG_INSTRETCFG, minstretcfg);
>             }
>             if (proc->extension_enabled_const(EXT_ZIHPM)) {
>               for (size_t j = 0; j < (SISELECT_SMCDELEG_END - SISELECT_SMCDELEG_HPMEVENT_3 + 1); j++)
>                 sireg->add_ireg_proxy(SISELECT_SMCDELEG_HPMEVENT_3 + j, csrmap[CSR_MHPMEVENT3H + j]);
>             }
>             break;
>           ......
>           case CSR_SIREG3:
>           case CSR_SIREG6:
>           default:
>             break;
>         }
>       }
> ```
>  

As referenced by @Steven-Li-Xiaogang I opened a separate issue for this.

 _Originally posted by @Steven-Li-Xiaogang in [#2188](https://github.com/riscv-software-src/riscv-isa-sim/issues/2188#issuecomment-3685947121)_

aswaterman:
This should be handled by #2204.

PR #2204, opened 2026-01-02
Mmhus/mireg sireg fixes #2188, #2202, #2203

### LINKED ISSUE #2188: Better implementation to raise illegal instruction exception while accessing un-implemented index CSR via sireg(really vsireg) at VS-mode
SPIKE code:
```
void sscsrind_reg_csr_t::verify_permissions(insn_t insn, bool write) const {
  if (proc->extension_enabled(EXT_SMSTATEEN)) {
    if ((state->prv < PRV_M) && !(state->mstateen[0]->read() & MSTATEEN0_CSRIND))
  }

  // Don't call base verify_permission for VS registers remapped to S-mode
  if (insn.csr() == address)
    csr_t::verify_permissions(insn, write);

  if (proc->extension_enabled(EXT_SMSTATEEN)) {
    if (state->v && !(state->hstateen[0]->read() & HSTATEEN0_CSRIND))
  }

  ......

  csr_t_p proxy_csr = get_reg();
  if (proxy_csr == nullptr) {
    if (!state->v) {
    } else {
    }
  }
  proxy_csr->verify_permissions(insn, write);
}

```

ISA Spec:
> The behavior upon accessing vsireg* from M-mode or HS-mode, or accessing sireg* (really vsireg*)
> from VS-mode, while vsiselect holds a value that is not implemented at HS level, is UNSPECIFIED.
>  It is recommended that implementations raise an illegal-instruction exception for
> such accesses, to facilitate possible emulation (by M-mode) of these accesses.

I think it's better to raise illegal instruction while `nullptr` returned by '`get_reg()`'

mmhus:
Fixes for #2188, #2202, #2203 

/cc @Steven-Li-Xiaogang Could you give your input, I wasn't able to test everything. Maybe you could run your tests on this branch?

### DIFF
--- a/riscv/csr_init.cc
+++ b/riscv/csr_init.cc
@@ -80,6 +80,28 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
     add_csr(CSR_MINSTRET, minstret);
     add_csr(CSR_MCYCLE, mcycle);
   }
+  if (proc->extension_enabled_const(EXT_SMCNTRPMF)) {
+    if (xlen == 32) {
+      add_csr(CSR_MCYCLECFG, std::make_shared<rv32_low_csr_t>(proc, CSR_MCYCLECFG, mcyclecfg));
+      add_csr(CSR_MCYCLECFGH, std::make_shared<rv32_high_csr_t>(proc, CSR_MCYCLECFGH, mcyclecfg));
+      add_csr(CSR_MINSTRETCFG, std::make_shared<rv32_low_csr_t>(proc, CSR_MINSTRETCFG, minstretcfg));
+      add_csr(CSR_MINSTRETCFGH, std::make_shared<rv32_high_csr_t>(proc, CSR_MINSTRETCFGH, minstretcfg));
+    } else {
+      add_csr(CSR_MCYCLECFG, mcyclecfg);
+      add_csr(CSR_MINSTRETCFG, minstretcfg);
+    }
+    if (proc->extension_enabled_const(EXT_SMCDELEG)) {
+      if (xlen == 32) {
+        cyclecfg = std::make_shared<counter_proxy_csr_t>(proc, CSR_MCYCLECFG, csrmap[CSR_MCYCLECFG]);
+        cyclecfgh = std::make_shared<counter_proxy_csr_t>(proc, CSR_MCYCLECFGH, csrmap[CSR_MCYCLECFGH]);
+        instretcfg = std::make_shared<counter_proxy_csr_t>(proc, CSR_MINSTRETCFG, csrmap[CSR_MINSTRETCFG]);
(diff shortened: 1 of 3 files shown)


mmhus:
The CI build for ubuntu is failing, but it seems like a clone issue, unrelated to the changes, perhaps someone can restart the job.

aswaterman (approved):
LGTM, thanks.

riscv/csr_init.cc
```diff
@@ -80,6 +80,28 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
...
+      add_csr(CSR_MCYCLECFGH, std::make_shared<rv32_high_csr_t>(proc, CSR_MCYCLECFGH, mcyclecfg));
+      add_csr(CSR_MINSTRETCFG, std::make_shared<rv32_low_csr_t>(proc, CSR_MINSTRETCFG, minstretcfg));
+      add_csr(CSR_MINSTRETCFGH, std::make_shared<rv32_high_csr_t>(proc, CSR_MINSTRETCFGH, minstretcfg));
+    } else {
+      add_csr(CSR_MCYCLECFG, mcyclecfg);
+      add_csr(CSR_MINSTRETCFG, minstretcfg);
+    }
+    if (proc->extension_enabled_const(EXT_SMCDELEG)) {
+      if (xlen == 32) {
+        cyclecfg = std::make_shared<counter_proxy_csr_t>(proc, CSR_MCYCLECFG, csrmap[CSR_MCYCLECFG]);
```
Steven-Li-Xiaogang: @mmhus 
I think `cyclecfg/instretcfg` should not be declared as `counter_proxy_csr_t`, access permission of ` mcyclecfg` and `minstretcfg` are not under control of` mcounteren`, also a fault indirect CSR index (with formula `address & 31`),  indirect access to CSR_MCYCLECFG and CSR_MINSTRETCFG will cause exception by calling `verify_permissions`

riscv/csrs.cc
```diff
@@ -1919,7 +1919,16 @@ void sscsrind_reg_csr_t::verify_permissions(insn_t insn, bool write) const {
 
   csr_t_p proxy_csr = get_reg();
   if (proxy_csr == nullptr) {
-    throw trap_illegal_instruction(insn.bits());
+    // An attempt from VS-mode to access any inaccessible sireg* causes virtual instruction exception when AIA is enabled
+    if (proc->extension_enabled_const(EXT_SMAIA)) {
+      if (state->prv == PRV_S && state->v) {
+        throw trap_virtual_instruction(insn.bits());
```
Steven-Li-Xiaogang: @mmhus @aswaterman @ved-rivos
There is still one question: should be all indices which are not proxied be treated as inaccessible?
1. AIA spec only defines 0x30-0x3F and 0x70-0xFF (if no IMSIC) as inaccessible, and others are reserved (unimplemented)
2. Could a virtual instruction exception be raised, while an attempt from VS-mode to `sireg`* while `vsiselect` holds a reserved value (ex. 0x100)?

PR #2205, opened 2026-01-03
correct smcdeleg indirect CSRs address accessed via sireg*

Steven-Li-Xiaogang:
1. `hpmevent3-hpmevent31` are accessed via `sireg2`
2. `hpmevent3h-hpmevent31h` are accessed via `sireg5`, `hpmcounter3-hpmcounter31` should not be accessed indirectly via `sireg5`
3. `hpmevent3h-hpmevent31h` proxies require `Sscofpmf` extension


aswaterman (commented):
@mmhus delegating this one to you again.  Thanks for the help, both of you.

mmhus:
I've checked it and the changes look good. 32 bit counters should be sireg5, and events should be sireg 2.

aswaterman (approved):
Thanks again.

PR #2207, opened 2026-01-05
Only set mstatus.VS for legal vector instructions

### LINKED ISSUE #2206: vmsof.m/vmsif.m dirty vs when exception happen
Image

Image

When executing the vmsof.m/vmsif.m instruction, the require_vector(true) logic will directly dirty the vs field of the status.

However, if a trap illegal instruction exception occurs due to require(insn.rd() != insn.rs2()), I understand that the vs should not be dirtied, which would otherwise lead to incorrect status updates.

aswaterman:
@jerryz123 I'd like your opinion on this one; it's a little messy and a little risky.

The problem is that we are setting the mstatus.VS field very early on in handling a vector instruction, possibly before other `require`ments might trip.  This isn't technically invalid under the ISA, but it's definitely an error of intent.

My proposed solution piggy-backs on the fact that most vector instructions don't raise any exceptions other than illegal-instruction exceptions, and the fact that most of them explicitly write zero to `vstart` at the end.  That `vstart` write sets VS to Dirty as a side effect.

A few instructions were avoiding explicitly zeroing `vstart`, e.g. `vcompress`, which skipped the write since it asserts `vstart` is 0 to begin with.  I used some hacky static analysis to make sure I didn't leave out any instructions.

Memory instructions require special handling, since they can raise exceptions in the middle, so we must make sure they write `vstart` in all code paths.

Fixes #2206

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1598,7 +1598,7 @@ void vector_csr_t::write_raw(const reg_t val) noexcept {
 
 bool vector_csr_t::unlogged_write(const reg_t val) noexcept {
   if (mask == 0) return false;
-  dirty_vs_state;
+  STATE.sstatus->dirty(SSTATUS_VS);
   return basic_csr_t::unlogged_write(val & mask);
 }
 
@@ -1612,7 +1612,7 @@ void vxsat_csr_t::verify_permissions(insn_t insn, bool write) const {
 }
 
 bool vxsat_csr_t::unlogged_write(const reg_t val) noexcept {
-  dirty_vs_state;
+  STATE.sstatus->dirty(SSTATUS_VS);
   return masked_csr_t::unlogged_write(val);
 }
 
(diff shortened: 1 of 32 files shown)


jerryz123:
Oops, missed the ping here. I'll take a look at this tonight

aswaterman:
@jerryz123 Thanks.  I had to fix it up a bit, but it should be good to go by the time you get around to it.

jerryz123 (approved):
I think these changes are safe, although I agree this is a bit risky. 

At one point I tried splitting up the instruction headers into a "pre-decode/verification" phase, and an execution phase. Although the motivation was to hoist logic out of the i-cache hit path, and move it to the i-cache fill path. https://github.com/riscv-software-src/riscv-isa-sim/compare/master...speed2

It didn't have as much of a perf benefit as I had hoped for, so I dropped it. But there is something useful about having the API to "check legality, do not mutate arch state."

aswaterman:
I thought about that, too, but didn't get around to trying it.  I suspect the reason it doesn't help as much as we'd prefer is that the legality checks feed into perfectly predictable branches and otherwise introduce no dependencies, making for cheaply exploitable ILP.

PR #637, opened 2021-02-02
Support RISC-V p-ext-proposal v0.9.2

hope51607:
See the previous PR #572 

Hello, we’re the team from Programming Langauge Lab, NTHU (National Tsing-Hua University), Taiwan and also from Peakhills group.

This PR implements all instructions in the P-Extension Proposal v0.9.2 ([link](https://github.com/riscv/riscv-p-spec/blob/master/P-ext-proposal.adoc)).

I also create riscv/riscv-opcodes#60 PR to add p-ext opcodes in riscv-opcodes.

PR #2210, opened 2026-01-16
Support RISC-V p-ext-proposal v0.9.12 for RV32

jason01180118:
See the previous PR #637 

Hello, we’re the team from Programming Langauge Lab, NTHU (National Tsing-Hua University), Taiwan.

This PR implements all instructions in the latest P-Extension specification ([link](https://www.jhauser.us/RISCV/ext-P/))

Branches are split for RV32 and RV64, since some instructions may conflict between architectures.


riscv/p_ext_macros.h
```diff
@@ -113,6 +138,16 @@
   } \
   WRITE_RD(sext_xlen(rd_tmp));
 
+#define P_REDUCTION_ULOOP_END(BIT, IS_SAT) \
+    } \
+    if (IS_SAT) { \
+      P_USAT(BIT, p_res); \
```
chihminchao: Should it be  
```
p_res = P_USAT(BIT, p_res); 
```
?

jason01180118: These instructions no longer have a saturation function in the new P extension spec, so I will remove the `if (IS_SAT)` condition and clear other build warnings. Thank you for the helpful feedback.

chihminchao:
Just a reminding,
based on https://www.jhauser.us/RISCV/ext-P/RVP-instrEncodings-019.pdf, there are some minor renaming 
"This version (019) diff ers from the previous one (018) in correcting the names of instructions MQWACC
and MQRWACC, which previously were miswritten with a ‘P’ as “PMQWACC” and “PMQRWACC”."

christian-herber-nxp:
The readme should be updated to reflect the exact version of P supported

PR #2216, opened 2026-01-21
csrs.cc: if no U-mode, mstatus.tw is read-only 0

### LINKED ISSUE #2215: TW is read-only 0 when there are no modes less privileged than M
According to the RISC-V Privileged Specification (page 37), the TW bit has the following property:

> The TW (Timeout Wait) bit is a WARL field that supports intercepting the WFI instruction (see Section
3.3.3). When TW=0, the WFI instruction may execute in modes less privileged than M when not prevented
for some other reason. When TW=1, then if WFI is executed in any less-privileged mode, and it does not
complete within an implementation-specific, bounded time limit, the WFI instruction causes an illegal-
instruction exception. An implementation may have WFI always raise an illegal-instruction exception in
modes less privileged than M when TW=1, even if there are pending globally-disabled interrupts when the
instruction is executed. **TW is read-only 0 when there are no modes less privileged than M.**

This seems to imply that when an implementation supports only M-mode (i.e., no S-mode or U-mode), the `TW` bit should not be writable and should always read as 0.

https://github.com/riscv-software-src/riscv-isa-sim/blob/efdaa0837137ddd4b4154196251d52daae3277bc/riscv/csrs.cc#L556-L566

From my understanding, `TW` appears to be treated as writable even when Spike is configured with no less-privileged modes than M. This made me wonder whether the current behavior fully matches the spec requirement quoted above.

I’d appreciate any correction if my understanding is wrong.

Thanks

pointerliu:
Fix https://github.com/riscv-software-src/riscv-isa-sim/issues/2215

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -556,7 +556,8 @@ bool mstatus_csr_t::unlogged_write(const reg_t val) noexcept {
   const reg_t mask = adj_write_mask
                    | MSTATUS_MIE | MSTATUS_MPIE
                    | (proc->extension_enabled('U') ? MSTATUS_MPRV : 0)
-                   | MSTATUS_MPP | MSTATUS_TW
+                   | MSTATUS_MPP
+                   | (proc->extension_enabled('U') ? MSTATUS_TW : 0)
                    | (proc->extension_enabled('S') ? MSTATUS_TSR : 0)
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
Thanks again for the fix!

PR #2218, opened 2026-01-21
Fix incorrect mip.SEIP handling by adding hardware SEIP latch & correcting PLIC aliasing

rmkhurana28:
This PR fixes a long-standing correctness issue in Spike’s handling of the
Supervisor External Interrupt Pending bit (mip.SEIP).

### Root Cause
SEIP was incorrectly read/written directly through `mip->val`,
causing mismatches when:
- `mvien.SEIP = 0` (SEIP is aliased to mvip)
- `mvien.SEIP = 1` (SEIP must come only from hardware/PLIC)
- PLIC needs to set/clear SEIP independently of software writes

### Fix
This PR introduces a **hardware SEIP latch** inside `mip_csr_t` and updates all
call sites to follow the correct read/write ordering.

#### Changes:
- Added `seip_hw_latch` field to `mip_csr_t`
- Updated `mip_csr_t::read()` to merge hvip + mvip alias + hardware latch
- Updated `mip_csr_t::backdoor_write_with_mask()` to route SEIP writes properly
- Fixed `plic_t::context_update()` to respect mvien.SEIP aliasing rules
- Added `set_seip_hw_latch()` for safe PLIC updates

### Result
This restores correct SEIP semantics according to privileged spec 1.12, improves
behavior under virtualization (H-extension), and eliminates SEIP mis-ordering.

This is a correctness fix (not a functional extension), and has no effect unless
SEIP is toggled via hardware or mvien.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -821,10 +821,29 @@ void mip_csr_t::write_with_mask(const reg_t mask, const reg_t val) noexcept {
 }
 
 reg_t mip_csr_t::read() const noexcept {
-  return val | state->hvip->basic_csr_t::read() | ((state->mvien->read() & MIP_SEIP) ? 0 : (state->mvip->basic_csr_t::read() & MIP_SEIP));
+  reg_t result = val;
+
+  // Merge hypervisor virtual interrupts (unchanged behavior)
+  result |= state->hvip->basic_csr_t::read();
+
+  // Merge hardware SEIP latch (external interrupt pending)
+  result |= seip_hw_latch & MIP_SEIP;
+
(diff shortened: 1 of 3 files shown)


rmkhurana28:
Updated the PR to address all follow-up issues:

• Correct MEIP bypass logic (MEIP no longer goes through SEIP/mvien/mvip path)
• SEIP is now fully isolated from software-writable mip.val
• PLIC update logic now cleanly separates machine-mode vs supervisor-mode paths
• Comments updated for clarity

This version reflects proper interrupt routing per RISC-V Privileged Spec v1.12.

PR #2221, opened 2026-01-24
Fix: Enforce virtual_instruction trap for VU-mode indirect CSR access

DymShanks:
This PR enforces a `virtual_instruction` exception when accessing `sireg`/`vsireg` from VU-mode.

**Implementation Details:**
Added a privilege check at the entry of `sscsrind_reg_csr_t::verify_permissions`.

**Reasoning:**
VU-mode cannot access HS-mode resources. Trapping early resolves ambiguity and ensures correct virtualization behavior. Verified build passes locally.

### DIFF
--- a/riscv/csrs.cc
+++ b/riscv/csrs.cc
@@ -1852,6 +1852,10 @@ sscsrind_reg_csr_t::sscsrind_reg_csr_t(processor_t* const proc, const reg_t addr
 }
 
 void sscsrind_reg_csr_t::verify_permissions(insn_t insn, bool write) const {
+  if (state->v && state->prv == PRV_U) {
+     throw trap_virtual_instruction(insn.bits());
+  }
+
   if (proc->extension_enabled(EXT_SMSTATEEN)) {
     if ((state->prv < PRV_M) && !(state->mstateen[0]->read() & MSTATEEN0_CSRIND))
       throw trap_illegal_instruction(insn.bits());


riscv/csrs.cc
```diff
@@ -1852,6 +1852,13 @@ sscsrind_reg_csr_t::sscsrind_reg_csr_t(processor_t* const proc, const reg_t addr
 }
 
 void sscsrind_reg_csr_t::verify_permissions(insn_t insn, bool write) const {
+  // [FIX START] Issue #2554: Trap VU-mode access to indirect CSRs
+  // VU-mode cannot access HS-mode resources. Trapping here removes ambiguity.
+  if (state->v && state->prv == PRV_U) {
+    throw trap_virtual_instruction(insn.bits());
+  }
+  // [FIX END]
```
aswaterman: Please remove the `[FIX START]` and `[FIX END]` comments.

riscv/csrs.cc
```diff
@@ -1484,7 +1484,7 @@ bool dcsr_csr_t::unlogged_write(const reg_t val) noexcept {
   v = proc->extension_enabled('H') ? get_field(val, CSR_DCSR_V) : false;
   mprven = get_field(val, CSR_DCSR_MPRVEN);
   pelp = proc->extension_enabled(EXT_ZICFILP) ?
-         static_cast<elp_t>(get_field(val, DCSR_PELP)) : elp_t::NO_LP_EXPECTED;
+          static_cast<elp_t>(get_field(val, DCSR_PELP)) : elp_t::NO_LP_EXPECTED;
```
aswaterman: Please remove the formatting changes to parts of the code you aren't otherwise modifying.

riscv/csrs.cc
```diff
@@ -2252,4 +2203,4 @@ void aia_csr_t::verify_permissions(insn_t insn, bool write) const {
   }
 
   basic_csr_t::verify_permissions(insn, write);
-}
+}
```
aswaterman: Undo

riscv/csrs.cc
```diff
@@ -1934,10 +1887,8 @@ void sscsrind_reg_csr_t::verify_permissions(insn_t insn, bool write) const {
       throw trap_illegal_instruction(insn.bits());
     }
   }
-  proxy_csr->verify_permissions(insn, write);
 }
 
-
```
aswaterman, line 1940: Undo

aswaterman (changes requested):
I'm struggling to see how this is correct, given that it removes the dependence on `MENVCFG_CDE`.

riscv/csrs.cc
DymShanks: I have reset the file to upstream and re-applied only the necessary trap logic for VU-mode. This restores the original formatting and the MENVCFG_CDE dependency. Thanks for the guidance

aswaterman (approved):
Thank you.

@ved-rivos can you sanity-check this before I merge it?

ved-rivos (approved):

ved-rivos:
@aswaterman this looks good to me.

aswaterman:
Thanks @ved-rivos @DymShanks

binno:
Access s-mode csrs on VU mode originally raise virtual instruction trap, am I right?

https://github.com/riscv-software-src/riscv-isa-sim/blob/98ccf030bb02a029944cd938d5bcb73275350df4/riscv/csrs.cc#L46

aswaterman:
@binno I think this is needed because it is an override of `csr_t::verify_permissions`, and indirect CSRs have different rules?

binno:
@aswaterman , you're right and my question isn't existed.
This rule should be for VS registers remapped to S-mode, it will not call **csr_t::verify_permissions**
https://github.com/riscv-software-src/riscv-isa-sim/blob/98ccf030bb02a029944cd938d5bcb73275350df4/riscv/csrs.cc#L1867

jameshippisley:
This change seems to contradict the spec:

https://github.com/riscv/riscv-isa-manual/blob/f30292bb486d735e186f06f705fcd7beff1e3e97/src/indirect-csr.adoc?plain=1#L308-L316

"If extension Smstateen is implemented together with Smcsrind, bit 60 of state-enable register mstateen0 controls access to siselect, sireg*, vsiselect, and vsireg*. **When mstateen0[60]=0, an attempt to access one of these CSRs from a privilege mode less privileged than M-mode results in an illegal-instruction exception**. As always, the state-enable CSRs do not affect the accessibility of any state when in M-mode, only in less privileged modes. For more explanation, see the documentation for extension [Smstateen](https://github.com/riscv/riscv-state-enable/releases/download/v1.0.0/Smstateen.pdf)."

This says that the mstateen[60] == 0 -> illegal instruction exception rule should have higher precedence than the virtual instruction check included in this PR, unless I am missing something.

aswaterman:
@jameshippisley You're right.  And that clause is handled correctly by the next block of code.  I'll submit a PR to interchange them.

PR #2227, opened 2026-02-04
Raise correct trap in U-mode on indirect CSRs when !mstateen.csrind

aswaterman:
Fixes a regression introduced by #2221

@ved-rivos can you confirm you agree this fixes @jameshippisley's comment while otherwise preserving the intent of the PR it's fixing? https://github.com/riscv-software-src/riscv-isa-sim/pull/2221#issuecomment-3848024136


ved-rivos:
@aswaterman this looks right.

aswaterman:
Thanks!

PR #2240, opened 2026-03-05
Fix vmv<n>r.v no elements are written if vstart>= evl

Steven-Li-Xiaogang:
Whole vector register move instructions  'vmv<n>r.v' has EEW=SEW and EMUL=NREG, effective length evl = NREG * VLEN / EEW. Spec also declares 

> The usual property that no elements are written if does not apply to these vstart ≥ vl instructions. Instead, no elements are written if vstart ≥ evl

.

### DIFF
--- a/riscv/insns/vmvnfr_v.h
+++ b/riscv/insns/vmvnfr_v.h
@@ -5,7 +5,8 @@ const reg_t vs2 = insn.rs2();
 const reg_t len = insn.rs1() + 1;
 require_align(vd, len);
 require_align(vs2, len);
-const reg_t size = len * P.VU.vlenb;
+const reg_t evl = len * P.VU.VLEN / P.VU.vsew;
+const reg_t size = evl * (P.VU.vsew >> 3);
 const reg_t start = P.VU.vstart->read() * (P.VU.vsew >> 3);
 
 //register needs one-by-one copy to keep commitlog correct


riscv/insns/vmvnfr_v.h
```diff
@@ -5,9 +5,13 @@ const reg_t vs2 = insn.rs2();
...
 require_align(vd, len);
 require_align(vs2, len);
-const reg_t size = len * P.VU.vlenb;
+const reg_t NREG = insn.v_zimm5() + 1;
+const reg_t evl = NREG * P.VU.VLEN / P.VU.vsew;
+const reg_t size = evl * (P.VU.vsew >> 3);
 const reg_t start = P.VU.vstart->read() * (P.VU.vsew >> 3);
 
+require(NREG == 1 || NREG == 2 || NREG == 4 || NREG == 8);
+
```
aswaterman: delete this; it is already enforced elsewhere

riscv/insns/vmvnfr_v.h
```diff
@@ -5,9 +5,13 @@ const reg_t vs2 = insn.rs2();
 const reg_t len = insn.rs1() + 1;
 require_align(vd, len);
 require_align(vs2, len);
-const reg_t size = len * P.VU.vlenb;
+const reg_t NREG = insn.v_zimm5() + 1;
+const reg_t evl = NREG * P.VU.VLEN / P.VU.vsew;
+const reg_t size = evl * (P.VU.vsew >> 3);
 const reg_t start = P.VU.vstart->read() * (P.VU.vsew >> 3);
```
aswaterman, line 10: I agree this looks right; here, `start` is in units of elements, and so `size` should also be in units of elements, but it in the old code, it is in units of bytes.

riscv/insns/vmvnfr_v.h
```diff
@@ -5,9 +5,13 @@ const reg_t vs2 = insn.rs2();
 const reg_t len = insn.rs1() + 1;
 require_align(vd, len);
 require_align(vs2, len);
-const reg_t size = len * P.VU.vlenb;
+const reg_t NREG = insn.v_zimm5() + 1;
```
aswaterman: delete this and use `len` instead below

aswaterman (commented):
Thanks for pointing this out.  I think the reason no one has encountered it is that Spike never takes interrupts in the middle of vector ALU instructions, so it is impossible to execute this code path unless some verification code attempts to test it by manually programming `vstart`.  And since most implementations trap on ALU ops when vstart != 0, no one is attempting to test it :)

aswaterman (changes requested):
I now no longer think this change is needed.

The old code computed size as `(# registers) * (# bytes per register)`.

The new code computes the size as `((# registers) * (# bits per register) / (# bits per element)) * (# bits per element / 8)`, which evaluates to the same thing.

The important thing is that both `start` and `size` use the same units.  It doesn't matter whether those units are bytes or elements, as long as they're consistent.

PR #2243, opened 2026-03-06
Fix scountovf read behavior in M-mode according to RISC-V Privileged SpecFix scountovf read behavior in M-mode according to RISC-V Privileged …

chillakalyan:
According to the RISC-V Privileged Specification, scountovf bits should
always be readable in M-mode.

The current implementation masks the value using mcounteren (and
hcounteren in VS-mode), which incorrectly restricts visibility in M-mode.

This patch updates scountovf_csr_t::read() so that counteren masking
is applied only when the current privilege mode is not M-mode.


aswaterman:
Sounds right to me.  Please fix the merge conflict, then force-push over the PR.

PR #2244, opened 2026-03-07
Add basic SPMP support

ybc-alkaid:
This PR contains SPMP address and configuration CSR support for running basic non-corner-case tests, as required by https://riscv.atlassian.net/browse/RVS-1831.


aswaterman:
> for running basic non-corner-case tests

Is the implication that you think there are bugs?  We expect PRs to this project to be complete.

ybc-alkaid:
Not bugs. What I meant is the implementation only provides SPMP CSR tests, so test cases such as resource sharing between PMP and SPMP are not supported.

aswaterman:
OK, thanks for the PR; I'll review it this week.

riscv/csr_init.cc
```diff
@@ -528,6 +538,26 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
...
+      for (size_t i = 0; i < proc->n_spmp; i++) {
+        // Create spmpaddr CSR for indirect access via sireg
+        proc->get_state()->spmpaddr[i] = std::make_shared<spmpaddr_csr_t>(proc, SISELECT_SPMP_START + i, i);
+
+        // Create spmpcfg CSR for indirect access via sireg2
+        auto spmpcfg = std::make_shared<spmpcfg_csr_t>(proc, SISELECT_SPMP_START + i, proc->get_state()->spmpaddr[i]);
+
+        // Register spmpaddr proxy on sireg
+        if (sireg_for_spmp)
+          sireg_for_spmp->add_ireg_proxy(SISELECT_SPMP_START + i, proc->get_state()->spmpaddr[i]);
```
aswaterman, line 554: You should just be able to use `sireg` here rather than declaring `sireg_for_spmp`, right?  And you can get rid of the `if` statement because the Sspmp extension requires the Sscsrind extension, and so you know that this cannot be null.

riscv/csr_init.cc
```diff
@@ -528,6 +538,26 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
...
+
+        // Register spmpcfg proxy on sireg2
+        if (sireg2_for_spmp)
+          sireg2_for_spmp->add_ireg_proxy(SISELECT_SPMP_START + i, spmpcfg);
```
aswaterman, line 558: Similarly, we know `sireg2_for_spmp` cannot be null here.  (Also, I would just rename the variable to `sireg2`.)

riscv/csrs.cc
```diff
@@ -2258,3 +2257,192 @@ void aia_csr_t::verify_permissions(insn_t insn, bool write) const {
...
+#define SPMP_W      0x02
+#define SPMP_X      0x04
+#define SPMP_A      0x18  // bits [4:3]
+#define SPMP_L      0x80
+
+// SPMP address matching modes (in A field)
+#define SPMP_A_OFF   0x00
+#define SPMP_A_TOR   0x08
+#define SPMP_A_NA4   0x10
+#define SPMP_A_NAPOT 0x18
```
aswaterman, line 2275: I would prefer that we just reuse all of the `PMP_` definitions, since they're the same.  If you don't like that approach, then please define the SPMP fields in terms of the PMP fields (e.g. `#define SPMP_W PMP_W`).

riscv/csrs.cc
```diff
@@ -2258,3 +2257,192 @@ void aia_csr_t::verify_permissions(insn_t insn, bool write) const {
...
+  return ~(mask & ~(mask + 1)) << PMP_SHIFT;
+}
+
+bool spmpaddr_csr_t::match4(reg_t addr) const noexcept {
+  uint8_t a_field = (cfg & SPMP_A) >> 3;
+  if (a_field == 0)  // OFF
+    return false;
+
+  bool is_tor = a_field == 1;
+  bool is_na4 = a_field == 2;
```
aswaterman, line 2327: The hard-coded `>> 3` and `== 1` etc. are kind of awkward.  It would be a little cleaner to do

```
if ((cfg & SPMP_A) == SPMP_A_OFF)
  return false;

bool is_tor = (cfg & SPMP_A) == SPMP_A_TOR;
bool is_na4 = (cfg & SPMP_A) == SPMP_A_NA4;
```

riscv/csrs.cc
```diff
@@ -2258,3 +2257,192 @@ void aia_csr_t::verify_permissions(insn_t insn, bool write) const {
...
+
+  return addr >= base && addr + 4 <= base + size;
+}
+
+bool spmpaddr_csr_t::subset_match(reg_t addr, reg_t len) const noexcept {
+  uint8_t a_field = (cfg & SPMP_A) >> 3;
+  if (a_field == 0)  // OFF
+    return false;
+
+  bool is_tor = a_field == 1;
```
aswaterman, line 2349: Same comment as https://github.com/riscv-software-src/riscv-isa-sim/pull/2244/changes#r2921185079

riscv/csrs.cc
```diff
@@ -2258,3 +2257,192 @@ void aia_csr_t::verify_permissions(insn_t insn, bool write) const {
...
+
+bool spmpaddr_csr_t::access_ok(access_type type, reg_t mode) const noexcept {
+  // Check if access is allowed based on spmpcfg permissions
+  // SPMP applies to S-mode and U-mode only
+  if (mode != PRV_S && mode != PRV_U)
+    return true;  // M-mode bypasses SPMP
+
+  uint8_t a_field = (cfg & SPMP_A) >> 3;
+  if (a_field == 0)  // OFF - entry disabled
+    return true;
```
aswaterman, line 2380: Same comment as https://github.com/riscv-software-src/riscv-isa-sim/pull/2244/changes#r2921185079

riscv/processor.h
```diff
@@ -166,6 +164,9 @@ struct state_t
   static const int max_pmp = 64;
   pmpaddr_csr_t_p pmpaddr[max_pmp];
 
+  static const int max_spmp = 64;
+  spmpaddr_csr_t_p spmpaddr[max_spmp];
+
```
aswaterman, line 169: Since there are two copies of the storage (one for PMP and one for SPMP), how are they kept in sync?  Isn't it the case that a write to `pmpaddr` should show up in the corresponding `spmpaddr` and vice-versa?  Would it not be easier to just have one copy of the state?

ybc-alkaid: Since mpmpdeleg is not included in this implementation, it essentially simulates a hardcoded separation.
If you feel mpmpdeleg support is necessary, I will add it later.

aswaterman: Ah, I see. Are you going to take the merged-state approach? My sense is that will be easier to maintain than synchronizing on every write, but I could be wrong.

I’d like this feature to be added as part of this current PR, rather than as a follow-up.

ybc-alkaid: > Are you going to take the merged-state approach? My sense is that will be easier to maintain than synchronizing on every write

Make sense.

> I’d like this feature to be added as part of this current PR, rather than as a follow-up.

Will do.

aswaterman: Cool, thanks for the help.

binno:
Hi @ybc-alkaid ,
    I developed SPMP implementation in spike from last year and verify my implementation with SPMP tests provided by Synopsys. Now I commit my commit changes at https://github.com/riscv-software-src/riscv-isa-sim/pull/2249. I believe your experience with SPMP will be of great help to my code correction and optimization. And I sincerely consult you to help review this PR, thanks

ybc-alkaid:
Hi @binno, thanks for implementing SPMP support to Spike.
We really appreciate it!
I can close this PR and proceed with #2249.

ybc-alkaid:
> and verify my implementation with SPMP tests provided by Synopsys.

@binno Is it possible to make these tests public and create a PR at https://github.com/riscv/riscv-arch-test?

binno:
> > and verify my implementation with SPMP tests provided by Synopsys.
> 
> @binno Is it possible to make these tests public and create a PR at https://github.com/riscv/riscv-arch-test?

Hi @ybc-alkaid , 
  These are SiFive internal tests, and we can't public them now.
FYI

PR #2249, opened 2026-03-16
Add SPMP extension implementation

binno:
* Refercee specification
  https://github.com/riscv/riscv-spmp/releases/tag/v0.8.18

* Key feature
  * Add 'spmp' in isa string parsing for SPMP extension
  * Add PMP_Entry CSRs of spmp
  * Implement SPMP matching logic
  * Update implementation to spec rc5 version

* Validation
  Sting test cases for SPMP extension released only to SiFive by Synopsys


aswaterman (commented):
@binno did you coordinate with @ybc-alkaid and his WIP on #2244?  Your version appears to be more complete as it includes delegation, but you might want to review that PR to see if there is anything that should be incorporated.  I also suggest reaching out to @ybc-alkaid to avoid duplicate work going forward.

riscv/mmu.cc
```diff
@@ -484,39 +490,70 @@ tlb_entry_t mmu_t::refill_tlb(reg_t vaddr, reg_t paddr, char* host_addr, access_
   return entry;
 }
 
+bool mmu_t::pmp_search(reg_t addr, reg_t len, pmpaddr_csr_t_p* pmpaddr, size_t pmp_num, int &index)
+{
+	index = NO_IDX;
```
aswaterman: two spaces of indentation, no tabs

riscv/processor.h
```diff
@@ -164,7 +164,10 @@ struct state_t
   csr_t_p mseccfgh;
 
   static const int max_pmp = 64;
-  pmpaddr_csr_t_p pmpaddr[max_pmp];
+  csr_t_p mpmpdeleg;
+  pmpaddr_csr_t_p mpmpaddr[max_pmp];
+  pmpaddr_csr_t_p spmpaddr[max_pmp];
```
aswaterman: Why two copies of the the state?  Since PMP registers can be read from M-mode even when delegated to S-mode, having two copies means writes to the M-mode variant need to be synchronized with writes to the S-mode variant.  That seems more cumbersome than just having a single copy of the state, as you would do in a hardware implementation.

riscv/csrs.cc
```diff
@@ -153,6 +118,11 @@ reg_t pmpaddr_csr_t::napot_mask() const noexcept {
 }
 
 bool pmpaddr_csr_t::match4(reg_t addr) const noexcept {
+  if (proc->extension_enabled_const(EXT_SSPMPEN) && is_spmp) {
+    if ((state->sspmpswitch->read() & (0x1 << pmpidx)) == false)
```
aswaterman: `0x1 << pmpidx` can overflow, since `pmpidx` can be bigger than the number of bits in `int`.  The `== false` pattern is also a little odd.  I suggest rewriting this as `if (!((state->sspmpswitch->read() >> pmpidx) & 1))`

riscv/csrs.cc
```diff
@@ -182,7 +157,54 @@ bool pmpaddr_csr_t::subset_match(reg_t addr, reg_t len) const noexcept {
...
+  const bool lock_bypass = state->mseccfg->get_rlb();
+  const bool locked = !lock_bypass && (cfg & PMP_L);
+
+  if (pmpidx < proc->n_pmp && !locked && !next_locked_and_tor()) {
+    this->val = val & ((reg_t(1) << (proc->paddr_bits() - PMP_SHIFT)) - 1);
+  }
+  else
+    return false;
+  proc->get_mmu()->flush_tlb();
+  return true;
```
aswaterman: I think this would be easier to read as

```
  if (pmpidx >= proc->n_pmp || locked || next_locked_and_tor())
    return false;

  this->val = val & ((reg_t(1) << (proc->paddr_bits() - PMP_SHIFT)) - 1);
  proc->get_mmu()->flush_tlb();
  return true;
```

riscv/csrs.cc
```diff
@@ -182,7 +157,54 @@ bool pmpaddr_csr_t::subset_match(reg_t addr, reg_t len) const noexcept {
   return !(is_tor ? tor_homogeneous : napot_homogeneous);
 }
 
-bool pmpaddr_csr_t::access_ok(access_type type, reg_t mode, bool hlvx) const noexcept {
```
aswaterman, line 185: Not sure I like renaming `pmpaddr` to `mpmpaddr` since `pmpaddr` is still the name of the CSR in the ISA manual.  (Maybe we should have used the name `mpmpaddr` initially, but we didn't, so I think we should stick with it.)

riscv/csr_init.cc
```diff
@@ -572,4 +572,32 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
...
   }
+
+  if (proc->extension_enabled_const(EXT_SMPMPDELEG)) {
+    add_csr(CSR_MPMPDELEG, mpmpdeleg = std::make_shared<mpmpdeleg_csr_t>(proc, CSR_MPMPDELEG, MPMPDELEG_PMPNUM, proc->n_pmp));
+  }
+
+  if (proc->extension_enabled_const(EXT_SSPMP)) {
+    const reg_t spmp_cfg_mask = (SPMP_SHARED | SPMP_U | PMP_L | PMP_A | PMP_X | PMP_W | PMP_R);
+    // add entry registers (index: 0x0..0x3F)
+    for (size_t i = 0; i < 64; ++i) {
```
aswaterman: `64` -> `max_pmp`

riscv/csrs.cc
```diff
@@ -286,6 +308,118 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
...
+  return write_success;
+}
+
+spmpen_csr_t::spmpen_csr_t(processor_t* const proc, const reg_t addr, const reg_t init):
+  basic_csr_t(proc, addr, init) {
+}
+
+bool spmpen_csr_t::unlogged_write(const reg_t val) noexcept {
+  reg_t write_val = 0;
+  for (reg_t i = 0; i < (64 - proc->n_pmp); ++i) {
```
aswaterman: `64` -> `max_pmp`

riscv/csrs.cc
```diff
@@ -286,6 +308,118 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
...
+}
+
+bool spmpaddr_csr_t::unlogged_write(const reg_t val) noexcept {
+  if ((state->prv == PRV_M || !(cfg & PMP_L)) && !next_locked_and_tor()) {
+    this->val = val & ((reg_t(1) << (proc->paddr_bits() - PMP_SHIFT)) - 1);
+  }
+  else
+    return false;
+  proc->get_mmu()->flush_tlb();
+  return true;
```
aswaterman: similar comment as https://github.com/riscv-software-src/riscv-isa-sim/pull/2249#discussion_r2950344378

riscv/csrs.cc
aswaterman: (I see now this is similar to the style in the original code, but I think it's worth improving the style while we are here)

disasm/isa_parser.cc
```diff
@@ -197,6 +197,13 @@ void isa_parser_t::add_extension(const std::string& ext_str, const char* str)
   } else if (ext_str == "zkt") {
   } else if (ext_str == "smepmp") {
     extension_table[EXT_SMEPMP] = true;
+  } else if (ext_str == "sspmp") {
+    if (!extension_table[EXT_SMCSRIND] || !extension_table[EXT_SSCSRIND])
+      bad_isa_string(str, "'SPMP' extension requires 'smcsrind/sscsrind'");
```
aswaterman: We have stopped using this pattern in Spike.  Instead, we just turn on the dependent extensions.  So, I'd replace this with `extension_table[EXT_SMCSRIND] = true` etc.

riscv/mmu.cc
binno: OK

riscv/csrs.cc
binno: Thanks for suggestion

riscv/csrs.cc
binno: Thanks for suggestion

riscv/csr_init.cc
binno: OK

riscv/csrs.cc
binno: OK

riscv/csrs.cc
binno: Thanks for suggestion

binno:
> @binno did you coordinate with @ybc-alkaid and his WIP on #2244? Your version appears to be more complete as it includes delegation, but you might want to review that PR to see if there is anything that should be incorporated. I also suggest reaching out to @ybc-alkaid to avoid duplicate work going forward.

I don't coordinate with @ybc-alkaid and will invite him to help review PR together, thanks

ybc-alkaid:
Hi @binno, thanks for implementing SPMP support to Spike.
We really appreciate it!
Hi @aswaterman, as this version is more complete, I have closed #2244. Let's move forward with #2249.

disasm/isa_parser.cc
binno: OK

riscv/csrs.cc
binno: OK, I will rename **pmpaddr_csr_t** to **base_pmpaddr_csr_t**, and ### mpmpaddr_csr_t to **pmpaddr_csr_t**

riscv/processor.h
binno: To set mpmpdeleg.**pmpnum** can dynamically rearrange PMP entries & SPMP entries at run time. To easy to achieve this operation, I declare two entry structures for both PMP and SPMP to handle variations in length. 

If you have other good ideas, please give me suggestions, thanks.

c.c. @chihminchao @ybc-alkaid

aswaterman: The ISA was designed so that a HW implementation can use a single set of registers to hold all of the state.  If that's the case for a HW implementation, it should be true for Spike, too.  So let's do it that way (and rename the field to `pmpaddr`).

riscv/csrs.cc
```diff
@@ -286,6 +307,117 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
...
+
+spmpen_csr_t::spmpen_csr_t(processor_t* const proc, const reg_t addr, const reg_t init):
+  basic_csr_t(proc, addr, init) {
+}
+
+bool spmpen_csr_t::unlogged_write(const reg_t val) noexcept {
+  reg_t write_val = 0;
+  for (reg_t i = 0; i < (state->max_pmp - proc->n_pmp); ++i) {
+    if (!state->spmpaddr[i]->is_locked() && (val & (0x1 << i)))
+      write_val |= (0x1 << i);
```
aswaterman: Another `int` range bug; use `write_val |= reg_t(1) << i;`

riscv/csrs.cc
```diff
@@ -286,6 +307,117 @@ bool pmpcfg_csr_t::unlogged_write(const reg_t val) noexcept {
...
+}
+
+spmpen_csr_t::spmpen_csr_t(processor_t* const proc, const reg_t addr, const reg_t init):
+  basic_csr_t(proc, addr, init) {
+}
+
+bool spmpen_csr_t::unlogged_write(const reg_t val) noexcept {
+  reg_t write_val = 0;
+  for (reg_t i = 0; i < (state->max_pmp - proc->n_pmp); ++i) {
+    if (!state->spmpaddr[i]->is_locked() && (val & (0x1 << i)))
```
aswaterman: Another `int` range bug; use `((val >> i) & 1)`

riscv/mmu.h
```diff
@@ -529,7 +529,10 @@ class mmu_t
   }
 
   reg_t pmp_homogeneous(reg_t addr, reg_t len);
+  #define NO_IDX -1
```
aswaterman: This is a global macro in a heavily-included file.  We should avoid `#define`s with short names that pollute the global namespace.  Can you make it a `static const int` member of this class instead?

riscv/processor.h
binno: I will need time to develop it, thanks for suggestion.

riscv/csrs.cc
binno: OK, thanks

riscv/csrs.cc
binno: OK, thanks

riscv/mmu.h
binno: OK, thanks

riscv/csr_init.cc
```diff
@@ -572,4 +572,32 @@ void state_t::csr_init(processor_t* const proc, reg_t max_isa)
...
+    add_csr(CSR_MPMPDELEG, mpmpdeleg = std::make_shared<mpmpdeleg_csr_t>(proc, CSR_MPMPDELEG, MPMPDELEG_PMPNUM, proc->n_pmp));
+  }
+
+  if (proc->extension_enabled_const(EXT_SSPMP)) {
+    const reg_t spmp_cfg_mask = (SPMP_SHARED | SPMP_U | PMP_L | PMP_A | PMP_X | PMP_W | PMP_R);
+    // add entry registers (index: 0x0..0x3F)
+    for (size_t i = 0; i < max_pmp; ++i) {
+      spmpaddr[i] = std::make_shared<spmpaddr_csr_t>(proc, i);
+      dynamic_cast<sscsrind_reg_csr_t*>(csrmap[CSR_MIREG].get())->add_ireg_proxy(i, spmpaddr[i]);
+      dynamic_cast<sscsrind_reg_csr_t*>(csrmap[CSR_SIREG].get())->add_ireg_proxy(i, spmpaddr[i]);
```
ybc-alkaid: `dynamic_cast<sscsrind_reg_csr_t*>(csrmap[CSR_SIREG].get())` and `dynamic_cast<sscsrind_reg_csr_t*>(csrmap[CSR_SIREG2].get())` obtain the underlying indirect CSR objects. However, when Sscsrind is enabled, CSR_SIREG / CSR_SIREG2 are wrapped in `virtualized_indirect_csr_t` before being stored into csrmap. The dynamic_cast to `sscsrind_reg_csr_t*` fails (returns nullptr), and the subsequent `add_ireg_proxy()` call dereferences it, causing a seg fault.

Note: CSR_MIREG / CSR_MIREG2 are not affected because they are stored directly as `sscsrind_reg_csr_t` without wrapping.

ybc-alkaid:
@binno 
When running this PR with Sspmp enabled, a segfault occurred on my machine.

The reason can be:
> dynamic_cast<sscsrind_reg_csr_t*>(csrmap[CSR_SIREG].get())->add_ireg_proxy(i, spmpaddr[i]);

`dynamic_cast<sscsrind_reg_csr_t*>(csrmap[CSR_SIREG].get())` obtains the underlying indirect CSR objects. However, when Sscsrind is enabled, CSR_SIREG / CSR_SIREG2 are wrapped in `virtualized_indirect_csr_t` before being stored into csrmap. The dynamic_cast to `sscsrind_reg_csr_t*` fails (returns nullptr), and the subsequent `add_ireg_proxy()` call dereferences it, causing a seg fault.

Note: CSR_MIREG / CSR_MIREG2 are not affected because they are stored directly as `sscsrind_reg_csr_t` without wrapping.

binno:
> @binno When running this PR with Sspmp enabled, a segfault occurred on my machine.
> 
> The reason can be:
> 
> > dynamic_cast<sscsrind_reg_csr_t*>(csrmap[CSR_SIREG].get())->add_ireg_proxy(i, spmpaddr[i]);
> 
> `dynamic_cast<sscsrind_reg_csr_t*>(csrmap[CSR_SIREG].get())` obtains the underlying indirect CSR objects. However, when Sscsrind is enabled, CSR_SIREG / CSR_SIREG2 are wrapped in `virtualized_indirect_csr_t` before being stored into csrmap. The dynamic_cast to `sscsrind_reg_csr_t*` fails (returns nullptr), and the subsequent `add_ireg_proxy()` call dereferences it, causing a seg fault.
> 
> Note: CSR_MIREG / CSR_MIREG2 are not affected because they are stored directly as `sscsrind_reg_csr_t` without wrapping.

Thanks for reminder, I will take a look.

riscv/processor.h
binno: Hi @aswaterman ,
    I have updated the implementation with single entry resources to hold all of the state between PMP & SPMP.
Please help to review it, thanks.

riscv/csrs.h
```diff
@@ -89,11 +89,11 @@ class basic_csr_t: public csr_t {
...
 
-class pmpaddr_csr_t: public csr_t {
+class base_pmpaddr_csr_t: public csr_t {
  public:
-  pmpaddr_csr_t(processor_t* const proc, const reg_t addr);
-  virtual void verify_permissions(insn_t insn, bool write) const override;
-  virtual reg_t read() const noexcept override;
+  base_pmpaddr_csr_t(processor_t* const proc, const reg_t addr, const size_t idx);
+
+  virtual reg_t read() const noexcept override { return 0; }
```
aswaterman: Is this a function that's expected to never be called?  I prefer `{ abort(); }` instead for better self-documentation.  (Alternatively, is there any reason this method can't be left unimplemented, making this an abstract class?)

riscv/mmu.cc
```diff
@@ -491,21 +497,19 @@ tlb_entry_t mmu_t::refill_tlb(reg_t vaddr, reg_t paddr, char* host_addr, access_
   return entry;
 }
 
-bool mmu_t::pmp_ok(reg_t addr, reg_t len, access_type type, reg_t mode, bool hlvx)
+bool mmu_t::pmp_search(reg_t addr, reg_t len, size_t start, size_t pmp_num, int &index)
```
aswaterman: Instead of returning a bool and passing an integer by reference, it would be cleaner to do make the return value an `std::optional<size_t>`, where the option is empty if the return value would have been false.

riscv/mmu.cc
```diff
@@ -515,15 +519,48 @@ bool mmu_t::pmp_ok(reg_t addr, reg_t len, access_type type, reg_t mode, bool hlv
...
+
+bool mmu_t::pmp_ok(reg_t addr, reg_t len, access_type type, reg_t mode, bool hlvx)
+{
+  if (!proc || proc->n_pmp == 0)
+    return true;
+
+  int pmp_idx = NO_IDX;
+  if (pmp_search(addr, len, 0, proc->n_pmp, pmp_idx)) {
+    if (pmp_idx != NO_IDX)
+      return dynamic_cast<pmpaddr_csr_t*>(proc->state.pmpaddr[pmp_idx].get())->access_ok(type, mode, hlvx);
```
aswaterman: And we can clean this up if we do https://github.com/riscv-software-src/riscv-isa-sim/pull/2249/changes#r3245876532 

something like:

```
  if (auto pmp = pmp_search(addr, len, 0, proc->n_pmp); pmp.has_value()))
    return dynamic_cast<pmpaddr_csr_t*>(proc->state.pmpaddr[*pmp].get())->access_ok(type, mode, hlvx);
  return false;
```

And then we can get rid of `NO_IDX`, too.

aswaterman (commented):
Thanks @binno, looks good to me, I'll merge it once you address my last round of feedback (and once you've fixed the CI failure).

riscv/csrs.h
binno: Thanks for suggestion. I make this an abstract class.

riscv/mmu.cc
binno: Thanks for suggestion

riscv/mmu.cc
binno: Thanks for suggestion

aswaterman (approved):
I'm going to play around with the `pmp_search` code a bit more to see if I can make it more concise, but I think this PR can move forward in the meantime.  Thanks @binno!

PR #2251, opened 2026-03-22
make the executable path relative from spike

hirooih:
Currently, the execution path (for pk and bbl) is determined by the values of the `--prefix` and `--with-target` options specified during the Spike build: `/<prefix>/<target-path>/bin`.
However, this does not work when the binary distribution is placed in a different directory.

This pull request (PR) introduces the use of a relative path to Spike: `.../bin/../<target-path>/bin`.

This PR is supported on Linux and macOS. On other platforms, the behaviour remains as it is currently.

I have avoided cluttering the code by not trying to support a number of non-uniform cases, as can be seen at https://stackoverflow.com/questions/1023306/finding-current-executables-path-without-proc-self-exe.
Note that `../configure ... CFLAGS=\"PROC_SELF_EXE=/proc/.../..\"` should works for FreeBSD variations.

### DIFF
--- a/fesvr/htif.cc
+++ b/fesvr/htif.cc
@@ -20,6 +20,11 @@
 #include <unistd.h>
 #include <signal.h>
 #include <getopt.h>
+#include <libgen.h>
+#include <limits.h>
+#ifdef __APPLE__
+#include <mach-o/dyld.h>
+#endif
 
 /* Attempt to determine the execution prefix automatically.  autoconf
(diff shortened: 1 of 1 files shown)


aswaterman:
Just to confirm: the behavior is unchanged if Spike is installed into the same directory as the `--prefix` setting?

hirooih:
Thank you for your reviewing.

> the behavior is unchanged if Spike is installed into the same directory as the --prefix setting?

Yes, it is unchanged.

aswaterman (approved):

PR #2253, opened 2026-03-24
Add Zvzip extension.

5265325:
This patch adds support for the zvzip extension.
See https://github.com/ved-rivos/riscv-isa-manual/blob/zvzip/src/zvzip.adoc


aswaterman (commented):
@nibrunieAtSi5 can you review this one?  Structurally, it looks OK to me.

nibrunieAtSi5 (approved):
@aswaterman PR LGTM so far, it seems all the spec constraints have been implemented and the functionality looks correct (I have not performed any functional testing yet).

@5265325, the only nit pick I have is why call the single-input macro `VZIP_V` (and not `VUNZIP_V` or extract or ...) when it is only used to implement unzip ? (I agree that the functionalities are symmetrical and it definitely makes sense to use the `VZIP_VV` macro for the VPAIR definition)

5265325:
@nibrunieAtSi5 Thanks. Now I modified VZIP_V prefix to ZVZIP_EXTRACT_V. This name is more reasonable.

aswaterman (approved):
Thank you, @5265325!

PR #2254, opened 2026-03-24
Add Svvptc extension.

5265325:
This patch adds support for the svvptc extension.
See https://github.com/riscv/riscv-svvptc/blob/main/svvptc.adoc


aswaterman (approved):
I agree with the approach--accept the ISA string, but take no action on it.  But you can make another simplification.  Delete `EXT_SVVPTC` and make this a one-liner.  It's just dead code.

Ping me when you've made the change.

aswaterman (approved):

PR #2256, opened 2026-03-28
Add support for Za64rs and Za128rs extensions.

5265325:
This patch adds support for Za64rs and Za128rs extensions.
See https://github.com/riscv/riscv-profiles/blob/main/src/rva23-profile.adoc

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Spike supports the following RISC-V ISA features:
   - Zabha extension, v1.0
   - Zacas extension, v1.0
   - Zawrs extension, v1.0
+  - Za64rs, Za128rs extension, v1.0
   - Zicfiss extension, v1.0
   - Zicfilp extension, v1.0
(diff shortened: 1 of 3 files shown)


aswaterman (changes requested):
This is another case where we can accept the ISA strings, but then not change the simulator behavior at all.  The reasoning is more subtle in this case, though.

Recall that the ISA allows for SC to fail if the reserved address does not exactly match the LR address--regardless of the reservation-set size.  So, Spike's behavior already conforms to both Za64rs and Za128rs, without any changes.

The ZaNrs extensions exist to describe when other harts will not interfere with a reservation.  Since Spike is single-threaded and stays on the same hart for a long time (longer than the longest possible constrained LR/SC sequence), it can't detectably run afoul of ZaNrs.

aswaterman (approved):

PR #2261, opened 2026-04-02
Bulk normalization granularity is based on VLMAX, not vl

aswaterman:


riscv/v_ext_macros.h
```diff
@@ -2245,8 +2245,8 @@ c_t generic_dot_product(const std::vector<a_t>& a, const std::vector<b_t>& b, c_
 }
 
 #define ZVLDOT_LOOP(a_t, b_t, c_t, dot) \
-  std::vector<a_t> a(P.VU.vl->read(), a_t()); \
-  std::vector<b_t> b(P.VU.vl->read(), b_t()); \
+  std::vector<a_t> a(P.VU.vlmax, a_t()); \
+  std::vector<b_t> b(P.VU.vlmax, b_t()); \
   for (reg_t i = 0; i < a.size(); i++) { \
```
nibrunieAtSi5: I think we need to replace `a.size()` with `P.VU.vl->read()`

PR #2265, opened 2026-04-05
Fix duplicate addition of mop_rr_7

bwv846:
Please review the commit.


aswaterman (approved):
I think this could probably be refined further to leverage the decoder's priority ordering.  We could just register the MOPs unconditionally, provided the extensions that redefine the MOPs precede them in this list.  But for now, this is OK.

PR #2266, opened 2026-04-06
Implement Zama16b extension

### LINKED ISSUE #2262: Regarding the priority between the trigger(store address) and the AMO access fault.
According to the documentation, the exception priority of a store address trigger should be higher than that of a store access fault; however, in the code implementation, the store access fault clearly has higher priority than the trigger.

Image

Image

Image
It checks the store exception first,  the trigger check is masked.

aswaterman:
Allow misaligned AMOs within 16-byte region when this extension is present.

Fixes #2262 along the way.


riscv/mmu.cc
```diff
@@ -377,8 +377,9 @@ void mmu_t::store_slow_path(reg_t original_addr, std::size_t len,
     auto [tlb_hit, host_addr, paddr] = access_tlb(tlb_store, original_addr, TLB_FLAGS & ~TLB_CHECK_TRIGGERS);
     bool intrapage = (original_addr % PGSIZE) + len <= PGSIZE;
     bool aligned = (original_addr & (len - 1)) == 0;
+    bool misaligned_ok = !require_alignment && intrapage && is_misaligned_enabled();
```
pmundkur, line 380: The `UNUSED` attribute on the `require_alignment` parameter could be removed.

PR #2274, opened 2026-04-13
feat: make chunk_max_size non-virtual

arromanoff:
This allows to completely eliminate virtual call overhead for functions that usually return a constant number

On my benchmarks this gives about 2-3% performance increase

### DIFF
--- a/fesvr/dtm.cc
+++ b/fesvr/dtm.cc
@@ -213,11 +213,6 @@ uint32_t dtm_t::run_abstract_command(uint32_t command,
 
 }
 
-size_t dtm_t::chunk_align()
-{
-  return xlen / 8;
-}
-
 void dtm_t::read_chunk(uint64_t taddr, size_t len, void* dst)
 {
   uint32_t prog[MAX_PROG_WORDS];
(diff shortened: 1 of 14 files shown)


aswaterman (commented):
I'm leery of this, because it suggests that the chunk size will always be known at `htif_t` construction time.  I think some downstream users might disagree.

Looking at the code, I see that there is ample opportunity to reduce the number of calls to these functions without changing the API.  For example, in `clear_chunk`, we call `chunk_max_size` 2C+1 times, where only 1 would suffice.  Maybe just factor out the calls?  I suspect this will obtain nearly all of the benefit.

As another alternative, we could cache the values inside of the `htif_t` object--e.g. something like

```
void htif_t::chunk_max_size_internal()
{
  if (!chunk_max_size_cached)
    chunk_max_size_cached = chunk_max_size();
  return chunk_max_size_cached;
}
```

but I think this is also inelegant and so shouldn't be done unless there's something problematic about my first suggestion.

arromanoff:
I see that there are indeed a lot of excessive virtual calls happening. After testing I see that this actually does improve performance as much. I'll create an MR as soon as I have time. I'll keep this one open for the time being but I guess it won't be necessary

arromanoff:
Created https://github.com/riscv-software-src/riscv-isa-sim/pull/2278

arromanoff:
I tested it with https://github.com/riscv-software-src/riscv-isa-sim/pull/2278 and it actually doesn't make any measurable performance improvement. Closing as redundant.

Thanks to @aswaterman for the idea

PR #2275, opened 2026-04-14
[#2273][bulknormdot] fixing spurious invalid flag in bulk_norm_dot_no_mult

### LINKED ISSUE #2273: question on invalid_flag for bulknormaldot
Hi,

I got a question on invalid flag calculation in the bulknormaldot.h file. Normally, a pos/neg infinite number should be non-NAN and non-Zero. Currently, I am performing instruction vfwldot.vv with sew=16 and lmul=1, and a=0xffd8 b=0x7f80 as one pair of the src。 For the src B, both any_neg_inf and nan are set to 1, which later resulting in invalid_flag being set(as there are other pos_inf srcs ). In my opinion, !NAN should be checked when setting any_pos/neg_inf.

I am not sure if I misunderstood this part. Please correct me if I am wrong.

[](url)

nibrunieAtSi5:
Fixes #2273 

Bulk normalized dot product is expected to raise the invalid flag if two products with infinite values and opposite signs are added. Spike function failed to consider the second operand of a product when determining whether the product had infinite value or not, and missed the exclusion of NaN and zero from this case.

The zero case is quiet, because it happens for inf * zero which is expected to raise the invalid flag. The qNaN case is not quiet and can cause the apparition of a spurious invalid flag when none were expected.

### DIFF
--- a/riscv/bulknormdot.h
+++ b/riscv/bulknormdot.h
@@ -203,13 +203,18 @@ template<typename ValueTypeLHS, typename ValueTypeRHS, typename SigProdType> bul
                          a[i].isZero() || b[i].isZero() ? (f32_exp_bias - (lhs_bias + rhs_bias)) : // minimalize exp of zero product
                          a[i].expSubFixed() + b[i].expSubFixed() + (f32_exp_bias - (lhs_bias + rhs_bias));
 
+    bool a_is_zero = (a[i].subOrZero() && cfg.flushSub) || a[i].isZero(); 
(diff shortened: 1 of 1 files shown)


aswaterman (approved):
LGTM; ping me again when it is ready to merge.

nibrunieAtSi5:
@Vivvvvvvv would you mind reviewing this PR ?

Vivvvvvvv:
> @Vivvvvvvv would you mind reviewing this PR ?

This looks good to me, thanks

aswaterman (approved):

PR #2278, opened 2026-04-15
refactor: remove excessive chunk_max_size calls

arromanoff:
These calls are virtual and cause notable overhead


aswaterman (approved):
LGTM, thanks.

PR #2279, opened 2026-04-20
change interactive input stderr to stdin

Happends:
interactive read from input from sterr, leads to input problems when reading from stderr


aswaterman:
The current scheme allows interactive mode to be used in programs that read from stdin (usually for piped I/O). Neither is perfect but we rely on the current scheme.

PR #2282, opened 2026-04-21
added is_client_connected method to remote_bitbang class

v-efimov:
This "is_client_connected" method can be checked from an external "sim" loop to see if there is still active "OpenOCD" tcp session exists. This check helps to decide when the simulation can be gracefully finished and no further polling of "remote_bitbang->tick()" needed.


aswaterman (approved):
LGTM.

PR #2287, opened 2026-04-30
Refactor ISA parser extension handling

jerryzj:
Replace the long hard-coded extension dispatch chain in isa_parser.cc with table-driven extension metadata. The table now records the extension name, the extension bits it enables, direct implied extensions, and simple XLEN requirements. This makes the parser behavior closer to LLVM, where extension parsing and implication are represented as data rather than as a large sequence of adhoc conditionals.

Move MISA single-letter extensions into the same table as multi-letter extensions. The parser now expands `g` to `imafd`, validates that the base extension is `i`, `e`, or `g`, and then lets the normal single-letter parsing loop consume the base extension through the table.  This also centralizes common MISA implications such as `a -> zaamo,zalrsc`, `b -> zba,zbb,zbs`, `d -> f`, `q -> d,f`, `c -> zca`, and `v -> zve64d,zvl128b`.

Model the Zve implication chain explicitly in the extension table

  zve32f -> zve32x, f
  zve32x -> zicsr, zvl32b
  zve64d -> zve64f, d
  zve64f -> zve32f, zve64x
  zve64x -> zve32x, zvl64b
  v      -> zve64d, zvl128b

Keep the Spike-specific Zve state updates out of the table by deriving VLEN, ELEN, zvf, and zvd from the Zve extension name in a small helper.

Add a separate combine pass for reverse combination rules that reconstruct combined extension bits from their component bits. This keeps combine logic separate from forward implication while preserving existing Spike behavior:

  zaamo + zalrsc -> A
  zba + zbb + zbs -> B
  C + D -> Zcd
  Zicfiss + Zca -> Zcmop

Retain explicit Spike-specific post-parse checks and derived state where they remain clearer than table entries, including RV32-only `C/Zce + F -> Zcf`, `Zfbfmin/Zfhmin -> EXT_INTERNAL_ZFH_MOVE`, incompatibility checks, privilege mode handling, and VLEN limit validation.

### DIFF
--- a/disasm/isa_parser.cc
+++ b/disasm/isa_parser.cc
@@ -1,6 +1,10 @@
 #include "isa_parser.h"
 #include <cstring>
+#include <initializer_list>
 #include <stdexcept>
+#include <vector>
+
+namespace {
 
 static std::string strtolower(const char* str)
 {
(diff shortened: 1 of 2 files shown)


aswaterman (commented):
@jerryzj are we confident this is semantically equivalent?

@jerryz123 can I ask you to take a first crack at reviewing this one?

jerryzj:
> @jerryzj are we confident this is semantically equivalent?
> 
> @jerryz123 can I ask you to take a first crack at reviewing this one?

I think I can build some assertion tests, maybe put them under `ci-tests`?

aswaterman:
@jerryzj I don't think adding more ci-tests is the best use of engineering effort in this case, provided you feel like you've done sufficient local testing.

jerryzj:
> @jerryzj I don't think adding more ci-tests is the best use of engineering effort in this case, provided you feel like you've done sufficient local testing.

Makes sense. I did local validation instead.

The local differential test to use ISA strings extracted from the `riscv-isa-tools` test corpus (which is SiFive's internal tool). That caught one real semantic regression in the refactor: `rv64ie` was accepted by the new parser but rejected by the old parser. I fixed that by restoring the old base-ISA parsing boundary: consume exactly one of I/E/G first, then parse the remaining ordered single-letter extensions.

After that fix, I reran the differential check against the pre-refactor parser:

- 427 unique generated/targeted candidate ISA strings checked
- 354 accepted by both parsers and compared by full parser state
- 73 rejected by both parsers
- 4 additional explicit invalid/conflict cases checked

For accepted strings, the comparison includes canonical ISA string, max_xlen, max_isa, vlen, elen, zvf/zvd, full extension_table bitset, and extension-name set.

In conclusion, I now have stronger local evidence that the refactor is semantically equivalent.

arromanoff:
Our snippy-based tests should check at least a part of ISA-parsing as it uses many different `--isa`s

aswaterman:
Thanks, that sounds thorough.  I'll give @jerryz123 a chance to chime in.

jerryz123:
Will take a look tonight

disasm/isa_parser.cc
```diff
@@ -48,315 +52,243 @@ static void bad_priv_string(const char* priv)
   abort();
 }
 
+struct extension_info_t {
+  const char* name;
+  std::vector<unsigned> enables;
+  std::vector<const char*> implies;
+
+  extension_info_t(const char* name,
```
jerryz123, line 62:
```suggestion
  // `enables` sets extension_table bits directly
  // `implies` recursively parses and applies other named extensions
  extension_info_t(const char* name,
```

disasm/isa_parser.cc
```diff
@@ -48,315 +52,243 @@ static void bad_priv_string(const char* priv)
...
+                   std::initializer_list<unsigned> enables = {},
+                   std::initializer_list<const char*> implies = {})
+    : name(name), enables(enables), implies(implies) {}
+};
+
+struct extension_combination_t {
```
jerryz123, line 69:
```suggestion
// if every extension in components is set, also set the enable
struct extension_combination_t {
```

disasm/isa_parser.cc
```diff
@@ -48,315 +52,243 @@ static void bad_priv_string(const char* priv)
...
+  unsigned enable;
+  std::vector<unsigned> components;
+
+  extension_combination_t(unsigned enable,
+                          std::initializer_list<unsigned> components = {})
+    : enable(enable), components(components) {}
+};
+
+// Entries without enables are accepted ISA names whose behavior is always present
+// in Spike, currently unmodeled, or handled by later parser checks.
```
jerryz123: Is this comment accurate? It looks like something like `zve32f` does turn on an extension through `implies`

jerryz123 (approved):
Pretty neat 👍 . Just some commentary nits

disasm/isa_parser.cc
jerryzj: ok, let me update the comments.

jerryzj:
> Pretty neat 👍 . Just some commentary nits

Thanks, I just pushed another fixup commit to address them

aswaterman (approved):
Expressing the extensions and their dependencies as a data structure rather than static code is much cleaner.  Thanks.  (And thanks for the review, @jerryz123.)

PR #2288, opened 2026-05-01
Build: make object and depfile outputs explicit

cmuellner:
The compile rules currently rely on the compiler to infer object and depfile names from the source pathname when -o and -MF are not passed explicitly.

That works for plain compiler invocations. However, older ccache versions can rewrite a symlinked source path to its target path before deriving the default output names. In that case, compiling spike_dasm_option_parser.cc may create option_parser.o and option_parser.d instead of the declared target filenames, which breaks the build.

Upstream ccache fixed this in v4.12 (in commit `8810fadab0b59`), but older versions remain common in distribution and CI environments (Ubuntu 22 or 24). Pass explicit `-o $@` and matching `-MF` paths in the generic C, C++, test, program, install-program, and precompiled-header rules.

This keeps each recipe output aligned with the declared make target and avoids depending on inferred filenames.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -219,11 +219,11 @@ $(2)_deps := $$(patsubst %.o, %.d, $$($(2)_objs))
 $(2)_deps += $$(patsubst %.o, %.d, $$($(2)_c_objs))
 $(2)_deps += $$(patsubst %.h, %.h.d, $$($(2)_precompiled_hdrs))
 $$($(2)_pch) : %.h.gch : %.h
-	$(COMPILE) -x c++-header $$< -c -o $$@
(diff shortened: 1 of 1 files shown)


cmuellner:
Context: this popped up when introducing ccache to the riscv-gnu-toolchain repo (PR in https://github.com/riscv-collab/riscv-gnu-toolchain/pull/1852 and error log in https://github.com/riscv-collab/riscv-gnu-toolchain/actions/runs/25214468625/job/73938971188?pr=1852).

aswaterman (approved):
LGTM, thanks @cmuellner

PR #2295, opened 2026-05-13
Add missing set_fp_exceptions in Zvdota/Zvbdota implementations

### LINKED ISSUE #2294: Unexpected fp exception flags in Zvdota/Zvbdota implementations
According to the Zvdota/Zvbdota isa spec (commit https://github.com/riscv/riscv-isa-manual/pull/2618/changes/d2702c74ac286fafcd9e8f33f1565d756fa451bd), RVBNA may raises the invalid operation and overflow exceptions.

I’ve created some simple test cases that follow the pattern specified in the spec `Invalid operation` section, expecting NV=1, but the NV value remains 0 in the current spike implementation (commit c567a19). 

> 32.4.6.1. Invalid operation
> The invalid operation flag must be raised if at least one of the following conditions is met:
> - Any of the operands is a signaling NaN
> - At least one of the following conditions:
>     - there are at least two products that are infinites with opposite sign.
>     - there is at least one product between a zero and an infinity.

rnax:
Fixes https://github.com/riscv-software-src/riscv-isa-sim/issues/2294

According to the Zvdota/Zvbdota isa spec, RVBNA may raises invalid/overflow exceptions.

This PR adds the missing set_fp_exceptions after RVBNA in ZVLDOT_LOOP and ZVBDOT_LOOP to propagate softfloat exception flags to the processor state. Without this, FP exceptions raised during RVBNA accumulation are silently dropped.

### DIFF
--- a/riscv/v_ext_macros.h
+++ b/riscv/v_ext_macros.h
@@ -2253,7 +2253,8 @@ c_t generic_dot_product(const std::vector<a_t>& a, const std::vector<b_t>& b, c_
     b[i] = P.VU.elt<b_t>(insn.rs2(), i); \
   } \
   auto& acc = P.VU.elt<c_t>(insn.rd(), 0, true); \
-  acc = dot(a, b, acc)
+  acc = dot(a, b, acc); \
+  set_fp_exceptions;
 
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

aswaterman (approved):
LGTM, thanks.

PR #2297, opened 2026-05-15
Report CBO.ZERO faults on effective address, not base address

### LINKED ISSUE #2296: mtval reports cache-block-aligned address on CBO access fault instead of faulting virtual address
When a CBO instruction (`cbo.zero`, `cbo.clean`, `cbo.flush`, `cbo.inval`) traps with a store/AMO access fault, Spike writes the cache-block-aligned address into `mtval` rather than the raw faulting address from `rs1`.

According to the RISC-V Privileged Specification:

> A cbo.zero instruction performs stores of zeros to the full set of bytes corresponding to the cache block that contains the address specified in rs1. It is not required that rs1 is aligned to the size of a cache block. On faults, the faulting virtual address is considered to be the value in rs1, rather than the base address of the cache block.

The "faulting address" for a CBO instruction is the address in `rs1`, not the derived cache-block-aligned address. The cache block alignment is an internal implementation detail of how the CBO instruction operates, but the faulting address that triggered the exception is the original value from `rs1`.

**Example:**

Instruction: `cbo.zero (x8)` where `x8 = 0xDF56FF76DF56FF76`
Cache block size: 64 bytes (`cache_block_size_exp = 6`)

**Spike:**
```
core   0: 0x80000100 (0x0044200f) cbo.zero (s0)
core   0: exception trap_store_access_fault, epc 0x80000100
core   0:           tval 0xdf56ff76df56ff40   <-- aligned to 64-byte boundary
```

**Sail:**
```
0x80000100 (0x0044200F) cbo.zero (x8)
trapping from M to M to handle store/amo-access-fault
mtval = 0xDF56FF76DF56FF76   <-- raw rs1 value
```

Spike computes `0xDF56FF76DF56FF76 & ~0x3F = 0xDF56FF76DF56FF40` and reports that in `mtval`. Sail reports the unmodified `rs1` value.

**Expected behavior:**

`mtval` should contain `0xDF56FF76DF56FF76` (the virtual address from `rs1`), not the cache-block-aligned address. The spec does not say the address should be aligned before being written to `mtval`.

aswaterman:
Also restricts CBO.ZERO to main-memory regions, as is allowed.

Fixes #2296

### DIFF
--- a/riscv/mmu.h
+++ b/riscv/mmu.h
@@ -227,10 +227,16 @@ public:
     auto access_info = generate_access_info(addr, STORE, {});
     reg_t transformed_addr = access_info.transformed_vaddr;
 
-    auto base = transformed_addr & ~(blocksz - 1);
     check_triggers(triggers::OPERATION_STORE, transformed_addr, false, blocksz);
-    for (size_t offset = 0; offset < blocksz; offset += 1)
-      store<uint8_t>(base + offset, 0);
+
+    reg_t paddr = translate(access_info, 1);
+    if (auto host_addr = sim->addr_to_mem(paddr)) {
+      if (tracer.interested_in_range(paddr, paddr + PGSIZE, STORE))
+        tracer.trace(paddr - (transformed_addr & (blocksz - 1)), blocksz, STORE);
+      memset(host_addr - (transformed_addr & (blocksz - 1)), 0, blocksz);
+    } else {
+      throw trap_store_access_fault((proc) ? proc->state.v : false, transformed_addr, 0, 0);
+    }
   }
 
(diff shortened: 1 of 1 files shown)


UmerShahidengr (approved):
This one has fixed #2296. Please merge this one

aswaterman:
Thanks @UmerShahidengr.

aap-sc:
@aswaterman , @UmerShahidengr it looks like that this commit introduce a regression , since now we match against non-aligned base, causing the comparison to be outside of the intended block. I've created an MR to address this: https://github.com/riscv-software-src/riscv-isa-sim/pull/2443 , could you please take a look?

PR #2299, opened 2026-05-15
Add SPMP

aswaterman:
This is a redux of #2249 with a bug fix and some tweaks.

@binno the debug tests were failing because of this change: https://github.com/riscv-software-src/riscv-isa-sim/pull/2249/changes#diff-83b9a4f31daff27d009b12bb8f572dd539c76354f506c638be46beb8c8c218b1R332

All PMPs need to be created at this point.

If there are any follow-up bug fixes, let's do them as part of a second PR.


aswaterman:
cc @binno

binno:
@aswaterman,
    Many thanks for your effort to debug the failure of debug test and refine code.

PR #2300, opened 2026-05-19
Fix Incorrect Sign Extension in vmulhu.vx, vmulhsu.vx, vdivu.vx, and vremu.vx

fourth-bit:
In vmulhu.vx, vmulhsu.vx, vdivu.vx, and vremu.vx, the scalar operand is being incorrectly sign-extended when xlen = 32, leading to the mismatches observed here: https://github.com/riscv/riscv-arch-test/issues/1474. This patch changes the casts in each of these operations so that the values are handled correctly in both xlen = 64 and xlen = 32 cases.

### DIFF
--- a/riscv/insns/vdivu_vx.h
+++ b/riscv/insns/vdivu_vx.h
@@ -3,6 +3,8 @@ VI_VX_ULOOP
 ({
   if (rs1 == 0)
     vd = -1;
+  else if (xlen == 64)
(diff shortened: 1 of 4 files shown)


fourth-bit:
Never mind, spec says that sign extension is valid here

aswaterman:
Yep, confirmed that sign extension is required.

PR #2301, opened 2026-05-20
Emit initrd addresses as 64-bit DT cells

5265325:
The root node advertises #address-cells = <2>, and memory nodes already encode addresses as high/low 32-bit cells. However, linux,initrd-start and linux,initrd-end were emitted as a single cell, which truncates initrd addresses when the memory region is placed above 4 GiB.

Emit these chosen properties using the same high/low cell representation so Linux can boot with initrd images located in high memory.


aswaterman (approved):
LGTM.

PR #2322, opened 2026-06-23
relax pmp checks for RWX accesses to debug ROM in debug mode

EccoTheDolphin:
Prior this patch it was impossible to enter debug mode when mseccfg.MML was set. The reason is that debug ROM requires both RW and X permissions for debug module to function properly (for example for progbuf execution).

This patch relaxes pmp checks and allows unconditional access to debug firmware area whenever the processor is in debug state.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -158,11 +158,10 @@ static reg_t reg_from_bytes(size_t len, const uint8_t* bytes)
   return res;
 }
 
-bool mmu_t::mmio_ok(reg_t paddr, access_type UNUSED type)
+bool mmu_t::mmio_ok(reg_t paddr, size_t len, access_type UNUSED type)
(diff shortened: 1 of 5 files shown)


EccoTheDolphin:
@aswaterman could you take a look, please?

Also, it seems that it can be feasible to drop `mmu_t::mmio_ok`-related checks. If no objections, I would do that in a subsequent commit.

@fkhaidari , @nadime15 , @wissygh FYI.

aswaterman (approved):
Seems like a reasonable solution to me.

aswaterman:
> Also, it seems that it can be feasible to drop `mmu_t::mmio_ok`-related checks. If no objections, I would do that in a subsequent commit.

@EccoTheDolphin Yeah, let's discuss that change separately.

PR #2324, opened 2026-06-24
Switch exception bits by correspond extensions

binno:
if zicfi extension is enabled, software check fault existed.
if zicntr extension is enabled, hardware error fault existed.


aswaterman:
What bug is this fixing?  I don't think Spike is doing anything wrong here.

binno:
> What bug is this fixing? I don't think Spike is doing anything wrong here.

I think this is not bug but more like the following operation:
  If the Zca extension isn't enabled,  misaligned fetch fault never happened
So the corresponding bit of medeleg is write ignored and mask out
```
bool medeleg_csr_t::unlogged_write(const reg_t val) noexcept {
  const reg_t mask = 0
    | (proc->extension_enabled(EXT_ZCA) ? 0 : 1 << CAUSE_MISALIGNED_FETCH)
```

aswaterman (approved):

aswaterman (approved):

Issue #2329, opened 2026-06-25
Add Ziccrse to disassembler


mslijepc:
As discussed in https://github.com/riscv-software-src/riscv-isa-sim/issues/2323

PR #2330, opened 2026-06-25
Add missing Ziccrse extension to disassembler fallback

mslijepc:
[Ziccrse](https://github.com/riscv/riscv-unified-db/blob/d5c88128ad7f3f3367fc58d5d85bf956c24870c3/spec/std/isa/ext/Ziccrse.yaml) is implemented in Spike unconditionally/


mslijepc:
Closes https://github.com/riscv-software-src/riscv-isa-sim/issues/2329

aswaterman (approved):

PR #2331, opened 2026-06-25
Fix name typo for zvqwbdota8i/zvqwbdota16i extension

binno:
As title


disasm/isa_parser.cc
```diff
@@ -193,7 +193,7 @@ static const extension_info_t extension_infos[] = {
   {"zvksed", {EXT_ZVKSED}},
   {"zvksh", {EXT_ZVKSH}},
   {"zvqdotq", {EXT_ZVQDOTQ}},
-  {"zvqbwdota8i", {EXT_ZVQWBDOTA8I}},
+  {"zvqwbdota8i", {EXT_ZVQWBDOTA8I}},
   {"zvqbwdota16i", {EXT_ZVQWBDOTA16I}},
```
chihminchao: I think it should be "zvqwbdota16i"

it is my fault in previous PR
reference
https://github.com/aswaterman/riscv-misc/blob/main/isa/ldot-bdot/ldot-bdot.adoc

aswaterman: @binno can you fix this one too and then ping me?

binno: @aswaterman , 
    Done. please help review it again, thanks

aswaterman (approved):

PR #2340, opened 2026-07-11
perf: add optimized Clang builds and vector fast paths

apullin:
Spike is faster when compiled with Clang 21 and ThinLTO. The build system is updated to prefer that configuration for native builds when it is available, with a fallback to the normal GCC build.

The vector fast-path change uses each generated wrapper's compile-time logging mode to bypass status, CSR, and destination logging when commit logging is disabled. Logged wrappers retain architectural commit tracking.

| Benchmark | Speedup |
|---|---:|
| `vector_mix` | 38.55% |
| `vector_mask_mix` | 31.06% |
| `branch_mix` | 7.92% |
| `int_mix` | 7.08% |

These are isolated measurements of the vector fast-path commit using 16 paired runs on an AMD EPYC 7F52 host.

This PR intentionally contains two commits: one build-system commit and one vector fast-path commit.
The Clang 21 + ThinLTO build substantially improves several scalar, FPU, and CSR lanes, but regresses the vector workloads. So this PR pairs it with a vector fast-path change that skips logging work when commit logging is already disabled. Logged execution remains unchanged. In isolated measurements, this change improved vector_mix by 38.55% and vector_mask_mix by 31.06%, offsetting the main Clang + ThinLTO regressions.

### DIFF
--- a/riscv/csrs.h
+++ b/riscv/csrs.h
@@ -789,6 +789,8 @@ class vector_csr_t: public basic_csr_t {
   virtual void verify_permissions(insn_t insn, bool write) const override;
   // Write without regard to mask, and without touching mstatus.VS
   void write_raw(const reg_t val) noexcept;
+  // Internal write for instruction implementations with commit logging disabled.
+  void write_without_logging(const reg_t val) noexcept;
  protected:
   virtual bool unlogged_write(const reg_t val) noexcept override;
  private:
--- a/riscv/csrs.h
+++ b/riscv/csrs.h
@@ -789,6 +789,17 @@ class vector_csr_t: public basic_csr_t {
   virtual void verify_permissions(insn_t insn, bool write) const override;
   // Write without regard to mask, and without touching mstatus.VS
   void write_raw(const reg_t val) noexcept;
+  // Internal write for instruction implementations with commit logging disabled.
+  void write_without_logging(const reg_t val) noexcept;
+  // Internal write for instruction implementations with a compile-time-known
+  // commit logging mode.
+  template <bool log>
+  void write_internal(const reg_t val) noexcept {
+    if constexpr (log)
+      write(val);
+    else
+      write_without_logging(val);
+  }
  protected:
   virtual bool unlogged_write(const reg_t val) noexcept override;
  private:
--- a/riscv/v_ext_macros.h
+++ b/riscv/v_ext_macros.h
@@ -7,6 +7,13 @@
 #include "zvbdot.h"
 #include <functional>
 
+#define WRITE_VSTART(value) do { \
+  if constexpr (DECODE_MACRO_USAGE_LOGGED) \
+    P.VU.vstart->write(value); \
+  else \
+    P.VU.vstart->write_without_logging(value); \
+} while (0)
+
 //
 // vector: masking skip helper
 //
@@ -240,7 +247,7 @@ static inline bool is_overlapped_widen(const int astart, int asize,
  }
 
 #define VECTOR_END \
-  P.VU.vstart->write(0)
+  WRITE_VSTART(0)
 
 #define VI_LOOP_END \
   VI_LOOP_END_BASE \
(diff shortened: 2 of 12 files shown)


riscv/v_ext_macros.h
```diff
@@ -7,6 +7,13 @@
 #include "zvbdot.h"
 #include <functional>
 
+#define WRITE_VSTART(value) do { \
+  if constexpr (DECODE_MACRO_USAGE_LOGGED) \
+    P.VU.vstart->write(value); \
+  else \
+    P.VU.vstart->write_without_logging(value); \
+} while (0)
+
```
aswaterman: I'm not wild about this design pattern.  I suggest backing this part out and only including the straightforward `DECODE_MACRO_USAGE_LOGGED` substitutions for the moment.

aswaterman:
@apullin Sorry for taking a long while to revisit this topic.  I'm cool with this PR modulo my comment above, and I'll try to review future ones quite a bit more quickly.

riscv/v_ext_macros.h
apullin: This wrapper is just to elide logging. Which is ~1/3 of the vector speedup.
But I get what you mean.

Alternate idea is to change to compile-time specialization,
```
template <bool log>
void write_internal(reg_t value) noexcept;
  
constexpr bool log = DECODE_MACRO_USAGE_LOGGED;
```

And then call sites can become:
```
P.VU.vstart->write_internal<log>(0);
P.VU.elt<T>(rd_num, i, log);
```  

I will try to rework it that way.

apullin:
Reworked the macro into compile-time specialization, rebased onto current master.
The vector speedup is recovered to within noise. Ready for review.

riscv/csrs.h
```diff
@@ -789,6 +789,17 @@ class vector_csr_t: public basic_csr_t {
   virtual void verify_permissions(insn_t insn, bool write) const override;
   // Write without regard to mask, and without touching mstatus.VS
   void write_raw(const reg_t val) noexcept;
+  // Internal write for instruction implementations with commit logging disabled.
+  void write_without_logging(const reg_t val) noexcept;
```
aswaterman, line 793: We already have the existing `unlogged_write` API.  Can't we just use that?

riscv/v_ext_macros.h
```diff
@@ -329,61 +329,61 @@ static inline bool is_overlapped_widen(const int astart, int asize,
 // vector: integer and masking operand access helper
 //
 #define VXI_PARAMS(x) \
-  type_sew_t<x>::type &vd = P.VU.elt<type_sew_t<x>::type>(rd_num, i, true); \
+  type_sew_t<x>::type &vd = P.VU.elt<type_sew_t<x>::type>(rd_num, i, log); \
```
aswaterman, line 332: TBH, I think I preferred the previous, more obvious pattern where instead of `log` we had `DECODE_MACRO_USAGE_LOGGED`.  The additional template parameter makes the code harder to follow and maintain.

riscv/csrs.h
apullin: `unlogged_write` is protected, I'd have to change API contracts to use that.  

Converting to a runtime flag gives back ~14-24 percentage points of the reported speedup, but is still a speedup.

I did bake in the assumptions for MY workflow, where I want a fast binary (to run millions of times, so speed == literal money for me) that doesn't ever need logging on this path.  
But LMK if that is the shape it should take for this PR, and if the `protected` can be removed.

riscv/v_ext_macros.h
apullin: What is the preferred shape here? non-template inline overload? 
e.g.
```
void write(reg_t value, bool logged) noexcept {
    if (logged)
        write(value);
    else
        unlogged_write(value);
}
```
then
```
P.VU.vstart->write(0, DECODE_MACRO_USAGE_LOGGED);
P.VU.elt<T>(rd, i, DECODE_MACRO_USAGE_LOGGED);
```

Please provide guidance.

PR #2343, opened 2026-07-14
get rid of mmio_ok checks

EccoTheDolphin:
before this patch, hart access to debug module region was controlled in two places:

1. pmp_ok performed the normal architectural PMP checks, but **bypassed** them for debug module region accesses while the hart was in debug mode.
2. mmio_ok **rejected** debug module region accesses from harts outside debug mode.  Processor-less debug_mmu accesses remained allowed.

This patch moves the control decision entirely to pmp_ok routine: **allow** debug module regoin accesses in debug mode and **deny** them otherwise.  This check is run before the no-PMP fast path, making mmio_ok redundant.

Additionally, we flush the TLB on processor reset because it may contain stale entries from the time while hart was in debug mode.

### DIFF
--- a/riscv/mmu.cc
+++ b/riscv/mmu.cc
@@ -158,20 +158,8 @@ static reg_t reg_from_bytes(size_t len, const uint8_t* bytes)
   return res;
 }
 
-bool mmu_t::mmio_ok(reg_t paddr, size_t len, access_type UNUSED type)
-{
-  // Disallow access to debug region when not in debug mode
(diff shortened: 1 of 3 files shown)


EccoTheDolphin:
@aswaterman could you take a look, please? This is follow up from https://github.com/riscv-software-src/riscv-isa-sim/pull/2322#issuecomment-4784607874  (sorry for the delay).

aswaterman (approved):
Yep, LGTM.  Thanks.

PR #2344, opened 2026-07-14
allow different clint size

mslijepc:
In some cases `clint` we use has different size. This PR allows setting `clint` size, rather than hacking `platform.h` file. Default value remains the same.

### DIFF
--- a/riscv/clint.cc
+++ b/riscv/clint.cc
@@ -6,8 +6,8 @@
 #include "sim.h"
 #include "dts.h"
 
-clint_t::clint_t(const simif_t* sim, uint64_t freq_hz, bool real_time)
-  : sim(sim), freq_hz(freq_hz), real_time(real_time), mtime(0)
+clint_t::clint_t(const simif_t* sim, uint64_t freq_hz, bool real_time, reg_t size)
+  : sim(sim), freq_hz(freq_hz), real_time(real_time), sz(size), mtime(0)
 {
   struct timeval base;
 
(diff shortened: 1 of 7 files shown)


riscv/clint.cc
```diff
@@ -137,10 +137,9 @@ std::string clint_generate_dts(const sim_t* sim, const std::vector<std::string>&
   for (size_t i = 0; i < sim->get_cfg().nprocs(); i++)
     s << "&CPU" << i << "_intc 3 &CPU" << i << "_intc 7 ";
   reg_t clintbs = CLINT_BASE;
-  reg_t clintsz = CLINT_SIZE;
   s << std::hex << ">;\n"
     "      reg = <0x" << (clintbs >> 32) << " 0x" << (clintbs & (uint32_t)-1) <<
-    " 0x" << (clintsz >> 32) << " 0x" << (clintsz & (uint32_t)-1) << ">;\n"
+    " 0x" << (sz >> 32) << " 0x" << (sz & (uint32_t)-1) << ">;\n"
```
aswaterman: `clint_generate_dts` is not a member function of `clint_t`, so `sz` is not defined here.

riscv/devices.h
```diff
@@ -93,10 +93,10 @@ class external_sim_device_t : public abstract_device_t {
 
 class clint_t : public abstract_device_t {
  public:
-  clint_t(const simif_t*, uint64_t freq_hz, bool real_time);
+  clint_t(const simif_t*, uint64_t freq_hz, bool real_time, reg_t size = CLINT_SIZE);
```
aswaterman: Rather than using a default argument here, I think this parameter should be passed to the `clint_t` constructor.

mslijepc: I will need to add `clint_size` to `cfg_t` and the command line argument

aswaterman:
I'm now questioning the wisdom of this PR.  Why don't we just change CLINT_SIZE to the constant 0xC000, which is the absolute smallest the CLINT can be?  Making it larger than that doesn't make sense, since there are no registers defined past that point.  Making it smaller doesn't make sense, because that would prevent access to the `mtime` CSR.

I suspect that the current size of 0xC0000 is a typo and it was always meant to be 0xC000.

mslijepc:
Make sense. Will close this PR and create a new one.

PR #2352, opened 2026-07-23
Fix PNCLIPUP.W by avoiding signed arithmetic

### LINKED ISSUE #2350: Bug with `PNCLIPUP.W` instruction
[PNCLIPUP.W](https://github.com/riscv/riscv-p-spec/blob/master/P-ext-proposal.adoc#pnclipupw-rv64): Packed Narrowing Unsigned Clip (64-bit to 32-bit).
Spike casts the 64-bit source registers to a signed 64-bit integer (`sreg_t`) before evaluating saturation. Because of this, any 64-bit input with the 63rd bit set (MSB = 1) is incorrectly evaluated as a negative number rather than as an unsigned number. This leads Spike to incorrectly saturate the result to `0x00000000` instead of the expected unsigned maximum `0xFFFFFFFF`.

> **Example:**
> * **Input:** `rs1 = 0x8ed30e0630d20a98`, `rs2 = 0x83b9d1d6ae9ffccb`
> * **Operation:**
> * **Lane 0 (rs1):** `0x8ed30e0630d20a98 >_u 0x00000000FFFFFFFF` (True, must saturate to `0xFFFFFFFF`)
> * **Lane 1 (rs2):** `0x83b9d1d6ae9ffccb >_u 0x00000000FFFFFFFF` (True, must saturate to `0xFFFFFFFF`)
> 
> 
> * **Expected result:** `0xFFFFFFFFFFFFFFFF`
> * **Spike result:** `0x0000000000000000`
> 
> 

**Possible Cause**: In the [header file](https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/insns/pnclipup_w.h), `s1` and `s2` are explicitly cast to `(sreg_t)`. When an input has the highest bit set to 1, the C++ cast interprets it as a negative value (`< 0`). The `P_USAT_FULL` macro detects a value less than zero and incorrectly clamps the final output to 0.

**Possible Fix**: Remove the `(sreg_t)` cast, or updated the clamping logic to explicitly handle 64-bit unsigned comparisons.

---
Spike hash : b2d6fbf0fdce4fe36de763a03ffb1798469507a2

aswaterman:
The poorly named P_USAT_FULL macro is only used by the unsigned clip instructions, so using signed arithmetic never made sense, even though it was benign in cases where the inputs were narrower than 64 bits.

Fixes #2350

@mattiamirigaldi @chihminchao

### DIFF
--- a/riscv/insns/pnclipiu_b.h
+++ b/riscv/insns/pnclipiu_b.h
@@ -1,4 +1,4 @@
 require_rv32;
 P_NARROW_RD_RS1_ULOOP(8, 16, {
-    p_rd = P_USAT_FULL(8, (sreg_t)(p_rs1 >> insn.shamth()));
+    p_rd = P_USAT_FULL(8, p_rs1 >> insn.shamth());
 })
--- a/riscv/insns/pnclipiu_h.h
+++ b/riscv/insns/pnclipiu_h.h
@@ -1,4 +1,4 @@
 require_rv32;
 P_NARROW_RD_RS1_ULOOP(16, 32, {
-    p_rd = P_USAT_FULL(16, (sreg_t)(p_rs1 >> insn.shamtw()));
+    p_rd = P_USAT_FULL(16, p_rs1 >> insn.shamtw());
 })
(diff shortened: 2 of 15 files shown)


aswaterman:
@mattiamirigaldi @chihminchao I'm gonna go ahead and merge this; let me know if it breaks anything.

PR #2355, opened 2026-07-28
Add Zilx extension

wangpc-pp:
Implement the Zilx extension (Indexed Integer Load Instructions, v0.1), which adds integer load instructions that compute their effective address from a base register (rs2) and an index register (rs1) using the AMO major opcode.

The extension provides three addressing modes selected by funct5:
- unscaled indexed load (lx*): EA = base + index
- scaled indexed load (lxs*): EA = base + (index << log2(access-size))
- scaled unsigned-word-index load (lxsuw*, RV64-only): EA = base + (zext32(index) << log2(access-size))

lxb/lxbu are provided as assembler pseudoinstructions (disassembler aliases) for lxsb/lxsbu.

Doc: https://github.com/riscv/riscv-zilx/blob/main/src/unpriv/zilx.adoc

Assisted-by: TRAE CLI (Opus 4.8)


wangpc-pp (commented):
Ping for reviews. :-)

riscv/encoding.h
```diff
@@ -1353,6 +1353,44 @@
 #define MASK_LW_AQ 0xfdf0707f
 #define MATCH_LWU 0x6003
 #define MASK_LWU 0x707f
+#define MATCH_LXD 0x9000302f
```
aswaterman, line 1356: This file is auto-generated from `riscv-opcodes` and should not be manually modified.  The comments at the top contain the commit hash, which should have changed.

aswaterman (commented):
@wangpc-pp Thanks, this looks great.  Ping me once my comment is addressed and I'll merge it.

riscv/encoding.h
wangpc-pp: I have regenerated `encoding.h` but it contained some changes from `Zabhlrsc` and something like `MCAUSE_INT`. I just kept the Zilx part here.

wangpc-pp:
> @wangpc-pp Thanks, this looks great. Ping me once my comment is addressed and I'll merge it.

I think it is done.

aswaterman (approved):

PR #2358, opened 2026-07-30
rvv: Add checks the src registers have different EEW

Yosiky:
According to the specification[1], chapter "11.5.2 Vector Operands":
> A vector register cannot be used to provide source operands
> with more than one EEW for a single instruction. A mask
> register source is considered to have EEW=1 for this constraint.

vrgatherei16.vv vd, vs2, vs1 -- vs2_eew=sew, vs1_eew=16
vcompress.vm    vd, vs2, vs1 -- vs2_eew=sew, vs1_eew=1

[1]: The RISC-V Instruction Set Manual v202660416

### DIFF
--- a/riscv/insns/vcompress_vm.h
+++ b/riscv/insns/vcompress_vm.h
@@ -3,7 +3,7 @@ require(P.any_vector_extensions());
 require(P.VU.vstart->read() == 0);
 require_align(insn.rd(), P.VU.vflmul);
 require_align(insn.rs2(), P.VU.vflmul);
-require(insn.rd() != insn.rs2());
+require(insn.rd() != insn.rs2() && insn.rs1() != insn.rs2());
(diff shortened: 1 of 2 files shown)


aswaterman (commented):
Let's hold off on this for now, for a few reasons:
- The vrgatherei16 check raises an exception in a case where it shouldn't: when SEW=16, the overlap is allowed because the EEWs are equal.
- These checks aren't sufficient, as vector register groups can span multiple vector registers.  Consider the case that LMUL=2, vs2=v0, and vs1=v1: vcompress is still reserved in that case.
- This PR covers only 2 instructions out of the 100s that currently allow some form of reserved overlap.  We should come up with a software engineering strategy that doesn't require so much code to be written in so many places.
- Trapping on reserved behavior isn't strictly required, so Spike doesn't technically have a bug.

So I agree we should do something eventually, but it isn't urgent, and I'd like to do it more systematically.

PR #2362, opened 2026-08-06
Use standard C++ initializers for fli_* constant tables

jerryzj:
C99 array designators ([index] = value) are not valid C++ and trigger -Wc99-designator under clang. C++20 designated initializers only cover named members, not array indices, so use sequential static constexpr tables and keep rs1 encodings in comments.


aswaterman (approved):
Thanks.  I think this is a stylistic improvement, too; using designated initializers when the array is sequentially and completely populated makes the code harder to read.

PR #2363, opened 2026-08-07
Fix linker warnings about duplicate libraries

jerryzj:
Stop appending dependency .a files twice when linking shared libraries, and drop the redundant AC_CHECK_LIB(pthread) in riscv.ac so -lpthread is only added once (still checked in fesvr.ac).


aswaterman (approved):

aswaterman:
Thanks, @jerryzj!

PR #2366, opened 2026-08-11
disasm: reject 6-bit shift immediates for RV32 in strict mode

### LINKED ISSUE #2030: Incorrect decoding of SLLI for rv32
The instruction
```
0x02029e93 (32'b0000001_00000_00101_001_11101_0010011)
```
is dissasembled to 
```
slli    t4, t0, 32
```
For RV32, this should be an illegal instruction
```
unknown
```
but it is allowed as it is a valid instruction for the RV64 dissasembly despite the dissasembler being set to strict.

phverg:
Fixes #2030.

On RV32 the shift-immediate instructions `slli`/`srli`/`srai` use a 5-bit shift amount; encodings with `shamt >= 32` (instruction bit 25 set) are reserved. The disassembler registered only the 6-bit (RV64) forms unconditionally, so in strict RV32 mode it wrongly disassembled e.g. `0x02029e93` as `slli t4, t0, 32` instead of `unknown`.

This registers the `*_rv32` forms (which require bit 25 = 0) when the disassembler is strict and the maximum XLEN is 32, mirroring the existing `xlen_eq` gating already used for `slliw`/`srliw`/`sraiw`.

Before:
```
$ echo 'DASM(0x02029e93)' | spike-dasm --isa=rv32i --strict
slli    t4, t0, 32
```

After:
```
$ echo 'DASM(0x02029e93)' | spike-dasm --isa=rv32i --strict
unknown
```

RV64 (where `slli t4, t0, 32` remains valid), non-strict RV32 (lenient, unchanged), and valid RV32 shifts (shamt 0-31) are all unaffected.

### DIFF
--- a/disasm/disasm.cc
+++ b/disasm/disasm.cc
@@ -937,6 +937,7 @@ void disassembler_t::add_instructions(const isa_parser_t* isa, bool strict)
   const uint32_t mask_amoop = 0x1fUl << 27;
   const uint32_t mask_width = 0x7Ul << 12;
   const uint32_t mask_shamt_msb = 0x1Ul << 25;
+  const uint32_t mask_rvc_shamt_msb = 0x1Ul << 12;
 
   #define DECLARE_INSN(code, match, mask) \
    const uint32_t match_##code = match; \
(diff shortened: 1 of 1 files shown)


aswaterman (commented):
I suggest defining a DEFINE_ITYPE_SHIFTW macro to avoid code duplication.

Also, this appears to be incomplete (consider `roriw`).

phverg:
Both done.

Added `DEFINE_ITYPE_SHIFTW` for shift-immediates with a 5-bit shift amount, and `DEFINE_ITYPE_SHIFT` now falls back to it when the maximum XLEN is 32. Every shift-immediate is handled at its existing registration rather than once per XLEN. That picks up `rori` and the Zbs shifts `bclri`/`binvi`/`bseti`/`bexti`, which have no `*_rv32` encodings and so were out of reach of the earlier approach. The result is exactly the set the interpreter already guards with `require(SHAMT < xlen)`.

`DEFINE_ITYPE_SHIFTW` is also used directly for `slliw`, `srliw`, `sraiw` and `roriw`, whose shift amount is 5 bits regardless of XLEN. Their masks in `encoding.h` already pin bit 25, so those four are unchanged.

Checking for other cases turned up `c.slli`, `c.srli` and `c.srai`, which have the same problem in bit 12, so strict rv32ic was decoding `c.slli ra, 32`. That is the second commit.

aswaterman (approved):

aswaterman:
@phverg LGTM now, thank you!

PR #2367, opened 2026-08-12
Fix crash when SSAIA is enabled without hypervisor

david-henley:
This logic in the trap handler was gated only on the SSAIA extension. However, the CSR it accesses is only present when hypervisor is enabled.


aswaterman:
Thanks for the fix, @david-henley.  I've tweaked this by using `extension_enabled` instead of the `_const` variant, as H can be dynamically disabled.  #2368 supersedes this PR.

PR #2373, opened 2026-08-18
Don't double-count WFIs while in WFI state

### LINKED ISSUE #2372: minstret over-counts by 1 for wfi + taken timer interrupt
**Summary:**
This test checks that minstret counts correctly when wfi is executed with mstatus.MIE=1 and stalls waiting for a machine timer interrupt. Before executing wfi, I read minstret into x13. The timer interrupt is armed to fire shortly after, so once wfi is executed, the core stalls, the interrupt comes in, we trap into the handler, the handler runs its instructions and returns via mret, and execution resumes right after wfi. At that point I read minstret again into x6 and take the difference (x6 - x13), which should be the total number of instructions retired across the whole wfi → trap → handler → mret sequence.

The expected delta of 0xa8 comes from the Sail reference model run on the same binary. Spike gives 0xa9 instead one instruction more than expected.

**Failure Log:**
```
RVCP-SUMMARY: TEST FAILED - Test File "InterruptsSm-00.S"

RVCP: DEBUG INFORMATION FOLLOWS
RVCP: Test Info: "test: 100; cg: InterruptsSm_cg; cp: cp_minstret_wfi_taken; bin: minstret_wfi_taken"
RVCP: Instruction: 0x30401073
RVCP: Approximate address (failure may be slightly after this): 0x0000000080003fc4
RVCP: Register: x13
RVCP: Bad Value:      0x00000000000000a9
RVCP: Expected Value: 0x00000000000000a8
RVCP: END OF DEBUG INFORMATION

*** FAILED *** (tohost = 1)
```

The test case, Spike trace log, and ELF are attached for reference.
[InterruptsSm.zip](https://github.com/user-attachments/files/31132837/InterruptsSm.zip)

The Sail reference model log is also attached.
[InterruptsSm-00.trace.zip](https://github.com/user-attachments/files/31132851/InterruptsSm-00.trace.zip)

aswaterman:
Fixes #2372

### DIFF
--- a/riscv/decode_macros.h
+++ b/riscv/decode_macros.h
@@ -251,13 +251,6 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
        STATE.pc = __npc; \
      } while (0)
 
-class wait_for_interrupt_t {};
-
-#define wfi() \
-  do { set_pc_and_serialize(npc); \
-       throw wait_for_interrupt_t(); \
-     } while (0)
-
 #define serialize() set_pc_and_serialize(npc)
 
 /* Sentinel PC values to serialize simulator pipeline */
--- a/riscv/execute.cc
+++ b/riscv/execute.cc
@@ -177,11 +177,6 @@ static inline reg_t execute_insn_logged(processor_t* p, reg_t pc, insn_fetch_t f
         commit_log_print_insn(p, pc, fetch.insn);
       }
      }
-  } catch (wait_for_interrupt_t &t) {
-      if (p->get_log_commits_enabled()) {
-        commit_log_print_insn(p, pc, fetch.insn);
-      }
-      throw;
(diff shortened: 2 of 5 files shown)


fkhaidari:
I guess this PR dropped the Debug Mode half of the old WFI handling. A WFI in Debug Mode used to end `step()`, and now it seems to retire as a nop while execution keeps going. The debug ROM parks a halted hart in a `wfi` loop, so the hart burns its whole INTERLEAVE budget before `sim_t` gets back to the debug module and remote bitbang. For us, OpenOCD against spike got about 10x slower.

Minimal repro: fill memory with WFIs, put hart 0 in Debug Mode at the first one, and call `step(1000)`. Before this PR pc advances by 4, now by 4000.

I have fixed that locally, so I will probably open a PR.

PR #2391, opened 2026-08-25
sign-extend zfinx single-precision results on RV64

### LINKED ISSUE #2390: RV64 Zfinx zero-extends a single-precision result instead of sign-extending it
With Zfinx on RV64, `WRITE_FRD_F` writes a single-precision result to an `x` register zero-extended. The Zfinx specification requires sign extension I believe.

In order to reproduce:

```
unsigned long in = 0xbf800000UL;   // raw single-precision -1.0f
unsigned long out;

asm volatile ("fsgnj.s %0, %1, %1" : "=r" (out) : "r" (in));
```

Logs shows

```
core 0: 3 0x0000000080000030 (0x205280d3) x1 0x00000000bf800000
```

the expected value of x1 should be `0xffffffffbf800000`

akifejaz:
Fixes #2390

### DIFF
--- a/riscv/decode_macros.h
+++ b/riscv/decode_macros.h
@@ -141,7 +141,7 @@ do { \
 #define WRITE_FRD_F(value) \
 do { \
   if (p->extension_enabled(EXT_ZFINX)) \
-    WRITE_REG(insn.rd(), sext_xlen((int32_t)((value).v))); \
+    WRITE_REG(insn.rd(), sext32((value).v)); \
   else { \
     WRITE_FRD(value); \
   } \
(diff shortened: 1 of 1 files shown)


riscv/decode_macros.h
```diff
@@ -141,7 +141,7 @@ do { \
 #define WRITE_FRD_F(value) \
 do { \
   if (p->extension_enabled(EXT_ZFINX)) \
-    WRITE_REG(insn.rd(), sext_xlen((value).v)); \
+    WRITE_REG(insn.rd(), sext_xlen((int32_t)((value).v))); \
```
aswaterman: `sext32((value).v)` is the preferred idiom

aswaterman (approved):

PR #2395, opened 2026-08-27
Zalasr address fault exception

### LINKED ISSUE #2392: Zalasr lw.aq and sw.rl are succeed instead of trapping when Zicclsm is enabled
The eight Zalasr instructions call the ordinary `MMU.load<T>` / `MMU.store<T>` with default `xlate_flags`. When `Zicclsm` is enabled, a misaligned access is therefore split at the page boundary and completed as up to two sub-accesses.

The access succeeds, instead of raising a misaligned-address exception:
`LoadAddressMisaligned` (cause 4) for `lb.aq`, `lh.aq`, `lw.aq` and `ld.aq`, and
`StoreAmoAddressMisaligned` (cause 6) for `sb.rl`, `sh.rl`, `sw.rl` and `sd.rl`.

like for `lw_aq.h`:

```c
require_extension(EXT_ZALASR);
WRITE_RD(MMU.load<int32_t>(RS1));
```
If we've
```c
static volatile unsigned char buf[64] __attribute__((aligned(64)));
const void *addr = (const void *)&buf[1];   // misaligned by one byte
unsigned long got;

// lw.aq rd, (rs1)   MATCH_LW_AQ 0x3400202f
asm volatile (".insn r 0x2f, 0x2, 0x1a, %0, %1, x0"
              : "=r" (got) : "r" (addr) : "memory");
```

Commit log for the misaligned access:

```
core 0: 3 0x0000000080000054 (0x3402a32f) x6 0xffffffffddaabbcc mem 0x0000000080003001
```

The expected was `mcause=4` I believe.

akifejaz:
Fixes #2392

### DIFF
--- a/riscv/insns/lb_aq.h
+++ b/riscv/insns/lb_aq.h
@@ -1,2 +1,2 @@
 require_extension(EXT_ZALASR);
-WRITE_RD(MMU.load<int8_t>(RS1));
+WRITE_RD(MMU.load_strict<int8_t>(RS1));
--- a/riscv/insns/ld_aq.h
+++ b/riscv/insns/ld_aq.h
@@ -1,3 +1,3 @@
 require_rv64;
 require_extension(EXT_ZALASR);
-WRITE_RD(MMU.load<int64_t>(RS1));
+WRITE_RD(MMU.load_strict<int64_t>(RS1));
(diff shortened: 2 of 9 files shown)


aswaterman (approved):
Thanks.  This looks like what we want functionally.  I'm going to do some refactoring after merging, as I'm trying to reduce the number of unique paths through the MMU code.

PR #2398, opened 2026-08-28
draft: debug module start address

### LINKED ISSUE #2397: Custom start address of debug module
In case in the system we have a debug address on address different than `0` (which is defined in `platform.h`), Spike should have a support for it. We should add support for CLI, and DTS parsing (if both are set, that should be an error)

mslijepc:
Closes https://github.com/riscv-software-src/riscv-isa-sim/issues/2397

### DIFF
--- a/ci-tests/custom-csr.cc
+++ b/ci-tests/custom-csr.cc
@@ -68,7 +68,8 @@ int main(int argc, char **argv) {
                                      .support_abstract_csr_access = true,
                                      .support_abstract_fpr_access = true,
                                      .support_haltgroups = true,
-                                     .support_impebreak = true};
+                                     .support_impebreak = true,
+                                     .debug_start = 0x0};
   std::vector<std::pair<reg_t, abstract_mem_t *>> mems =
       make_mems(cfg.mem_layout);
   bool dtb_discovery=false;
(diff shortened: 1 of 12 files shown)


chihminchao (commented):
I am ok to use dts node to adjust debut module base.

fkhaidari:
Hi @mslijepc! Configurable debug module base sounds like a good idea. When updating the debug ROM, please keep in mind:
1. The generated abstract commands also need to account for the base, including the DEBUG_ROM_SAVED_STATE accesses ([scratch storage](https://github.com/riscv-software-src/riscv-isa-sim/commit/0e801337e73ae3536958f4fbd675b62c662d9db5), [register commands](https://github.com/riscv-software-src/riscv-isa-sim/commit/b4e66312e2dbc0e5e2ce52788a8f74bad6abcdcc), [memory commands](https://github.com/riscv-software-src/riscv-isa-sim/commit/1860bc73601125801087e6c918a51c39c5cc56ae)).
2. The extra instructions must fit in the abstract command buffer.
3. The sequences already use `s0` and `s1`, with their original values saved in `dscratch0/1`. Address calculations must preserve the values needed on both the normal and exception paths.

PR #2399, opened 2026-08-29
Fix mip SEIP read-modify-write semantics

### LINKED ISSUE #2360: csrrs/csrrc on mip read-modify-write the wrong SEIP bit, latching the interrupt-controller signal
RISCV spec says [link](https://github.com/riscv/riscv-isa-manual/blob/310a111489a0bad6e60ef4cbfba574417c6f825f/src/priv/machine.adoc#L1435-L1440):
When csr:mip[] is read with a CSR instruction, the value of the csr::[seip] bit returned in the rd 
destination register is the logical-OR of the software-writable bit and the interrupt signal from 
the interrupt controller, but the signal from the interrupt controller is not used to calculate the 
value written to csr::[seip]. Only the software-writable csr::[seip] bit participates in the 
read-modify-write sequence of a insn:csrrs[] or insn:csrrc[] instruction.

However, the bug in spike is that a read-modify-write can write the interrupt signal from the interrupt controller into the software-writable SEIP bit.

Spike hash : [3d8eb0](https://github.com/riscv-software-src/riscv-isa-sim/commit/3d8eb089bd289c59dcb506f197a172e02beb7b5b)

Example test case: 
```
// build and run
// riscv64-unknown-elf-gcc -O2 -specs=htif.specs -o seip-rmw seip-rmw.c
// spike --isa=rv64gc seip-rmw
#include <stdint.h>
#include <stdio.h>

/* ------------------------------------------------------------------ platform */

#define PLIC_BASE           0x0c000000UL
#define PLIC_PRIORITY(src)  (PLIC_BASE + 0x000000UL + 4UL * (src))
#define PLIC_PENDING(src)   (PLIC_BASE + 0x001000UL + 4UL * ((src) / 32))
#define PLIC_ENABLE(ctx, src) \
	(PLIC_BASE + 0x002000UL + 0x80UL * (ctx) + 4UL * ((src) / 32))
#define PLIC_THRESHOLD(ctx) (PLIC_BASE + 0x200000UL + 0x1000UL * (ctx))

#define UART_BASE           0x10000000UL
#define UART_IER            1
#define UART_IER_THRI       0x02

#define UART_IRQ            1

// hart 0 supervisor context index 1
#define SUPERVISOR_CTX      1

// Any priority above the threshold (0) will do
#define IRQ_PRIORITY        7

#define MIP_STIP            (1UL << 5)
#define MIP_SEIP            (1UL << 9)

/* --------------------------------------------------------------- primitives */

static inline void mmio_w32(uintptr_t a, uint32_t v) { *(volatile uint32_t *)a = v; }

static inline void mmio_w8(uintptr_t a, uint8_t v) { *(volatile uint8_t *)a = v; }

static inline uint32_t mmio_r32(uintptr_t a) { return *(volatile uint32_t *)a; }

static inline int plic_pending(unsigned src) { return (mmio_r32(PLIC_PENDING(src)) >> (src % 32)) & 1; }

static inline uint64_t read_mip(void)
{
	uint64_t v;
	__asm__ volatile ("csrr %0, mip" : "=r"(v));
	return v;
}

static inline void clear_mip(uint64_t bits) { __asm__ volatile ("csrc mip, %0" :: "r"(bits)); }

/* ------------------------------------------------------------------ reporting */

/*
 * Lower 16 bits of mip, laid out as in the privileged spec:
 *
 *   15 14 13 12   11   10    9    8    7    6    5    4    3    2    1    0
 *    0  0  0  0 MEIP    0 SEIP    0 MTIP    0 STIP    0 MSIP    0 SSIP    0
 */
static const char *const MIP_FIELD[16] = {
	"0",    "SSIP", "0",    "MSIP", "0",    "STIP", "0",    "MTIP",
	"0",    "SEIP", "0",    "MEIP", "0",    "0",    "0",    "0"
};

static void print_mip_header(void)
{
	printf("%-14s", "  mip bit");
	for (int i = 15; i >= 0; i--)
		printf("%5d", i);
	printf("\n%-14s", "  field");
	for (int i = 15; i >= 0; i--)
		printf("%5s", MIP_FIELD[i]);
	printf("\n");
}

static void print_mip(const char *label, uint64_t mip)
{
	printf("  %-12s", label);
	for (int i = 15; i >= 0; i--)
		printf("%5d", (int)((mip >> i) & 1));
	printf("     (0x%lx)\n", mip);
}

/* ------------------------------------------------------------------- the test */

int main(void)
{
	// Program the PLIC
	mmio_w32(PLIC_PRIORITY(UART_IRQ), IRQ_PRIORITY);
	mmio_w32(PLIC_THRESHOLD(SUPERVISOR_CTX), 0);
	mmio_w32(PLIC_ENABLE(SUPERVISOR_CTX, UART_IRQ), 1u << (UART_IRQ % 32));

	uint64_t mip = read_mip();
	printf("seip-rmw: lower 16 bits of mip at each step\n");
	print_mip_header();
	print_mip("initial", mip);

	/*
	* SEQUENCE: assert the interrupt, which should be reflected in mip.SEIP.
	* When a csrrc clears mip.STIP, mip.SEIP should not be updated.
	* Then when the interrupt is cleared, mip.SEIP should also clear, but the bug
	* in spike shows that mip.SEIP is actually set
	*/

	// assert the interrupt
	mmio_w8(UART_BASE + UART_IER, UART_IER_THRI);
	printf("  %-12s PLIC reports source %u pending = %d\n",
	       "", UART_IRQ, plic_pending(UART_IRQ));

	uint64_t asserted = read_mip();
	print_mip("IRQ asserted", asserted);

	// THE INSTRUCTION UNDER TEST. A read-modify-write of a DIFFERENT mip bit
	clear_mip(MIP_STIP);

	uint64_t after = read_mip();
	print_mip("after csrc", after);

	// deassert the interrupt
	mmio_w8(UART_BASE + UART_IER, 0);
	printf("  %-12s PLIC reports source %u pending = %d\n",
	       "", UART_IRQ, plic_pending(UART_IRQ));

	// Software never wrote mip.SEIP and the interrupt is deasserted so SEIP should be 0 now.
	// but it's not
	uint64_t final_mip = read_mip();
	print_mip("IRQ cleared", final_mip);

	if (final_mip & MIP_SEIP) {
		printf("seip-rmw: FAIL - mip.SEIP stuck set after the interrupt"
		       " controller deasserted.\n"
		       "seip-rmw: the csrrc read-modify-write copied the interrupt"
		       " signal into the\n"
		       "seip-rmw: software-writable SEIP bit, which nothing can"
		       " clear.\n");
		return 1;
	}

	printf("seip-rmw: PASS -- mip.SEIP returned to 0.\n");
	return 0;
}
```

Result: 
```
seip-rmw: lower 16 bits of mip at each step
  mip bit        15   14   13   12   11   10    9    8    7    6    5    4    3    2    1    0
  field           0    0    0    0 MEIP    0 SEIP    0 MTIP    0 STIP    0 MSIP    0 SSIP    0
  initial         0    0    0    0    0    0    0    0    1    0    0    0    0    0    0    0     (0x80)
               PLIC reports source 1 pending = 1
  IRQ asserted    0    0    0    0    0    0    1    0    1    0    0    0    0    0    0    0     (0x280)
  after csrc      0    0    0    0    0    0    1    0    1    0    0    0    0    0    0    0     (0x280)
               PLIC reports source 1 pending = 0
  IRQ cleared     0    0    0    0    0    0    1    0    1    0    0    0    0    0    0    0     (0x280)
seip-rmw: FAIL - mip.SEIP stuck set after the interrupt controller deasserted.
seip-rmw: the csrrc read-modify-write copied the interrupt signal into the
seip-rmw: software-writable SEIP bit, which nothing can clear.
*** FAILED *** (tohost = 1)
```

SEIP after `PLIC reports source 1 pending = 0` should be 0.

sinhaparth5:
Fixes #2360.

`mip.SEIP` reads as the OR of its software-writable bit and the external interrupt-controller signal. CSRRS and CSRRC were feeding that combined read value back into the write, which could latch an asserted external SEIP signal into the software bit.

This adds a separate read value for CSR read-modify-write operations. Most CSRs use their normal read value; `mip` excludes the externally driven SEIP contribution while retaining the software-writable bit. The value returned to `rd` is unchanged.

The regression raises SEIP through the PLIC, performs a CSRRC operation on an unrelated `mip` bit, clears the PLIC source, and checks that SEIP does not remain set.

Tested on a clean Debian 12 VM:

- Regression fails against unmodified `master` (exit 1)
- Regression passes with this patch (exit 0)
- `make check` passes

### DIFF
--- a/ci-tests/seip-rmw.S
+++ b/ci-tests/seip-rmw.S
@@ -1,0 +1,63 @@
+.section .text.init
+.globl _start
+_start:
+  li t0, 0x0c000004
+  li t1, 7
+  sw t1, 0(t0)
+
+  li t0, 0x0c201000
+  sw zero, 0(t0)
+
+  li t0, 0x0c002080
+  li t1, 2
+  sw t1, 0(t0)
+
+  li t0, 0x10000001
+  li t1, 2
+  sb t1, 0(t0)
+
+  li t0, 0x0c001000
+1:
+  lw t1, 0(t0)
+  andi t1, t1, 2
+  beqz t1, 1b
+
+  li t0, 1 << 5
+  csrrc t1, mip, t0
+  andi t1, t1, 1 << 9
+  beqz t1, fail
+
+  li t0, 0x10000001
+  sb zero, 0(t0)
+
+  li t0, 0x0c001000
+2:
+  lw t1, 0(t0)
+  andi t1, t1, 2
+  bnez t1, 2b
+
+  csrr t1, mip
+  andi t1, t1, 1 << 9
+  bnez t1, fail
+
+  li t1, 1
+  j exit
+
+fail:
+  li t1, 3
+
+exit:
+  lla t0, tohost
+  sd t1, 0(t0)
+3:
+  j 3b
+
+.section .tohost,"aw",@progbits
+.balign 64
+.globl tohost
+tohost:
+  .dword 0
+.balign 64
+.globl fromhost
+fromhost:
+  .dword 0
--- a/ci-tests/test-spike
+++ b/ci-tests/test-spike
@@ -27,6 +27,9 @@ riscv64-linux-gnu-gcc -static -O2 -o hello $CI/hello.c
 riscv64-linux-gnu-gcc -static -O2 -o dummy-slliuw $CI/dummy-slliuw.c
 riscv64-linux-gnu-gcc -static -O2 -o customcsr $CI/customcsr.c
 riscv64-linux-gnu-gcc -static -O2 -o atomics $CI/atomics.c
+riscv64-linux-gnu-gcc -nostdlib -nostartfiles -static -no-pie -fno-pie \
+  -Wl,--build-id=none -Wl,-Ttext-segment=0x80000000 \
+  -Wl,--section-start=.tohost=0x81000000 -o seip-rmw $CI/seip-rmw.S
 riscv64-linux-gnu-gcc -static -O2 -march=rv64gcv -o vector-sum $CI/vector-sum.c
 
 # run snippy-based tests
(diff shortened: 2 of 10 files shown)


aswaterman (approved):
Thanks for the fix!

PR #2402, opened 2026-08-31
triggers: avoid oversized shifts for NAPOT matching

neosys007:
Build the NAPOT mask with the target register type and avoid shifting by the target XLEN. This removes undefined host-integer shifts and handles a full-width NAPOT range with a zero comparison mask.


aswaterman (approved):
Thanks for the fix!  I'm going to make a stylistic tweak as a follow-on PR.

PR #2408, opened 2026-09-01
[NFC] Use standard c++ style system headers

darthscsi:
Add clang-tidy rules to check.

Also, move the more restrictive ci build flags into the normal build flow.


aswaterman:
@darthscsi oops, I didn't see you had fixed the merge conflict, and now it's conflicting again :) I've fixed it this time.

PR #2409, opened 2026-09-02
[NFC] Use std::string::starts_with() instead of substr/compare/find idioms

darthscsi:
Add .clang-tidy configuration to enforce this.

.clang-tidy will conflict depending on order of merges.  Will update this or other PRs based on order.


aswaterman (approved):
Much cleaner, thanks for the sanitization.

PR #2423, opened 2026-09-16
[NFC]  Prefer move to copy for passed-by-value parameters

darthscsi:
Lock in behavior with .clang-tidy


darthscsi:
Reduces text segment size by a bit.

aswaterman (approved):

PR #2433, opened 2026-09-18
Ignore debug-entry requests while already in Debug Mode

fkhaidari:
Trap handling can enter Debug Mode before single-step completion is processed. The later STEP request previously overwrote dcsr.cause and dpc. enter_debug_mode() now returns if the hart is already in Debug Mode.


aswaterman (commented):
@fkhaidari I'm worried this change hides a problem rather than fixing it.  What exact situation gives rise to it, and why do you think it's better to allow `enter_debug_mode` to be called from within debug mode, rather than not calling it and taking some other action?

After all, this is basically saying "ignore the exception and re-execute the faulting instruction", since the effect is that the PC doesn't get updated.

fkhaidari:
@aswaterman Moved the checks to the call sites and removed the guard from `enter_debug_mode()`.

The test single-steps an illegal instruction with Smdbltrp enabled and `mstatus.MDT=1`. The corresponding scenario is:

```asm
    la   t0, trap_handler
    csrw mtvec, t0
    ecall
1:  j    1b

    .balign 4
trap_handler:
    .word 0
```

Start in M-mode with MDT cleared. Set a hw bp at `trap_handler`, then single-step `.word 0`. The `ecall` sets MDT, so the illegal instruction causes a double trap.

Critical-error handling enters Debug Mode and sets PC to `DEBUG_ROM_ENTRY`. Single-step completion then enters again, overwriting `dcsr.cause` and `dpc`. Skipping the second call leaves PC at the debug ROM entry; it does not re-execute the faulting instruction.

aswaterman (approved):
Thanks, the version with the checks at the call sites is less scary :)

PR #2439, opened 2026-09-21
Implement hedelegh, gated by mstateen0.P1P13

### LINKED ISSUE #2431: hedelegh not implemented
Related to #2382, hedelegh is not implemented.  Once implemented, it should be controlled by mstateen0.P1P13.

Harshit6-b:
Fixes #2431

**Issue:** On RV32, `hedelegh` was not implemented, so accessing it always
raised illegal instruction. `mstateen0.P1P13`, which should control it, could
not be set.

**Fix:**
- Add `hedelegh` on RV32 as the upper half of `hedeleg`, like `medelegh`.
- Below M-mode, access requires `mstateen0.P1P13` when Smstateen is present.
- Make P1P13 writable when H is present.

**Tested:** In M, HS, VS and U modes, with and without Smstateen, on RV32 and
RV64. The test fails on current master and passes with this change.


aswaterman (approved):
LGTM, thanks for fixing this.

PR #2441, opened 2026-09-23
Return to the outer simulation loop after a WFI in Debug Mode

fkhaidari:
Since #2373 a WFI in Debug Mode no longer ends `step()`. The debug ROM parks halted harts in a WFI loop, so every remote bitbang tick now costs each halted hart a full INTERLEAVE slice, and OpenOCD sessions slow down by an order of magnitude.

WFI now calls a new `yield()` macro instead of `serialize()`. It retires the instruction and ends the step through a `PC_YIELD` sentinel. `in_wfi` is left alone.

`ci-tests/test-debug-wfi` checks that `step(1000)` on a halted hart retires exactly one WFI. It fails on current master.


riscv/insns/wfi.h
```diff
@@ -10,3 +10,7 @@ if (get_field(STATE.mstatus->read(), MSTATUS_TW)) {
 }
 STATE.in_wfi = !STATE.debug_mode && !p->get_cfg().wfi_as_nop;
 serialize();
+// In Debug Mode WFI is a nop, but it still ends the step, so that the debug
+// ROM's idle loop does not eat the whole instruction budget.
+if (STATE.debug_mode)
+  npc = PC_END_STEP;
```
aswaterman: Two suggestions:
- There doesn't appear to be any reason to restrict this control-flow change to debug mode; you can just get rid of the `if`, I think.
- Messing with `npc` and using magic PCs inside of the instruction definitions breaks an abstraction barrier.  Instead, refactor as follows, and call `yield()` instead of `serialize()`:

```
diff --git a/riscv/decode_macros.h b/riscv/decode_macros.h
index 18aee685..b60fcfa3 100644
--- a/riscv/decode_macros.h
+++ b/riscv/decode_macros.h
@@ -236,12 +236,16 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
        npc = sext_xlen(x); \
      } while (0)
 
-#define set_pc_and_serialize(x) \
+#define set_special_pc(x, special) \
   do { reg_t __npc = (x) & p->pc_alignment_mask(); \
-       npc = PC_SERIALIZE_AFTER; \
+       npc = (special_pc); \
        STATE.pc = __npc; \
      } while (0)
 
+#define set_pc_and_serialize(x) set_special_pc(x, PC_SERIALIZE_AFTER)
+
+#define yield() set_special_pc(npc, PC_END_STEP)
+
 #define serialize() set_pc_and_serialize(npc)
 
 /* Sentinel PC values to serialize simulator pipeline */
```

riscv/decode_macros.h
```diff
@@ -247,6 +247,7 @@ static inline bool is_aligned(const unsigned val, const unsigned pos)
 /* Sentinel PC values to serialize simulator pipeline */
 #define PC_SERIALIZE_BEFORE 3
 #define PC_SERIALIZE_AFTER 5
+#define PC_END_STEP 7
```
aswaterman: `PC_YIELD`?

fkhaidari: Addressed

riscv/insns/wfi.h
fkhaidari: Done, thanks

aswaterman (approved):
Thanks!

PR #2442, opened 2026-09-23
Fix action=0 trigger re-entrancy after delegated traps

### LINKED ISSUE #2422: Spike: action=0 etrigger fires re-entrantly in S-mode when the exception and breakpoint are both delegated to S-mode
**Description**

A re-entrancy issue is observed in Spike with an etrigger configured for an exception that occurs in U-mode and is delegated to S-mode, while breakpoint exception is also delegated to S-mode.

In this scenario, the hart is executing in U-mode and an exception occurs. The exception is delegated to S-mode, so trap entry occurs into the S-mode trap handler. The etrigger is configured for the corresponding exception with action=0, and medeleg[3]=1, so the breakpoint exception generated by the trigger is also delegated to S-mode.

Since SIE is cleared when entering the S-mode trap handler, the action=0 etrigger should not fire again in this situation.

**Observed behavior**

The trace shows:

> core   0: 0x0000000080000098 (0x100513ffffffff) unknown
core   0: exception trap_illegal_instruction, epc 0x0000000080000098
core   0:                 tval 0x00000000ffffffff
core   0: trigger action 0
core   0: exception trap_breakpoint, epc 0x0000000080001980
core   0:                tval 0x0000000000000000
core   0: >>>>  Strampoline

The trace shows the original exception followed by trigger action 0 and then a trap_breakpoint, indicating that the etrigger fires re-entrantly during the S-mode trap handling.
This results in the trap handler being entered again and eventually causes the ACT4 trap signature to overflow:

> *** FAILED *** (tohost = 1)
    RVCP-SUMMARY: TEST FAILED - Test File "SdtrigSm_Etrigger-00.S"
    RVCP: DEBUG INFORMATION FOLLOWS
    RVCP: Trap signature overflow in sig mode. Increase TRAP_SIGUPD_COUNT for this test.
    RVCP: END OF DEBUG INFORMATION

The RISC-V Debug Specification states under the action=0 re-entrancy Solution 1:

> "If medeleg[3]=1 then it prevents triggers with action=0 from matching or firing while in S-mode and while SIE in sstatus is 0."

Therefore, for an exception delegated from U-mode to S-mode with medeleg[3]=1, shouldn't an action=0 etrigger be prevented from firing while in S-mode with SIE=0?

arshiarifat-10x:
`trap_common_t::detect_trap_match()` intentionally calls:

```cpp
common_match(proc, true)
```

so that trap triggers are matched using the privilege and virtualization state of the event that caused the trap.

Consequently, `common_match()` uses `state->prev_prv` and `state->prev_v` for this path.

However, the action=0 re-entrancy restriction applies to the privilege mode in which the processor is currently executing. Using the selected `prv`/`v` values for this check therefore evaluates the restriction against the pre-trap privilege state.

For a U-mode exception delegated to S-mode, the processor state is:

```text
state->prev_prv = PRV_U
state->prv      = PRV_S
```

The existing S-mode re-entrancy check consequently evaluates against `PRV_U` and does not prevent the action=0 trigger from firing again.

### Fix

Use the current processor privilege and virtualization state (`state->prv` and `state->v`) for the action=0 re-entrancy checks.

The existing `prev_prv`/`prev_v` behavior is retained for normal trap-trigger matching.

This separates the two requirements:

* `prev_prv`/`prev_v`: identify the privilege context of the event being matched.
* `state->prv`/`state->v`: determine whether action=0 triggering is currently prohibited by the re-entrancy rules.

Fixes #2422


aswaterman (approved):
Thanks!

